PR #14, opened 2016-09-23
Added demo.

balajismaniam:
- Added helper scripts for running the apps.
- Added templates to express affinity and anti-affinity.
- Updated/Added readme appropriately.


demo/helper-scripts/demo-pod-with-discovery.json.parsec.template
```diff
@@ -0,0 +1,25 @@
+{
+    "apiVersion": "v1",
+    "kind": "Pod",
+    "metadata": {
+        "name": "demo-parsec-NUM"
+    },
+    "spec": {
+        "containers": [
+            {
+                "image": "intelsdi/node-feature-discovery-APP",
```
ConnorDoyle: Could we host these container images together with the main NFD image on quay.io? e.g. in the incubator org

ConnorDoyle: It would be nice to also include the image sources (e.g. as a git-submodule) so anyone can rebuild /publish them and it's not just a random container from the internet.

ConnorDoyle: If you want, we can create a separate issue to do that. It doesn't need to block merging the rest of the demo resources.

balajismaniam: I will move the images to quay.io and add the dockerfile(s) in a separate PR.

Issue #27, opened 2016-10-25
Consider updating the label namespace


ConnorDoyle:
Right now NFD publishes labels that begin with `node.alpha.intel.com`.
Does it make sense to change this to `node.alpha.kubernetes-incubator.io`?

cc @davidopp @timothysc @balajismaniam

davidopp:
@bgrant0607

timothysc:
I don't have strong opinions here, just know that someone is going to hard match on that label.

ConnorDoyle:
@timothysc same here. We want to tag our first release soon but also want to avoid label churn (even though we will be in the `0.x.y` series.)

fabiand:
I'd say that we can start with vendor labels - which got the "vendor freedom".
That allows us to have a quick start, and get vendors/sservices providing their labels for discovered functionality.

But on the mid-term run we maybe want to provide labels which have a defined name and values.
These labels should get the `io.kubernetes` prefix (I'm considering a reverse fqdn notation).

ConnorDoyle:
@fabiand IMO reverse-fqdn is not advisable since it goes against precedent for alpha/beta prefixing already present in core Kubernetes.

As for vendor labels, so far there's a common prefix so that we can always track down labels that were added by this discovery pod. Each source (implementations of the `Source` interface in this project, there are a few already) has a name that goes into the label. Effectively the vendor namespace could be just `<common-prefix>/<vendor>`.

ConnorDoyle:
Since there aren't too many objections here I'd like to get #27 merged (it's better IMO than the status quo: Intel's name on every label, that makes no sense.) We can continue to iterate if necessary.

fabiand:
Yes, I don't have strong objections - my comment was merely a small comment.

Issue #25, opened 2016-10-25
node-feature-discovery should delete non-discovered feature labels


ConnorDoyle:
For example, if NFD runs when turbo-boost is turned on and then later runs when turbo-boost is turned off, the label should no longer indicate that turbo-boost is enabled on that node.

cc @balajismaniam

PR #28, opened 2016-10-25
Removed version from published label keys.

### LINKED ISSUE #24: Remove version string from published label key
There has been feedback from many people that including the version in the published label is bad because it causes affinity annotations in pod specs to be invalidated upon upgrading the version of node-feature-discovery.

ConnorDoyle:
Fixes #24

### DIFF
--- a/main.go
+++ b/main.go
@@ -28,7 +28,7 @@ const (
 
 var (
 	version = "" // Must not be const, set using ldflags at build time
-	prefix  = fmt.Sprintf("%s/%s", Namespace, version)
+	prefix  = fmt.Sprintf("%s/nfd", Namespace)
 )
 
 // Labels are a Kubernetes representation of discovered features.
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -28,7 +28,7 @@ const (
 
 var (
 	version = "" // Must not be const, set using ldflags at build time
-	prefix  = fmt.Sprintf("%s/%s", Namespace, version)
+	prefix  = fmt.Sprintf("%s/", Namespace)
```
ConnorDoyle: I'm leaning toward adding some more identifying info to the prefix so we can implement #25. Maybe something simple like `node.alpha.intel.com/nfd-` as the prefix would help us know what we can safely delete. It's probably not reasonable for NFD to assume it owns the whole `node.alpha.intel.com` label namespace. Related to #27.

ConnorDoyle: OK, will make this change.

ConnorDoyle:
@balajismaniam updated

PR #39, opened 2016-10-29
Fixes #25: Deletes non-discovered nfd labels

### LINKED ISSUE #25: node-feature-discovery should delete non-discovered feature labels

IRCody:
Took a stab at #25. Appreciate any feedback.

Modifies AddLabels() to first remove any non-discovered nfd labels from
the node labels before adding newly discovered labels.

Adds tests to main_test.go to test the behavior of AddLabels.

edit:

This deletes non-discovered labels including those that may not be "discovered" because of the whitelist option. Wanted to make sure I called that out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,11 @@ the only label value published for features is the string `"true"`._
 
 The `--sources` flag controls which sources to use for discovery.
 
+_Note: Consecutive runs of node-feature-discovery will update the labels on a
+given node. If features are not discovered on a consecutive run, the corresponding
+label will be removed. This includes any restrictions placed on the consecutive run,
+such as restricting discovered features with the --label-whitelist option._
+
 ### Intel Resource Director Technology (RDT) Features
 
 | Feature name   | Description                                                                         |
--- a/main.go
+++ b/main.go
@@ -43,7 +43,12 @@ type APIHelpers interface {
 	// GetNode returns the Kubernetes node on which this container is running.
 	GetNode(*client.Client) (*api.Node, error)
 
-	// addLabels modifies the supplied node's labels collection.
+	// RemoveLabels removes labels from the supplied node that contain the
+	// search string provided. In order to publish the changes, the node must
+	// subsequently be updated via the API server using the client library.
+	RemoveLabels(*api.Node, string)
+
+	// AddLabels modifies the supplied node's labels collection.
 	// In order to publish the labels, the node must be subsequently updated via the
 	// API server using the client library.
 	AddLabels(*api.Node, Labels)
@@ -181,6 +186,8 @@ func advertiseFeatureLabels(helper APIHelpers, labels Labels) error {
 		return err
 	}
 
+	// Remove labels with our prefix
+	helper.RemoveLabels(node, prefix)
 	// Add labels to the node object.
 	helper.AddLabels(node, labels)
 
@@ -229,6 +236,16 @@ func (h k8sHelpers) GetNode(cli *client.Client) (*api.Node, error) {
 	}
 
 	return node, nil
+}
+
+// RemoveLabels searches through all labels on Node n and removes
+// any where the key contain the search string.
+func (h k8sHelpers) RemoveLabels(n *api.Node, search string) {
+	for k := range n.Labels {
+		if strings.Contains(k, search) {
+			delete(n.Labels, k)
+		}
+	}
 }
 
 func (h k8sHelpers) AddLabels(n *api.Node, labels Labels) {
--- a/main_test.go
+++ b/main_test.go
@@ -56,6 +56,7 @@ func TestDiscoveryWithMockSources(t *testing.T) {
 			mockAPIHelper.On("GetClient").Return(mockClient, nil)
 			mockAPIHelper.On("GetNode", mockClient).Return(mockNode, nil).Once()
 			mockAPIHelper.On("AddLabels", mockNode, fakeFeatureLabels).Return().Once()
+			mockAPIHelper.On("RemoveLabels", mockNode, prefix).Return().Once()
 			mockAPIHelper.On("UpdateNode", mockClient, mockNode).Return(nil).Once()
 			err := advertiseFeatureLabels(testHelper, fakeFeatureLabels)
 
@@ -89,6 +90,7 @@ func TestDiscoveryWithMockSources(t *testing.T) {
 			expectedError := errors.New("fake error")
 			mockAPIHelper.On("GetClient").Return(mockClient, nil)
 			mockAPIHelper.On("GetNode", mockClient).Return(mockNode, nil).Once()
+			mockAPIHelper.On("RemoveLabels", mockNode, prefix).Return().Once()
 			mockAPIHelper.On("AddLabels", mockNode, fakeFeatureLabels).Return().Once()
 			mockAPIHelper.On("UpdateNode", mockClient, mockNode).Return(expectedError).Once()
 			err := advertiseFeatureLabels(testHelper, fakeFeatureLabels)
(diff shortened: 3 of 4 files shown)


main.go
```diff
@@ -232,6 +232,15 @@ func (h k8sHelpers) GetNode(cli *client.Client) (*api.Node, error) {
 }
 
 func (h k8sHelpers) AddLabels(n *api.Node, labels Labels) {
+	// Remove nfd node labels that no longer apply
```
ConnorDoyle: Could you move this stuff out to another function? Removing labels in a function called `AddLabels` is just asking for trouble :)

ConnorDoyle:
Thanks for taking this on @IRCody! @balajismaniam I unassigned myself since I'll be out. Would you mind reviewing & shepherding it?

main.go
IRCody: Would it make sense to rename the function to UpdateLabels?

balajismaniam:
@ConnorDoyle: Sure I can do that.

main.go
balajismaniam: I would prefer changing `AddLabels` -> `AddorUpdateLabels` and moving the remove labels part to another function as Connor mentioned.

IRCody: If you move the part that removes labels I don't think you would want to rename `AddLabels` as it's behavior hasn't changed.

Does it make sense to split this into another function? We'd be changing the set of calls to update labels to:
```go
helper.RemoveNFDLabels(node)
helper.AddLabels(node, labels)
```
instead of just doing:
```go
helper.UpdateNodeLabels(node, labels)
```

You can definitely do it that way but I'm not sure what value you gain by splitting it up since the behavior will always be done together.

ConnorDoyle:
Testability seems better if they are split, so it gets my vote.

main.go
balajismaniam: Yes. I think it is better if we split it into another function. I would prefer calling it `RemoveLabels` similar to `AddLabels`.

main.go
```diff
@@ -232,6 +232,15 @@ func (h k8sHelpers) GetNode(cli *client.Client) (*api.Node, error) {
...
+	for k := range n.Labels {
+		// If the label is not in the labels we are adding
+		// and it is an nfd label, delete it.
+		if _, ok := labels[k]; !ok && strings.Contains(k, Namespace) {
```
balajismaniam: s/Namespace/prefix/. See https://github.com/kubernetes-incubator/node-feature-discovery/pull/28#discussion_r85028520 for context.

main_test.go
```diff
@@ -100,3 +100,47 @@ func TestDiscoveryWithMockSources(t *testing.T) {
...
+		Convey("If no labels are passed", func() {
+			helper.AddLabels(n, labels)
+
+			Convey("None should be added", func() {
+				So(len(n.Labels), ShouldEqual, 0)
+			})
+		})
+
+		Convey("They should be added to the node.Labels", func() {
+			test1 := Namespace + ".test1"
```
balajismaniam: Change `test1 := Namespace + ".test1"` to `test1 := prefix + "-test1"`.

main_test.go
```diff
@@ -100,3 +100,47 @@ func TestDiscoveryWithMockSources(t *testing.T) {
 
 	})
 }
+
+func TestAddLabels(t *testing.T) {
+
```
balajismaniam: Nit: Delete this line.

main_test.go
```diff
@@ -100,3 +100,47 @@ func TestDiscoveryWithMockSources(t *testing.T) {
 
 	})
 }
+
+func TestAddLabels(t *testing.T) {
```
balajismaniam: This should change appropriately as we are going to split remove and add into two functions.

main_test.go
```diff
@@ -100,3 +100,47 @@ func TestDiscoveryWithMockSources(t *testing.T) {
...
+
+			Convey("And they should be overwritten by subsequent calls", func() {
+				labels[test1] = "false"
+				helper.AddLabels(n, labels)
+				So(n.Labels, ShouldContainKey, test1)
+				So(n.Labels[test1], ShouldEqual, "false")
+			})
+
+			Convey("And they should be removed if not present on subsequent calls", func() {
+				test2 := Namespace + ".test2"
```
balajismaniam: Change `test1 := Namespace + ".test2"` to `test1 := prefix + "-test2"`.

main_test.go
```diff
@@ -100,3 +100,47 @@ func TestDiscoveryWithMockSources(t *testing.T) {
...
+		})
+
+		Convey("They should be added to the node.Labels", func() {
+			test1 := Namespace + ".test1"
+			labels[test1] = "true"
+			helper.AddLabels(n, labels)
+			So(n.Labels, ShouldContainKey, test1)
+
+			Convey("And they should be overwritten by subsequent calls", func() {
+				labels[test1] = "false"
```
balajismaniam: Right now we don't support key values other than `true` or updates to already exisiting labels (other than removal after this PR gets merged).  Can we delete this test case?

main.go
```diff
@@ -232,6 +232,15 @@ func (h k8sHelpers) GetNode(cli *client.Client) (*api.Node, error) {
 }
 
 func (h k8sHelpers) AddLabels(n *api.Node, labels Labels) {
+	// Remove nfd node labels that no longer apply
+	for k := range n.Labels {
+		// If the label is not in the labels we are adding
+		// and it is an nfd label, delete it.
```
balajismaniam: Nit: Consider changing comment to `If the label was not discovered in the current run of node-feature-discovery and if it has the prefix, delete it from the list of node labels` or something similar.

balajismaniam:
@IRCody As per our offline discussion, this PR will remove the node labels if the `--label-whitelist` changes between subsequent runs of the PR. I think we should update the readme to reflect this. We can do this in a separate PR if you prefer.

main.go
```diff
@@ -232,6 +232,15 @@ func (h k8sHelpers) GetNode(cli *client.Client) (*api.Node, error) {
...
+		if _, ok := labels[k]; !ok && strings.Contains(k, Namespace) {
+			delete(n.Labels, k)
```
balajismaniam: Can we log the labels we are deleting here?

main.go
balajismaniam: Nit on the comment: We don't use `nfd` anywhere in the repo right now. Consider deleting `nfd` from the comment.

IRCody:
@balajismaniam: Updated this addressing the comments. A summary of what's changed:

* Split the Removal of labels into it's own function, `RemoveLabels` that takes in a the api.Node and a search string and removes all keys that contain the search string. This is called from the advertiseFeatureLabels() func with the `prefix` as you pointed out in your comments.
* Modified the test for AddLabels and added some for RemoveLabels
* Added a note in the README about the behavior around label removal and it's interaction with whitelist (or potentially any other label-modifying action that is added in the future).

main_test.go
```diff
@@ -100,3 +102,66 @@ func TestDiscoveryWithMockSources(t *testing.T) {
...
+			ObjectMeta: api.ObjectMeta{
+				Labels: map[string]string{
+					"single":     "123",
+					"multiple_A": "a",
+					"multiple_B": "b",
+				},
+			},
+		}
+
+		Convey("a unique label should be remove", func() {
```
balajismaniam: s/remove/removed

README.md
```diff
@@ -80,6 +80,11 @@ the only label value published for features is the string `"true"`._
 
 The `--sources` flag controls which sources to use for discovery.
 
+_Note: consecutive runs of node-feature-discovery will update the labels on a
+given node. If labels are not discovered on a consecutive run they will be removed
```
balajismaniam: s/If labels are not discovered on a consecutive run they will be removed/If features are not discovered on a consecutive run, the corresponding label will be removed

README.md
```diff
@@ -80,6 +80,11 @@ the only label value published for features is the string `"true"`._
...
+from the node. This includes any restrictions placed on the consecutive run, such
+as restricting discovered nodes with the --label-whitelist option._
```
balajismaniam: s/discovered nodes/discovered features/

balajismaniam:
@IRCody Left some minor comments. Otherwise, LGTM.

IRCody:
@balajismaniam: Made those updates.

balajismaniam:
@IRCody Thanks for the updates.

PR #43, opened 2016-11-12
[WIP] Use out-of-tree k8s go client.

### LINKED ISSUE #38: Consider using client-go instead of the unversioned client.
Right now NFD uses the `unversioned` client to talk to the K8s cluster. We should consider using the `client-go` (https://github.com/kubernetes/client-go) instead.

ConnorDoyle:
Fixes #38.

Builds and tests pass, but still needs manual testing (since we don't have e2e set up yet.)

This also speeds up the build quite a bit.


ConnorDoyle:
Manual testing in our bare metal dev cluster worked. Think this is OK to merge after your LGTM @balajismaniam

main.go
```diff
@@ -8,8 +8,9 @@ import (
 	"strings"
 
 	"github.com/docopt/docopt-go"
-	"k8s.io/kubernetes/pkg/api"
-	client "k8s.io/kubernetes/pkg/client/unversioned"
+	"k8s.io/client-go/kubernetes"
```
balajismaniam: Can we alias this as `client` or `k8sclient`?

main.go
```diff
@@ -8,8 +8,9 @@ import (
...
+	api "k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/rest"
```
balajismaniam: Alias this as `restclient`?

main.go
ConnorDoyle: Will update.

main.go
ConnorDoyle: Will update.

ConnorDoyle: updated

PR #49, opened 2016-11-25
Adding SR-IOV capability discovery to node-feature-discovery

swatisehgal:
PR in reference to https://github.com/kubernetes-incubator/node-feature-discovery/issues/35

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ The current set of feature sources are the following:
 - [CPUID][cpuid] for x86 CPU details
 - [Intel Resource Director Technology][intel-rdt]
 - [Intel P-State driver][intel-pstate]
-- [Network][network]
+- Network
 
 ### Feature labels
 
@@ -116,7 +116,7 @@ such as restricting discovered features with the --label-whitelist option._
 
 | Feature name   | Description                                                                         |
 | :------------: | :---------------------------------------------------------------------------------: |
-| SRIOV          | Single Root Input/Output Virtualization (SR-IOV) enabled Network Interface Card
+| [SRIOV][sriov] | Single Root Input/Output Virtualization (SR-IOV) enabled Network Interface Card
 
 ## Getting started
 ### System requirements
@@ -137,7 +137,7 @@ repo that demonstrates how to deploy the job to unlabeled nodes.
 
 The discovery script will launch a job on each unlabeled node in the
 cluster. When the job runs, it contacts the Kubernetes API server to add labels
-to the node to advertise hardware features (initially, from `cpuid`, RDT ,p-state and network).
+to the node to advertise hardware features (initially, from `cpuid`, RDT, p-state and network).
 
 [![asciicast](https://asciinema.org/a/11wir751y89617oemwnsgli4a.png)](https://asciinema.org/a/11wir751y89617oemwnsgli4a)
 
@@ -234,7 +234,7 @@ A demo on the benefits of using node feature discovery can be found in [demo](de
 [cpuid]: http://man7.org/linux/man-pages/man4/cpuid.4.html
 [intel-rdt]: http://www.intel.com/content/www/us/en/architecture-and-technology/resource-director-technology.html
 [intel-pstate]: https://www.kernel.org/doc/Documentation/cpu-freq/intel-pstate.txt
-[network]: http://www.intel.com/content/www/us/en/pci-express/pci-sig-sr-iov-primer-sr-iov-technology-paper.html
+[sriov]: http://www.intel.com/content/www/us/en/pci-express/pci-sig-sr-iov-primer-sr-iov-technology-paper.html
 [docker-down]: https://docs.docker.com/engine/installation
 [golang-down]: https://golang.org/dl
 [gcc-down]: https://gcc.gnu.org
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ node-feature-discovery.
   -h --help                   Show this screen.
   --version                   Output version and exit.
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,rdt,pstate,netid]
+                              [Default: cpuid,rdt,pstate,network]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
@@ -57,7 +57,7 @@ The current set of feature sources are the following:
 - [CPUID][cpuid] for x86 CPU details
 - [Intel Resource Director Technology][intel-rdt]
 - [Intel P-State driver][intel-pstate]
-- [PCI SR-IOV NIC][pci-sriov-nic]
+- [Network][network]
 
 ### Feature labels
 
@@ -79,7 +79,7 @@ the only label value published for features is the string `"true"`._
   "node.alpha.kubernetes-incubator.io/nfd-cpuid-<feature-name>": "true",
   "node.alpha.kubernetes-incubator.io/nfd-rdt-<feature-name>": "true",
   "node.alpha.kubernetes-incubator.io/nfd-pstate-<feature-name>": "true",
-  "node.alpha.kubernetes-incubator.io/nfd-netid-<feature-name>": "true"
+  "node.alpha.kubernetes-incubator.io/nfd-network-<feature-name>": "true"
 }
 ```
 
@@ -112,12 +112,11 @@ such as restricting discovered features with the --label-whitelist option._
 | SSE4.2         | Streaming SIMD Extensions 4.2 (SSE4.2)
 | SGX            | Software Guard Extensions (SGX)
 
-
 ### Network Features
 
 | Feature name   | Description                                                                         |
 | :------------: | :---------------------------------------------------------------------------------: |
-| SRIOV          | PCI Single Root Input/Output Virtualization (SR-IOV) Network Interface Card
+| SRIOV          | Single Root Input/Output Virtualization (SR-IOV) enabled Network Interface Card
 
 ## Getting started
 ### System requirements
@@ -138,7 +137,7 @@ repo that demonstrates how to deploy the job to unlabeled nodes.
 
 The discovery script will launch a job on each unlabeled node in the
 cluster. When the job runs, it contacts the Kubernetes API server to add labels
-to the node to advertise hardware features (initially, from `cpuid`, RDT and p-state).
+to the node to advertise hardware features (initially, from `cpuid`, RDT ,p-state and network).
 
 [![asciicast](https://asciinema.org/a/11wir751y89617oemwnsgli4a.png)](https://asciinema.org/a/11wir751y89617oemwnsgli4a)
 
@@ -235,7 +234,7 @@ A demo on the benefits of using node feature discovery can be found in [demo](de
 [cpuid]: http://man7.org/linux/man-pages/man4/cpuid.4.html
 [intel-rdt]: http://www.intel.com/content/www/us/en/architecture-and-technology/resource-director-technology.html
 [intel-pstate]: https://www.kernel.org/doc/Documentation/cpu-freq/intel-pstate.txt
-[pci-sriov-nic]: http://www.intel.com/content/www/us/en/pci-express/pci-sig-sr-iov-primer-sr-iov-technology-paper.html
+[network]: http://www.intel.com/content/www/us/en/pci-express/pci-sig-sr-iov-primer-sr-iov-technology-paper.html
 [docker-down]: https://docs.docker.com/engine/installation
 [golang-down]: https://golang.org/dl
 [gcc-down]: https://gcc.gnu.org
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ node-feature-discovery.
   -h --help                   Show this screen.
   --version                   Output version and exit.
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,rdt,pstate]
+                              [Default: cpuid,rdt,pstate,netid]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
@@ -57,6 +57,7 @@ The current set of feature sources are the following:
 - [CPUID][cpuid] for x86 CPU details
 - [Intel Resource Director Technology][intel-rdt]
 - [Intel P-State driver][intel-pstate]
+- [PCI SR-IOV NIC][pci-sriov-nic]
 
 ### Feature labels
 
@@ -77,7 +78,8 @@ the only label value published for features is the string `"true"`._
   "node.alpha.kubernetes-incubator.io/node-feature-discovery.version": "v0.1.0",
   "node.alpha.kubernetes-incubator.io/nfd-cpuid-<feature-name>": "true",
   "node.alpha.kubernetes-incubator.io/nfd-rdt-<feature-name>": "true",
-  "node.alpha.kubernetes-incubator.io/nfd-pstate-<feature-name>": "true"
+  "node.alpha.kubernetes-incubator.io/nfd-pstate-<feature-name>": "true",
+  "node.alpha.kubernetes-incubator.io/nfd-netid-<feature-name>": "true"
 }
 ```
 
@@ -109,6 +111,13 @@ such as restricting discovered features with the --label-whitelist option._
 | SSE4.1         | Streaming SIMD Extensions 4.1 (SSE4.1)
 | SSE4.2         | Streaming SIMD Extensions 4.2 (SSE4.2)
 | SGX            | Software Guard Extensions (SGX)
+
+
+### Network Features
+
+| Feature name   | Description                                                                         |
+| :------------: | :---------------------------------------------------------------------------------: |
+| SRIOV          | PCI Single Root Input/Output Virtualization (SR-IOV) Network Interface Card
 
 ## Getting started
 ### System requirements
@@ -226,6 +235,7 @@ A demo on the benefits of using node feature discovery can be found in [demo](de
 [cpuid]: http://man7.org/linux/man-pages/man4/cpuid.4.html
 [intel-rdt]: http://www.intel.com/content/www/us/en/architecture-and-technology/resource-director-technology.html
 [intel-pstate]: https://www.kernel.org/doc/Documentation/cpu-freq/intel-pstate.txt
+[pci-sriov-nic]: http://www.intel.com/content/www/us/en/pci-express/pci-sig-sr-iov-primer-sr-iov-technology-paper.html
 [docker-down]: https://docs.docker.com/engine/installation
 [golang-down]: https://golang.org/dl
 [gcc-down]: https://gcc.gnu.org
--- a/node-feature-discovery-job.json.template
+++ b/node-feature-discovery-job.json.template
@@ -17,6 +17,14 @@
         }
       },
       "spec": {
+        "volumes": [
+          {
+            "name": "nfd-node-volume",
+            "hostPath": {
+              "path": "/sys"
+            }
+          }
+        ],
         "containers": [
           {
             "env": [
@@ -39,6 +47,13 @@
             ],
             "image": "quay.io/kubernetes_incubator/node-feature-discovery:v0.1.0",
             "name": "node-feature-discovery",
+            "volumeMounts": [
+              {
+                "mountPath": "/hostsys",
+                "name": "nfd-node-volume",
+		"readOnly": true
+              }
+            ],
             "ports": [
               {
                 "containerPort": 2233,
@@ -52,3 +67,4 @@
     }
   }
 }
+
--- a/node-feature-discovery-job.json.template
+++ b/node-feature-discovery-job.json.template
@@ -17,14 +17,7 @@
         }
       },
       "spec": {
-        "volumes": [
-          {
-            "name": "nfd-node-volume",
-            "hostPath": {
-              "path": "/sys"
-            }
-          }
-        ],
+        "hostNetwork": true,
         "containers": [
           {
             "env": [
@@ -47,16 +40,9 @@
             ],
             "image": "quay.io/kubernetes_incubator/node-feature-discovery:v0.1.0",
             "name": "node-feature-discovery",
-            "volumeMounts": [
-              {
-                "mountPath": "/hostsys",
-                "name": "nfd-node-volume",
-		"readOnly": true
-              }
-            ],
             "ports": [
               {
-                "containerPort": 2233,
+                "containerPort": 7156,
                 "hostPort": 7156
               }
             ]
@@ -67,4 +53,3 @@
     }
   }
 }
-
--- a/node-feature-discovery-job.json.template
+++ b/node-feature-discovery-job.json.template
@@ -17,7 +17,7 @@
         }
       },
       "spec": {
-        "hostNetwork": true,
+	"hostNetwork": true,
         "containers": [
           {
             "env": [
--- a/main.go
+++ b/main.go
@@ -10,6 +10,7 @@ import (
 	"github.com/kubernetes-incubator/node-feature-discovery/source"
 	"github.com/kubernetes-incubator/node-feature-discovery/source/cpuid"
 	"github.com/kubernetes-incubator/node-feature-discovery/source/fake"
+	"github.com/kubernetes-incubator/node-feature-discovery/source/network"
 	"github.com/kubernetes-incubator/node-feature-discovery/source/panic_fake"
 	"github.com/kubernetes-incubator/node-feature-discovery/source/pstate"
 	"github.com/kubernetes-incubator/node-feature-discovery/source/rdt"
@@ -145,6 +146,7 @@ func configureParameters(sourcesArg []string, whiteListArg string) (sources []so
 		cpuid.Source{},
 		rdt.Source{},
 		pstate.Source{},
+		network.Source{},
 		fake.Source{},
 		panic_fake.Source{},
 	}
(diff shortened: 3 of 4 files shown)


swatisehgal:
I signed it!

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+
+// Implements main.FeatureSource.
+type networkSource struct{}
+
+func (s networkSource) Name() string { return "netid" }
+func (s networkSource) Discover() ([]string, error) {
+	features := []string{}
+	var total_num_vfs int = 0
+	// sysfs of the node mounted as /hostsys inside the pod
+	netInterfaces, err := ioutil.ReadDir("/hostsys/class/net") //does not return . and ..
```
balajismaniam: Nit: delete the comment at the end.

balajismaniam: We may need to change the sysfs path if we decide not to mount the host sysfs. (See previous comment on voulumemounts for details.)

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("can't obtain the network interfaces: %s", err.Error())
```
balajismaniam: Nit: Change error to `can't obtain network interface details`.

node-feature-discovery-job.json.template
```diff
@@ -17,6 +17,14 @@
         }
       },
       "spec": {
+        "volumes": [
```
balajismaniam: cc @ConnorDoyle.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+	}
+	for _, netInterface := range netInterfaces {
+		//forming the file name : /hostsys/class/net/<network interface>/device/sriov_numvfs
```
balajismaniam: We can delete this comment.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+		filename := "/hostsys/class/net/" + netInterface.Name() + "/device/sriov_numvfs"
+		bytes_received, err := ioutil.ReadFile(filename)
+		if err != nil {
+			// interface does not support SRIOV
+			continue // proceeding to next iteration for next network interface
+		}
+		//removing white spaces
```
balajismaniam: Delete this comment.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+		num := bytes.TrimSpace(bytes_received)
+		//checking if 0 bytes read
```
balajismaniam: Delete this comment.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+		if l := len(bytes_received); l == 0 {
+			// Zero bytes read for this file
+			continue // proceeding to next iteration for next network interface
+		}
+		//converting bytes to integer and checking for error
```
balajismaniam: Delete this comment.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+// Network Source
+
+// Implements main.FeatureSource.
+type networkSource struct{}
+
+func (s networkSource) Name() string { return "netid" }
+func (s networkSource) Discover() ([]string, error) {
+	features := []string{}
+	var total_num_vfs int = 0
+	// sysfs of the node mounted as /hostsys inside the pod
```
balajismaniam: Can we add a single comment which explains what you are doing? Example: `read the network card details from sysfs and determining if SR-IOV is enabled for each of the interfaces.`

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+	netInterfaces, err := ioutil.ReadDir("/hostsys/class/net") //does not return . and ..
+	if err != nil {
+		return nil, fmt.Errorf("can't obtain the network interfaces: %s", err.Error())
+	}
+	for _, netInterface := range netInterfaces {
```
balajismaniam: May be add comment above to explain what you are doing in this loop.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+		//forming the file name : /hostsys/class/net/<network interface>/device/sriov_numvfs
+		filename := "/hostsys/class/net/" + netInterface.Name() + "/device/sriov_numvfs"
+		bytes_received, err := ioutil.ReadFile(filename)
+		if err != nil {
+			// interface does not support SRIOV
+			continue // proceeding to next iteration for next network interface
```
balajismaniam: Delete the comment at the end.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+	netInterfaces, err := ioutil.ReadDir("/hostsys/class/net") //does not return . and ..
+	if err != nil {
+		return nil, fmt.Errorf("can't obtain the network interfaces: %s", err.Error())
+	}
+	for _, netInterface := range netInterfaces {
+		//forming the file name : /hostsys/class/net/<network interface>/device/sriov_numvfs
+		filename := "/hostsys/class/net/" + netInterface.Name() + "/device/sriov_numvfs"
+		bytes_received, err := ioutil.ReadFile(filename)
+		if err != nil {
+			// interface does not support SRIOV
```
balajismaniam: Delete the comment. But log the error using `fmt.Errorf("SR-IOV not supported by the interface: %s", netInterface)`.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+		}
+		//converting bytes to integer and checking for error
+		n, err := strconv.Atoi(string(num))
+		if err != nil {
+			// Receiving non-integral value
+			continue // proceeding to next iteration for next network interface
+		}
+		total_num_vfs += n
+	}
+	if total_num_vfs > 0 {
```
balajismaniam: We can delete this `if` block if we decide not to count the VFs.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+		filename := "/hostsys/class/net/" + netInterface.Name() + "/device/sriov_numvfs"
+		bytes_received, err := ioutil.ReadFile(filename)
+		if err != nil {
+			// interface does not support SRIOV
+			continue // proceeding to next iteration for next network interface
+		}
+		//removing white spaces
+		num := bytes.TrimSpace(bytes_received)
+		//checking if 0 bytes read
+		if l := len(bytes_received); l == 0 {
```
balajismaniam: Delete this `if` block.

sources.go
```diff
@@ -102,3 +103,47 @@ func (s pstateSource) Discover() ([]string, error) {
...
+			// Zero bytes read for this file
+			continue // proceeding to next iteration for next network interface
+		}
+		//converting bytes to integer and checking for error
+		n, err := strconv.Atoi(string(num))
```
balajismaniam: Can we do this here instead?

```go
n, err := strconv.Atoi(string(num)) 
if err != nil && n > 0 {
 features = append(features, "SRIOV")
 break
}
```

This way we need not check the other interfaces and also avoid counting the number of VFs.

sources.go
balajismaniam: Can we loop over the interfaces that are `up`? By doing this, we only label nodes where SR-IOV is enabled for an interface that is `up`. WDYT?

CC @ConnorDoyle

balajismaniam:
@swatisehgal I left some comments. PTAL.

balajismaniam:
I just realized you made this PR from `master`. It is better to submit it from a feature branch. See http://blog.scottlowe.org/2015/01/27/using-fork-branch-git-workflow/ and this will help you fix it: https://coding.abel.nu/2015/03/fixing-a-pull-request-from-master.

sources.go
thed00de: +1

sources.go
thed00de: I think this looking for the network interfaces at this stage.

README.md
```diff
@@ -38,7 +38,7 @@ node-feature-discovery.
   -h --help                   Show this screen.
   --version                   Output version and exit.
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,rdt,pstate]
+                              [Default: cpuid,rdt,pstate,netid]
```
thed00de: +1

sources.go
```diff
@@ -1,12 +1,13 @@
...
 import (
+	"bytes"
 	"fmt"
+	"github.com/klauspost/cpuid"
 	"io/ioutil"
 	"os/exec"
 	"path"
-
-	"github.com/klauspost/cpuid"
+	"strconv"
```
swatisehgal: Thanks for letting me know. Will take care of this in the future.

sources.go
swatisehgal: Will look into this and see how this can be implemented.

README.md
```diff
@@ -107,6 +109,13 @@ such as restricting discovered features with the --label-whitelist option._
 | SSE4.2         | Streaming SIMD Extensions 4.2 (SSE4.2)
 | SGX            | Software Guard Extensions (SGX)
 
+
+### Network Features
+
+| Feature name   | Description                                                                         |
+| :------------: | :---------------------------------------------------------------------------------: |
+| SRIOV          | PCI Single Root Input/Output Virtualization (SR-IOV) Network Interface Card
```
swatisehgal: yes that sounds better! Thanks

node-feature-discovery-job.json.template
swatisehgal: also the mnt namespace gives the container its own view of the “mounts” on the system resulting in a different views of the filesystem hierarchy

sources.go
swatisehgal: I see you are using  stderrLogger in main.go but in the above comment you are suggesting to use fmt.Errorf(). Can you please let me know your preference.

node-feature-discovery-job.json.template
ConnorDoyle: Supplying `--net=host` should be sufficient. No need for additional bind-mounts or `--privileged`. See below.

```
root@d1k8s:~/sriov-demo/docker-iperf3-dpdk# ls /sys/class/net/
docker0  eth0  eth1  flannel.1  lo  veth8b16ae5  vethb3234d4
root@d1k8s:~/sriov-demo/docker-iperf3-dpdk# docker run -t debian bash -c "ls /sys/class/net"
eth0  lo
root@d1k8s:~/sriov-demo/docker-iperf3-dpdk# docker run -t --net=host debian bash -c "ls /sys/class/net"               
docker0  eth0  eth1  flannel.1  lo  veth8b16ae5  vethb3234d4
```

sources.go
ConnorDoyle: Will need to be formatted before merge.

sources.go
ConnorDoyle: Please use the provided stderrLogger instance. The idea is to output discovered features (only) to stdout to facilitate processing by other programs.

balajismaniam: Yes. Use the `stderrLogger`.

sources.go
balajismaniam: Ok. But I would like the error message to be specific to be helpful for debugging. May be something more specific like `can't obtain network interface details from sysfs` might be better?

ConnorDoyle:
@swatisehgal for the CNCF bot to authorize you, I think you need to add your Intel address as your public address in your GitHub profile.

node-feature-discovery-job.json.template
thed00de: @ConnorDoyle Agreed, but how do you do that via k8s?
We do not want to impact how the networking is setup for the pod, we just want to discover what node networking capabilities exist.
Seems to me that going forward, as we wish to discover more capabilities of the host, we are going to run into this type of concern again and again.
If the pod is completely sandboxed then it will be very limited in what it can discover.
To support it being completely open (i.e. privileged) then that opens other doors...maybe this is ok?
Compromise is required, but what is acceptable?
The kernel api via libnl might(?) be an alternative in this case, but we cannot always depend on such a solution being available.

ConnorDoyle:
I think we will need to clearly document any system or pod configuration requirements on a per-feature-source basis.

HostNetwork: true goes into the pod spec (see https://github.com/kubernetes/kubernetes/pull/5886/files). Users would also have to modify the allowable pod sources for host net on the Kubelet.

I agree feature sources should avoid requiring privileged mode.

swatisehgal:
@ConnorDoyle @thed00de I tested with HostNetwork=true in the file node-feature-discovery-job.json.template  and seems to work fine. I ensured to set --allow-privileged=false in kubernetes config file and in the kubelet configuration.  But, when hostNetwork is set to be true, I had to ensure that the hostPort is same as the containerPort.

node-feature-discovery-job.json.template
ConnorDoyle: @thed00de responded via email but it didn't get linked to this thread (see comments below)

ConnorDoyle: https://github.com/kubernetes-incubator/node-feature-discovery/pull/49#issuecomment-263903853

ConnorDoyle:
> have to ensure that hostPort is same as container port

What are the semantics of that in host networking mode? Not sure why this is required.

> Is there a reason we are specifying containerPort=2233 and hostPort=7156?

I don't think it's necessary; the kubernetes client should just acquire an ephemeral port for connecting to the API server. (@balajismaniam please correct this if I misunderstood why it's there)

balajismaniam:
@swatisehgal The `containerPort` and the `hostPort` can be the same. 

The NFD job deploys pods (= number of nodes) in parallel (https://github.com/kubernetes-incubator/node-feature-discovery/blob/master/node-feature-discovery-job.json.template#L12). Specifying the `hostPort` is a workaround to ensure only one pod lands in a node (i.e., when a pod lands on a node it consumes a hostPort and other pods can't land on the same node). It is required. Otherwise, multiple pods can land on a node.

Side-note: We don't deploy NFD as a DaemonSet (which ensures only one pod lands on each node) since the `restartPolicy` cannot be set to `Never`. This should resolved after we make a decision and implement https://github.com/kubernetes-incubator/node-feature-discovery/issues/34.

ConnorDoyle:
@balajismaniam yep, now I remember. We need to find a more robust solution for it, but that's another issue.

swatisehgal:
> for the CNCF bot to authorize you, I think you need to add your Intel address as your public address in your GitHub profile.

I have my github profile setup with my intel email added as my public email. The problem still persists.

ConnorDoyle:
The address isn't visible on your profile, you have to explicitly make it public in the account settings. Contrast your profile (@swatisehgal) with either @ConnorDoyle or @balajismaniam. The email should show up in the left-hand sidebar.

balajismaniam:
@ConnorDoyle and I were discussing about using `sriov_totalvfs` to determine if sriov is supported and use some other label for advertising if `sriov_numvfs > 0`. We ended-up with the idea of using two labels (`sriov` and `sriov-confgured`) to advertise sriov capabilities. If if `sriov_totalvfs > 0`, then set label`sriov = true` and if `sriov_numvfs > 0`, then `sriov-numvfs = true`. WDYT?

sources.go
```diff
@@ -102,3 +111,44 @@ func (s pstateSource) Discover() ([]string, error) {
...
+func (s networkSource) Discover() ([]string, error) {
+	features := []string{}
+	netInterfaces, err := net.Interfaces()
+
+	if err != nil {
+		return nil, fmt.Errorf("can't obtain the network interfaces from sysfs: %s", err.Error())
+	}
+	// iterating through network interfaces to obtain their respective number of virtual functions
+	for _, netInterface := range netInterfaces {
+		if strings.Contains(netInterface.Flags.String(), "up") {
```
balajismaniam: Can we also ignore loopback devices?

sources.go
```diff
@@ -102,3 +111,44 @@ func (s pstateSource) Discover() ([]string, error) {
...
+			}
+			num := bytes.TrimSpace(bytes_received)
+			n, err := strconv.Atoi(string(num))
+			if err != nil {
+				stderrLogger.Printf("Error in obtaining the number of virtual functions for network interface: %s", netInterface.Name)
+				continue
+			}
+			if n > 0 {
+				stdoutLogger.Printf("%d virtual functions configured on network interface: %s", n, netInterface.Name)
+				features = append(features, "SRIOV")
```
balajismaniam: `SRIOV` -> `sriov`.

README.md
```diff
@@ -54,6 +54,7 @@ The current set of feature sources are the following:
 - [CPUID][cpuid] for x86 CPU details
 - [Intel Resource Director Technology][intel-rdt]
 - [Intel P-State driver][intel-pstate]
+- [Network][network]
```
swatisehgal: I tried to search for a generic network link but couldn't come with a suitable one. I will remove this link and add the SRIOV link in features.

README.md
```diff
@@ -126,7 +134,7 @@ repo that demonstrates how to deploy the job to unlabeled nodes.
 
 The discovery script will launch a job on each each unlabeled node in the
 cluster. When the job runs, it contacts the Kubernetes API server to add labels
-to the node to advertise hardware features (initially, from `cpuid`, RDT and p-state).
+to the node to advertise hardware features (initially, from `cpuid`, RDT ,p-state and network).
```
swatisehgal: ah... Thanks for this. Will update.

sources.go
```diff
@@ -102,3 +111,44 @@ func (s pstateSource) Discover() ([]string, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("can't obtain the network interfaces from sysfs: %s", err.Error())
+	}
+	// iterating through network interfaces to obtain their respective number of virtual functions
+	for _, netInterface := range netInterfaces {
+		if strings.Contains(netInterface.Flags.String(), "up") {
+			netInterfaceVfPath := pathNet + netInterface.Name + deviceSriovNumvfs
+			bytes_received, err := ioutil.ReadFile(netInterfaceVfPath)
+			if err != nil {
+				stderrLogger.Printf("SR-IOV not supported for network interface: %s", netInterface.Name)
```
swatisehgal: Will log error too.

sources.go
swatisehgal: sure we can. Will update. Thanks for the suggestion.

swatisehgal:
> using sriov_totalvfs to determine if sriov is supported and use some other label for advertising if sriov_numvfs > 0. We ended-up with the idea of using two labels (sriov and sriov-confgured) to advertise sriov capabilities. If sriov_totalvfs > 0, then set labelsriov = true and if sriov_numvfs > 0, then sriov-configured = true. WDYT?

Yes, that's a good idea. This would provide the ability to determine nodes which are capable of supporting SR-IOV even if virtual functions are not configured allowing them to be distinguished from nodes that don't have SR-IOV capable NICs at all. But, thinking alternatively, its like assigning a label to node to indicate that it "can" support SR-IOV but "doesn't" do that. Would that information be beneficial as the virtual functions are not configured? The ultimate objective of assigning labels to the nodes is to assign  virtual functions of nodes to pods. A node that supports SR-IOV and have total_numvfs > 0 but not configured with any virtual functions (sriov = true but no label for sriov-configured = true ) would probably never be used unless the information is conveyed back to the user informing them that they **"have"** SR-IOV capable NICs in the system and are not using that capability. If we create another label for this, I don't completely understand what use case this might serve or is it just for bookeeping?  Moving forward to OIR, we would be using information from sriov_numvfs for different interfaces as the capacity and would track allocatable. Don't see how total_numvfs would be used here either. What do you guys think? Am I missing something that you have in mind?

cc @thed00de Any comments?

balajismaniam:
@swatisehgal Like you mention, the two labels will distinguish between SR-IOV support and configuration and allow us to advertise those separately. The operator can decide to configure the nodes with VFs (i.e., set `sriov_numvfs > 0`). 

> A node that supports SR-IOV and have total_numvfs > 0 but not configured with any virtual functions (sriov = true but no label for sriov-configured = true ) would probably never be used unless the information is conveyed back to the user informing them that they "have" SR-IOV capable NICs in the system and are not using that capability. 

When the label `sriov` is set but the label `sriov-configured` is not set, it conveys the above. 

Finally if the user desires, the labels can be filtered out using the `--label-whitelist` flag.

sources.go
balajismaniam: It seems like the import ordering is messed-up here. Can you please order them? See https://github.com/golang/go/wiki/CodeReviewComments#imports for details.

sources.go
```diff
@@ -102,3 +105,61 @@ func (s pstateSource) Discover() ([]string, error) {
...
+				continue
+			}
+			total := bytes.TrimSpace(totalBytes)
+			t, err := strconv.Atoi(string(total))
+			if err != nil {
+				stderrLogger.Printf("Error in obtaining total number of virtual functions for network interface: %s: %s", netInterface.Name, err.Error())
+				continue
+			}
+			if t > 0 {
+				stdoutLogger.Printf("%d total number of virtual functions on network interface: %s", t, netInterface.Name)
```
balajismaniam: Can you also log something like `"SR-IOV capability is detected on the network interface: %s, netInterface.Name"`.

sources.go
```diff
@@ -102,3 +105,61 @@ func (s pstateSource) Discover() ([]string, error) {
...
+				features = append(features, "sriov")
+				numVfsPath := "/sys/class/net/" + netInterface.Name + "/device/sriov_numvfs"
+				numBytes, err := ioutil.ReadFile(numVfsPath)
+				if err != nil {
+					stderrLogger.Printf("SR-IOV not supported for network interface: %s: %s", netInterface.Name, err.Error())
```
balajismaniam: s/not supported/not configured/

sources.go
```diff
@@ -102,3 +105,61 @@ func (s pstateSource) Discover() ([]string, error) {
...
+					continue
+				}
+				num := bytes.TrimSpace(numBytes)
+				n, err := strconv.Atoi(string(num))
+				if err != nil {
+					stderrLogger.Printf("Error in obtaining the number of virtual functions for network interface: %s: %s", netInterface.Name, err.Error())
```
balajismaniam: s/number/configured number/

sources.go
```diff
@@ -102,3 +105,61 @@ func (s pstateSource) Discover() ([]string, error) {
...
+					continue
+				}
+				if n > 0 {
+					stderrLogger.Printf("%d virtual functions configured on network interface: %s", n, netInterface.Name)
+					features = append(features, "sriov-os-configured")
```
balajismaniam: We decided on `sriov-configured` for the feature name las time. Is there any reason you prefer `sriov-os-configured`?

sources.go
balajismaniam: s/total number/maximum supported number/

sources.go
```diff
@@ -102,3 +105,61 @@ func (s pstateSource) Discover() ([]string, error) {
...
+			totalVfsPath := "/sys/class/net/" + netInterface.Name + "/device/sriov_totalvfs"
+			totalBytes, err := ioutil.ReadFile(totalVfsPath)
+			if err != nil {
+				stderrLogger.Printf("SR-IOV not supported for network interface: %s: %s", netInterface.Name, err.Error())
+				continue
+			}
+			total := bytes.TrimSpace(totalBytes)
+			t, err := strconv.Atoi(string(total))
+			if err != nil {
+				stderrLogger.Printf("Error in obtaining total number of virtual functions for network interface: %s: %s", netInterface.Name, err.Error())
```
balajismaniam: s/total number/maximum supported number/

sources.go
```diff
@@ -102,3 +105,61 @@ func (s pstateSource) Discover() ([]string, error) {
...
+				n, err := strconv.Atoi(string(num))
+				if err != nil {
+					stderrLogger.Printf("Error in obtaining the number of virtual functions for network interface: %s: %s", netInterface.Name, err.Error())
+					continue
+				}
+				if n > 0 {
+					stderrLogger.Printf("%d virtual functions configured on network interface: %s", n, netInterface.Name)
+					features = append(features, "sriov-os-configured")
+					break
+				} else {
```
balajismaniam: Question: Can `sriov_numvfs < 0`?

sources.go
ConnorDoyle: gofmt will FTFY (fix this for you)

sources.go
swatisehgal: We would at some stage need a label or a mechanism to check if sriov networking is configured in kubernetes. We might go for something like sriov-k8s-configured. Using sriov-configured might be ambiguous and misleading.

sources.go
swatisehgal: I saw [this](http://goinbigdata.com/goimports-vs-gofmt/) article that indicates goimports is a superset of gofmt, it groups and orders the imports correctly in addition to format correction. I'll use goimports if it seems okay to you both?

ConnorDoyle:
Yep, please do that for now. My editor does this automatically. We should really put this into the Makefile in the future to make it easier.

sources.go
swatisehgal: this is for the case where the file cannot be obtained i.e. numVfsPath does not exits for a specific interface and hence the interface does not support sr-iov. do you have something else in mind?

balajismaniam: I'm just trying to delineate between `support` and `configuration` in the logs. `sriov_totalvfs` determines if there is support and `sriov_numvfs` determines if sriov is configured. We should use `not configured` here since we are checking for `sriov_numvfs`.

sources.go
balajismaniam: For now, we can use `sriov-configured`. We can change it to something else when we add the feature for sriov networking support in kubernetes.

flyingcougar:
@swatisehgal LGTM - please just rebase

PR #71, opened 2017-08-25
change label namespace

andyxning:
Change nfd label to `nfd.alpha.kubernetes.io` according to [Kubernetes Community #300](https://github.com/kubernetes/community/issues/300) when in alpha phase.

How about we change nfd namespace according to nfd available phase like below:
* alpha 
  * `nfd.alpha.kubernetes.io`
* beta
  * `nfd.beta.kubernetes.io`
* GA(graduate from incubator) 
  * `nfd.kubernetes.io`

/cc @balajismaniam @ConnorDoyle @davidopp

### DIFF
--- a/main.go
+++ b/main.go
@@ -18,7 +18,7 @@ const (
 	ProgramName = "node-feature-discovery"
 
 	// Namespace is the prefix for all published labels.
-	Namespace = "node.alpha.kubernetes-incubator.io"
+	Namespace = "nfd.alpha.kubernetes.io"
 
 	// PodNameEnv is the environment variable that contains this pod's name.
 	PodNameEnv = "POD_NAME"


ConnorDoyle:
I can get to this pose v1.8

ConnorDoyle:
I'm not totally convinced that we need to evolve the alpha names, but the pattern for beta and GA proposed here look great. Could you start a tracking issue for graduation to beta that includes the beta naming scheme? Then we can get momentum moving towards graduating this repo.

cc @balajismaniam for another opinion here

andyxning:
@ConnorDoyle @balajismaniam @dchen1107 

The reason for using `nfd` as the subdomain name as `kubernetes.io` in alpha, beta and GA is mainly:
* more smaller and at the same time it is also clear that it represents `node-feature-discovery`. Smaller labels is more friendly to ETCD.
* the unification of alpha, beta and GA is mainly to keep the subdomain clearly targeted for `node-feature-discovery`. Maybe it is just a placeholder for alpha.

I also suggest we  add node-feature-discovery labels to the [well-known kubernetes labels](https://kubernetes.io/docs/reference/labels-annotations-taints/) when we make an agreement on this.

ConnorDoyle:
I guess my concern is adding unnecessary label churn beyond what's required by alpha => beta and beta => ga. On board with picking up the convention upon move to beta. I don't think it's worthwhile to rev the naming scheme before that.

andyxning:
> Could you start a tracking issue for graduation to beta that includes the beta naming scheme? Then we can get momentum moving towards graduating this repo.

Will fire an issue to track the jobs we need to do for a beta release later.

andyxning:
@ConnorDoyle I have added an issue #74 and list something i think we need to do before we reach beta release. PTAL.

balajismaniam:
I agree with @ConnorDoyle. We should not change label namespaces before graduating to beta. Moreover, moving to `kubernetes.io` namespace might require discussion with the community. I like the label conventions mentioned for beta and GA.

andyxning:
@balajismaniam Agree on `moving to kubernetes.io namespace might require discussion with the community.`. What about send an email to kubernetes-sig-node and cc sponsor @dchen1107 and champion @davidopp .

balajismaniam:
Closing this for now. Let's wait till beta to change this.

PR #72, opened 2017-08-27
refactor sources to source pkg

andyxning:
This PR refactor different sources to `sources` package in case we will add more sources in the future.

/cc @davidopp @dchen1107 @balajismaniam @ConnorDoyle

PR #73, opened 2017-08-27
Add net feature source

andyxning:
This PR adds net feature souce `net_speed_xxxxxx`. 

For 10000Mb/s, `net_speed_10000` label is set to `true`.
For 1000Mb/s, `net_speed_1000` label is set to `true`.

This PR is based on #72 and must be accepted after #72 .

/cc @davidopp @dchen1107 @balajismaniam @ConnorDoyle

### DIFF
--- a/source/net/net.go
+++ b/source/net/net.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package net
+
+import (
+	"net"
+
+	"fmt"
+
+	"io/ioutil"
+	"strings"
+
+	"strconv"
+
+	"github.com/golang/glog"
+)
+
+var physicalInterfacePrefixes = []string{"eth", "em"}
+
+// var instead of const just for unit test only.
+var interfaceSpeedSysFilePattern = "/sys/class/net/%s/speed"
+
+const (
+	speedFeaturePattern = "speed_%d"
+)
+
+// Source implements FeatureSource.
+type Source struct{}
+
+func (s Source) Name() string { return "net" }
+
+// Discover returns feature names for net related features such as network device bandwidth.
+func (s Source) Discover() ([]string, error) {
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		glog.Errorf("get available network interfaces error. %v", err)
+		return nil, fmt.Errorf("get available network interfaces error")
+	}
+
+	speedFeatures := s.getSpeedFeatures(interfaces)
+
+	return speedFeatures, nil
+}
+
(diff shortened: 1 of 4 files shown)


ConnorDoyle:
I can get to this post v1.8

ConnorDoyle:
Could you rebase on top of #72 after we merge it? That should fix the tests.

source/net/net.go
```diff
@@ -0,0 +1,78 @@
...
+func (s NetSource) Discover() ([]string, error) {
+	features := []string{}
+
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		glog.Errorf("get available network interfaces error. %v", err)
+		return nil, fmt.Errorf("get available network interfaces error")
+	}
+
+	for _, iFace := range interfaces {
```
ConnorDoyle: Moving this loop into a function that takes a slice of interfaces as input would allow adding unit test coverage. Could you make this change and add some test cases?

andyxning: Will do.

ConnorDoyle:
Needs rebase

andyxning:
@ConnorDoyle comments addressed. I have also add some unit test for net source. PTAL

ConnorDoyle:
Do you have docs to support the "em" prefix? Besides the old "eth" names it looks like we should consider supporting all of these: https://github.com/systemd/systemd/blob/master/src/udev/udev-builtin-net_id.c#L29

Stepping back a bit, what is the prefix list trying to prevent? You've already filtered lo and down interfaces via flags. IMO there's no harm in iterating other NICs like virtual functions.

Sorry for the late review on that part, I missed it first time through.

andyxning:
@ConnorDoyle FYI about `em`
* https://en.wikipedia.org/wiki/Consistent_Network_Device_Naming
* http://fedoraproject.org/wiki/Features/ConsistentNetworkDeviceNaming#User_Experience
* http://tdt.rocks/linux_network_interface_naming.html
* http://tdt.rocks/linux_network_interface_naming.html

andyxning:
> IMO there's no harm in iterating other NICs like virtual functions.

In the previous impl, i only consider the physical network interface which must be run through for an CNI network scheme. 

What is the reason to consider virtual function network interfaces as they seems have unlimited speed(correct me if i am wrong).

andyxning:
@ConnorDoyle It seems that there are three possible physical network interface naming schemes:
* original `eth`
* biosdevname `em`
* systemd udev `en`

andyxning:
@ConnorDoyle @balajismaniam PTAL.

ConnorDoyle:
> virtual function network interfaces as they seems have unlimited speed

I can only speak for VFs on Intel NICs which do report speed.

I would like to know the answer to the previous question I asked, which is: what specific problem does the prefix check solve?

andyxning:
> I would like to know the answer to the previous question I asked, which is: what specific problem does the prefix check solve?

@ConnorDoyle For now, i think we should only consider up and none localhost physical network interfaces speed and as far as i known physical network interface name may start with `eth` and `em`. The reason for why i suggest to only collect physical network interface speed is that all network flow must be sent and receive through physical NIC. It is the bottleneck. The common question asked by developer is that "does this machine 1000M-based ethernet or 10000M-based ethernet".

> virtual function network interfaces as they seems have unlimited speed
> I can only speak for VFs on Intel NICs which do report speed.

First, i am not sure if VFs stands for virtual network interface, like `docker0`. If then, actually, i ask one of my colleagues, he told me that virtual network interface actually has unlimited speed. IIRC, i have encountered an situation that virtual network interface has a speed stated to 1000M, but can actually exceeded that easily.

balajismaniam:
Closing due to inactivity.

Issue #74, opened 2017-09-06
[WIP] moving to beta release 


andyxning:
Before we make node-feature-discovery to a beta release, we need to do things below finished.

- [ ] move import path from `github.com/kubernetes-incubator` to `k8s.io`.
- [ ] make an agreement on the label schema about node-feature-discovery. #71
- [ ] adjust node-feature-discovery to meet the requirements to be a Kubernetes repo described in [kubernetes-template-project](https://github.com/kubernetes/kubernetes-template-project).

andyxning:
@marquiz Any docs or record video about the discussion.

marquiz:
Hmm, it seems that the last recording from SIG-node meeting is from April, so I'm afraid the answer is no. But, this was still a preliminary discussion. I think there needs to more evidence on the health of the nfd project and its benefits before the move under sig-node will be considered

andyxning:
> I think there needs to more evidence on the health of the nfd project and its benefits before the move under sig-node will be considered

Agreed. We need first make nfd more common, imo, the support for the hardware detect is somewhat too detail. and collect more user cases.

marquiz:
> As for the detailed collect, it do not mean that it is not necessary, but maybe some common useful ones are more demanded.

Hmm, I don't know if I quite understand what you mean here. Do you have some specific features in mind?

PR #75, opened 2017-09-06
move to k8s.io namespace

andyxning:
This PR is part of #74.

This PR will move node-feature-discovery import path namespace from `github.com/kubernetes-incubaotor` to `k8s.io`.


balajismaniam:
As discussed in https://github.com/kubernetes-incubator/node-feature-discovery/pull/71#issuecomment-327364231, let's do this in beta.

andyxning:
We are moving forward to make beta release, i think it is fine to make this change for now. When beta release is make, it should already move to `k8s.io`, imo. WDYT @dchen1107 @davidopp

PR #76, opened 2017-09-06
Migrate from glide to godep

andyxning:
This PR is part of #74 .

This PR keep node-feature-discovery skeleton in sync with [kubernetes-template-project](https://github.com/kubernetes/kubernetes-template-project).

Note: This PR also migrate vendor management tool from glide to godep.

This PR should be accepted after #75 .


balajismaniam:
@andyxning For graduating to beta, I don't think it is necessary to move from glide to godep. Closing this PR for now.

andyxning:
Make node-feature-discovery's skeleton suitable for an kubernetes repo should be done ASAP, imo.

WDYT @dchen1107 @balajismaniam

ConnorDoyle:
I think @balajismaniam's point is that the template project is not prescriptive at all about go dependency manager tooling, and therefore isn't relevant to beta graduation.

andyxning:
@ConnorDoyle 

That was my mistake. I mixed two logical different commits into one. The reason for doing this is that as stated in the template project `godep` is suggested(maybe it is not a strong restrict because moving to glide has actually been discussed in Kubernetes.).

PR #77, opened 2017-09-06
Move nfd image to quay under kubernetes

andyxning:
This PR is part of #74 .

This PR move default image for node-feature-discovery to `quay.io` under the namespace of `kubernetes`.

This PR should be accepted after #76 

Note: This will need to contact the guys who have write access to `quay.io/kubernetes`.
 .


ConnorDoyle:
Closing: 1) this patch is based on another PR which we already closed. 2) we don't want to host the image in the kubernetes org in quay.io until after graduation.

andyxning:
@ConnorDoyle What is the status about the graduation scheduler? or, what is the current step for NFD?

marquiz:
The opinion in SIG-node was inclined towards moving nfd under kubernetes-sigs instead of kubernetes  core. There's no schedule, yet. I think we need to do some homework (e.g. provide data on the benefits of nfd, possibly improving testing), after which we can bring this up to SIG-node, again.

PR #13, opened 2016-09-20
Fixed vulnerabilities in Docker image.

balajismaniam:
Fixes #10. Update resolves all fixable vulnerabilities. Tested Manually.
- Updated to debian testing (stretch).
- Updated to go v1.7.
- Added .dockerignore file.

PR #88, opened 2017-12-20
enable nfd framework on Arm64 platform

lubinsz:
currently, nfd framework can't be work on Arm64 platform.
   Fix points:
   1, removed rdt on Arm64 platform
   2, optimized the code related to multi-arch

Change-Id: If605041f6d2243ae2afb3248edbf102083c6dcb4
Jira: ENTOS-432

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,4 @@
-# Taken from https://github.com/docker-library/golang/blob/master/1.6/Dockerfile 
-# to build our golang image with debian testing (stretch).
-FROM buildpack-deps:stretch-scm
-
-# gcc for cgo.
-RUN apt-get update && apt-get install -y --no-install-recommends \
-		g++ \
-		gcc \
-		libc6-dev \
-		make \
-	&& rm -rf /var/lib/apt/lists/*
-
-ENV GOLANG_VERSION 1.7.1
-ENV GOLANG_DOWNLOAD_URL https://golang.org/dl/go$GOLANG_VERSION.linux-amd64.tar.gz
-ENV GOLANG_DOWNLOAD_SHA256 43ad621c9b014cde8db17393dc108378d37bc853aa351a6c74bf6432c1bbd182
-
-RUN curl -fsSL "$GOLANG_DOWNLOAD_URL" -o golang.tar.gz \
-	&& echo "$GOLANG_DOWNLOAD_SHA256  golang.tar.gz" | sha256sum -c - \
-	&& tar -C /usr/local -xzf golang.tar.gz \
-	&& rm golang.tar.gz
-
-ENV GOPATH /go
-ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
-
-RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
-WORKDIR $GOPATH
+FROM golang:1.8 
 
 # Build node feature discovery and set it as entrypoint.
 ADD . /go/src/github.com/kubernetes-incubator/node-feature-discovery
@@ -31,9 +6,18 @@ ADD . /go/src/github.com/kubernetes-incubator/node-feature-discovery
 WORKDIR /go/src/github.com/kubernetes-incubator/node-feature-discovery
 
 ARG NFD_VERSION
-RUN git clone --depth 1 https://github.com/01org/intel-cmt-cat.git
-RUN cd intel-cmt-cat/lib; make install
-RUN cd rdt-discovery; make
+
+RUN case $(dpkg --print-architecture) in \
+        arm64) \
+                echo "skip rdt on Arm64 platform" \
+                ;; \
+        *) \
+                git clone --depth 1 https://github.com/01org/intel-cmt-cat.git \
+                && cd intel-cmt-cat/lib; make install \
+                && cd ../../rdt-discovery; make \
+                ;; \
+        esac
+
 RUN go get github.com/Masterminds/glide
 RUN glide install --strip-vendor
 RUN go install \


lubinsz:
@jaypipes
Hi, 
Please check it.
Thanks.

Dockerfile
```diff
@@ -1,39 +1,23 @@
...
-	&& echo "$GOLANG_DOWNLOAD_SHA256  golang.tar.gz" | sha256sum -c - \
-	&& tar -C /usr/local -xzf golang.tar.gz \
-	&& rm golang.tar.gz
-
-ENV GOPATH /go
-ENV PATH $GOPATH/bin:/usr/local/go/bin:$PATH
-
-RUN mkdir -p "$GOPATH/src" "$GOPATH/bin" && chmod -R 777 "$GOPATH"
-WORKDIR $GOPATH
+FROM golang:1.8
```
jaypipes: Does the above change break builds for x86 platforms? Specifically, does the above change break the Intel RDT/CAT library build dependencies?

Dockerfile
```diff
@@ -1,39 +1,23 @@
...
 WORKDIR /go/src/github.com/kubernetes-incubator/node-feature-discovery
 
 ARG NFD_VERSION
-RUN git clone --depth 1 https://github.com/01org/intel-cmt-cat.git
-RUN cd intel-cmt-cat/lib; make install
-RUN cd rdt-discovery; make
+
+RUN case $(dpkg --print-architecture) in \
+        arm64) \
+                echo "skip rdt on Arm64 platform" \
```
balajismaniam: Nit: s/skip rdt/skipping RDT feature source

Dockerfile
lubinsz: Hi @jaypipes 
The content of this Dockerfile is too redundant.
To make it clean, I used the multi-arch supported image to replace it.
golang:1.8 support multi-arch.
And also it not break the Intel RDT/CAT library build dependencies. 
I have tested it.  No problems were found in x86 platform.

Dockerfile
lubinsz: Hi @balajismaniam ,
1, The extension instructions in Arm64 platform are different with x86. So I will also update the docs in an independent patch.
Here is a extension instruction list in Arm64 which we collected: For your reference: https://docs.google.com/document/d/13Huvi5bTyqhWp2BQtacaFTaKmeimXLT3wze72MbgkVk/edit

2, For multi-arch issues, I will define the labels with new format, and it is only for arm platform.
Add arm64 in the label,  such as nfd-cpuid-arm64-aes. Just like the format in Openstack.
For your reference: https://docs.openstack.org/os-traits/latest/user/index.html

3, RDT doesn't support arm. We have a similar project in ARM, we called it MPAM.
So for Arm platform, RDT will be skipped, and then we will enable MPAM for NFD.

Dockerfile
jaypipes: OK, thanks for checking on that @lubinsz. I still think that you should separate out the "get multi-platform Dockerfile working" changes from the "introduce support for NFD" changes, though. Can you do that?

Dockerfile
jaypipes: @lubinsz: is arm64 AES somehow different from x86 AES? AFAIU, AES support in arm64 just means support for the AES x86 instruction set extension on arm64 platforms. It's not a completely new capability. So, I would advise the label/annotation be simply "nfd-x86-aes" since it's an x86 instruction set extension, not an arm64 instruction set extension.

Dockerfile
balajismaniam: @lubinsz The content you deleted was added to reduce vulnerabilities in the Docker image. See https://github.com/kubernetes-incubator/node-feature-discovery/pull/13.

Dockerfile
balajismaniam: > 1, The extension instructions in Arm64 platform are different with x86. So I will also update the docs in an independent patch.
Here is a extension instruction list in Arm64 which we collected: For your reference: https://docs.google.com/document/d/13Huvi5bTyqhWp2BQtacaFTaKmeimXLT3wze72MbgkVk/edit

Thanks.

> 2, For multi-arch issues, I will define the labels with new format, and it is only for arm platform.
Add arm64 in the label,  such as nfd-cpuid-arm64-aes. Just like the format in Openstack.
For your reference: https://docs.openstack.org/os-traits/latest/user/index.html

See https://github.com/kubernetes-incubator/node-feature-discovery/issues/86#issuecomment-353184147. Adding the architecture as a part of the label will be redundant.

Dockerfile
jaypipes: @balajismaniam I see the code from issue #13, but I actually see no indication of what specific vulnerabilities in the Docker image were addressed by those changes. I think it's probably good to remove those lines and go with the straight FROM golang:1.8 base unless there's an important reason to diverge.

lubinsz: @balajismaniam 
I see. 
I have updated the code by following  #13 
Currently, starting from golang version 1.85,  there is an official binary package for arm64 available for download.
So the golang version should be updated into 1.8.5 at least.
Please see the code as more reference.

Dockerfile
lubinsz: @balajismaniam @jaypipes
I see.
Currently, the architecture will not be a part of the label for arm64 platform

lubinsz:
@jaypipes @balajismaniam 
Any questions?

Dockerfile
balajismaniam: @jaypipes I added the comment just for reference. You see the differences in the vulnerabilities in quay.io. There were some packages in golang base image, which introduced some vulnerabilities. I'm fine with using just `FROM golang:1.8`. 
@lubinsz Please revert back to using `FROM golang:1.8`. Sorry for the confusion.

balajismaniam:
@lubinsz See https://github.com/kubernetes-incubator/node-feature-discovery/pull/88#discussion_r160212796.

lubinsz:
@balajismaniam 
Done. Thanks.

PR #91, opened 2018-02-01
Export proxy env vars inside docker build

marquiz:
Without correct proxy settings propagated to docker build it is not possible to
build node-feature-discovery from behind a proxy server, e.g. in corporate
networks. This patch fixes the issue by exporting all http(s) proxy related
environment variables as build-time variables.


Makefile
```diff
@@ -12,4 +12,10 @@ all: docker
 # QUAY_REGISTRY_USER=<my-username> make docker -e.
 docker:
 	docker build --build-arg NFD_VERSION=$(VERSION) \
+		--build-arg http_proxy=$(http_proxy) \
```
balajismaniam: I'm not sure if this change is required. Usually developers use software mechanisms to bypass firewall and proxy settings.

marquiz: Do you see that this change would cause some problems? There are developers, like me, who do not use these software mechanisms or do not want to deploy those in every development environment. These proxy settings are ubiquitous all over kubernetes so why couldn't nfd build support these, as well?

marquiz:
ping @balajismaniam, any further comments?

marquiz:
ping @balajismaniam! This is biting me all the time

PR #94, opened 2018-02-16
Remove hardcoding of nfd source path

### LINKED ISSUE #52: Parameterize RDT detection binary location.
NFD shouldn't necessarily assume it runs inside the container we build by default. In `sources.go`:

```go
RDTBin = "/go/src/github.com/kubernetes-incubator/node-feature-discovery/rdt-discovery"
```

The program could get value from an environment variable instead (and the build can inject the value above when building the container).

marquiz:
Makes it possible to build/run nfd outside the hardcoded path, e.g. locally. Fixes #52


balajismaniam:
@marquiz please update this branch.

marquiz:
updated

PR #95, opened 2018-02-19
Added nonrotational storage detection

### LINKED ISSUE #85: Discover nodes with SSD/Rotational hard drives
It is often important for I/O intensive workloads to utilize SSDs. Being able to select nodes with an SSD available will make operating kubernetes clusters much easier.
For other workloads a rotational hard drive may be good enough.

okartau:
Add label if any non-rotational block device is present in the node.
The label will be: nfd-storage-nonrotational=true
This change also adds feature source "storage".

Fixes #85 

Note that during label creation it is not known how devices are actually
used by pods, so the approach is simplistic, and more sophisticated
logic will be needed at some level to make precise decisions.
But labeling "if any SSD present" should cover simple cases reasonably.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ node-feature-discovery.
   -h --help                   Show this screen.
   --version                   Output version and exit.
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,rdt,pstate,network]
+                              [Default: cpuid,rdt,pstate,network,storage]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
@@ -58,6 +58,7 @@ The current set of feature sources are the following:
 - [Intel Resource Director Technology][intel-rdt]
 - [Intel P-State driver][intel-pstate]
 - Network
+- Storage
 
 ### Feature labels
 
@@ -79,7 +80,8 @@ the only label value published for features is the string `"true"`._
   "node.alpha.kubernetes-incubator.io/nfd-cpuid-<feature-name>": "true",
   "node.alpha.kubernetes-incubator.io/nfd-rdt-<feature-name>": "true",
   "node.alpha.kubernetes-incubator.io/nfd-pstate-<feature-name>": "true",
-  "node.alpha.kubernetes-incubator.io/nfd-network-<feature-name>": "true"
+  "node.alpha.kubernetes-incubator.io/nfd-network-<feature-name>": "true",
+  "node.alpha.kubernetes-incubator.io/nfd-storage-<feature-name>": "true"
 }
 ```
 
@@ -117,6 +119,12 @@ such as restricting discovered features with the --label-whitelist option._
 | Feature name   | Description                                                                         |
 | :------------: | :---------------------------------------------------------------------------------: |
 | [SRIOV][sriov] | Single Root Input/Output Virtualization (SR-IOV) enabled Network Interface Card
+
+### Storage Features
+
+| Feature name       | Description                                                                         |
+| :--------------:   | :---------------------------------------------------------------------------------: |
+| nonrotationaldisk  | Non-rotational disk, like SSD, is present in the node
 
 ## Getting started
 ### System requirements
--- a/source/storage/storage.go
+++ b/source/storage/storage.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package storage
+
+import (
+	"fmt"
+	"io/ioutil"
+)
+
+// Source implements FeatureSource.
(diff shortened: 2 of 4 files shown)


okartau:
I have verified this change in a mixed cluster with few VMs (no SSD disk presented to SW) and one physical host with SSD, The SSD detection worked. K8s components were 1.9.3 versions.

okartau:
travis-ci check complete by log, but GH status still "in progress" for few days. Comms problem?

marquiz:
Hmm, strange. Seems like a glitch as travis shows green status. Repo admins should be able to re-trigger the build. Or, you could try closing and re-opening this PR.

okartau:
force pushed after --amend, re-tested and state OK

source/storage/storage.go
```diff
@@ -0,0 +1,41 @@
...
+
+// Name returns an identifier string for this feature source.
+func (s Source) Name() string { return "storage" }
+
+// Discover returns feature names for storage: nonrotational if any SSD drive present.
+func (s Source) Discover() ([]string, error) {
+	features := []string{}
+
+	// Check is there zero in any device rotational status
+	cmd := exec.Command("/bin/sh", "-c", "grep 0 /sys/block/*/queue/rotational")
```
kad: Generally, instead of spawning shell and using grep, it is more safe to write small function to traverse directories and read files.

okartau: Thanks for suggestion, I rewrote that piece to walk directories instead of grep (previous variant remained visible in kad's comment as "outdated").

README.md
```diff
@@ -118,6 +120,12 @@ such as restricting discovered features with the --label-whitelist option._
 | :------------: | :---------------------------------------------------------------------------------: |
 | [SRIOV][sriov] | Single Root Input/Output Virtualization (SR-IOV) enabled Network Interface Card
 
+### Storage Features
+
+| Feature name   | Description                                                                         |
+| :------------: | :---------------------------------------------------------------------------------: |
+| nonrotational  | Non-rotational disk, like SSD-disk, is present in the node
```
balajismaniam: May be change to `nonrotationaldisk`?

balajismaniam: s/SSD-disk/SSD

source/storage/storage.go
```diff
@@ -0,0 +1,51 @@
...
+	if err == nil {
+		for _, bdev := range blockdevices {
+			fname := "/sys/block/" + bdev.Name() + "/queue/rotational"
+			bytes, err := ioutil.ReadFile(fname)
+			if err != nil {
+				return nil, fmt.Errorf("can't read rotational status: %s", err.Error())
+			}
+			if bytes[0] == byte('0') {
+				// Non-rotational storage is present, add label.
+				features = append(features, "nonrotational")
```
balajismaniam: May be change to `nonrotationaldisk`?

source/storage/storage.go
```diff
@@ -0,0 +1,51 @@
...
+	// Check is there zero in any device rotational status
+	blockdevices, err := ioutil.ReadDir("/sys/block/")
+	if err == nil {
+		for _, bdev := range blockdevices {
+			fname := "/sys/block/" + bdev.Name() + "/queue/rotational"
+			bytes, err := ioutil.ReadFile(fname)
+			if err != nil {
+				return nil, fmt.Errorf("can't read rotational status: %s", err.Error())
+			}
+			if bytes[0] == byte('0') {
```
balajismaniam: Do you know if this is the standard way of figure out whether a disk is non-rotational or not? Are there any other ways to figure it out?

balajismaniam: Will this work in both a VM and bare-metal environment?

source/storage/storage.go
```diff
@@ -0,0 +1,51 @@
...
+type Source struct{}
+
+// Name returns an identifier string for this feature source.
+func (s Source) Name() string { return "storage" }
+
+// Discover returns feature names for storage: nonrotational if any SSD drive present.
+func (s Source) Discover() ([]string, error) {
+	features := []string{}
+
+	// Check is there zero in any device rotational status
```
balajismaniam: Nit: replace with `Check if there is any non-rotational block devices attached to the node.`

source/storage/storage.go
```diff
@@ -0,0 +1,51 @@
...
+		for _, bdev := range blockdevices {
+			fname := "/sys/block/" + bdev.Name() + "/queue/rotational"
+			bytes, err := ioutil.ReadFile(fname)
+			if err != nil {
+				return nil, fmt.Errorf("can't read rotational status: %s", err.Error())
+			}
+			if bytes[0] == byte('0') {
+				// Non-rotational storage is present, add label.
+				features = append(features, "nonrotational")
+				break // don't repeat add of same label
```
balajismaniam: Nit: You can remove the comment or change it to `avoid adding the same labels`.

source/storage/storage.go
okartau: Reading /sys/block/DEV/queue/rotational is AFAIK the only universal way without using more specialized disk tools which may be specific to technology or vendor. Some sources recommend lsblk but this utility simply reads same location under /sys/block. If someone knows better or more "standard" way, please speak up.

About VM: situation is not very good from detection POV. If a VM runs on top of bare metal where non-rotational disk is used, VM likely cant detect it easily, as (at least typically I think?) disk device is virtualized for VM, and "non-rotational" status gets lost in VM even if phys device has it. It is possible that some virtualization methods allow exposing device directly, then situation can be better.
In similar fashion, non-rotational capability gets hidden/lost by RAID controller if RAID controller presents virtual storage to host.

okartau:
I addressed recent mod requests (nonrotationaldisk, SSD-disk, and comments), force-pushed

source/storage/storage.go
balajismaniam: Thanks for looking at this.

PR #96, opened 2018-02-21
Add memory source and NUMA detection.

### LINKED ISSUE #84: Discover nodes with NUMA architecture
Software such as MongoDB or MySQL have known problems with NUMA architectures. See https://docs.mongodb.com/manual/administration/production-notes/#mongodb-and-numa-hardware and https://blog.jcole.us/2010/09/28/mysql-swap-insanity-and-the-numa-architecture/ for example.
It'd be useful to be able to select nodes without such architecture for such deployments.

okartau:
If multiple nodes are marked online in
/sys/devices/system/node/online, it's a sign of NUMA
architecture. Mark it using nfd-memory-numa label.

Fixes #84 

Here is one way of detecting NUMA architecture based on number of online nodes.
Please comment if you see some problems with such detection.
I have verified it on single-node and 2- and 4-node NUMA systems.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ node-feature-discovery.
   -h --help                   Show this screen.
   --version                   Output version and exit.
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,rdt,pstate,network,storage]
+                              [Default: cpuid,rdt,pstate,memory,network,storage]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
@@ -57,6 +57,7 @@ The current set of feature sources are the following:
 - [CPUID][cpuid] for x86 CPU details
 - [Intel Resource Director Technology][intel-rdt]
 - [Intel P-State driver][intel-pstate]
+- Memory
 - Network
 - Storage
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -112,6 +114,12 @@ such as restricting discovered features with the --label-whitelist option._
 | SSE4.2         | Streaming SIMD Extensions 4.2 (SSE4.2)
 | SGX            | Software Guard Extensions (SGX)
 
+### Memory Features
+
+| Feature name   | Description                                                                         |
+| :------------: | :---------------------------------------------------------------------------------: |
+| multinode      | Multiple nodes detected, sign of NUMA architecture
```
balajismaniam: Can we call this feature `numa`?

source/memory/memory.go
```diff
@@ -0,0 +1,51 @@
...
+	"strings"
+)
+
+// Source implements FeatureSource.
+type Source struct{}
+
+// Name returns an identifier string for this feature source.
+func (s Source) Name() string { return "memory" }
+
+// Discover returns feature names for memory: multinode if more than one node present.
```
balajismaniam: Nit: `multinode if more than one node present` -> `numa if more than one memory node is present`.

README.md
okartau: The reason why I called it multinode is to keep the naming more neutral, i.e. NFD does not associate automatically "multiple nodes == numa", but instead express what NFD can read from system, i.e. "there is more than one node".
But I admit, this may in reality add confusion. In practice numa equals multinode everywhere, right, so lets leave separation to future when there may be multiple mappings. I will change the name to numa

okartau: should it be NUMA or numa ?

okartau:
force pushed addressing latest comments. Using numa (note: lowercase) for label

README.md
balajismaniam: `numa` is fine.

balajismaniam:
@okartau needs rebase.

okartau:
rebased and force-pushed

PR #99, opened 2018-02-22
Make intel-cmt-cat a git submodule

marquiz:
This makes the build more deterministic. Previously, nfd just blindly
took the tip revision from intel-cmt-cat master branch which could brake
the build without any changes in nfd itself.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ node-feature-discovery-job.json
 vendor/
 demo/helper-scripts/*.pdf
 demo/helper-scripts/*.log
-intel-cmt-cat/
 rdt-discovery/l2-alloc-discovery
 rdt-discovery/l3-alloc-discovery
 rdt-discovery/mon-discovery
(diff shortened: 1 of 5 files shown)


kad:
current location is: https://github.com/intel/intel-cmt-cat

marquiz:
Ping @balajismaniam, any comments

balajismaniam:
@marquiz This is a small project. Adding a submodule to it makes it more complicated. My suggestion would be to pick one of the releases (https://github.com/intel/intel-cmt-cat/releases) in order to increase the determinism of the builds.

marquiz:
> @marquiz This is a small project. Adding a submodule to it makes it more complicated.

I don't necessarily see how the size of the project relates to this. In my view git-submodule is a way to do deterministic vendoring for "git dependencies", similarly how glide is used for go vendoring.

But, I also understand why many people dislike git submodules as they have their problems.

marquiz:
I created PR #112 that solves this in the way you suggested.

balajismaniam:
Thanks. Closing in favor of #112.

PR #103, opened 2018-02-22
 Simplify dockerfile

marquiz:
This change requires #128, and, it has been cherry-picked in this PR


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "intel-cmt-cat"]
+	path = intel-cmt-cat
+	url = https://github.com/01org/intel-cmt-cat.git
```
kad: it is https://github.com/intel/intel-cmt-cat now.

marquiz:
ping @balajismaniam, any review comments?

kad:
@marquiz @okartau do we need to keep those open ? what is parth forward with it ?

marquiz:
> do we need to keep those open ? what is parth forward with it ?

@kad: Which "those" are you referring to?

kad:
I've seen few PRs that are quite old and still in open state, and as well, needs rebase. That was the reason of question.

marquiz:
> I've seen few PRs that are quite old and still in open state, and as well, needs rebase. That was the reason of question.

I've been rebasing these periodically. But with no review there is no progress.

ping @balajismaniam

PR #104, opened 2018-02-23
Add, correct some notes in documentation

### LINKED ISSUE #92: Validated NFD with recent kubernetes version (1.9.2), could add some doc.notes
Here are some notes from  "a newcomer run attempt" validating NFD on top of
current k8s (1.9.2). Some doc (code?) changes could be made based on findings.

Systems configuration:
4 nodes as KVM-style VMs on top of x86_64 HW
Distro on all nodes: CentOS 7, stock "minimal server" config
Added repositories for: 
- Kubernetes: packages.cloud.google.com/yum/
-  Docker: download.docker.com/linux/centos/7

Kernel: 3.10.0-693.17.1.el7.x86_64
Docker: docker-ce-17.12.0.ce-1.el7.centos.x86_64
kube- ctl,adm,let versions: 1.9.2
Cluster initial creation by "kubeadm init" and "kubeadm join"

1. NFD Building does not work behind proxy. This can be improved by PR #91, thanks to @marquiz for creating this PR.

2. Tried stand-alone run as in README "Command Line Interface". This chapter does not describe how to run NFD as standalone. By checking Dockerfile it becomes clear that container activation is enough. Mentioning this in README would add clarity.

3. Tried to run as cluster job. label-nodes.sh schedules a pod run on every node which is in "Ready" state. In a default-config cluster, the master node is tainted:NoSchedule which results in one of scheduled pods doomed to fail. If a clean run is desired, we should either schedule NUM-1 jobs leaving master out (master will not get labels, but if it remains NoSchedule that will be consistent, right), or Master config needs to be modified. We could add note in README, and/or comment in label-nodes.sh?

4. A job run on a node will detect features as designed, but labels-add part will fail in a default-config cluster, caused by RBAC (Role-based Access Control) blocking get,set operations. RBAC was added after NFD was written? The first error will be:

> can't get pods: pods "node-feature-discovery-f2v8m" is forbidden: User "system:serviceaccount:default:default" cannot get pods in the namespace "default"

To solve RBAC blockers:
Role access rights can be added using [audit2rbac](https://github.com/liggitt/audit2rbac) tool. Use of audit2rbac tool requires auditing configured. Roles adding needs 3 iterations to cover all stages of similar errors:

- User cannot get pods in the namespace "default"
- User cannot get nodes at the cluster scope
- User cannot update nodes at the cluster scope

With role rights added, the NFD label adding starts to work.
The main level README could mention roles rights dependency.

5. README states that "The discovery script will launch a job on each unlabeled node". But that seems not so: label-nodes.sh script does not check labeled vs. unlabeled nodes. It launches a cluster job with NumNodes desired completions. That means, job is always restarted on any Ready node.

6. Mixed success case: If some node is not able to complete the successful run, for example if RBAC state (see point 4 above) is incorrect on some nodes and correct on others (happens for example if roles have been added using audit2rbac, followed by new node join), cluster will run repeated NFD jobs on "good" nodes in order to fulfill the number of desired completions, and marks overall job progress as good. This is misleading and incorrect as some nodes will remain unlabeled. To cover such mixed cases, more sophisticated logic like "give me exactly one good completion per node" will be required instead of simple "give me N overall completions". Doc change or separate issue based on this? Not sure does k8s provide easy API for such scheduling.

okartau:
A "newbie style" deployment attempt was made on a recent
cluster, and some added notes in README and in label-nodes.sh
could help someone who just started with node-feature-discovery.
Correcting also description of label-nodes.sh which does not
deal with unlabeled nodes as was promised in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,10 @@ This project uses GitHub [milestones](https://github.com/kubernetes-incubator/no
 
 ## Command line interface
 
+To try out stand-alone, one can run a Docker container where node-feature-discovery is already set as entry point.
+Such run is useful for checking features-detection part, but labeling part is expected to fail.
+It is recommended to use --no-publish and --oneshot to achieve clean run in stand-alone case.
+
 ```
 node-feature-discovery.
 
@@ -182,6 +186,9 @@ kubectl create -f rbac.yaml
 kubectl create -f node-feature-discovery-daemonset.json.template
 ```
 
+When the job runs, it contacts the Kubernetes API server to add labels
+to the node to advertise hardware features.
+
 If you have RBAC authorization enabled (as is the default e.g. with clusters initialized with kubeadm) you need to configure the appropriate ClusterRoles, ClusterRoleBindings and a ServiceAccount in order for NFD to create node labels. The provided templates will configure these for you.
 
 When run as a daemonset, nodes are re-labeled at an interval specified using
@@ -189,16 +196,15 @@ the `--sleep-interval` option. In the [template](https://github.com/kubernetes-i
 which is also the default when no `--sleep-interval` is specified.
 
 Feature discovery can alternatively be configured as a one-shot job. There is
-an example script in this repo that demonstrates how to deploy the job to
-unlabeled nodes.
+an example script in this repo that demonstrates how to deploy the job in the cluster.
 
 ```
 ./label-nodes.sh
 ```
 
-The discovery script will launch a job on each unlabeled node in the
-cluster. When the job runs, it contacts the Kubernetes API server to add labels
-to the node to advertise hardware features (initially, from `cpuid`, RDT, p-state and network).
+The label-nodes.sh script tries to launch as many jobs as there are Ready nodes.
+Note that this approach does not guarantee running once on every node.
+For example, if some node is tainted NoSchedule or fails to start a job for some other reason, then some other node will run extra job instance(s) to satisfy the request and the tainted/failed node does not get labeled.
 
 [![asciicast](https://asciinema.org/a/11wir751y89617oemwnsgli4a.png)](https://asciinema.org/a/11wir751y89617oemwnsgli4a)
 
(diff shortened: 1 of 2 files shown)


marquiz:
This is partly outdated, e.g. because RBAC template was added recently. @okartau, please update the patch after which we can merge it

okartau:
rebased and polished a bit more. Note that I moved sentence about "contacts API server" up where it fits better

README.md
```diff
@@ -182,23 +184,25 @@ kubectl create -f rbac.yaml
...
+an example script in this repo that demonstrates how to deploy the job in the cluster.
 
 ```
 ./label-nodes.sh
 ```
 
-The discovery script will launch a job on each unlabeled node in the
-cluster. When the job runs, it contacts the Kubernetes API server to add labels
-to the node to advertise hardware features (initially, from `cpuid`, RDT, p-state and network).
+The label-nodes.sh script tries to launch as many jobs as there is Ready nodes.
```
marquiz: should be "as there **are**"

README.md
```diff
@@ -182,23 +184,25 @@ kubectl create -f rbac.yaml
...
+Note that this approach does not guarantee running once on every node.
+For example, if some node is tainted NoSchedule or fails to start a job for some other reason, then some other node will run extra job instance(s) to satisfy the request.
```
marquiz: should this be clarified a bit more, by adding something like
*...and the tainted/failed node does not get labeled.*

label-nodes.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Get the node count in the Kubernetes cluster
+# Get the count of nodes in Ready state in the Kubernetes cluster
```
marquiz: **number** of nodes(?)

README.md
```diff
@@ -29,6 +29,8 @@ This project uses GitHub [milestones](https://github.com/kubernetes-incubator/no
 
 ## Command line interface
 
+One way to run stand-alone is by running a Docker container where node-feature-discovery is set as entry point.
```
marquiz: Hmm, what usage are you thinking here? First of all, in nfd images node-feature-discovery is already set as the default entry point. Of course, you can run the nfd image directly on docker, but then the labeling fails because the k8s APIs of the cluster are not available. I think this should be mentioned, at least, probably with a recommendation to use --no-publish to avoid the related errors.

PR #106, opened 2018-03-07
Add label to advertise selinux status

chaitanyaenr:
This commit:
- enables node-feature-dicovery to advertise selinux status
  on the node by adding a label.
- update the template to mount /sys into the container, this is
  needed to know about the selinux status on the host
- adds selinux source for unit tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ node-feature-discovery.
   -h --help                   Show this screen.
   --version                   Output version and exit.
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,rdt,pstate,memory,network,storage]
+                              [Default: cpuid,rdt,pstate,memory,network,storage,selinux]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
@@ -60,6 +60,7 @@ The current set of feature sources are the following:
 - Memory
 - Network
 - Storage
+- Selinux
 
 ### Feature labels
 
@@ -83,7 +84,8 @@ the only label value published for features is the string `"true"`._
   "node.alpha.kubernetes-incubator.io/nfd-pstate-<feature-name>": "true",
   "node.alpha.kubernetes-incubator.io/nfd-memory-<feature-name>": "true",
   "node.alpha.kubernetes-incubator.io/nfd-network-<feature-name>": "true",
-  "node.alpha.kubernetes-incubator.io/nfd-storage-<feature-name>": "true"
+  "node.alpha.kubernetes-incubator.io/nfd-storage-<feature-name>": "true",
+  "node.alpha.kubernetes-incubator.io/nfd-selinux-<feature-name>": "true"
 }
 ```
 
@@ -133,6 +135,12 @@ such as restricting discovered features with the --label-whitelist option._
 | Feature name       | Description                                                                         |
 | :--------------:   | :---------------------------------------------------------------------------------: |
 | nonrotationaldisk  | Non-rotational disk, like SSD, is present in the node
+
+### Selinux Features
+
+| Feature name       | Description                                                                         |
+| :--------------:   | :---------------------------------------------------------------------------------: |
+| selinux            | selinux is enabled on the node
 
 ## Getting started
 ### System requirements
--- a/node-feature-discovery-job.json.template
+++ b/node-feature-discovery-job.json.template
@@ -17,7 +17,7 @@
         }
       },
       "spec": {
-	"hostNetwork": true,
+        "hostNetwork": true,
         "containers": [
           {
             "env": [
@@ -37,10 +37,24 @@
                 "containerPort": 7156,
                 "hostPort": 7156
               }
+            ],
+            "volumeMounts": [
+              {
+                "name": "host-sys",
+                "mountPath": "/host-sys"
+              }
             ]
           }
         ],
-        "restartPolicy": "Never"
+        "restartPolicy": "Never",
+        "volumes": [
+          {
+            "name": "host-sys",
+            "hostPath": {
+              "path": "/sys"
+            }
+          }
+        ]
       }
     }
   }
--- a/source/selinux/selinux.go
+++ b/source/selinux/selinux.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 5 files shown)


okartau:
CI test fails because you did not add new source in main_test.go

README.md
```diff
@@ -80,6 +81,7 @@ the only label value published for features is the string `"true"`._
   "node.alpha.kubernetes-incubator.io/nfd-rdt-<feature-name>": "true",
   "node.alpha.kubernetes-incubator.io/nfd-pstate-<feature-name>": "true",
   "node.alpha.kubernetes-incubator.io/nfd-network-<feature-name>": "true"
+  "node.alpha.kubernetes-incubator.io/nfd-selinux-<feature-name>": "true"
```
okartau: added line without comma creates bad json syntax here as there are now 2 lines without comma. 
This is README so no actual error however will happen.
Is it also the reason why GH diff display highlights one colon character here with red?

okartau:
Mounting /sys explicitly: is there setup where /sys is not already implicitly mounted ?
In existing code, /sys is already used by source/network/network.go, without explicit mount.
I have created two new PRs that also read files under /sys and I did not have problem running that code.

chaitanyaenr:
@okartau I think the source/network/network.go was able to get the info as the host network is set to  true in the template. I couldn't get the selinux status without mounting /sys.

chaitanyaenr:
@okartau I have signed the cncf-cla. Any changes needed for this PR?

okartau:
I'm not authority here (joined just recently myself and proposing my first PRs), so I dont have good visibility from that POV. Code looks OK, but let's wait for approving review(s). Automatic mounting of /sys still slightly puzzles me, as in my PRs I deal with other areas (than network), like block devices, and I am able to read /sys without explicit mount. I wonder can this be platform dependent?

okartau:
@chaitanyaenr can you verify that if you run NFD job without your patch (i.e. without explicit /sys mount), how the pstate detection goes? (the code in source/pstate/pstate.go). The idea is to clarify is /sys just never mounted from the start in your case.

okartau:
I cross-checked other way trying change of this PR in my cluster, and that supports what you say, i.e. /sys/fs/selinux becomes visible when /sys is explicitly mounted. But other (most?) areas under /sys are readable without explicit mount. Same can be observed leaving k8s out of picture, i.e. by starting NFD container manually using 'docker'

chaitanyaenr:
@okartau thanks for trying it out.

okartau:
In code, you print an error msg if /sys/fs/selinux/enforce can't be read. But such state may be normal condition on a system without SELinux, so msg could be log/info without indicating error?

If we want to get into detail in detection and reporting (which I believe may help), I see 4 possible states (there may be even more):
1. /sys/fs/selinux/enforce readable and contains "1": we add label
2. /sys/fs/selinux/enforce readable and contains "0": we don't add label
In cases 1. and 2. we can emit info-level msg that file was readable.

3. /sys/fs/selinux exists but enforce unreadable: This happens on a SELinux-enabled system where docker runs without mounting /sys. We can't detect SELinux enforcement state, so we don't add label. Here error msg is justified, with a hint about mounting /sys
4. /sys/fs/selinux does not exist: system is not SELinux-enabled, no label, no error. We can indicate finding with log message. Note that presence /sys/fs/selinux can be read without explicit mount of /sys to Docker.

jeremyeder:
@chaitanyaenr thoughts on @okartau feedback?

chaitanyaenr:
@jeremyeder @okartau I have changed the log level to info and also included info about checking if the  system supports selinux, mounting /sys to read the selinux status.

chaitanyaenr:
@okartau did you get a chance to review the changes?

fabiand:
Cool - This is nice.

chaitanyaenr:
@fabiand makes sense, got rid of the privileged securitycontext.

fabiand:
Argh - Yes, you don't even need privileged :) /me could also read it locally as a regular user.

chaitanyaenr:
@balajismaniam PTAL.

README.md
```diff
@@ -60,6 +60,7 @@ The current set of feature sources are the following:
 - Memory
 - Network
 - Storage
+- Selinux
```
balajismaniam: Please add some details related to this feature similar to https://github.com/kubernetes-incubator/node-feature-discovery#storage-features.

node-feature-discovery-job.json.template
```diff
@@ -37,10 +37,24 @@
                 "containerPort": 7156,
                 "hostPort": 7156
               }
+            ],
+            "volumeMounts": [
+              {
+                "name": "sys",
```
balajismaniam: s/sys/host-sysfs

node-feature-discovery-job.json.template
```diff
@@ -37,10 +37,24 @@
...
+                "mountPath": "/sys"
```
balajismaniam: Mount it in `/host-sys` to differentiate between sys fs inside container and host sys fs?

node-feature-discovery-job.json.template
```diff
@@ -37,10 +37,24 @@
...
+              }
             ]
           }
         ],
-        "restartPolicy": "Never"
+        "restartPolicy": "Never",
+        "volumes": [
+          {
+            "name": "sys",
```
balajismaniam: s/sys/host-sysfs

source/selinux/selinux.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/golang/glog"
+)
+
+type Source struct{}
+
+func (s Source) Name() string { return "selinux" }
+
+func (s Source) Discover() ([]string, error) {
+	features := []string{}
+	status, err := ioutil.ReadFile("/sys/fs/selinux/enforce")
```
balajismaniam: If we mount this in a different path as per my comment above, this should be modified.

source/selinux/selinux.go
```diff
@@ -0,0 +1,39 @@
...
+	if err != nil {
+		glog.Info("Failed to detect the status of selinux, please check if the system supports selinux and make sure /sys is mounted into the container: %s", err)
```
balajismaniam: You need to return here similar to https://github.com/kubernetes-incubator/node-feature-discovery/blob/master/source/pstate/pstate.go#L37.

source/selinux/selinux.go
balajismaniam: Is checking the contents of `enforce` sufficient to determine the status of selinux? Do we need to examine the contents of `status` as well?

chaitanyaenr:
@balajismaniam thanks for the review, I have updated the PR to address the comments. PTAL.

balajismaniam:
Thanks @chaitanyaenr. 

Is checking the contents of `/sys/fs/selinux/enforce` sufficient to determine the status of selinux? Do we need to examine the contents of `/sys/fs/selinux/status` as well?

Can you add details on how you tested this end-to-end?

chaitanyaenr:
@balajismaniam checking the contents of /sys/fs/selinux/enforce is sufficient to determine the status of selinux. 
About the e2e testing, I have built a docker image with the selinux feature, modified the template to use the test image and checked if the selinux label is being added or not.

PR #108, opened 2018-03-15
Convert templates and examples from json to yaml

marquiz:
I find YAML a lot more (human-)user-friendly and less error prone to edit than json. Just to point that out, the json example in README had one syntax error (one extra comma)


marquiz:
One additional benefit with yaml is that it allows embedding comments

marquiz:
ping @balajismaniam!

marquiz:
Rebased. Ping @balajismaniam, this has been open without review for over 4 months now

PR #110, opened 2018-03-19
Fix json syntax errors in the node-discovery template

chaitanyaenr:


marquiz:
Those values are placeholders that are supposed to be replaced with numbers so I think this PR is invalid.

okartau:
this change actually breaks label-nodes.sh as replaced string in quotes becomes different data type:

./label-nodes.sh
error: error validating "node-feature-discovery-job.json": error validating data: [ValidationError(Job.spec.completions): invalid type for io.k8s.api.batch.v1.JobSpec.completions: got "string", expected "integer"

chaitanyaenr:
makes sense, closing the PR.

PR #111, opened 2018-03-23
check frequency boost support on Arm

lubinsz:


balajismaniam:
@lubinsz Can you please update this branch?

PR #119, opened 2018-04-10
Implement systemd service detection for 'irqbalance'

lubinsz:
When we deploy DPDK related pods, we should check some reuqired
system configuration information to get best performance for DPDK application.

Such as:
enable iommu, disable irqbalance, disable hyperthreading,
disable thp...

In this patch, I enabled detection for irqbalance.
For other information, I will enable them later.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,12 @@ Kernel config file to use, and, the set of config options to be detected are
 configurable.
 See [configuration options](#configuration-options) for more information.
 
+### System Features
+
+| Feature | Attribute  | Description                                           |
+| ------- | ---------- | ----------------------------------------------------- |
+| systemd | irqbalance | Check the service of irqbalance is active or not
+
 ### Local (User-specific Features)
 
 NFD has a special feature source named *local* which is designed for running
--- a/main.go
+++ b/main.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/node-feature-discovery/source/rdt"
 	"sigs.k8s.io/node-feature-discovery/source/selinux"
 	"sigs.k8s.io/node-feature-discovery/source/storage"
+	"sigs.k8s.io/node-feature-discovery/source/system"
 	api "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	k8sclient "k8s.io/client-go/kubernetes"
@@ -182,7 +183,7 @@ func argsParse(argv []string) (args Args) {
                               will override settings read from the config file.
                               [Default: ]
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,iommu,kernel,local,memory,network,pci,pstate,rdt,selinux,storage]
+                              [Default: cpuid,iommu,kernel,local,memory,network,pci,pstate,rdt,selinux,storage,system]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
@@ -270,6 +271,7 @@ func configureParameters(sourcesWhiteList []string, labelWhiteListStr string) (e
 		rdt.Source{},
 		selinux.Source{},
 		storage.Source{},
+		system.Source{},
 		// local needs to be the last source so that it is able to override
 		// labels from other sources
 		local.Source{},
--- a/source/system/system.go
+++ b/source/system/system.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package system
+
+import (
+	"log"
+
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+// Source implements FeatureSource.
+type Source struct{}
+
+// Name returns an identifier string for this feature source.
+func (s Source) Name() string { return "system" }
+
+// Discover returns feature names for system configuration: irqbalance, swap, ksm, memory compaction, iommu...
+func (s Source) Discover() (source.Features, error) {
+	features := source.Features{}
+	active, err := SystemctlStatusIsActive("irqbalance")
+
+	if err != nil {
+		log.Printf("ERROR: Failed to check irqbalance: %s", err)
+	} else {
+		features["systemd.irqbalance"] = active
+	}
+
+	return features, nil
+}
--- a/source/system/systemd.go
+++ b/source/system/systemd.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package system
+
+import (
+	"os/exec"
+	"strings"
+)
+
+func runSystemctlCmd(args []string) (string, error) {
+	output, err := exec.Command("systemctl", args...).CombinedOutput()
+	return strings.TrimRight(string(output), "\n"), err
+}
+
+func SystemctlStatusIsActive(service string) (string, error) {
+	args := []string{"is-active"}
+	args = append(args, service)
+
+	output, err := runSystemctlCmd(args)
+	if err != nil {
+		return "", err
+	}
+
+	return output, nil
+}
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,7 @@ FROM debian:stretch-slim
 
 COPY --from=builder /usr/local/bin /usr/local/bin
 COPY --from=builder /usr/local/lib /usr/local/lib
+COPY --from=builder /bin/systemctl /bin/systemctl
 COPY --from=builder /etc/kubernetes/node-feature-discovery /etc/kubernetes/node-feature-discovery
 RUN ldconfig
 COPY --from=builder /go/bin/node-feature-discovery /usr/bin/node-feature-discovery
--- a/main_test.go
+++ b/main_test.go
@@ -152,7 +152,7 @@ func TestArgsParse(t *testing.T) {
 				So(args.sleepInterval, ShouldEqual, 60*time.Second)
 				So(args.noPublish, ShouldBeTrue)
 				So(args.oneshot, ShouldBeTrue)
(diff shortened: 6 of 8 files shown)


lubinsz:
@balajismaniam @marquiz 
Hi, 
Please check it.
Thanks.

marquiz:
I don't see how this would work. NFD is run inside a container and irqbalance is a host (container host) level service. And, I don't think systemctl even works inside the container.

BTW, info about hyperthreading is already available throught CPUID

okartau:
I am skeptical about systemctl as well (nothing of systemd runs in container, right?)
Attempt to run systemctl in a nfd container (started interactive with entry:/bin/bash) fails here like this:
Failed to get D-Bus connection: Unknown error -1
Was code in this PR verified in a real NFD environment?

lubinsz:
@marquiz 
@okartau 
Hi all,
Sorry, I forgot to deliver the patch in 'node-feature-discovery-job.json.template'.
In container, we can detect the status of services in host by mounting the volume of '/run/systemd'.
Please use following command to check it:
docker run -it -v /run/systemd:/run/systemd  golang:1.8 /bin/bash

Thanks.

okartau:
confirmed systemd mount,
I can run systemctl in NFD container if -v /run/systemd:/run/systemd is used on docker cmd line

okartau:
CI result (which seems not moving ATM due to some problem) fails for current sate of this PR
because you changed source list in main.go, but did not make similar change in main_test.go.
These two should be in sync, see some recent PR like #96 for example, ShouldResemble line.
Also, needs rebasing onto top of recent merges.

source/sysconf/systemd.go
```diff
@@ -0,0 +1,48 @@
...
+	if len(cmd) < 1 {
+		return "", fmt.Errorf("Need command")
+	}
+
+	output, err := exec.Command(cmd, args...).CombinedOutput()
+
+	return strings.TrimRight(string(output), "\n"), err
+}
+
+func SystemctlStatus(services string) (success bool) {
```
fabiand: `SystemctlStatusIsActive` might be a more appropriate considering the scope of this function.

lubinsz:
@okartau 
Thank you for your review.  So nice.

I have fixed related issues. Please check more.

Thanks.

lubinsz:
@marquiz @balajismaniam @okartau @fabiand 
Any idea ?
Thanks.

lubinsz:
ping @marquiz

marquiz:
@lubinsz could you rebase this?

I was pondering about the name of the feature source and the labels. I start to dislike the ever-growing number of separate feature sources. In #149 I implement a new `os` feature source (with labels `nfd-os-release.VERSION_ID`). Would `os` make sense for this feature, too? I.e. the label could be e.g.
```
nfd-os-systemd.irqbalance = true
```
What do you think? Or would there be some better name for these host os features?

Also, is there need for separate `disabled` and `enabled` flags? Just `irqbalance: true` should be enough(?) If it's present, we know that it's been enabled, otherwise it's disabled.

lubinsz:
@marquiz 
I have had a very long vacation. Sorry for delay.
Got it. I will do it.

marquiz:
> I have had a very long vacation. Sorry for delay.
> Got it. I will do it.

No problem 😄

I'm still a bit undecided about the name of the feature source (`os` vs. `system` vs. something else, see #149)

I also started to ponder the `true`/`false` values. False values might actually be useful in some cases where users want to specifically detect that a feature is absent/disabled. The label may be missing not because a feature is absent/disabled, but, because the detection failed altogether (i.e. it might be true or false, we just don't know). Thus, a `false` value is probably justifiable, if we're certain that we were able to detect it this state correctly (i.e. the feature truly is absent/disabled). However, the documentations (i.e. README) must be changed to reflect this.

fabiand:
> I'm still a bit undecided about the name of the feature source (`os` vs. `system` vs. something else, see #149)

`system` imho.

`os` can be about package management, versions, config files, …
`system` is more along the lines of kernel features, hardware, …

marquiz:
I probably also like `system` better. There is a already a separate `kernel` feature source for kernel-only features, like kernel version and kernel config options (#146).

@kad, @lmdaly any thoughts?

lubinsz:
@marquiz @balajismaniam
Any comments on it?
Thanks.

source/system/systemd.go
```diff
@@ -0,0 +1,45 @@
...
+func runCmd(cmd string, args []string) (string, error) {
+	if len(cmd) < 1 {
+		return "", fmt.Errorf("Need command")
+	}
+
+	output, err := exec.Command(cmd, args...).CombinedOutput()
+	return strings.TrimRight(string(output), "\n"), err
+}
+
+func SystemctlStatusIsActive(services string) (success bool) {
```
marquiz: I think this should return status, and, error. Return error if the command fails

source/system/system.go
```diff
@@ -0,0 +1,41 @@
...
+// Source implements FeatureSource.
+type Source struct{}
+
+// Name returns an identifier string for this feature source.
+func (s Source) Name() string { return "system" }
+
+// Discover returns feature names for system configuration: irqbalance, swap, ksm, memory compaction, iommu...
+func (s Source) Discover() (source.Features, error) {
+	features := source.Features{}
+	active := SystemctlStatusIsActive("irqbalance")
```
marquiz: SystemctlStatusIsActive() should return error if it is not able to detect the status of the service. In case of error, no label should be set  as we don't really know the status, if it's *false* or *true*. This might happen e.g. if the mounts are missing from the container spec.

main.go
```diff
@@ -257,6 +258,7 @@ func configureParameters(sourcesWhiteList []string, labelWhiteListStr string) (e
 	// Configure feature sources.
 	allSources := []source.FeatureSource{
 		cpuid.Source{},
+		system.Source{},
```
marquiz: These should be in alphabetical order

source/system/system.go
```diff
@@ -0,0 +1,41 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package system
+
+import (
+	"github.com/kubernetes-incubator/node-feature-discovery/source"
```
marquiz: Use sigs.k8s.io as the import namespace, now that the project has been migrated into a new location.

source/system/systemd.go
```diff
@@ -0,0 +1,45 @@
...
+
+package system
+
+import (
+	"fmt"
+	"os/exec"
+	"strings"
+)
+
+func runCmd(cmd string, args []string) (string, error) {
```
marquiz: This could possibly be more specific to systemd, runSystemctlCmd(?)

lmdaly:
Agree on system naming convention. Also agreed that an explicit false flag may be useful for the case you mentioned where people want to land on a node where something is disabled.

Also is there any additional drawbacks to adding an additional mount point to the nfd job? As I guess this is the first feature that has needed this.

source/system/system.go
lubinsz: Done.

source/system/systemd.go
lubinsz: Done.

main.go
lubinsz: Done.

source/system/system.go
lubinsz: Done.

source/system/systemd.go
lubinsz: Done.

lubinsz:
@marquiz 
Please check it.
Thanks.

source/system/system.go
marquiz: Please rebase on top of latest master. There, the import paths for other files have been changed, already

source/system/system.go
```diff
@@ -0,0 +1,46 @@
...
+// Source implements FeatureSource.
+type Source struct{}
+
+// Name returns an identifier string for this feature source.
+func (s Source) Name() string { return "system" }
+
+// Discover returns feature names for system configuration: irqbalance, swap, ksm, memory compaction, iommu...
+func (s Source) Discover() (source.Features, error) {
+	features := source.Features{}
+	active, err := SystemctlStatusIsActive("irqbalance")
```
marquiz: As I commented before, I think the list of services to detect should be configurable, allowing the detection of other system services, too. irqbalance can be the default. You can take a look at kernel config detection to see an example.

README.md
```diff
@@ -178,6 +178,12 @@ such as restricting discovered features with the --label-whitelist option._
 | <br>    | minor     | Second component of the kernel version (e.g. '5')
 | <br>    | revision  | Third component of the kernel version (e.g. '6')
 
+### System Features
+
+| Feature name              | Description                                            |
+| :---------------------:   | :----------------------------------------------------: |
+| systemd.irqbalance        | Check the service of irqbalance is active or not
```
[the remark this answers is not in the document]
lubinsz: Done.

source/system/systemd.go
```diff
@@ -0,0 +1,48 @@
...
+	if len(cmd) < 1 {
+		return "", fmt.Errorf("Need command")
+	}
+
+	output, err := exec.Command(cmd, args...).CombinedOutput()
+	return strings.TrimRight(string(output), "\n"), err
+}
+
+func SystemctlStatusIsActive(service string) (bool, error) {
+	cmd := "systemctl"
```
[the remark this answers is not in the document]
lubinsz: Done. Please check it. Thanks.

source/system/system.go
lubinsz: Hmmm...

README.md
```diff
@@ -179,6 +179,12 @@ Kernel config file to use, and, the set of config options to be detected are
 configurable.
 See [configuration options](#configuration-options) for more information.
 
+### System Features
+
+| Feature  | Attribute  | Description                                           |
+| -------  | ---------- | ----------------------------------------------------- |
+| systemd  | irqbalance | Check the service of irqbalance is active or not
```
marquiz: Nit: extra white space between the first and second columns

source/system/system.go
```diff
@@ -0,0 +1,46 @@
...
+package system
+
+import (
+	"log"
+	"os"
+
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+var logger = log.New(os.Stderr, "", log.LstdFlags)
```
marquiz: Similar to what I commented in #195, I think it's better just to use `log.Printf()` instead of initializing a logger with default settings.

source/system/systemd.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"os/exec"
+	"strings"
+)
+
+func runSystemctlCmd(args []string) (string, error) {
+	cmd := "systemctl"
+
+	output, err := exec.Command(cmd, args...).CombinedOutput()
```
marquiz: Nit: Just drop the `cmd` variable.

source/system/system.go
lubinsz: Can we just use this method:
for _, attrib := range []string{"irqbalance", "openswitch"} {
...
Just like the same method in #195  :)

source/system/systemd.go
lubinsz: Done.

source/system/system.go
lubinsz: Done.

README.md
lubinsz: Done.

jeremyeder:
@lubinsz I would like you to check out this project:  https://github.com/redhat-performance/tuned

Especially the cpu-partitioning profile and it's parents.  I think you will like what you see as it handles all of the tuning for the highest performance DPDK.  If you're planning on running DPDK apps on Kubernetes, that's another topic that both Intel and Red Hat are already working together on.  Perhaps ARM folks would like to join so that we can have a cohesive approach (which is what tuned provides) rather than re-solving all known issues in NFD.

zvonkok:
@lubinsz Have you tried the cluster-tuned-operator to implement the needed tuning for the DPDK case? 
Any objections closing this one and #195?

PR #126, opened 2018-05-23
Add RBAC policy

poussa:
* Add rbac.yaml which contains needed rules for NFD to work on a RBAC
  enabled cluster
* Add lines to templates to use the RBAC policy
* Update README

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,8 +178,11 @@ Feature discovery is preferably run as a Kubernetes DaemonSet. There is an
 example spec that can be used as a template, or, as is when just trying out the
 service:
 ```
+kubectl create -f rbac.yaml
 kubectl create -f node-feature-discovery-daemonset.json.template
 ```
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -175,6 +175,7 @@ Feature discovery is preferably run as a Kubernetes DaemonSet. There is an
 example spec that can be used as a template, or, as is when just trying out the
 service:
 ```
+kubectl create -f rbac.yaml
```
marquiz: It would probably be good to add a bit more explanation. E.g. something like
If you have RBAC authorization enabled (as is the default e.g. with clusters initialized with kubeadm) you need to configure the appropriate ClusterRoles, ClusterRoleBindings and a ServiceAccount in order for NFD to create node labels. The provided templates will configure these for you.

poussa:
Fixes #92

mythi:
I hit this too and the PR helped. +1, verified.

poussa:
@marquiz name changed from `nfd` to `node-feature-discovery`

PR #128, opened 2018-05-25
Add intel-cmt-cat a git-subtree

marquiz:
The purpose of this change is to make management of the sources and
building NFD easier. In particular, it makes sure that you always build
in docker the same source code you have locally. And, it makes way for
future improvements for building NFD in a similar manner, independent
whether you build it locally or under docker. Local builds greatly
speed-up the development cycle of NFD. Lastly, this change slightly
simplifies the Dockerfile.

Future versions of intel-cmt-cat can be pulled into the repo with
git-subtree command, e.g.
$ git subtree pull --squash --prefix intel-cmt-cat \
    https://github.com/intel/intel-cmt-cat.git master


marquiz:
@balajismaniam ping!

marquiz:
Dependency to intel-cmt-cat was dropped in #189

PR #134, opened 2018-06-12
Template fixes

marquiz:


marquiz:
Ping @balajismaniam 

Without this fix the selinux feature source is not working in DaemonSet

node-feature-discovery-job.json.template
```diff
@@ -34,12 +34,6 @@
             "image": "quay.io/kubernetes_incubator/node-feature-discovery:v0.1.0",
             "name": "node-feature-discovery",
             "args": ["--oneshot"],
-            "ports": [
```
balajismaniam: @marquiz This is probably still required to make sure only one pod from the job lands on a node. Can you please manually e2e test it so that it doesn't break anything even while using the Job. See https://github.com/kubernetes-incubator/node-feature-discovery/pull/49#issuecomment-263951742.

marquiz: > This is probably still required to make sure only one pod from the job lands on a node. Can you please manually e2e test it so that it doesn't break anything even while using the Job.

Hmm, good point. I previously checked this briefly, but, need to double-check this. I dropped this patch from the PR, for now. I'll create a new PR for this after the release.

balajismaniam:
We can wait till the CI passes and then merge.

Issue #125, opened 2018-05-04
Deployment script for NFD


marquiz:
A deployment/initialization script would make NFD deployment easier and hopefully less error-prone, especially for new users.

**Tasks it could handle:**
- DaemonSet deployment
- Job deployment
- RBAC configuration
- Configuration options
  - Common
    - container image which to use
    - k8s namespace
    - feature source whitelist flag (`--sources`)
    - feature label whitelist flag (`--label-whitelist`)
  - DaemonSet
    - sleep interval
- Reset / de-initialization

marquiz:
I already started to write a shell script but then decided to scrap that. Now I'm pondering whether golang and cobra would be the way to go.

marquiz:
Closing this as not to be implemented

PR #135, opened 2018-06-14
Implement nfdadm

marquiz:
Implement a helper tool for deploying NFD (see issue #125). The script is currently cabable of deploying NFD as a DaemonSet or Job, configuring RBAC, creating target namespace and also undeploying NFD.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,7 +25,7 @@ RUN case $(dpkg --print-architecture) in \
 RUN go get github.com/Masterminds/glide
 RUN glide install --strip-vendor
 RUN go install \
-  -ldflags "-s -w -X main.version=$NFD_VERSION" \
+  -ldflags "-s -w -X github.com/kubernetes-incubator/node-feature-discovery/version.version=$NFD_VERSION" \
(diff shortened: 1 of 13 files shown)


balajismaniam:
@marquiz At a high-level, I'm not sure if this PR is required. NFD is deployed as a simple daemonset or a run to completion job. While this kind of a tool might be required for a complicated installation for Kubernetes (e.g., kubeadm), NFD can be deployed with a much more simple script or `kubectl create -f`.

marquiz:
I find it useful as it automatically configures (and deconfigures) RBAC, sets up a namespace and provides an easy way to customize probably the most commonly used settings (like image name). But, it's true that NFD is a simple service and this might be an overkill. It'd be nice to get some comments from others, as well – whether or not this would be something useful and make NFD more approachable.

balajismaniam:
Ok. Still my opinion is that this tool is unnecessary. We can think about helm charts. However, even that might not be required for such a simple project and deployment.

balajismaniam:
Closing this. If there is a necessity, we can use helm charts.

Issue #136, opened 2018-07-05
Detect IOMMU


marquiz:
Detect if IOMMU is present (supported and enabled, by HW and the kernel).

balajismaniam:
@marquiz what is the use-case for this?

marquiz:
AFAIU, detecting correctly configured nodes for certain workloads. But, this is a fair question, I don't know have any concrete use case. @lmdaly: would you be able to elaborate on this?

lmdaly:
That's correct Markus, the workload to be deployed should only deployed where IOMMU is enabled, if not we don't want the workload to land on that node.

marquiz:
Also, DPDK workloads were mentioned in #119

marquiz:
Implemented in #137. Closing this issue

PR #137, opened 2018-07-05
Implement detection of IOMMU (#136)

marquiz:
Add a new feature source, i.e. 'iommu', which detects if an IOMMU is
present and enabled in the kernel. The new node label is
  node.alpha.kubernetes-incubator.io/nfd-iommu-present

Implements #136


marquiz:
I verified this, getting "node.alpha.kubernetes-incubator.io/nfd-iommu-enabled=true" label when iommu was enabled in hw and in kernel

main.go
```diff
@@ -136,7 +137,7 @@ func argsParse(argv []string) (args Args) {
   -h --help                   Show this screen.
   --version                   Output version and exit.
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,rdt,pstate,memory,network,storage,selinux]
+                              [Default: cpuid,iommu,rdt,pstate,memory,network,storage,selinux]
```
kad: this list probably should be sorted alphabetically ?

main.go
```diff
@@ -186,6 +187,7 @@ func configureParameters(sourcesWhiteList []string, labelWhiteListStr string) (e
 	// Configure feature sources.
 	allSources := []source.FeatureSource{
 		cpuid.Source{},
+		iommu.Source{},
```
kad: here is practically same question: is the order of the elements matter ? if not, should it be added in the end then ?

main.go
marquiz: Yeah, I actually did this, but decided to move that into a separate patch. The order here is a leftover from that. There's no PR about that, yet.

main.go
marquiz: Same here

marquiz:
There's now a separate PR (#139) about arranging the feature sources alphabetically

main.go
balajismaniam: In the end, labels are represented as `map[string]string` and maps are unordered in Go. But I agree that we should sort the sources for readability.

source/iommu/iommu.go
```diff
@@ -0,0 +1,46 @@
...
+type Source struct{}
+
+func (s Source) Name() string { return "iommu" }
+
+func (s Source) Discover() ([]string, error) {
+	features := []string{}
+
+	// Check if any iommu devices are available
+	devices, err := ioutil.ReadDir("/sys/class/iommu/")
+	if err == nil {
```
balajismaniam: The idiomatic way in Go is to check `if err != nil` here. It also removes the dangling else.

source/iommu/iommu.go
```diff
@@ -0,0 +1,46 @@
...
+		if len(devices) > 0 {
+			glog.Errorf("Detected %v IOMMU devices", len(devices))
```
balajismaniam: This is not an error. Use `Infof` instead?

source/iommu/iommu.go
```diff
@@ -0,0 +1,46 @@
...
+			features = append(features, "enabled")
+		} else {
+			glog.Errorf("No IOMMU devices detected")
```
balajismaniam: This is not an error. Use Infof instead? May be it is not necessary to log this. It is sufficient if we log when we find a device.

source/iommu/iommu.go
marquiz: Infof are not printed and thus not shown in logs

source/iommu/iommu.go
marquiz: As with the previous case, Infof are not shown at all. It is debatable if this should be logged. There are two kinds of feature sources in nfd, some do not log anything if they fail to detect, some print something in any case. I kind of find it useful to see at least something in the log, to verify that the module was run successfully.

marquiz:
@balajismaniam see the new version

I hope I addressed your concerns
- I rebased on top of #139 so you see that change, too
- changed the error handling

However, I didn't change the `glog.Errorf()` usage as `glog.Infof` messages are not shown at all. I think changing the logging of the feature sources is worth a separate PR. I think glog is not a good choice for NFD because you cannot change the default log level without bringing in all the logging-related command line flags, which is an overkill considering the simple needs of NFD, IMO.

marquiz:
Added explanation to README

source/iommu/iommu.go
```diff
@@ -0,0 +1,48 @@
...
+	features := []string{}
+
+	// Check if any iommu devices are available
+	devices, err := ioutil.ReadDir("/sys/class/iommu/")
+	if err != nil {
+		return nil, fmt.Errorf("Failed to check for IOMMU support: %v", err)
+	}
+
+	if len(devices) > 0 {
+		glog.Errorf("Detected %v IOMMU devices", len(devices))
```
balajismaniam: Ok. But I think this log is not required. Can we remove this?

source/iommu/iommu.go
```diff
@@ -0,0 +1,48 @@
...
+		features = append(features, "enabled")
+	} else {
+		glog.Errorf("No IOMMU devices detected")
```
balajismaniam: Ok. But I think this log is not required. Can we remove this?

marquiz:
> Please rebase and remove the comments. LGTM after that.

@balajismaniam done

marquiz:
@balajismaniam: btw, please do not squash the commits, if possible. I find it useful to preserve the commit messages

Issue #140, opened 2018-07-09
Detect Kernel and OS version


marquiz:
These would help in placement of workloads which are dependent on certain OS or kernel version. 

I proposed implementing these in kubelet ([Issue #65850](https://github.com/kubernetes/kubernetes/issues/65850)), but, that request seems unlikely to be accepted. This feature could be implemented in NFD, instead.

Kernel version can be read from `/proc/sys/kernel/osrelease`. Kernel version would be split into multiple labels, in order to make it more practical to use as labelselectors need exact values, and, no comparison operators (i.e. "greater than" etc.) can be used. Suggested labels, with example values:
```
node.alpha.kubernetes-incubator.io/nfd-kernel-version=4.4.0-112-generic
node.alpha.kubernetes-incubator.io/nfd-kernel-version-x=4
node.alpha.kubernetes-incubator.io/nfd-kernel-version-xy=4.4
node.alpha.kubernetes-incubator.io/nfd-kernel-version-xyz=4.4.0
```

An initial RFD implementation of kernel version detection is available at:
https://github.com/marquiz/node-feature-discovery/tree/feature/kernel-version

OS version would be read from `/etc/os-release`. At least `ID` and `VERSION_ID` should be advertised, I think:
```
node.alpha.kubernetes-incubator.io/nfd-os-release-ID=ubuntu
node.alpha.kubernetes-incubator.io/nfd-os-release-VERSION_ID=16.04
```

An initial RFD implementation of OS version detection is available at:
https://github.com/marquiz/node-feature-discovery/tree/feature/os-release

balajismaniam:
What is the use case for this feature?

jeremyeder:
For example, whether or not the kernel has realtime capabilities.

balajismaniam:
@jeremyeder I think that use case will be satisfied by #138?

jeremyeder:
Interesting - ok. Probably either would work. For RH, we have 'rt' as a string in the kernel uname -r output, so for us it is a boolean, (either RT or not, one version or not).  I can see why #138 makes sense outside of my little ecosystem of kernels :-)

marquiz:
> I think that use case will be satisfied by #138?

There are some cases I could think of:
* some ehancements/fixes that are not visible via kconfig flags
* user-specific custom kernel
* security

But, it is a fair question if the kconfig flags would be enough. I'm not necessarily super happy about these version number labels as they can cause problems because of the limitations of label selectors.

[...]

PR #146, opened 2018-07-17
Discover kconfig flags

marquiz:
Implement kernel feature source and add support for config file

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ the only label value published for features is the string `"true"`._
   "feature.node.kubernetes.io/node-feature-discovery.version": "v0.3.0",
   "feature.node.kubernetes.io/nfd-cpuid-<feature-name>": "true",
   "feature.node.kubernetes.io/nfd-iommu-<feature-name>": "true",
+  "feature.node.kubernetes.io/kernel-config.<option-name>": "true",
   "feature.node.kubernetes.io/nfd-kernel-version.<version component>": "<version number>",
   "feature.node.kubernetes.io/nfd-memory-<feature-name>": "true",
   "feature.node.kubernetes.io/nfd-network-<feature-name>": "true",
@@ -169,12 +170,16 @@ such as restricting discovered features with the --label-whitelist option._
 
 ### Kernel Features
 
-| Feature | Attribute | Description                                            |
-| ------- | --------- | ------------------------------------------------------ |
-| version | full      | Full kernel version as reported by `/proc/sys/kernel/osrelease` (e.g. '4.5.6-7-g123abcde')
-| <br>    | major     | First component of the kernel version (e.g. '4')
-| <br>    | minor     | Second component of the kernel version (e.g. '5')
-| <br>    | revision  | Third component of the kernel version (e.g. '6')
+| Feature | Attribute   | Description                                          |
+| ------- | ----------- | ---------------------------------------------------- |
+| config  | NO_HZ       | Kernel config option is enabled
+| <br>    | NO_HZ_FULL  |
+| <br>    | NO_HZ_IDLE  |
+| <br>    | PREEMPT     |
+| version | full        | Full kernel version as reported by `/proc/sys/kernel/osrelease` (e.g. '4.5.6-7-g123abcde')
+| <br>    | major       | First component of the kernel version (e.g. '4')
+| <br>    | minor       | Second component of the kernel version (e.g. '5')
+| <br>    | revision    | Third component of the kernel version (e.g. '6')
 
 ### Memory Features
 
--- a/README.md
+++ b/README.md
@@ -170,16 +170,17 @@ such as restricting discovered features with the --label-whitelist option._
 
 ### Kernel Features
 
-| Feature | Attribute   | Description                                          |
-| ------- | ----------- | ---------------------------------------------------- |
-| config  | NO_HZ       | Kernel config option is enabled
-| <br>    | NO_HZ_FULL  |
-| <br>    | NO_HZ_IDLE  |
-| <br>    | PREEMPT     |
-| version | full        | Full kernel version as reported by `/proc/sys/kernel/osrelease` (e.g. '4.5.6-7-g123abcde')
-| <br>    | major       | First component of the kernel version (e.g. '4')
-| <br>    | minor       | Second component of the kernel version (e.g. '5')
-| <br>    | revision    | Third component of the kernel version (e.g. '6')
+| Feature | Attribute           | Description                                  |
+| ------- | ------------------- | -------------------------------------------- |
+| config  | &lt;option name&gt; | Kernel config option is enabled (set 'y' or 'm').<br> Default options are `NO_HZ`, `NO_HZ_IDLE`, `NO_HZ_FULL` and `PREEMPT`
+| version | full                | Full kernel version as reported by `/proc/sys/kernel/osrelease` (e.g. '4.5.6-7-g123abcde')
+| <br>    | major               | First component of the kernel version (e.g. '4')
+| <br>    | minor               | Second component of the kernel version (e.g. '5')
+| <br>    | revision            | Third component of the kernel version (e.g. '6')
+
+Kernel config file to use, and, the set of config options to be detected are
+configurable.
+See [configuration options](#configuration-options) for more information.
 
 ### Memory Features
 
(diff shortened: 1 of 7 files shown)


marquiz:
Some modifications: added README sections, changed the default location of the config file, made the (empty) example config available in the Docker image

marquiz:
Based on discussion in #140, I changed the label names so that there now is a dot (instead of dash) used as a separator between the feature name (i.e. `config`) and the attribute (i.e. the config option name). Thus, the labels now look like this:
```
nfd-kernel-config.NO_HZ = true
```

node-feature-discovery.conf.example
```diff
@@ -0,0 +1,7 @@
+#sources:
```
przemek-lal: I don't think there's a need to comment out all options in the example config

main.go
```diff
@@ -91,6 +105,18 @@ func main() {
 	// Parse command-line arguments.
 	args := argsParse(nil)
 
+	// Read the config file
+	err := configParse(args.configFile)
+	if err != nil {
+		stderrLogger.Printf("Failed to read config file: %v", err)
```
przemek-lal: Error of parsing function is a "read error" instead of a "parsing error"? I know it's a nit-picking, but there's also small code duplication issue there in the parsing - see other comments

main.go
```diff
@@ -91,6 +105,18 @@ func main() {
...
+	}
+
+	// Parse config overrides from command line
+	err = yaml.Unmarshal([]byte(args.options), &config)
```
przemek-lal: so you're unmarshalling identically formatted data using the same function twice: here and [here](https://github.com/kubernetes-incubator/node-feature-discovery/pull/146/files#diff-7ddfb3e035b42cd70649cc33393fe32cR224) - maybe there's a way to refactor this to make it more "dry"?

README.md
```diff
@@ -231,6 +251,28 @@ For example, if some node is tainted NoSchedule or fails to start a job for some
...
+specifying the`--config` command line flag. The file is read inside the Docker
+image. Thus, Volumes and VolumeMounts are needed, or, a custom Docker image
+must be built in order to specify your settings via a configuration file.
+The (empty) [example config](https://github.com/kubernetes-incubator/node-feature-discovery/blob/master/node-feature-discovery.conf.example)
+is available in the NFD Docker image, which can be used as base in custom-built
+images.
+
+Configuration options can also be specified via the `--options` command line
+flag, in which case no mounts need to be used. The same format as in the config
+file must be used, i.e. YAML or JSON.
```
przemek-lal: Command line argument in yaml format sounds like a bad idea to me in terms of usability and user experience, imagine how confusing would be probably doing something this:
```
/usr/bin/node-feature-discovery --config="sources:\
  kernel:\
    kconfigFile: \"/path/to/kconfig\"\
    configOpts:\
      - \"NO_HZ\"\
      - \"X86\"\
      - \"DMI\"" --another-options="blah-blah" ...
```
especially that one option is to pass this as a yaml, where you have to take care of number of spaces, indentations, etc.
Now imagine something even more crazy: setting this option inside pod specification, where you would end up with yaml config in bash command in yaml config :)

I understand that the purpose of this command line argument is to be able to quickly test a new configuration and it's great to have this options, but I just wanted to point out that there may be some problems on the user side.

main.go
```diff
@@ -129,15 +155,23 @@ func argsParse(argv []string) (args Args) {
...
   Options:
   -h --help                   Show this screen.
   --version                   Output version and exit.
+  --config=<path>             Config file to use.
+                              [Default: /etc/kubernetes/node-feature-discovery/node-feature-discovery.conf]
+  --options=<config>          Specify config options from command line. Config
+                              options are specified in the same format as in the
+                              config file (i.e. json or yaml). These options
+                              will override settings read from the config file.
+                              [Default: ]
```
przemek-lal: same comment as above

README.md
```diff
@@ -231,6 +251,28 @@ For example, if some node is tainted NoSchedule or fails to start a job for some
 
 [![asciicast](https://asciinema.org/a/11wir751y89617oemwnsgli4a.png)](https://asciinema.org/a/11wir751y89617oemwnsgli4a)
 
+### Configuration options
+
+NFD supports a configuration file. The default location is
+`/etc/kubernetes/node-feature-discovery/node-feature-discovery.conf`
+,but, this can be changed by
+specifying the`--config` command line flag. The file is read inside the Docker
+image. Thus, Volumes and VolumeMounts are needed, or, a custom Docker image
```
przemek-lal: Suggesting user to build a custom image doesn't seem like the best idea to me. I think the better alternative would be to do this with ConfigMap - there could be even an example of it provided and mounted by default in the pods deployed by job/daemonset. ConfigMap are mutable during runtime, so it kinda also solves the issue with the `--options` argument below. If I'd like to rerun NFD with new config, I could just:
1. edit ConfigMap
2. restart process (by killing a pod)

and that's it.
The only downside of it, is that the config is shared between all nodes in the cluster, so if that's the problem we can always fall back to using regulat hostPath volume/volumemounts, which is still way more convenient than rebuilding custom Docker image for each node separately, right?

node-feature-discovery.conf.example
marquiz: It needs to be commented out because the example conf is actually copied to be a "real" conf inside the docker image. This probably should be better documented, e.g. stated in the readme.

main.go
marquiz: I'm not sure how to do what you're after. We want to separate the cases where config file parsing and --options parsing fails. That is, we want to parse twice, and have two distinct errors

main.go
marquiz: I get the point, I'll move the error message generation to configParse()

README.md
marquiz: Well, the idea of course is for people to use json. But using ghodss/yaml gives you both json and yaml so...

main.go
marquiz: Same answer

README.md
marquiz: Yeah, I totally agree that this is the preferred and best way. I'll reword the readme to more clearly point this out.

main.go
przemek-lal: It's a minor issue, but maybe something like that (pseudo-code):
```
configFileContent ,err := readConfigFile(path)
//if err
err = parseConfig(configFileContent)
//if err
err = parseConfig([]byte(args.options)
//if err
```
in order to have a single place in the code where you parse the config []byte. And you still handle errors from both sources. Imagine something in the future changes and you have to recreate the config parsing logic from YaML to INI (bizarre turn of events :wink:) - this way you would need to rework only single function instead of two.

README.md
przemek-lal: Yea, if you would decide to go with the ConfigMap as the main option, you wouldn't even need to add the default config file to the docker image (currently commented out). The less stuff in the docker image the better, right? From the user perspective there's not much difference between configMap and hostPath - it's only where the config file is stored - single place in cluster vs on each node. Both has pros and cons, so it depends on the use-case probably.

node-feature-discovery.conf.example
przemek-lal: If that's the case, then it makes sense to have it commented out. But maybe there's no need to have this file added to the docker image in the first place? If we could always load it from volume (hostPath/configMap/whatever) and tell the user to do that if they want to load custom config. Or even give them example how it's done? Just thinking.

README.md
przemek-lal: isn't json as ugly as yaml in this scenario? :wink: but as I said earlier it's always better to have an option. Also, I can't think of any significantly better alternative, except maybe instead of `--config=<yaml_or_json>` split it to `--kernel-config-file=<filepath> --kernel-config-opts=NO_HZ,X86,DMI,...`. Not sure how it's gonna look in the future tho, if more plugins will decide to add their own kernel flags.

marquiz: Yeah, I thought something like this but forgot the idea because it makes everything more complex, it isn't future proof and expressive enough to cover all scenarios. JSON on the command line is not beatiful but I think this is is sometimes a handy way specifying one single config option, for example.

marquiz:
@przemek-lal thanks for your review and comments!

Let's continue discussion on the config/option support in #169 because this same mechanism is actually used in other open PRs (e.g. #168). I pushed a new version there, trying to address all your comments.

marquiz:
@przemek-lal any further comments?

Config support was merged in #169 so there's only the kconfig detection part to review.

PR #147, opened 2018-07-17
Implement cpu feature source

marquiz:
Currently, it only discovers one feature label, i.e. 'logical-cpus',
which basically tells if hyper-threading has been enabled. However, the
label is not named 'hyper-threading' because the setting is not detected
directly, and, the heuristics can be mislead. NFD determines the thread
siblings of each (logical, online) cpu in the system. If any cpu has
thread siblings the feature is set to true. Thus, hyper-threading could
be effectively disabled by putting all sibling cpus offline (even if HT
would be enabled in hardware).

### DIFF
--- a/source/cpu/cpu.go
+++ b/source/cpu/cpu.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2018 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cpu
+
+import (
+	"fmt"
+	"io/ioutil"
+	"path"
+
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+// Implement FeatureSource interface
+type Source struct{}
+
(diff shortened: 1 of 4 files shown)


okartau:
Perhaps we should add notes in README about this new source, and how it's different from existing cpuid source.
Those README additions should here be in same PR?
Code itself looks ok (but I haven't run it)

marquiz:
> Perhaps we should add notes in README about this new source, and how it's different from existing cpuid source

You're right, those should be added. I'll update the PR

marquiz:
I updated the PR, with explanations in README

source/cpu/cpu.go
```diff
@@ -0,0 +1,70 @@
...
+
+// Implement FeatureSource interface
+type Source struct{}
+
+func (s Source) Name() string { return "cpu" }
+
+func (s Source) Discover() ([]string, error) {
+	features := []string{}
+
+	// Check if hyper-threading seems to be enabled
```
lmdaly: Seems? Does this mean that this is not a guaranteed method?

marquiz: There is no "guaranteed" method. Even if HT is enabled in HW, it can be effectively soft-disabled (with the same effect as with hard-disable) by offlining all sibling cpus

marquiz: Maybe the wording in the comment could be changed to be more self-evident(?) We just rely on the topology that the kernel sees. If some core has siblings, we deduce that HT seems to be enabled.

marquiz:
I really dislike the label `logical-cpus` that the patch currently advertises. Any thoughts of a better name? What do you think of e.g. `hardware-multithreading`, `simultaneous-multithreading` or `smt`?

lmdaly:
+1 on underscores for naming convention. Easy to read and distinguished from the . notation

marquiz:
Pushed a new version, simplifying the code and returning an error if any read fails.

Issue #144, opened 2018-07-12
Plugin mechanism for customer-specific labels


marquiz:
**What**: provide a way for customers to write their own feature detectors, creating custom labels.

Originally, I thought, it would be sufficient to implement custom feature sources as user-specific modules under source/ directory in nfd source code. The reasoning behind this idea was that the source modules and their API to "nfd core" are simple, so writing and maintaining a custom plugin would not be too cumbersome. However, this would effectively be forking nfd, requiring maintenance of your own code base and also building and using custom Docker images.

So, I was thinking about providing some relatively simple mechanism to have custom feature detectors, without the need to fork the code or use custom Docker images.

The first, obvious, idea would be to use hooks. E.g. look under `/etc/kubernetes/node-feature-discovery/source.d/` and run all executable binaries from there. The hooks would return the detected features to `stdout` and have logging to `stderr`.

E.g. 
running `/etc/kubernetes/node-feature-discovery/source.d/my-source` would have `stdout` output:
```
FEATURE_1
FEATURE_2
``` 
which, in turn, would cause nfd to create labels
```
node.alpha.kubernetes-incubator.io/nfd-my-source-FEATURE_1=true
node.alpha.kubernetes-incubator.io/nfd-my-source-FEATURE_2=true
```

This could be easily extended to non-binary labels, too

lmdaly:
+1 on a pluggable mechanism for specific features that are not acceptable for upstream NFD.

marquiz:
An initial version implemented in PR #148

MarSik:
The proposed label name seems very limiting. The custom hooks might be used to detect functionality that has been provided using manual labels so far or that integrates with other existing tools and the proposed format would not allow the use of expected label names.

I would rather let the hook scripts return the full label names to make the label independent on the project/script reporting it completely. That way the hooks can even replace a detection method that might not be appropriate for some environments or hosts and still provide the same feature labels.

marquiz:
NFD is supposed to advertise node features, thus, I would restrict the hook to *nfd* labels, and not give control over the label namespace. This also relates to re-labeling which has been discussed in #176. I think it's an important feature of NFD that it is able to remove out-of-date/incorrect labels. NFD wouldn't be able to remove any old labels if the hooks had full control over the label name.

Replacing or extending existing feature sources is probably a valid use case, though. So, dropping the feature source part, i.e. `local`, from the label name in the current implementation makes sense.

[...]

PR #148, opened 2018-07-18
Support feature detector hooks

marquiz:
Implement a new feature source named 'local' whose only purpose is to
run feature source hooks found under
/etc/kubernetes/node-feature-discovery/source.d/ It tries to execute all
files found under the directory in alphabetical order.

This provides a mechanism for users to implement custom feature sources
in a pluggable way, without modifying nfd source code or Docker images.

The hooks are supposed to print all discovered features in stdout, one feature per line. Normally, the name of the feature label is composed of the hook and feature names, separated by a dash. However, if the feature name (that the hook prints) starts with a slash (`/`) it is used as is in the label name.

The value of the label is either `true` (for binary labels) or `<value>`        
(for non-binary labels).                                                        

Consider an example where hook `/etc/kubernetes/node-feature-discovery/source.d/my-source` has the following `stdout` output:
```
MY_FEATURE_1
MY_FEATURE_2=myvalue
/override_source-OVERRIDE_BOOL
/override_source-OVERRIDE_VALUE=123
```
This will translate into the following node labels:
```
feature.node.kubernetes.io/my-source-MY_FEATURE_1=true
feature.node.kubernetes.io/my-source-MY_FEATURE_2=myvalue
feature.node.kubernetes.io/override_source-OVERRIDE_BOOL=true
feature.node.kubernetes.io/override_source-OVERRIDE_VALUE=123
```

Stderr from the hooks is propagated to nfd log.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ node-feature-discovery.
                               will override settings read from the config file.
                               [Default: ]
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,iommu,kernel,memory,network,pci,pstate,rdt,selinux,storage]
+                              [Default: cpuid,iommu,kernel,local,memory,network,pci,pstate,rdt,selinux,storage]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
@@ -80,6 +80,7 @@ The current set of feature sources are the following:
 - [CPUID][cpuid] for x86/Arm64 CPU details
 - IOMMU
 - Kernel
+- Local (user-specific features)
 - Memory
 - Network
 - Pstate ([Intel P-State driver][intel-pstate])
@@ -122,7 +123,8 @@ the only label value published for features is the string `"true"`._
   "feature.node.kubernetes.io/nfd-pstate-<feature-name>": "true",
   "feature.node.kubernetes.io/nfd-rdt-<feature-name>": "true",
   "feature.node.kubernetes.io/nfd-selinux-<feature-name>": "true",
-  "feature.node.kubernetes.io/nfd-storage-<feature-name>": "true"
+  "feature.node.kubernetes.io/nfd-storage-<feature-name>": "true",
+  "feature.node.kubernetes.io/<hook name>-<feature name>": "true"
 }
 ```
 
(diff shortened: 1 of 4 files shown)


okartau:
Perhaps not very serious threat, but every time previously statically built code creates a method to run something which is not known at build time, like dynamic hooks proposed here, we should at least make an attempt to analyze and document the security implications.
In the documentation it needs to be clearly stated that such feature creates a way to run whatever is in hooks code, and responsibility to ensure security is thus extended to maintainer(s) of those hooks, as none of hooks code will be reviewed, checked in the NFD project

marquiz:
I added a disclaimer about hook security.

fromanirh:
Hey, I'm interested in this PR as well, any update?
Thanks for the great work!

marquiz:
> Hey, I'm interested in this PR as well, any update?

Good to hear that.

Comments/review would be valuable.

What is still clearly missing from the implementation is a validator for the label names.

marquiz:
Added label name validation (I also created separate PR #171 for that)

marquiz:
Removed RFC status from the title.

There changes in the label  naming scheme adopted in #168 (already merged), #146, #147, #149, #157 and #173: using dot to separate attributes within one feature, and, using underscores (instead of dashes) inside feature names. I started to think for the label name pattern to use here. The currently used `nfd-local-<hook name>-<feature name>` does not quite adhere to these. We could use e.g. dot or underscore between the hook and feature names in the label
- `nfd-local-<hook name>.<feature name>`
- `nfd-local-<hook name>_<feature name>`

Of course, the `local` source could be seen an exception and just use the current scheme.

ksimon1:
Hi, great work on this feature. What should be done to get this merged?

marquiz:
> Hi, great work on this feature. What should be done to get this merged?

Code review would be a good start 😊

marquiz:
Thanks for the review!

I'm holding back this PR until NFD has been migrated to the new repo under kubernetes-sigs (kubernetes/org#172), and, the PR for changing the label namespaces (#176) has been merged.

The reason is that there are some changes planned to the implementation of the hook support that will require the namespace change (see the discussion in #144). Basically, the idea is to give the hooks more control over the label names.

marquiz:
@MarSik @ksimon1, I now pushed an updated version of the PR.

The hooks are now able to fully control the label name (namespace cannot be overridden, it's always `feature.node.kubernetes.io`) by using `/` prefix.

ksimon1:
@marquiz Hi, thank you for this plugin. It works as expected :)
cpus

marquiz:
@MarSik, @ksimon1 , thank you for the review and verification

PR #149, opened 2018-07-20
OS version detection

marquiz:
Implement new 'system' feature source. It now detects OS release                
information from the os-release file, assumed to be available at                
/host-etc/os-release. It currently creates two labels (assuming that the        
corresponding fields are found in the os-release file), with example            
values:                                                                         
  feature.node.kubernetes.io/system-os_release.ID=opensuse                      
  feature.node.kubernetes.io/system-os_release.VERSION_ID=42.3                  
                                                                                
Also, update the template spec to mount /etc/os-release file from the           
host inside the container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ node-feature-discovery.
                               will override settings read from the config file.
                               [Default: ]
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpuid,iommu,kernel,local,memory,network,pci,pstate,rdt,selinux,storage]
+                              [Default: cpuid,iommu,kernel,local,memory,network,pci,pstate,rdt,selinux,storage,system]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
@@ -85,6 +85,7 @@ The current set of feature sources are the following:
 - RDT ([Intel Resource Director Technology][intel-rdt])
 - Selinux
 - Storage
+- System
 
 ### Feature labels
 
@@ -120,6 +121,7 @@ the only label value published for features is the string `"true"`._
   "feature.node.kubernetes.io/rdt-<feature-name>": "true",
   "feature.node.kubernetes.io/selinux-<feature-name>": "true",
   "feature.node.kubernetes.io/storage-<feature-name>": "true",
+  "feature.node.kubernetes.io/system-<feature name>": "<feature value>",
   "feature.node.kubernetes.io/<hook name>-<feature name>": "<feature value>"
 }
 ```
@@ -289,6 +291,13 @@ for more information on NFD config.
 | Feature name       | Description                                                                         |
 | :--------------:   | :---------------------------------------------------------------------------------: |
 | nonrotationaldisk  | Non-rotational disk, like SSD, is present in the node
+
+### System Features
+
+| Feature     | Attribute  | Description                                       |
+| ----------- | ---------- | --------------------------------------------------|
+| os_release  | ID         | Operating system identifier
+| <br>        | VERSION_ID | Operating system version identifier
 
 ## Getting started
 ### System requirements
--- a/main.go
+++ b/main.go
@@ -12,6 +12,10 @@ import (
 
 	"github.com/docopt/docopt-go"
 	"github.com/ghodss/yaml"
+	api "k8s.io/api/core/v1"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	k8sclient "k8s.io/client-go/kubernetes"
+	restclient "k8s.io/client-go/rest"
 	"sigs.k8s.io/node-feature-discovery/source"
 	"sigs.k8s.io/node-feature-discovery/source/cpuid"
 	"sigs.k8s.io/node-feature-discovery/source/fake"
@@ -26,10 +30,7 @@ import (
 	"sigs.k8s.io/node-feature-discovery/source/rdt"
 	"sigs.k8s.io/node-feature-discovery/source/selinux"
 	"sigs.k8s.io/node-feature-discovery/source/storage"
-	api "k8s.io/api/core/v1"
-	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	k8sclient "k8s.io/client-go/kubernetes"
-	restclient "k8s.io/client-go/rest"
+	"sigs.k8s.io/node-feature-discovery/source/system"
 )
 
 const (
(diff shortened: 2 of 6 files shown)


okartau:
Is it really justified to make os-release a main-level feature source? On such path we will end up with big number of main-level sources, and unclear structure.
I would imagine labels of os-release being part of some more generic main-level feature source.

okartau:
Note that this is one of these features which work correctly only when host path is mounted into Docker.
There is "oneshot trial" recommendation in README (to try to run just docker image directly) which may miss all/some of such mounts, if not specified explicitly on docker command line. 
Should we maintain the set of mounts in docker command example? This is not robust approach as it risks to become out-of-sync. Perhaps better is to point out the the need to specify some -v options, with short example?

marquiz:
> Should we maintain the set of mounts in docker command example? This is not robust approach as it risks to become out-of-sync. Perhaps better is to point out the the need to specify some -v options, with short example?

I think we should not maintain these mounts in the docker command line example. But, it is worth mentioning that all features are not correctly detected unless the correct mount options are used (with a pointer to the template spec). This would probably be worth of a separate PR as there are already existing mounts.

marquiz:
> Is it really justified to make os-release a main-level feature source? On such path we will end up with big number of main-level sources, and unclear structure.
> I would imagine labels of os-release being part of some more generic main-level feature source.

Good point. My original intention was actually to make it *os* feature source, like *kernel* in the other PRs. I then hastily wrote the implementation and added *release* to the feature source name, although I shouldn't have.

marquiz:
Updated PR. I changed the feature source name from `os-release` to simply `os`, which, I think, would be general (or high level) enough for a feature source.

Comments?

marquiz:
> Note that this is one of these features which work correctly only when host path is mounted into Docker.
> There is "oneshot trial" recommendation in README (to try to run just docker image directly) which may miss all/some of such mounts, if not specified explicitly on docker command line.

A created a separate PR (#156) to address this.

pmossakx:
Verified #149 and #157 on generic CentOS:
```
[root@pmossakx-centos node-feature-discovery]# git log -1
commit 6ebb52937fc810b327e3ed3844848483eb97d9fa
Merge: 05a8162 9994063
Author: Piotr Mossakowski <piotrx.mossakowski@intel.com>
Date:   Wed Sep 26 07:30:05 2018 +0000

    Merge remote-tracking branch 'origin/feature/os-release' into feature/kernel-version

[root@pmossakx-centos node-feature-discovery]# uname -r
4.4.156-1.el7.elrepo.x86_64

[root@pmossakx-centos node-feature-discovery]# cat /etc/redhat-release
CentOS Linux release 7.5.1804 (Core)

[root@pmossakx-centos node-feature-discovery]# kubectl get node pmossakx-centos -o json | jq .metadata.labels
{
<output omitted>
  "node.alpha.kubernetes-incubator.io/nfd-kernel-version-full": "4.4.156-1.el7.elrepo.x86_64",
  "node.alpha.kubernetes-incubator.io/nfd-kernel-version-major": "4",
  "node.alpha.kubernetes-incubator.io/nfd-kernel-version-minor": "4",
  "node.alpha.kubernetes-incubator.io/nfd-kernel-version-revision": "156",
  "node.alpha.kubernetes-incubator.io/nfd-os-release-ID": "centos",
  "node.alpha.kubernetes-incubator.io/nfd-os-release-VERSION_ID": "7",
  "node.alpha.kubernetes-incubator.io/node-feature-discovery.version": "v0.2.0-10-g6ebb529"
}
```
and custom ubuntu kernel:
```
root@pmossakx-ubuntu:~/node-feature-discovery# git log -1
commit 27e6443e7eff65c8f275fb94bd87c0d3da30e9f4
Merge: 26cae36 089ca10
Author: Piotr Mossakowski <piotrx.mossakowski@intel.com>
Date:   Thu Sep 27 10:45:06 2018 +0100

    Merge remote-tracking branch 'origin/feature/os-release' into feature/kernel-version

root@pmossakx-ubuntu:~/node-feature-discovery# uname -r
4.18.0-pmossakx

root@pmossakx-ubuntu:~/node-feature-discovery# lsb_release -a
Distributor ID: Ubuntu
Description:    Ubuntu 18.04 LTS
Release:        18.04
Codename:       bionic

root@pmossakx-ubuntu:~/node-feature-discovery# kubectl get node pmossakx-ubuntu -o json | jq .metadata.labels
{
<output omitted>
  "node.alpha.kubernetes-incubator.io/nfd-kernel-version-full": "4.18.0-pmossakx",
  "node.alpha.kubernetes-incubator.io/nfd-kernel-version-major": "4",
  "node.alpha.kubernetes-incubator.io/nfd-kernel-version-minor": "18",
  "node.alpha.kubernetes-incubator.io/nfd-kernel-version-revision": "0",
  "node.alpha.kubernetes-incubator.io/nfd-os-release-ID": "ubuntu",
  "node.alpha.kubernetes-incubator.io/nfd-os-release-VERSION_ID": "18.04",
  "node.alpha.kubernetes-incubator.io/node-feature-discovery.version": "v0.2.0-15-g27e6443-dirty"
}
```

marquiz:
Based on discussion in #140, I changed the label names so that there now is a dot (instead of dash) between `release` and the name of the os release parameter, i.e. we would now have:
```
nfd-os-release.ID=opensuse
nfd-os-release.VERSION_ID=42.3
```

lmdaly:
LGTM! New naming convention +1

marquiz:
I'm still pondering (or bikeshedding) the name of feature source. Is `os` good or would something else like `system` be more descriptive? I'm thinking the future where we might add other host-os related labels (like #119).

lmdaly:
+1 for system

marquiz:
> +1 for system

Agree, I made the change and pushed an update.

Please review, basically only the naming has changed since the last version (in addition to rebasing on latest master)

README.md
```diff
@@ -121,6 +122,7 @@ the only label value published for features is the string `"true"`._
   "feature.node.kubernetes.io/selinux-<feature-name>": "true",
   "feature.node.kubernetes.io/storage-<feature-name>": "true",
   "feature.node.kubernetes.io/<hook name>-<feature name>": "<feature value>"
```
lmdaly: Should there a comma after  feature.node.kubernetes.io/< hook name >-< feature name >: < feature value > before the system-< feature-name> line?

marquiz: Correct, well spotted! I had pasted the new line in wrong, as the "hook" line should be the last on the list. Fixed now

zvonkok:
We have a POC where different containers are run depending on the major rhel version. 
Would it make sense to add this additional label to the os-release feature? 

In the meantime I have added a hook that extracts the major version: 
```
      feature.node.kubernetes.io/system-os_release.ID=rhel
      feature.node.kubernetes.io/system-os_release.MAJOR_VERSION=7
      feature.node.kubernetes.io/system-os_release.VERSION_ID=7.6
```
If we add a MINOR_VERSION too, users would have finer granularity on steering workkloads
to specific os-release versions, where there are strict restrictions on the versions running. 

With the VERSION_ID we have to be too explicit in versioning selection (With MAJOR_VERSION coarse grained selection and additionally with MINOR_VERSION fine grained selection of nodes).

@marquiz WDYT? 

/cc @jeremyeder

marquiz:
> We have a POC where different containers are run depending on the major rhel version.
> Would it make sense to add this additional label to the os-release feature?

@zvonkok, could you open a separate issue about this?

zvonkok:
@marquiz Done, have a look at #210

PR #150, opened 2018-07-20
Simplify logging

marquiz:
Simplify logging of feature sources. Drop glog which was used in some feature
sources as it was too complicated for the simple logging needs of
NFD.

### DIFF
--- a/source/kernel/kernel.go
+++ b/source/kernel/kernel.go
@@ -34,8 +34,6 @@ type NFDConfig struct {
 	ConfigOpts  []string `json:"configOpts,omitempty"`
 }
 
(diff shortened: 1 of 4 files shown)


okartau:
In what measurable way is use of glog too complicated? Does it make NFD big, slow, less readable etc?
Note that at source level, current change makes code bigger.

marquiz:
> In what measurable way is use of glog too complicated?

It's *proper* use is too complicated, IMO. With the current setup, if you want anything to show in the k8s logs you need to use Errorf which is rather unintuitive if you just would want print out info level messages. Also, messages from different sources (log/glog) look different (e.g. main.go does not use glog).

PR #131, opened 2018-05-28
Utilize multi-stage build to reduce docker image size

marquiz:
This reduces the image basically to size of the debian stretch minimal base image. I.e. size drops from ~1.2GB to ~75MB. Targeted against #129

Issue #129, opened 2018-05-28
Container image size is huge


mythi:
With the build instructions given, the resulting docker container becomes 1.2GB in size. This is huge.

Improve the ```Dockerfile``` to only ship the resulting ```node-feature-discovery``` and whatever else may be needed and drop the build environment parts.

okartau:
not sure is this same or different issue, but docker container also gets bigger when build is repeated. I have seen 1.2 GB, 2.5 GB, 4.3GB, 6.3 GB images, (possibly just missed 3 and 5 GB steps) just like there is some approx 1GB increase step happening. Is this caused by some docker layer added when contents change?

marquiz:
Apparently there's a new (to me) "multi-stage builds" feature in Docker. By utilizing that the image size can be easily reduced. I'm writing a patch to do just that

marquiz:
> not sure is this same or different issue, but docker container also gets bigger when build is repeated.

I haven't encountered this problem. It would be interesting see what causes this

okartau:
growing image appears my local issue: I used to save image into file which resided in work dir,
and docker ADD directive adds this dir into container. And as the first saved image was approx 1.2 GB, and I had repeated the cycle, It seemed that newer images became 1G bigger. Docker ADD is risky if you repeat build in unclean directory.

okartau:
#131 is big improvement, but can we delete more of devel-time stuff from image?
The runtime /usr/bin/node-feature-discovery seems not to depend on much libs,
and is the 289MB in /usr/local/go seems all about development?
If it is so, can we just wipe it (or some of it) from final image without affecting runtime ability,
or is there risk to harm go runtime?

marquiz:
> #131 is big improvement, but can we delete more of devel-time stuff from image?

Deleting stuff from the image during docker build is not a solution because all the layers stack up on top of each other. You need to use smaller base. #131 now drops the image size to about 75MB which should be small enough, I think.

I tried using even smaller alpine linux as the base but it isn't possible because the helper binaries are not buildable/runnable there.

marquiz:
PR #154 provides a bit different, more portable solution

marquiz:
Fixed by #131

PR #154, opened 2018-08-15
Use separate builder image to reduce docker image size

marquiz:
Another approach (different from #131) to fixing #129. Not as simple and neat as using multi-stage build, but, this should work on all systems independent of the Docker version used.

Here, we build NFD image in two phases. First, build the required binaries in a separate builder image and then copy these to a much slimmer production image.


marquiz:
The issue this PR was addressing (#129) was fixed in an alternative way, by #131

PR #156, opened 2018-08-17
Mention volume mounts in README

marquiz:
Add a note about required volume mounts when running NFD directly with
docker run.


okartau:
Looks good!

PR #160, opened 2018-08-31
travis: add deploy step to automatically push docker image

marquiz:
Configure .travis.yml to automatically push builds on the master branch
to the container image registry (quay.io by default). This will automatically
make the latest "experimental" version of NFD, built from the tip of the
master branch, available in the upstream container image repository.

IMAGE_REPO_USER and IMAGE_REPO_PASSWORD environment variables must be defined in
the travis repositorys settings in order for the deployment step to be
triggered.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 .FORCE:
 
 IMAGE_BUILD_CMD := docker build
+IMAGE_BUILD_EXTRA_OPTS :=
 IMAGE_PUSH_CMD := docker push
 
 VERSION := $(shell git describe --tags --dirty --always)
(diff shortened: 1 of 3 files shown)


marquiz:
Rebased. Renamed DOCKER_* configuration variables to IMAGE_REPO_*

marquiz:
Our travis-ci and quay.io projects should now be ready for this.

marquiz:
There still seems to be some problems with the travis
/hold

marquiz:
Should be good to go, now

/hold cancel

Makefile
```diff
@@ -24,3 +27,6 @@ mock:
 
 test:
 	go test ./cmd/... ./pkg/...
+
+push:
```
Ethyling: Should the `push` target depend on `image`?

marquiz: I probably wouldn't add image as a dependency, because then `make push` would re-build the image every time.

PR #161, opened 2018-09-03
Use pod anti-affinity in the worker Job template spec

marquiz:
Use pod anti-affinity to explicitly indicate that pods should be
scheduled on separate nodes. And, remove the obscure and non-obvious
ports definition whose only purpose was to reach the same effect.

### DIFF
--- a/nfd-worker-job.yaml.template
+++ b/nfd-worker-job.yaml.template
@@ -15,6 +15,16 @@ spec:
     spec:
       hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - topologyKey: kubernetes.io/hostname
+              labelSelector:
+                matchExpressions:
+                  - key: app
+                    operator: In
+                    values:
+                      - node-feature-discovery
       containers:
         - env:
           - name: NODE_NAME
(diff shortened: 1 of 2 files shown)


marquiz:
Rebased on latest master, adapted to the master/worker split

/remove-lifecycle stale

nfd-worker-job.yaml.template
```diff
@@ -14,6 +14,16 @@ spec:
     spec:
       hostNetwork: true
       dnsPolicy: ClusterFirstWithHostNet
+      affinity:
+        podAntiAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            - topologyKey: kubernetes.io/hostname
+              labelSelector:
+              matchExpressions:
```
Ethyling: You need to add one more indent on the `matchExpressions` field

Ethyling: @marquiz I don't know if you missed this one

marquiz: > @marquiz I don't know if you missed this one

Yup, seems I did. Thanks for spotting this! Fixed, now

Ethyling:
@marquiz Sorry I missed that, but you also need to remove the comment in the `label-nodes.sh` script

marquiz:
> @marquiz Sorry I missed that, but you also need to remove the comment in the `label-nodes.sh` script

Np, thanks for spotting this, also! Now fixed. I just removed the comment, instead of replacing/updating it in order to avoid similar "out-of-sync" issues in the future

PR #162, opened 2018-09-05
Include the RBAC configuration in workload template specs

marquiz:
Makes the deployment of default/example configuration even simpler, only
requiring the application of one yaml file. The downside is the that the
rbac configuration is duplicated, found in both the DaemonSet and the
Job template spec. However, the default RBAC rules should be rather stable
and need to be touched very rarely, so, this shouldn't be an issue.

Cherry picks one commit from #108 to convert template specs into yaml


marquiz:
Closing this as RBAC configuration is now included int the nfd-master template spec

PR #163, opened 2018-09-05
README: minor correction of wording

marquiz:


README.md
```diff
@@ -194,8 +194,8 @@ kubectl create -f rbac.yaml
 kubectl create -f node-feature-discovery-daemonset.json.template
 ```
 
-When the job runs, it contacts the Kubernetes API server to add labels
-to the node to advertise hardware features.
+When the NFD pods run, they contact the Kubernetes API server to add labels
+to the nodes to advertise hardware features.
```
lmdaly: Nit: node(s) to indicate 1 or more?

marquiz: Pods/nodes pods/node(s), I don't know if it is better now, but, I added the parenthesis

lmdaly: Oh okay you're right, that makes more sense. Actually better as either pods/nodes or pod(s)/node(s) - pods/node(s) doesn't make it fit to 1 or more.

marquiz: Heh, changed it back how it was. Giving another thought I think it's easiest to read and should be accurate enough (as production clusters prolly usually have more than one node)

PR #164, opened 2018-09-05
README: point to deployment template specs directly from github

marquiz:
User can directly use the (remote) template specs from github. No need
to download any files to do a deployment with the default setup as a
DaemonSet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,7 +435,7 @@ possible security implications.
 Nfd-master runs as a DaemonSet, by default in the master node(s) only. You can
 use the template spec provided to deploy nfd-master:
 ```
-kubectl create -f nfd-master.yaml.template
+kubectl create -f https://raw.githubusercontent.com/kubernetes-sigs/node-feature-discovery/master/nfd-master.yaml.template
 ```
 Nfd-master listens for connections from nfd-worker(s) and connects to the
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -190,8 +190,8 @@ Feature discovery is preferably run as a Kubernetes DaemonSet. There is an
 example spec that can be used as a template, or, as is when just trying out the
 service:
 ```
-kubectl create -f rbac.yaml
-kubectl create -f node-feature-discovery-daemonset.json.template
+kubectl create -f https://raw.githubusercontent.com/kubernetes-incubator/node-feature-discovery/master/rbac.yaml
```
lmdaly: Will the files be downloaded by default when cloning the repo/getting a release?

marquiz: Yes, they are part of the repo and will be downloaded. But, the catch here is that you don't need to clone anything to deploy the latest release

marquiz:
Rebased in order to adjust to the recent master/worker split

Ethyling:
Should we point to the files on master or on the last release? As the docker images published do not match the current YAML files

Ethyling:
Do we still need this PR?

marquiz:
> Do we still need this PR?

No, I think this is outdated, now. Thanks for pointing this out

PR #166, opened 2018-09-12
Configurable image build tool

marquiz:
Make it possible to specify an image build tool other than docker - a
limitation is that the build tool must be compatible with docker files,
of course. This makes it possible to build an NFD image without the
Docker daemon, for example.

The image build command is specified in a makefile variable and can be
overridden from command line, for example:
$ make IMAGE_BUILD_CMD="buildah bud"

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,17 +1,19 @@
 .PHONY: all
 
+IMAGE_BUILD_CMD := docker build
+
 QUAY_DOMAIN_NAME := quay.io
 QUAY_REGISTRY_USER := kubernetes_incubator
 DOCKER_IMAGE_NAME := node-feature-discovery
 
 VERSION := $(shell git describe --tags --dirty --always)
 
-all: docker
+all: image
 
 # To override QUAY_REGISTRY_USER use the -e option as follows:
 # QUAY_REGISTRY_USER=<my-username> make docker -e.
-docker:
-	docker build --build-arg NFD_VERSION=$(VERSION) \
+image:
+	$(IMAGE_BUILD_CMD) --build-arg NFD_VERSION=$(VERSION) \
 		--build-arg http_proxy=$(http_proxy) \
 		--build-arg HTTP_PROXY=$(HTTP_PROXY) \
 		--build-arg https_proxy=$(https_proxy) \
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ Download the source code.
 git clone https://github.com/kubernetes-incubator/node-feature-discovery
 ```
 
-**Build the Docker image:**
+**Build the container image:**
 
 ```
 cd <project-root>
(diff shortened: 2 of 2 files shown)


marquiz:
I was able to verify this on Centos 7 with `sudo make IMAGE_BUILD_CMD="buildah bud"`

However, running as a regular user fails with:
```
$ make IMAGE_BUILD_CMD="buildah bud"
buildah bud --build-arg NFD_VERSION=v0.2.0-6-g6ae5433 \
	--build-arg http_proxy=http://proxy.ir.intel.com:911 \
	--build-arg HTTP_PROXY= \
	--build-arg https_proxy=http://proxy.ir.intel.com:911 \
	--build-arg HTTPS_PROXY= \
	--build-arg no_proxy=localhost,127.0.0.1,.intel.com,intel.com,10.96.0.0/12,192.168.122.11 \
	--build-arg NO_PROXY= \
	-t quay.io/kubernetes_incubator/node-feature-discovery:v0.2.0-6-g6ae5433 ./
Error during unshare(CLONE_NEWUSER): Invalid argument
ERRO[0000] error parsing PID "": strconv.Atoi: parsing "": invalid syntax 
ERRO[0000] (unable to determine exit status)            
make: *** [image] Error 1
```

I guess this is expected, because of outdated shadow-utils or smth(?) Any comments from @jeremyeder @zvonkok

zvonkok:
Which minor release of centos 7 are you using? Kernel-version? 

Per default the max number of user namespaces is zero. 

`echo 10000 > /proc/sys/user/max_user_namespaces`

Then try: 

`unshare -fU -- echo foo`

This is where this error `Error during unshare(CLONE_NEWUSER): Invalid argument` came from.

If you're running an older kernel you may have to activate per kernel cmdline the unsernamespace.

AFAIK newgidmap and newuidmap are still missing in the latest Centos releases. So rootless build
will be working when new shadow-utils with newgidmap and newuidmap appear.

marquiz:
> Per default the max number of user namespaces is zero.
> 
> `echo 10000 > /proc/sys/user/max_user_namespaces`

Oh, that was one issue, thanks for pointing that out.


> Then try:
> 
> `unshare -fU -- echo foo`

This now works as expected


> AFAIK newgidmap and newuidmap are still missing in the latest Centos releases. So rootless build
> will be working when new shadow-utils with newgidmap and newuidmap appear.

Yes, that seems to be the case, because now the build fails with:
```
Error: error running newgidmap: exec: "newgidmap": executable file not found in $PATH: 
ERRO[0000] error running newgidmap: : exec: "newgidmap": executable file not found in $PATH
```

@zvonkok : Any (review) comments to the commit itself?

zvonkok:
@marquiz LGTM that's the approach I suggested on https://github.com/kubernetes-incubator/node-feature-discovery/pull/131#issuecomment-410995238

The root build works for me, tested with current master on rhel7.5.
The rootless build works on F29, Fedora has newgidmap and newuidmap.

Beware that rootless builds do not end up in `/var/lib/containers`. They are in `~/.local/share/containers`. 
The `root` user will not see the images built by an user. 

```
$ buildah images
IMAGE ID             IMAGE NAME                                               CREATED AT             SIZE
0d283eb41a92         docker.io/library/golang:1.8                             Feb 17, 2018 11:54     732 MB
dcfc056413ba         quay.io/kubernetes_incubator/node-feature-discovery:v0.2.0-5-gdb929b8-dirty Sep 19, 2018 15:27     1.25 GB
$ sudo buildah images
```

marquiz:
Cool, thanks.

I verified root build on CentOS 7.5.1804

Issue #167, opened 2018-09-20
Detect PCI devices


marquiz:
A new feature source for detecting the presence of PCI devices.

By default, it would detect Network devices, GPUs and accelerator cards (device classes 0x02, 0x03 and 0x12). The label format would be:
```
nfd-pci-device-<DEV_LABEL>-present
```
And, by default, the `<DEV_LABEL>` part would consist of:
```
<CLASS_ID>-<VENDOR_ID>
```

The feature source would need to be configurable, with the following settings available:
1. device class white list:
   - not all device classes are interesting to the user
   - limit the number of labels
1. label format:
   - users might want to have more granularity
   - the `DEV_LABEL` part of the node label would be configurable

The use case for this would be e.g. deploying device plugins only on the nodes that have compatible hardware.

kad:
it might be good that label will not in form "present", but also exposing quantity of such devices. 
E.g. 2 GPU of Vendor1, 1 GPU of Vendor2.

marquiz:
> @kad hmm, qty should be handled by device plugins.

And, the labels would be simpler to use, by just utilising nodeSelector. It requires the label value to be specified so numerical values make it virtually unusable.

marquiz:
I added network controllers to the default list of device classes to detect

zvonkok:
Yes, as a first step only discovering `VENDOR_ID=0x10DE` (NVIDIA) `CLASS_ID=0x0302` 3D display controller where `0x03` is a Display Controller and `0x03` + `02 is a 3D display controller.

I am using the following script [find_nvidia_display_adapter.sh](https://github.com/zvonkok/node-feature-discovery/blob/5781cf7fdffcb72b77304880efa6c2ddd5c895b9/source/gpu/find_nvidia_display_adapter.sh#L1) , that can be easily generalized for any VENDOR_ID and CLASS_ID.

marquiz:
I implemented a more generic solution, please see #168.

> I am not using any Linux commands like `lshw, lspci` because there are e.g Containerized Linux Distributions that do not ship these tools and only a shell is definitely available.

I disregarded those for the same reaseons.


> AFAIU a label for a 3D display controller from NV would look like:
> 
> `nfd-pci-device-0302-10DE-GPU-present` ?
> 
> where users in some kind of database could define the complete DEV_LABEL or only
> the last part of it?

[...]

PR #168, opened 2018-09-21
Implement PCI feature source

marquiz:
This feature source detects the presence of PCI devices (#167).

This PR also adds support for configurability via a config file and/or the `--options` command line flag.

The PCI feature source has two configuration options:
- device classes to detect
- PCI Id fields to present in the node label name

### DIFF
--- a/source/pci/pci.go
+++ b/source/pci/pci.go
@@ -29,7 +29,13 @@ type pciDeviceInfo map[string]string
 
 var logger = log.New(os.Stderr, "", log.LstdFlags)
 
-var deviceClassWhitelist = []string{"03", "0b40", "12"}
+type NFDConfig struct {
+	DeviceClassWhitelist []string `json:"deviceClassWhitelist,omitempty"`
+}
+
+var Config = NFDConfig{
+	DeviceClassWhitelist: []string{"03", "0b40", "12"},
+}
 
 // Implement FeatureSource interface
 type Source struct{}
@@ -47,8 +53,8 @@ func (s Source) Discover() ([]string, error) {
 	}
 
 	for class, classDevs := range devs {
-		for _, white := range deviceClassWhitelist {
-			if strings.HasPrefix(class, white) {
+		for _, white := range Config.DeviceClassWhitelist {
+			if strings.HasPrefix(class, strings.ToLower(white)) {
 				for _, dev := range classDevs {
 					features[fmt.Sprintf("%s_%s.present", dev["class"], dev["vendor"])] = true
 				}
--- a/source/pci/pci.go
+++ b/source/pci/pci.go
@@ -31,11 +31,15 @@ var logger = log.New(os.Stderr, "", log.LstdFlags)
 
 type NFDConfig struct {
 	DeviceClassWhitelist []string `json:"deviceClassWhitelist,omitempty"`
+	DeviceLabelFields    []string `json:"deviceLabelFields,omitempty"`
 }
 
 var Config = NFDConfig{
 	DeviceClassWhitelist: []string{"03", "0b40", "12"},
+	DeviceLabelFields:    []string{"class", "vendor"},
 }
+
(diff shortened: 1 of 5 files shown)


source/pci/pci.go
```diff
@@ -0,0 +1,144 @@
...
+
+var logger = log.New(os.Stderr, "", log.LstdFlags)
+
+type NFDConfig struct {
+	DeviceClassWhitelist []string `json:"deviceClassWhitelist,omitempty"`
+	DeviceLabelFields    []string `json:"deviceLabelFields,omitempty"`
+}
+
+var Config = NFDConfig{
+	DeviceClassWhitelist: []string{"03", "12"},
```
kad: I would suggest to add as well 0b40

marquiz: Added in the new version of this pr

marquiz:
I submitted a new version, adding class 0x0b40 to the defaults.

In the spirit of #173 and latest version of #147, the label naming (default) pattern was changed to
```
nfd-pci-<class id>_<vendor id>.present
```

marquiz:
Verified this on Fedora 27:
```
$ docker run marquiz-s-2.fi.intel.com/node-feature-discovery:test --oneshot --no-publish --options='{"sources": {"pci": {"deviceLabelFields": ["class", "vendor", "device"] } } }'
2018/10/04 14:14:27 node.alpha.kubernetes-incubator.io/node-feature-discovery.version = v0.3.0-8-g40661048
...
2018/10/04 14:14:27 node.alpha.kubernetes-incubator.io/nfd-pci-030000_102b_0534.present = true
...
```

@zvonkok verified on RHEL7.5, details found in #167

marquiz:
Verified on openSUSE Leap 15 from command line:
```
$ node-feature-discovery --oneshot --no-publish --sources=pci --options='{"sources": {"pci": {"deviceClassWhitelist": ["03", "0403"], "deviceLabelFields": ["class", "vendor", "device"] } } }'
2018/10/05 10:49:48 node.alpha.kubernetes-incubator.io/node-feature-discovery.version = v0.3.0-8-gba84e930
2018/10/05 10:49:48 node.alpha.kubernetes-incubator.io/nfd-pci-0403_8086_8d20.present = true
2018/10/05 10:49:48 node.alpha.kubernetes-incubator.io/nfd-pci-0403_10de_0fba.present = true
2018/10/05 10:49:48 node.alpha.kubernetes-incubator.io/nfd-pci-0300_10de_1401.present = true
```

source/pci/pci.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	// Iterate over all device classes
+	for class, classDevs := range devs {
+		for _, white := range Config.DeviceClassWhitelist {
+			if strings.HasPrefix(class, strings.ToLower(white)) {
+				for _, dev := range classDevs {
+					devLabel := ""
+					for i, attr := range deviceLabelFields {
+						// Take four four first characters, so that the programming
```
kad: "four four" ?

source/pci/pci.go
```diff
@@ -0,0 +1,149 @@
...
+						// interface identifier gets stripped from the raw class code
+						devLabel += dev[attr][0:4]
```
kad: maybe check size before slice operation ? if for some reasons it will be shorter, it will crash with runtime error.

marquiz: All the fields are min 2 byte IDs, presented by kernel as ascii text. I think we can rely on the kernel interface not to change(?)

source/pci/pci.go
marquiz: Woops :smile: Fixed in new revision of this PR

source/pci/pci.go
kad: Hopefully not, but if you expect to read some external data, I'd rather make sure that sizes of what was read is correct. It can be done in `readDevInfo()` then here it can be kept as is.
BTW, there in `readDevInfo()` will be also good to use `TrimPrefix()` for stripping `0x` and then do length check (4-6 symbols?)

PR #169, opened 2018-09-22
Configurability via a config file and command line flags

marquiz:
Adds support for config options that can be specified via a config file and/or the --options command line flag. This PR does not actually use (i.e. add any options) the new feature yet, but only adds the mechanism for adding config options. However, the mechanism is being used in #146 and #168, and, this PR was created to more easily review the new configurability feature separately.


marquiz:
Pushed a new version, based on the comments in #146. Improving documentation (e.g. adding an example how to use ConfigMap), plus some code refactoring and

marquiz:
@przemek-lal any comments on this?

przemek-lal:
Looks OK to me :+1:

PR #180, opened 2018-11-07
Migrate from glide to dep

marquiz:
Glide is not actively developed anymore, and, its documentation
recommends migrating to dep. Also, dep is widely used in other k8s
projects.

Migrating to dep dramatically reduces the size of the populated vendor/
directory from 75MB down to about 15MB.


Ethyling:
I think this MR can be closed as the change has been merged in #209

marquiz:
> I think this MR can be closed as the change has been merged in #209

Yup, thanks!

PR #184, opened 2018-11-19
travis: run in minimal environment and drop sudo

marquiz:
Default language is ruby and we don't need that. The minimal environment
is enough. Also, no su privileges are needed to run the build.


marquiz:
ping @zvonkok, @kad 

/remove-lifecycle stale

PR #181, opened 2018-11-07
Add simple e2e test

marquiz:
This patchset adds simple e2e test which utilises e2e framework of kubernetes. It vendors test/e2e from kubernetes, and, directly copies the e2e wireframe (`test/e2e/e2e*go`) from there, with some modifications aiming at code simplification.

It also incorporates migration from `glide` to `dep` for managing dependencies (PR #180). The reason for this is to keep the disk usage of the vendor directory reasonable.

The test itself is similar to #158. The difference is using kubernetes e2e framework which provides a huge amount of functionality that should make further development of NFD e2e tests a lot easier (than re-writing all the helper functionality in nfd).

Implements #33

PR #186, opened 2018-11-20
e2e test integration in Travis CI

marquiz:
Integrate end-to-end tests (#181) to Travis.

First, it reworks Makefile, adding targets for pushing images to remote registry and running e2e tests in a (remote) test cluster. Then it adds travis integration (in .travis.yml) for running e2e tests in AWS EKS.

Basically, the following environment variables need to be configured in travis in order to make it work:
* `AWS_ACCESS_KEY_ID`
* `AWS_SECRET_ACCESS_KEY`
* `AWS_REGION`
* `KUBECONFIG_AWS`
* `IMAGE_REPO`
* `IMAGE_REPO_PASSWORD`
* `IMAGE_REPO_USER`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,28 @@
-language: minimal
+language: go
+go:
+  - "1.12"
 
 env:
   global:
     # Sanitize git branch name into a valid docker tag name
     - IMAGE_TAG_NAME=$(echo -n $TRAVIS_BRANCH | tr -c "[a-zA-Z0-9._'" "_")
+    - GO111MODULE=on
 
 services:
   - docker
 
 script:
   - make image -e
+  - if [ "$TRAVIS_BRANCH" == master -a -n "$IMAGE_REPO_USER" ]; then
+      set -e;
+      echo "$IMAGE_REPO_PASSWORD" | docker login -u "$IMAGE_REPO_USER" --password-stdin quay.io;
+      make push -e;
+      curl -o $HOME/bin/aws-iam-authenticator --create-dirs https://amazon-eks.s3-us-west-2.amazonaws.com/1.10.3/2018-07-26/bin/linux/amd64/aws-iam-authenticator;
+      chmod a+x $HOME/bin/aws-iam-authenticator;
+      export PATH=$PATH:$HOME/bin;
+      echo "$KUBECONFIG_AWS" > kubeconfig_aws;
+      make e2e-test -e KUBECONFIG=`pwd`/kubeconfig_aws;
+    fi
 
 deploy:
   - on:


marquiz:
Rebased on the latest version of #181

marquiz:
This should work, now, removing RFC status. The whole pipeline is demonstated in my [personal repo](https://github.com/marquiz/node-feature-discovery/tree/feature/e2e-tests-framework-travis-debug)

/hold cancel

.travis.yml
```diff
@@ -9,7 +11,18 @@ services:
   - docker
 
 script:
-  - make image -e
+  - make image -e IMAGE_BUILD_EXTRA_OPTS="--label quay.expires-after=1h"
```
Ethyling: Not sure about this one, but as we build images for e2e tests with `--label quay.expires-after=1h`, in deploy stage we will deploy images with this label and it will expire no? We may need to rebuild images without the label

marquiz: > Not sure about this one, but as we build images for e2e tests with `--label quay.expires-after=1h`, in deploy stage we will deploy images with this label and it will expire no? We may need to rebuild images without the label

Woops, yes this is now badly out-of-sync with the automatic image publishing. Need to fix this (not done, yet)

Thanks for looking at this!

marquiz: I ended up fixing this by just removing `IMAGE_BUILD_EXTRA_OPTS`.

Now, we basically push the `master` image twice, once as part of e2e test and once in the deploy step, but, I don't think that matters as the second push is almost a no-op. I left it like this so that the "official" image deployment would stay the same even if running e2e test was done differently.

Ethyling:
@marquiz Hey, I was looking at this PR again and I was questioning myself about the fact that we push an image named `quay.io/kubernetes_incubator/node-feature-discovery:master` before doing e2e tests. Shouldn't we first build and push an image named `quay.io/kubernetes_incubator/node-feature-discovery:e2e`, run e2e tests against it and then tag it master and push it only if it is valid?

marquiz:
Thanks for reviewing and thinking this through!

> Hey, I was looking at this PR again and I was questioning myself about the fact that we push an image named `quay.io/kubernetes_incubator/node-feature-discovery:master` before doing e2e tests. Shouldn't we first build and push an image named `quay.io/kubernetes_incubator/node-feature-discovery:e2e`, run e2e tests against it and then tag it master and push it only if it is valid?

I was pondering things back in the days, too. Eventually, my thinking was that it is better to push the image, even if the tests fails. Some reasoning:
-  it's a "devel" image that might not be perfect
- it's good to be deterministically always in sync with the actual state of master branch (instead of the master image representing some "random" commit from the master branch)
- puts real pressure for fixing things fast in case e2e-tests on master fail

WDYT? @zvonkok, @kad ?

Ethyling:
Ok, LGTM. It could be nice to have another review before merging this

marquiz:
/hold

Need fixes after switching to go modules

marquiz:
Now fixed. E2E in action is demonstrated in my [fork repo](https://github.com/marquiz/node-feature-discovery/tree/feature/e2e-tests-framework-travis-debug), see
https://travis-ci.org/marquiz/node-feature-discovery/builds/626193175 for CI results

/hold cancel

PR #188, opened 2018-11-23
Makefile: drop docker build-args for https(s) proxy

marquiz:
Docker v17.07 and later supports configuring proxy servers via the
docker client configuration (https://docs.docker.com/network/proxy/).

This is better than using --build-args for passing the proxy settings to
the build environment. Previously, we could end up with empty variables
values which could cause the build to fail. E.g. if you had
http_proxy=<myproxy> defined but HTTP_PROXY unset in the host
environment, you ended up with http_proxy=<myproxy> and HTTP_PROXY=""
(i.e. empty value) inside the build which caused problems in some cases.
In addition, this makes builds via make and directly with docker more
similar.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,10 +14,4 @@ all: image
 # QUAY_REGISTRY_USER=<my-username> make docker -e.
 image:
 	$(IMAGE_BUILD_CMD) --build-arg NFD_VERSION=$(VERSION) \
-		--build-arg http_proxy=$(http_proxy) \
-		--build-arg HTTP_PROXY=$(HTTP_PROXY) \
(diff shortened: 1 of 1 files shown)


mythi:
@marquiz With Docker >= 17.07 the proxies can be configured in the client config.json so that build-args for them can be completely dropped. https://docs.docker.com/network/proxy/

marquiz:
> With Docker >= 17.07 the proxies can be configured in the client config.json so that build-args for them can be completely dropped. https://docs.docker.com/network/proxy/

Thanks for pointing that out. I changed the PR completely, now dropping the --build-args

PR #189, opened 2018-11-23
Implement RDT detection in go

marquiz:
Get rid of the dependency on intel-cmt-cat library and rdt helper
binaries written in C. Significantly simplifies the build procedure.

Implements minimal support (in assembler) for getting the raw data from
the CPUID instruction. Also, implement a stub so that the code works on
other architectures than amd64, too.


source/rdt/cpuid.go
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rdt
+
+type CpuidRet struct {
+	Eax, Ebx, Ecx, Edx uint32
```
kad: maybe `EAX`, `EBX` and so on ? or gofmt complaining about them?

source/rdt/rdt.go
```diff
@@ -33,52 +30,43 @@ func (s Source) Name() string { return "rdt" }
...
-	if err := cmd.Run(); err != nil {
-		glog.Errorf("support for RDT L2 allocation was not detected: %v", err)
-	} else {
-		// RDT L2 cache allocation detected.
-		features["RDTL2CA"] = true
+	// Read cpuid information
+	leaf07h := Cpuid(0x7, 0)
+	leaf0fh := Cpuid(0xf, 0)
+	leaf10h := Cpuid(0x10, 0)
+	leaf07h_1 := Cpuid(0xf, 0x1)
```
kad: I would rename variables to something more readable.
Constants like 0x7, 0xf and so on, maybe would be better to declare instead of hardcoding?

source/rdt/rdt.go
```diff
@@ -33,52 +30,43 @@ func (s Source) Name() string { return "rdt" }
...
+
+	// Detect RDT monitoring capabilities
+	if leaf07h.Ebx&(1<<12) != 0 {
+		if leaf0fh.Edx&(1<<1) != 0 {
```
kad: bit numbers (here and below) probably also good to declare as constants, instead of hardcoding.

source/rdt/cpuid_stub.go
```diff
@@ -0,0 +1,23 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rdt
+
+func cpuidAsm(eax_arg, ecx_arg uint32) (eax, ebx, ecx, edx uint32) {
+	return
```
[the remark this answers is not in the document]
marquiz: Good comment. I refactored the arch specific parts of the patch. In the new version there cpuid() is only present in amd64 specific code.

source/rdt/cpuid.go
marquiz: Changed

source/rdt/rdt.go
marquiz: Yes, these are ugly, but, I didn't change this as I couldn't come up with a better way to rename them. Leaf 07h et al. are directly from the CPU documentation which does not provide much better naming, just *leaf 07h* or *sub-leaf 0*. For the *main* leafs there are some descriptions like "structured extended feature flags enumeration leaf" but I'm not sure how to make something nice out of that

source/rdt/rdt.go
marquiz: Hard to come up with nice names as the documentation doesn't provide those. I added names for bitmasks in the latest version....

source/rdt/rdt.go
marquiz: I changed the names and used consts in the latest version of this PR

source/rdt/rdt_amd64.go
```diff
@@ -0,0 +1,98 @@
...
+
+func discoverRDT() []string {
+	features := []string{}
+
+	// Read cpuid information
+	extFeatures := cpuid(LEAF_EXT_FEATURE_FLAGS, 0)
+	rdtMonitoring := cpuid(LEAF_RDT_MONITORING, 0)
+	rdtL3Monitoring := cpuid(LEAF_RDT_MONITORING, RDT_MONITORING_SUBLEAF_L3)
+	rdtAllocation := cpuid(LEAF_RDT_ALLOCATION, 0)
+
```
kad: Looks better now. One last concern that I have is separation of Intel/AMD CPUs within amd64 build target. Maybe it should be one more "safety net" to check for vendor id before trying to execute cpuid assembly instruction?

marquiz: The patch replicates the current logic. This should work fine on AMD, too (i.e. not report any RDT features) as those are Intel-specific leafs, reserved/not used in AMD cpus

marquiz:
@kad ping

kad:
ok, if you validated that it works on older/other vendor CPUs, then it is lgtm.

PR #195, opened 2018-12-04
Memory: detect transparent hugepages

lubinsz:
Memory: detect transparent hugepages

This patch adds a new feature: detect "transparent_hugepage" in memory module.
For best performance, before deploying the DPDK pods, 
we need to verify that the system's THP is disabled.
So I think, a new memory feature to detect THP is suitable.

The new node label is
feature.node.kubernetes.io/memory-transparent_hugepage.enabled
feature.node.kubernetes.io/memory-transparent_hugepage.defrag

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,9 +234,11 @@ possible security implications.
 
 ### Memory Features
 
-| Feature name   | Description                                                                         |
-| :------------: | :---------------------------------------------------------------------------------: |
-| numa           | Multiple memory nodes i.e. NUMA architecture detected
+| Feature              | Attribute  | Description                                             |
+| -------------------- | ---------- | ------------------------------------------------------- |
+| numa                 | <br>       | Multiple memory nodes i.e. NUMA architecture detected
+| transparent_hugepage | enabled    | Check Transparent hugepage is enabled or not
+| <br>                 | defrag     | Check the defrag efforts for THP is enabled or not
 
 ### Network Features
 
--- a/source/memory/memory.go
+++ b/source/memory/memory.go
@@ -19,6 +19,7 @@ package memory
 import (
 	"fmt"
 	"io/ioutil"
+	"log"
 	"strings"
 
 	"sigs.k8s.io/node-feature-discovery/source"
@@ -30,6 +31,20 @@ type Source struct{}
 // Name returns an identifier string for this feature source.
 func (s Source) Name() string { return "memory" }
 
+func GetChoice(path string) (string, error) {
+	val, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("failed to read '%s': %v", path, err)
+	}
+
+	for _, choice := range strings.Fields(string(val)) {
+		if len(choice) > 2 && choice[0] == '[' && choice[len(choice)-1] == ']' {
+			return choice[1 : len(choice)-1], nil
+		}
+	}
+	return "", nil
+}
+
 // Discover returns feature names for memory: numa if more than one memory node is present.
 func (s Source) Discover() (source.Features, error) {
 	features := source.Features{}
@@ -38,8 +53,9 @@ func (s Source) Discover() (source.Features, error) {
 	// Multiple nodes is a sign of NUMA
 	bytes, err := ioutil.ReadFile("/sys/devices/system/node/online")
 	if err != nil {
-		return nil, fmt.Errorf("can't read /sys/devices/system/node/online: %s", err.Error())
+		log.Printf("ERROR: can't read /sys/devices/system/node/online: %s", err)
 	}
+
 	// File content is expected to be:
 	//   "0\n" in one-node case
 	//   "0-K\n" in N-node case where K=N-1
@@ -47,6 +63,18 @@ func (s Source) Discover() (source.Features, error) {
 	if strings.TrimSpace(string(bytes)) != "0" {
 		// more than one node means NUMA
 		features["numa"] = true
+	}
+
+	// Check transparent_hugepage
+	for _, attrib := range []string{"enabled", "defrag"} {
+		choice, err := GetChoice(fmt.Sprintf("/host-sys/kernel/mm/transparent_hugepage/%s", attrib))
+
+		if err != nil {
+			log.Printf("ERROR: can't read /host-sys/kernel/mm/transparent_hugepage/%s: %s", attrib, err)
+		} else {
+			features[fmt.Sprintf("transparent_hugepage.%s", attrib)] = choice
+		}
+
 	}
 
 	return features, nil


lubinsz:
@marquiz @balajismaniam 
Any comments on this?
Thanks.

source/memory/memory.go
```diff
@@ -30,6 +31,24 @@ type Source struct{}
...
+var consecutiveSpaces = regexp.MustCompile("[[:space:]]+")
+
+func GetChoice(parameter string) (string, error) {
+	val, err := ioutil.ReadFile(parameter)
+	if err != nil {
+		return "", fmt.Errorf("failed to read '%s': %v", parameter, err)
+	}
+
+	// Split up the choices
+	allChoices := consecutiveSpaces.Split(string(val), -1)
```
marquiz: Shouldn't you just use bytes.Fields() (or strings.Fields()) here?

source/memory/memory.go
```diff
@@ -30,6 +31,24 @@ type Source struct{}
 // Name returns an identifier string for this feature source.
 func (s Source) Name() string { return "memory" }
 
+var consecutiveSpaces = regexp.MustCompile("[[:space:]]+")
+
+func GetChoice(parameter string) (string, error) {
```
marquiz: Add a docstring to the function. Also, arg name 'parameter' is misleading/vague, IMO. Could you use e.g. 'path' or smth?

source/memory/memory.go
```diff
@@ -49,5 +69,15 @@ func (s Source) Discover() (source.Features, error) {
...
 
+	// Check kernel memory compaction
+	choice, err := GetChoice("/sys/kernel/mm/transparent_hugepage/enabled")
+	if err != nil {
+		return nil, fmt.Errorf("can't read /sys/kernel/mm/transparent_hugepage/enabled: %s", err.Error())
+	}
+
+	// File content is supposed to be:
+	// [always] madvise never
+	features["transparent_hugepage"] = choice
```
marquiz: I would use transparent_hugepage.enabled here. There might be need to be other attributes in the future.

source/memory/memory.go
```diff
@@ -49,5 +69,15 @@ func (s Source) Discover() (source.Features, error) {
...
 	}
 
+	// Check kernel memory compaction
+	choice, err := GetChoice("/sys/kernel/mm/transparent_hugepage/enabled")
+	if err != nil {
+		return nil, fmt.Errorf("can't read /sys/kernel/mm/transparent_hugepage/enabled: %s", err.Error())
+	}
+
+	// File content is supposed to be:
+	// [always] madvise never
```
marquiz: Unintended comment??

lubinsz: Done.

source/memory/memory.go
lubinsz: Done.

source/memory/memory.go
lubinsz: Done.

source/memory/memory.go
lubinsz: Done.

README.md
```diff
@@ -233,9 +233,10 @@ possible security implications.
 
 ### Memory Features
 
-| Feature name   | Description                                                                         |
-| :------------: | :---------------------------------------------------------------------------------: |
-| numa           | Multiple memory nodes i.e. NUMA architecture detected
+| Feature name          | Description                                              |
+| -------------------   | -------------------------------------------------------- |
+| numa                  | Multiple memory nodes i.e. NUMA architecture detected
+| transparent_hugepage  | Check Transparent hugepage is enabled or not
```
marquiz: Please list the attributes (enabled, defrag) in the table, too. Similar table as "Network Features" has, for example. "numa" would have no attribute, of course (i.e. the cell will be empty)

source/memory/memory.go
```diff
@@ -30,6 +30,23 @@ type Source struct{}
...
+func GetChoice(path string) (string, error) {
+	val, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "", fmt.Errorf("failed to read '%s': %v", path, err)
+	}
+
+	// Split up the choices
+	allChoices := strings.Fields(string(val))
+
+	for _, choice := range allChoices {
```
marquiz: nit: you don't necessarily need allChoises. Just
for _, choice := range strings.Fields(string(val))
??

source/memory/memory.go
```diff
@@ -49,5 +67,20 @@ func (s Source) Discover() (source.Features, error) {
 		features["numa"] = true
 	}
 
+	// Check transparent_hugepage
+	choice, err := GetChoice("/host-sys/kernel/mm/transparent_hugepage/enabled")
+	if err != nil {
+		return nil, fmt.Errorf("can't read /host-sys/kernel/mm/transparent_hugepage/enabled: %s", err.Error())
```
marquiz: Hmm, I think returning "nil, err" on error is not the best thing to do. Even if detection of THP fails, we could advertise numa status. So, I think you should change all these returns to just log.Printf("ERROR:...") – also the numa case above

README.md
lubinsz: Done.

source/memory/memory.go
lubinsz: Done. Thanks.

source/memory/memory.go
lubinsz: Done.

lubinsz:
@marquiz 
Hi,
Please check it.
Thanks.

README.md
```diff
@@ -234,9 +234,11 @@ possible security implications.
 
 ### Memory Features
 
-| Feature name   | Description                                                                         |
-| :------------: | :---------------------------------------------------------------------------------: |
-| numa           | Multiple memory nodes i.e. NUMA architecture detected
+| Feature | Attribute  | Description                                                                        |
+| ------- | ---------- | ---------------------------------------------------------------------------------- |
```
marquiz: Please align the header columns with the data columns below, so that it "renders" nicely in a text editor, too

source/memory/memory.go
```diff
@@ -19,17 +19,35 @@ package memory
...
 	"fmt"
 	"io/ioutil"
+	"log"
+	"os"
 	"strings"
 
 	"sigs.k8s.io/node-feature-discovery/source"
 )
 
+var logger = log.New(os.Stderr, "", log.LstdFlags)
```
marquiz: No need to initialize a logger. You can just use log.Printf() which utilizes the default logger. I did a similar change for the other sources in #150

source/memory/memory.go
```diff
@@ -49,5 +68,20 @@ func (s Source) Discover() (source.Features, error) {
...
+		features["transparent_hugepage.enabled"] = choice
+	}
+
+	choice, err = GetChoice("/host-sys/kernel/mm/transparent_hugepage/defrag")
+	if err != nil {
+		logger.Printf("ERROR: can't read /host-sys/kernel/mm/transparent_hugepage/defrag: %s", err)
+	} else {
+		features["transparent_hugepage.defrag"] = choice
+	}
+
```
marquiz: nit: I think you could reduce the LOC (and possibly easier to add new attributes) by doing a for loop like
```
for _, attrib := range []string{"enabled", "defrag"} {
...
```

What do you think

marquiz:
> Does this lead us down the path of having every tuning knob in /sys or /proc exposed as a label? Seems unsustainable.

I don't think we're there yet, but, I can see the point. And, I've also thought this, how to make it more flexible and scalable. E.g. by making it possible to more directly specify the directly/easily matchable knobs/properties, like sysfs knobs, kernel modules (and other info under /proc), system services etc (even device nodes?). Either by making it possible directly in the Pod spec, or probably in a more controllable way, making it possible for the admin to define node classes or similar.

> We have some other ideas on how to prove a node is "fit for service" for use-cases like DPDK that we will be discussing at KubeCon, hope to report back to the community with some more sustainable/scalable approaches.

I'd love to join the discussion, but, unfortunately I'm not able to participate KubeCon 😞

README.md
lubinsz: Done.

source/memory/memory.go
lubinsz: Done.

source/memory/memory.go
lubinsz: Done.

lubinsz:
@marquiz 
I have updated all. Please check it.
Thanks.

lubinsz:
ping @marquiz

marquiz:
>We have some other ideas on how to prove a node is "fit for service" for use-cases like DPDK that we will be discussing at KubeCon, hope to report back to the community with some more sustainable/scalable approaches.

@jeremyeder did you discuss this subject area in KubeCon, is there something to report?

jeremyeder:
@marquiz unfortunately we did not get around to discussing it, but here you go:  https://github.com/openshift/cluster-node-tuning-operator

This operator is basically a reconciliation loop for tuning options; if there is any config/tuning drift, it will fix it.

marquiz:
@lubinsz please take a look at the tuning operator above. I think it should match your needs. It works the other way around, when compared to NFD. That is, you set labels (node or pod) and the operator correctly tunes the node according to your configuration profiles/rules.

I've quickly tried it on upstream Kubernetes (v1.11) and it worked fine. Just needed minor edit to the operator rbac yaml in order to be able to apply it.

@jeremyeder @jmencak can chime in and supplement/correct my limited understanding.

jeremyeder:
Correct, that's the way RH is headed.

zvonkok:
@lubinsz Have you tried the cluster-tuned-operator to implement the needed tuning for the DPDK case?
Any objections closing this one?

PR #197, opened 2018-12-04
Support non-binary kconfig options in kernel feature source

marquiz:
Discover other than bool or tristate kconfig options, too. For bool and tristate the node label is still binary (i.e. set to "true" if the kconfig option has been enabled). For other kconfig types (e.g. string or int) the value of the label directly corresponds to the value of the kconfig flag, e.g. "32", "elf64-x86-64" etc.

### DIFF
--- a/source/kernel/kernel.go
+++ b/source/kernel/kernel.go
@@ -25,6 +25,7 @@ import (
 	"regexp"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/util/validation"
 	"sigs.k8s.io/node-feature-discovery/source"
 )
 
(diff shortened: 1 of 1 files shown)


source/kernel/kernel.go
```diff
@@ -170,7 +170,9 @@ func parseKconfig() (map[string]bool, error) {
 	for _, line := range lines {
 		if m := re.FindStringSubmatch(string(line)); m != nil {
 			if m[2] == "y" || m[2] == "m" {
-				kconfig[m[1]] = true
+				kconfig[m[1]] = "true"
+			} else {
+				kconfig[m[1]] = strings.Trim(m[2], `"`)
```
kad: Generally ok, however here it might be good to add a bit more validation for value, beside just stripping `"`

marquiz: Validation of label values implemented in #199. I think that should be enough, WYDT?

kad: in theory should be enough. my comment was more about general practice of validating input close to the point of source of information.

marquiz: I wouldn't want to do duplicate regexp matching in multiple places. Checking the length makes sense so I added that.

PR #199, opened 2018-12-05
Add validation for feature label values

marquiz:
Prevents NFD errors e.g. in the case custom hooks happen to output
invalid label values.


kad:
lgtm

PR #203, opened 2018-12-17
Makefile: add multiarch building support

lubinsz:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,21 @@
+# Get qemu-user-static
+ARG IMAGEARCH
+FROM alpine:3.9.2 as qemu
+RUN apk add --no-cache curl
+ARG QEMUVERSION=2.9.1
+ARG QEMUARCH
+
+SHELL ["/bin/ash", "-o", "pipefail", "-c"]
+
+RUN curl -fsSL https://github.com/multiarch/qemu-user-static/releases/download/v${QEMUVERSION}/qemu-${QEMUARCH}-static.tar.gz | tar zxvf - -C /usr/bin
+RUN chmod +x /usr/bin/qemu-*
+
 # Build node feature discovery
-FROM golang:1.10 as builder
+FROM ${IMAGEARCH}golang:1.12.7-stretch as builder
+ENV GO_FLAGS=${GO_FLAGS:-"-tags netgo"}
+
+ARG QEMUARCH
+COPY --from=qemu /usr/bin/qemu-${QEMUARCH}-static /usr/bin/
 
 ADD . /go/src/sigs.k8s.io/node-feature-discovery
 
@@ -7,7 +23,7 @@ WORKDIR /go/src/sigs.k8s.io/node-feature-discovery
 
 ARG NFD_VERSION
 
-RUN go get github.com/golang/dep/cmd/dep
+RUN curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
 RUN dep ensure -v
 RUN go install \
   -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$NFD_VERSION" \
@@ -16,9 +32,8 @@ RUN install -D -m644 nfd-worker.conf.example /etc/kubernetes/node-feature-discov
 
 RUN make test
 
-
 # Create production image for running node feature discovery
-FROM debian:stretch-slim
+FROM ${IMAGEARCH}debian:stretch-slim
 
 # Use more verbose logging of gRPC
 ENV GRPC_GO_LOG_SEVERITY_LEVEL="INFO"
--- a/Makefile
+++ b/Makefile
@@ -18,12 +18,52 @@ KUBECONFIG :=
 yaml_templates := $(wildcard *.yaml.template)
 yaml_instances := $(patsubst %.yaml.template,%.yaml,$(yaml_templates))
 
+ARCH    ?= amd64
+ALL_ARCH = amd64 arm64
+
+IMAGEARCH ?=
+QEMUARCH  ?=
+
+ifneq ($(ARCH),amd64)
+    IMAGE_TAG = $(IMAGE_REPO)-$(ARCH):$(IMAGE_TAG_NAME)
+endif
+
+ifeq ($(ARCH),amd64)
+    IMAGEARCH =
+    QEMUARCH  = x86_64
+else ifeq ($(ARCH),arm)
+    IMAGEARCH = arm32v7/
+    QEMUARCH  = arm
+else ifeq ($(ARCH),arm64)
+    IMAGEARCH = arm64v8/
+    QEMUARCH  = aarch64
+else ifeq ($(ARCH),ppc64le)
+    IMAGEARCH = ppc64le/
+    QEMUARCH  = ppc64le
+else ifeq ($(ARCH),s390x)
+    IMAGEARCH = s390x/
+    QEMUARCH  = s390x
+else
+    $(error unknown arch "$(ARCH)")
+endif
+
 all: image
 
 image: yamls
 	$(IMAGE_BUILD_CMD) --build-arg NFD_VERSION=$(VERSION) \
+		--build-arg IMAGEARCH=$(IMAGEARCH) \
+		--build-arg QEMUARCH=$(QEMUARCH) \
 		-t $(IMAGE_TAG) \
 		$(IMAGE_BUILD_EXTRA_OPTS) ./
+
+pre-cross:
+	docker run --rm --privileged multiarch/qemu-user-static:register --reset
+
+# Building multi-arch docker images
+images-all: pre-cross
+	for arch in $(ALL_ARCH); do \
+		$(MAKE) image ARCH=$$arch; \
+	done
 
 yamls: $(yaml_instances)
 
--- a/README.md
+++ b/README.md
@@ -593,6 +593,19 @@ cd <project-root>
 make
 ```
 
+**Build non-amd64 image on amd64:**<br>
+Optional, for example.
+```
+cd <project-root>
+docker run --rm --privileged multiarch/qemu-user-static:register --reset
+make image ARCH=arm64
+```
+Or, you can build all supported amd64/non-amd64 images together.
+```
+cd <project-root>
+make images-all
+```
+
 **Push the container image:**<br>
 Optional, this example with Docker.
 
--- a/source/cpu/pstate.go
+++ b/source/cpu/pstate.go
@@ -26,8 +26,7 @@ import (
 func turboEnabled() (bool, error) {
 	// On Arm platform, the frequency boost mechanism is software-based.
 	// So skip pstate detection on Arm.
-	switch runtime.GOARCH {
-	case "arm64":
+	if runtime.GOARCH != "amd64" {
 		return false, nil
 	}
 


lubinsz:
@marquiz 
Any comments?

marquiz:
I understand the motivation behind this. But, there are some details I don't fully understand, and, I think the current implementation is a bit too hackish.

Some quick thoughts and questions:
- First, what are the Docker requirements for this? At least my docker CE v17.09 does not support this?
- What about buildah?
- I think there should be more detailed checks to make sure that the build tool (only Docker?) supports multiple platforms.
- If the build steps start to be complex, maybe those should be moved into a separate `build.sh` script or smth(?)
- What is the idea behind using qemu-static? I dislike the blind "wgeting" of these and creating per-arch Dockerfiles, leaving the repo dirty.
- Makefile could have one target per arch, possibly utilizing make's pattern rules?
- Better error handling would be needed. I think it should error out if one build fails.

lubinsz:
@marquiz 
I have reorganized the code according to the compilation method in Kubernetes.
Any comments?
Thanks.

Makefile
```diff
@@ -8,10 +13,73 @@ DOCKER_IMAGE_NAME := node-feature-discovery
...
+	BASEIMAGE1 ?= s390x/golang:1.11-stretch
+	BASEIMAGE2 ?= s390x/debian:stretch-slim
+	QEMUARCH = s390x
+endif
+
+check-docker:
+	@if [ -z $$(which docker) ]; then \
+          echo "Missing \`docker\` client which is required for development"; \
+          exit 2; \
+        fi
```
marquiz: I think this check should be removed. The build tool is supposed to be available, and, it might be something else than docker

Dockerfile
```diff
@@ -1,5 +1,7 @@
 # Build node feature discovery
-FROM golang:1.8 as builder
+FROM BASEIMAGE1 as builder
```
marquiz: How about getting rid of `sed` magic and using build args, instead? I.e. something like:
```
ARG  BASEIMAGE1_ARCH
FROM ${BASEIMAGE1_ARCH}golang:1.11-stretch
```
Then, the Dockerfile could be used directly for the basic use case, without any mangling.

Dockerfile
```diff
@@ -1,5 +1,7 @@
...
+
+CROSS_BUILD_COPY qemu-QEMUARCH-static /usr/bin/
```
marquiz: Could this be commented out, so that no mangling would be needed for amd64. In addition, you can pass `QEMUARCH` as a build arg

Dockerfile
```diff
@@ -32,12 +34,16 @@ RUN go test .
 
 
 # Create production image for running node feature discovery
-FROM debian:stretch-slim
+FROM BASEIMAGE2
+
+CROSS_BUILD_COPY qemu-QEMUARCH-static /usr/bin/
```
marquiz: Again, could this be commented out, and, QEMUARCH turned into a build arg

Dockerfile
```diff
@@ -32,12 +34,16 @@ RUN go test .
 
 
 # Create production image for running node feature discovery
-FROM debian:stretch-slim
+FROM BASEIMAGE2
```
marquiz: Similar to the build stage, would it be possible to use ARG here?

Makefile
```diff
@@ -8,10 +13,73 @@ DOCKER_IMAGE_NAME := node-feature-discovery
...
 
 # To override QUAY_REGISTRY_USER use the -e option as follows:
 # QUAY_REGISTRY_USER=<my-username> make docker -e.
-image:
+image: check-docker
+	cat Dockerfile \
+		| sed "s|BASEIMAGE1|$(BASEIMAGE1)|g" \
+		| sed "s|BASEIMAGE2|$(BASEIMAGE2)|g" \
+		| sed "s|QEMUARCH|$(QEMUARCH)|g" \
+		> Dockerfile.build
```
marquiz: Using build args, you should be able to drop this sed magic

Makefile
```diff
@@ -8,10 +13,73 @@ DOCKER_IMAGE_NAME := node-feature-discovery
...
+
+ifeq ($(ARCH),amd64)
+        # When building "normally" for amd64, remove the whole line, it has no part in the amd64 image
+	sed "/CROSS_BUILD_/d" Dockerfile.build > Dockerfile.build.tmp
```
marquiz: No need for `sed` if `CROSS_BUILD_` is in comments in the Dockerfile

Makefile
```diff
@@ -8,10 +13,73 @@ DOCKER_IMAGE_NAME := node-feature-discovery
...
+else
+        # When cross-building, only the placeholder "CROSS_BUILD_" should be removed
+        # Register /usr/bin/qemu-ARCH-static as the handler for non-x86 binaries in the kernel
+	docker run --rm --privileged multiarch/qemu-user-static:register --reset
+	curl -sSL https://github.com/multiarch/qemu-user-static/releases/download/$(QEMUVERSION)/x86_64_qemu-$(QEMUARCH)-static.tar.gz | tar -xz -C ./ 
+        # Ensure we don't get surprised by umask settings
+	chmod 0755 qemu-$(QEMUARCH)-static
```
marquiz: Would it be possible to run this inside the Dockerfile, or, put the temporary files in a separate build directory in order to minimize the dirtying of the git work dir?

Makefile
```diff
@@ -8,10 +13,73 @@ DOCKER_IMAGE_NAME := node-feature-discovery
...
+	sed "s/CROSS_BUILD_//g" Dockerfile.build > Dockerfile.build.tmp
+endif
+
+ifeq ($(ARCH),amd64)
+	$(IMAGE_BUILD_CMD) --build-arg NFD_VERSION=$(VERSION) \
+		-t $(QUAY_DOMAIN_NAME)/$(QUAY_REGISTRY_USER)/$(DOCKER_IMAGE_NAME):$(VERSION) -f Dockerfile.build.tmp ./
+endif
 	$(IMAGE_BUILD_CMD) --build-arg NFD_VERSION=$(VERSION) \
-		-t $(QUAY_DOMAIN_NAME)/$(QUAY_REGISTRY_USER)/$(DOCKER_IMAGE_NAME):$(VERSION) ./
+		-t $(QUAY_DOMAIN_NAME)/$(QUAY_REGISTRY_USER)/$(DOCKER_IMAGE_NAME)-$(ARCH):$(VERSION) -f Dockerfile.build.tmp ./
```
marquiz: I think you could just mangle `DOCKER_IMAGE_NAME` earlier, where `QEMUARCH` and friends get set, in order to get rid of the if-else-endif here

Makefile
lubinsz: Done.

Dockerfile
lubinsz: Done

Makefile
lubinsz: Done

Makefile
lubinsz: Done.

Makefile
lubinsz: Done.
I added a new layer in Dockerfile to get the qemu file.

Makefile
lubinsz: Done.

Dockerfile
lubinsz: Done.

lubinsz:
@marquiz 
Any comments?

Dockerfile
```diff
@@ -1,12 +1,26 @@
...
+ARG QEMUARCH
+
+SHELL ["/bin/ash", "-o", "pipefail", "-c"]
+
+RUN curl -fsSL https://github.com/multiarch/qemu-user-static/releases/download/v${QEMUVERSION}/qemu-${QEMUARCH}-static.tar.gz | tar zxvf - -C /usr/bin
+RUN chmod +x /usr/bin/qemu-*
+
 # Build node feature discovery
-FROM golang:1.10 as builder
+ARG IMAGEARCH
```
marquiz: I think this ARG needs to be specified at the top of the file, before the first `FROM`

Makefile
```diff
@@ -1,22 +1,59 @@
...
+ALL_ARCH = amd64 arm arm64 ppc64le s390x
+
 .PHONY: all test
 
 IMAGE_BUILD_CMD := docker build
 
 VERSION := $(shell git describe --tags --dirty --always)
 
+IMAGEARCH ?=
+QEMUARCH ?= x86_64
```
marquiz: Even though probably not a big problem in reality, I would put these inside an `ifeq` block, too. Otherwise you will happily get an (x86_64) image when running something like `make ARCH=mips64`.

Makefile
```diff
@@ -1,22 +1,59 @@
...
+endif
+ifeq ($(ARCH),ppc64le)
+        IMAGEARCH = ppc64le/
+        QEMUARCH = ppc64le
+endif
+ifeq ($(ARCH),s390x)
+        IMAGEARCH = s390x/
+        QEMUARCH = s390x
+endif
+
```
marquiz: Should we have a construct something like
```
ifeq
...
else ifeq
...
else ifeq
...
else
$(error unknown arch "$(ARCH)")
endif
```

Makefile
lubinsz: Done.

Makefile
lubinsz: Done.

Makefile
lubinsz: Done.

Dockerfile
lubinsz: Done.

Makefile
lubinsz: Done.

Makefile
```diff
@@ -1,22 +1,61 @@
...
+        IMAGEARCH = arm64v8/
+        QEMUARCH  = aarch64
+else ifeq ($(ARCH),ppc64le)
+        IMAGEARCH = ppc64le/
+        QEMUARCH  = ppc64le
+else ifeq ($(ARCH),s390x)
+        IMAGEARCH = s390x/
+        QEMUARCH  = s390x
+else
+	$(error unknown arch "$(ARCH)")
```
marquiz: Change the tab indentation here to spaces

Makefile
```diff
@@ -1,22 +1,61 @@
...
+endif
+
 IMAGE_REGISTRY := quay.io/kubernetes_incubator
 IMAGE_NAME := node-feature-discovery
 IMAGE_TAG_NAME := $(VERSION)
 IMAGE_REPO := $(IMAGE_REGISTRY)/$(IMAGE_NAME)
 IMAGE_TAG := $(IMAGE_REPO):$(IMAGE_TAG_NAME)
 
+ifneq ($(ARCH),amd64)
+	IMAGE_TAG = $(IMAGE_REPO)-$(ARCH):$(IMAGE_TAG_NAME)
```
marquiz: Tab indentation should be changed to spaces

marquiz:
Oh, forgot to mention that please also update the documentation accordingly.

Makefile
lubinsz: Done.

Makefile
lubinsz: Done.

lubinsz:
> Needs to be rebased on latest master.
> 
> I tried doing an arm64 build (`make ARCH=arm64`) but for some reason go commands are getting stuck. The build gets stuck at:
> 
> ```
> RUN go get github.com/golang/dep/cmd/dep
> ```
> 
> I could work around that by doing `curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh` instead, but, in this case the build get stuck at:
> 
> ```
> RUN go install ...
> ```
> 
> Any idea what the problem might be? Does this work for you as expected?

Hi @marquiz 
I tested it in travis, the result seems OK.
The only one problem is, the speed is not so good.
Please see the following link as reference:
https://travis-ci.org/lubinsz/node-feature-discovery/builds/560796224

My test script is:

script:
  - "travis_wait 30 sleep 1800 &"
  - docker run --rm --privileged multiarch/qemu-user-static:register --reset
  - make image ARCH=arm64

README.md
```diff
@@ -593,6 +593,19 @@ cd <project-root>
 make
 ```
 
+**Build non-amd64 image on amd64:**<br>
+Optional, for example.
+```
+cd <project-root>
+docker run --rm --privileged multiarch/qemu-user-static:register --reset
```
marquiz: Would `make pre-cross` be better?

README.md
```diff
@@ -593,6 +593,19 @@ cd <project-root>
...
+make image ARCH=arm64
+```
+Or, you can build all supported amd64/non-amd64 images together.
```
marquiz: I'd suggest `Or, you can build all supported architectures with one command:` or similar, drop the amd64 here...

Makefile
```diff
@@ -18,13 +18,53 @@ KUBECONFIG :=
...
+		--build-arg IMAGEARCH=$(IMAGEARCH) \
+		--build-arg QEMUARCH=$(QEMUARCH) \
 		-t $(IMAGE_TAG) \
 		$(IMAGE_BUILD_EXTRA_OPTS) ./
 
+pre-cross:
+	docker run --rm --privileged multiarch/qemu-user-static:register --reset
+
+# Building multi-arch docker images
+images-all: pre-cross
```
marquiz: I think `pre-cross` should be listed as a phony target, or, removed from here. With the former, we always would run `pre-cross` when building `images-all` which prolly is ok

README.md
```diff
@@ -593,6 +593,19 @@ cd <project-root>
 make
 ```
 
+**Build non-amd64 image on amd64:**<br>
```
marquiz: NIT: this just looks a bit clumsy heading :smile: 

How about `Multiarch builds` or `Build for target architectures` or smth? If AMD64 is mentioned I think it should be capital case.

marquiz:
Hmm, now my test build fails with:
```
Step 18/24 : RUN go install   -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$NFD_VERSION"   ./cmd/*
 ---> Running in f0534f7d6494
qemu: uncaught target signal 11 (Segmentation fault) - core dumped
Segmentation fault (core dumped)
```

Need to investigate this. I'd like to be able to run the build myself before merging :smile:

marquiz:
I still have issues running `make image` on any other `ARCH` than amd64. I've tried with multiple different hardware and mutiple different OSes and on most systems the build gets stuck in the go install phase. However, I was able to successfully build it on Ubuntu 16.04 with Docker.io v18.09. I haven't had the time to look at the problem closer.

What system and docker version are you using to test this on?

Could you rebase this PR?

/remove-lifecycle stale

Dockerfile
```diff
@@ -1,13 +1,29 @@
+# Get qemu-user-static
+ARG IMAGEARCH
+FROM alpine:3.9.2 as qemu
+RUN apk add --no-cache curl
+ARG QEMUVERSION=2.9.1
```
marquiz: QEMU version needs to be update
```suggestion
ARG QEMUVERSION=4.1.1-1
```

marquiz:
OK, I think I cracked this one. QEMU version was too old and buggy. Used `QEMUVERSION=4.1.1-1` and it started to work.

@lubinsz please rebase and apply that change and I think we can merge this.

marquiz:
Please rebase

/remove-lifecycle rotten

marquiz:
ping @lubinsz

pmundt:
Since this PR was opened, Docker has also changed the way in which it does multiarch builds, now preferring the [buildx extension](https://docs.docker.com/buildx/working-with-buildx/), which amongst other things, takes care of wrapping the backing QEMU instances automatically. I suspect much of this could now be simplified by wrapping `TARGETPLATFORM` in the Dockerfile.

pmundt:
Following up on `buildx` builds, I can confirm that there are no source-level modifications required in order to do multi-arch builds. I've just successfully built and run amd64 and arm64 containers on the current git HEAD as-is:

https://hub.docker.com/repository/docker/adaptant/node-feature-discovery/tags?page=1

The change to the `pstate` file should be split out and applied separately.

pmundt:
Actually, I see that the pstate workaround has already been applied, so it's possible to close this PR.

marquiz:
> Following up on `buildx` builds, I can confirm that there are no source-level modifications required in order to do multi-arch builds. I've just successfully built and run amd64 and arm64 containers on the current git HEAD as-is:

Thanks for the investigation and testing. This could be an option if amd64 builds are not affected, not on older docker versions or other builders like podman (or docker ce).

If you have time and wish to do so you could open another PR utilising buildx(?)

pmundt:
As I said, there are no source-level modifications required for buildx - it just works out of the box. The only thing I had to change was the addition of 32-bit ARM support, which was already merged in #322 . I could make a PR that documents it in the README, but to be honest, there is nothing special about its invocation within the NFD context, so the standard buildx instructions are more than sufficient.

Issue #204, opened 2018-12-18
Only masters should label nodes, not worker nodes themselves


zvonkok:
### Kind: Feature request
Labeling in a k8s cluster is a powerful capability.  If a process/pod misbehaves with labeling capabilities, a worker node can label itself as a master and overtake the cluster. 

Only a master should be able to label nodes in a cluster. 

NFD could use some kind of client/server architecture/paradigm to address this. 
What do you think?

@marquiz @balajismaniam @jeremyeder as discussed on our f2f@kubecon.

marquiz:
Makes sense. Even if it would be NFD-specific server, it would be simple, easy to review, and could be scheduled on master node(s), only.

You mentioned that you had some plans to take labeling capabilities from kubelet itself, too(?) Any details, could the same mechanism be used for NFD??

MarSik:
Well, I would say a daemonset running on a node should be allowed to expose labels for that node as well.

marquiz:
I can see why someone would possibly like to limit the labeling capabilities to a more restricted service, possibly only running on the master.

I played with the idea and implemented one proof-of-concept in #209. It splits NFD into a server and client and uses gRPC for the communication between the parties.

Ethyling:
What is the current state of this issue and the associated PR? Could we hope it to be merged soon?

marquiz:
Basically waiting  for the final review and +1 to indicate that everybody's happy with the PR. I'm basically  looking at @zvonkok and @kad here :wink:

zvonkok:
The master/worker split works, I am not done testing the TLS connection but almost there.

marquiz:
Closing this as #209 was merged

PR #209, opened 2019-01-16
Split NFD into client and server

marquiz:
Refactor NFD into a simple server-client system. Labeling is now done by
a separate 'nfd-master' server. It is a simple service with small
codebase, designed for easy isolation. The feature discovery part is
implemented in a 'nfd-worker' client which sends labeling requests to
nfd-server, thus, requiring no access/permissions to the Kubernetes API
itself.

Client-server communication is implemented by using gRPC. The protocol
currently consists of only one request, i.e. the labeling request.

The spec templates are converted to the new scheme. The nfd-master
server can be deployed using the nfd-master.yaml.template which now also
contains the necessary RBAC configuration. NFD workers can be deployed
by using the nfd-worker-daemonset.yaml.template or
nfd-worker-job.yaml.template (most easily used with the label-nodes.sh
script)

Also, migrates from Glide to dep (#180) in dependency handling. This was necessary because Glide was unable to handle the dependencies of the new codebase.

This PR aims at solving #204.

Testing this should be relatively straightforward:
1. Build a custom image by running `make`, tag and push it: `docker tag <IMAGE_ID> <TARGET_IMAGE> && docker push <TARGET_IMAGE>`
2. Convert the templates to use your custom image: `sed s'!quay.io/kubernetes_incubator/node-feature-discovery:v0.3.0!<TARGET_IMAGE>!' -i nfd-*template`
3. Deploy master and workers: `kubectl apply -f nfd-master.yaml.template && kubectl apply -f nfd-worker-daemonset.yaml.template`

**TODO:**
- [x] Add TLS authentication
- [x] Fix unit tests
- [x] Write unit tests for new functionality
- [x] Update documentation


nfd-worker-daemonset.yaml.template
```diff
@@ -2,29 +2,31 @@ apiVersion: apps/v1
...
               fieldRef:
                 fieldPath: spec.nodeName
           image: quay.io/kubernetes_incubator/node-feature-discovery:v0.3.0
-          name: node-feature-discovery
+          name: nfd-worker
+          command:
+            - "nfd-worker"
           args:
             - "--sleep-interval=60s"
+            - "--server=$(NFD_MASTER_SERVICE_HOST):$(NFD_MASTER_SERVICE_PORT)"
```
zvonkok:
```suggestion
            - "--server=${NFD_MASTER_SERVICE_HOST}:${NFD_MASTER_SERVICE_PORT}"
```
Shouldn't those be curly brackets? I do not think you meant command substitution here. Those are the env vars of the service.

cmd/nfd-worker/main.go
```diff
@@ -34,21 +49,14 @@ import (
...
-	// ProgramName is the canonical name of this discovery program.
-	ProgramName = "node-feature-discovery"
-
-	// Namespace is the prefix for all published labels.
-	labelNs = "feature.node.kubernetes.io/"
-
-	// Namespace is the prefix for all published labels.
-	annotationNs = "nfd.node.kubernetes.io/"
+	// ProgramName is the canonical name of this program
+	ProgramName = "nfd-slave"
```
zvonkok:
```suggestion
	ProgramName = "nfd-worker"
```
This would be consistent with the executable name. 
```
nfd-worker -h
```
prints out nfd-slave

Dockerfile
```diff
@@ -38,6 +38,6 @@ COPY --from=builder /usr/local/bin /usr/local/bin
 COPY --from=builder /usr/local/lib /usr/local/lib
 COPY --from=builder /etc/kubernetes/node-feature-discovery /etc/kubernetes/node-feature-discovery
 RUN ldconfig
-COPY --from=builder /go/bin/node-feature-discovery /usr/bin/node-feature-discovery
+COPY --from=builder /go/bin/* /usr/bin/
```
zvonkok:
```suggestion
COPY --from=builder /go/bin/nfd-* /usr/bin/
```
We need only the nfd-master and nfd-client binaries, other binaries shouldn't be copied into the final image. On my build systems there are other files in this directory.

Dockerfile
```diff
@@ -21,14 +21,14 @@ RUN case $(dpkg --print-architecture) in \
                 ;; \
         esac
 
-RUN go get github.com/Masterminds/glide
-RUN glide install --strip-vendor
+RUN go get github.com/golang/dep/cmd/dep
+RUN dep ensure
```
zvonkok: Nice, this makes life a lot of easier.

cmd/nfd-master/main.go
```diff
@@ -0,0 +1,241 @@
...
+
+	// Create server listening for TCP connections
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", args.port))
+	if err != nil {
+		stderrLogger.Fatalf("failed to listen: %v", err)
+	}
+	grpcServer := grpc.NewServer()
+	pb.RegisterLabelerServer(grpcServer, &labelerServer{args: args, apiHelper: helper})
+	grpcServer.Serve(lis)
+}
```
zvonkok: Can we add here more info like 
```suggestion
	stdoutLogger.Printf("Node Feature Discover Master gRPC Server serving on port: %d", args,port)
}
```
In case someone uses a different port per commandline but forgets to update the Service.

nfd-master.yaml.template
```diff
@@ -0,0 +1,79 @@
...
+          image: quay.io/kubernetes_incubator/node-feature-discovery:v0.3.0
+          name: nfd-master
+          command:
+            - "nfd-master"
+---
+apiVersion: v1
+kind: Service
+metadata:
+  name: nfd-master
+spec:
```
zvonkok:
```suggestion
spec:
  type: ClusterIP
```
More verbosity just to make sure people understand, that we're here creating an intra-cluster service.
Dunno how many people know that a Service is per default ClusterIP.

nfd-worker-daemonset.yaml.template
marquiz: Env variable substitution inside k8s spec works with "normal" round brackets

cmd/nfd-worker/main.go
marquiz: Thanks for spotting this. I did some bikeshedding around the naming and this was a leftover, as you can guess...

Dockerfile
marquiz: You're right, thanks for pointing this out. Fixed!

cmd/nfd-master/main.go
marquiz: Makes sense. Added in the latest version of this PR.

Other suggestions around logging/verbosity are also welcome :smile:

nfd-master.yaml.template
marquiz: A fair suggestion. Fixed in the latest version of this PR

zvonkok:
I have a POC running with #209 , #211 , #189. Still some testing to do. 

```
#211 
feature.node.kubernetes.io/kernel-version.full=3.10.0-957.1.3.el7.x86_64
feature.node.kubernetes.io/kernel-version.major=3
feature.node.kubernetes.io/kernel-version.minor=10
feature.node.kubernetes.io/kernel-version.revision=0
```

```
#209 
NAME               READY     STATUS    RESTARTS   AGE
nfd-master-69h6x   1/1       Running   0          56m
nfd-master-l86qm   1/1       Running   0          56m
nfd-master-x8tw7   1/1       Running   0          56m
nfd-worker-sz4nr   1/1       Running   0          56m
```

nfd-master.yaml.template
```diff
@@ -0,0 +1,80 @@
...
+  selector:
+    matchLabels:
+      app: nfd-master
+  template:
+    metadata:
+      labels:
+        app: nfd-master
+    spec:
+      hostNetwork: true
```
zvonkok: Does the master need access to the host network? If we use hostNetwork we have to open a port on the nodes where the nfd-masters are running. Ommiting hostNetwork uses the internal k8s network.

marquiz: Thanks! No, we don't need that for master. Fixed...

marquiz:
> I have a POC running with #209 , #211 , #189. Still some testing to do.

Thanks you  for testing these.

One major missing piece of functionality (in my backlog) regarding this PR is the TLS encryption. That could be added in a separate PR a bit later, too, of course, if we want to get this merged soon.

One thing that could also be added would be a spec template of running the master and worker in the same pod on the nodes. In case somebody would not want to deploy the split architecture. Do you think this would make sense?

zvonkok:
Yeah makes sense, there are folks who run k8s only on one node, and this would help deploy NFD on a single host.  This could be a configuration option for the operator specifying a CR with a special keyword option, have no idea, but something like `working-mode: multi-node` or `working-mode: single-node`

nfd-master.yaml.template
```diff
@@ -0,0 +1,79 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: nfd-master
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - pods
```
zvonkok: Why does the master needs RBAC rules for pods?

marquiz: Hmm, looks like an ancient  leftover/mistake. It shouldn't really be needed. Thanks for spotting this! Fixed in the latest version of the PR

marquiz:
> Yeah makes sense, there are folks who run k8s only on one node, and this would help deploy NFD on a single host.

OK, I now added `nfd-daemonset-combined.yaml.template`. I'm not entirely sure about the naming of this, but, couldn't come up with anything better...

I also implemented (mutual) TLS authentication, i.e. both nfd-master and nfd-worker(s) are authenticated. Comments on this are welcome, too.

marquiz:
I wrote a quick update to the documentation. Comments to that are more than welcome :smile:

marquiz:
Unit tests have now been updated

marquiz:
Changed name of worker version annotation from `nfd.node.kubernetes.io/slave.version` to `nfd.node.kubernetes.io/worker.version`

I hope we finally got rid of the name *slave* completely...

zvonkok:
A `git grep slave` did not show any occurences of `slave`. Currently testing mutual TLS for master and worker, will get back to you this week. The master/worker split works for me. I might have some minor things to add but nothing drastic.

marquiz:
> A `git grep slave` did not show any occurences of `slave`. Currently testing mutual TLS for master and worker, will get back to you this week.

@zvonkok thanks for the update 


> The master/worker split works for me. I might have some minor things to add but nothing drastic.

Good. Any feedback is certainly welcome!

jeremyeder:
Is this PR waiting on any other tasks, or is it good to merge?

marquiz:
> Is this PR waiting on any other tasks, or is it good to merge?

We're good to go as soon as people are happy with this. Including the documentation 😉

marquiz:
Removed RFC status of the PR

marquiz:
Pushed a new rebased version. With one new patch: logging NodeName, and reading it only once, at at startup

PR #212, opened 2019-01-18
README: improve documentation of local hooks

marquiz:
Add a paragraph about hook config files.


marquiz:
> Good point to put the conf in a separate directory, otherwise those errors will show up in the logs. I see more and more customers interested in providing an own hook to NFD. Would be nice if this gets merged soon so I can point to the upstream documentaion.

I regarded this as an lgtm

Issue #178, opened 2018-10-10
Discovery of PMEM / NVDIMM devices


kad:
It would be good if NFD will be able to detect and report presence of PMEM devices on the node.

Minimal way to detect that can be presence of entries under `/sys/class/nd/` hierarchy.

Label can be something like `nfd-pmem.present`

marquiz:
I was thinking about the feature source. Would it be more logical to have a separate source (like *pmem* that you suggested), with a label like `nfd-pmem-present`, or, put it under currently existing feature source such as *memory* or *storage*, having label like `nfd-storage-pmem.present`?

marquiz:
Detection of nvdimm hardware was merged in #214. Closing this issue

PR #214, opened 2019-01-25
source/memory: detect presence of NVDIMM devices

marquiz:
Add a new (binary) label indicating the presence of non-volatile DIMM devices:
```
feature.node.kubernetes.io/memory-nv.present
```

Implements #178

### DIFF
--- a/README.md
+++ b/README.md
@@ -253,9 +253,10 @@ possible security implications.
 
 ### Memory Features
 
-| Feature name   | Description                                                                         |
-| :------------: | :---------------------------------------------------------------------------------: |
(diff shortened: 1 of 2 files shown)


marquiz:
I put that under the `memory` feature source, but I'm not sure if that is logical / the right place for it. Any thoughts on this?

msivosuo:
Between `storage` and `memory` I'd also lean towards `memory`.  @pohly, @avalluri, @okartau - any opinion?

pohly:
`memory` sounds good to me.

PR #215, opened 2019-02-12
Fix APIHelpers

marquiz:
Change RemoveLabelsWithPrefix() to do what its name suggests.
Previously, it did not test (only) for prefix, but it accepted any
substring match.


marquiz:
ping @zvonkok @kad

marquiz:
Fixed in #230

PR #217, opened 2019-02-19
sources: move all cpu related features under the cpu source

marquiz:
Remove 'cpuid', 'pstate' and 'rdt' feature sources and move their
functionality under the 'cpu' source. The goal is to have a more
systematic organization of feature sources and labels. After this change
we now basically have one source per type of hw, one for kernel and one
for userspace sw.

Related feature labels are changed, correspondingly, new labels being:
```
    feature.node.k8s.io/cpu-cpuid.<cpuid flag>
    feature.node.k8s.io/cpu-pstate.turbo
    feature.node.k8s.io/cpu-rdt.<rdt feature>
```


marquiz:
Rebased.

Any comments or thoughts @zvonkok, @kad, @Ethyling ?

Ethyling:
I don't know if it makes sense, but as the `cpuid` and the `pstate` format is `cpu-<source>.<value>=true` (for example `cpu-cpuid.AVX2=true`), maybe the `rdt` format should be `cpu-rdt.<value>=true`, for example `cpu-rdt.RDTMBA=true`. I think it could be good to try to unify the format labels.

marquiz:
> I don't know if it makes sense, but as the `cpuid` and the `pstate` format is `cpu-<source>.<value>=true` (for example `cpu-cpuid.AVX2=true`), maybe the `rdt` format should be `cpu-rdt.<value>=true`, for example `cpu-rdt.RDTMBA=true`. I think it could be good to try to unify the format labels.

Woops(!) Thanks for spotting this – yes, the naming should be as you described. The commit message and README have the correct naming, but, I had obviously broken the code in some earlier rebase. Now fixed.

Ping @zvonkok, @kad

PR #219, opened 2019-02-20
Use k8s apimachinery for label validation

marquiz:


main.go
```diff
@@ -376,6 +376,10 @@ func getFeatureLabels(source source.FeatureSource) (labels Labels, err error) {
 
 		// Validate label value
 		value := fmt.Sprintf("%v", v)
+		if len(value) > 63 {
```
kad: I would declare constant like in `pkg/apis/core/validation/validation.go`, so it will be understandable where this 63 value came from.
```go
const qualifiedNameMaxLength int = 63
```
Otherwise, lgtm.

marquiz: Hmm, actually why not use validation from apimachinery directly.

I'll get back to this...

marquiz: I changed this patch to use apimachinery directly for all validation

marquiz:
@kad ping, I rewrote the patch

kad:
lgtm

PR #220, opened 2019-02-22
Add kad as reviewer in OWNERS file

kad:
I have some time to help with reviews for @marquiz


jeremyeder:
I think having more reviewers would be very useful.  Having worked with @kad for a while, I think he would be a great addition.

PR #221, opened 2019-03-04
Dockerfile: change the base image

marquiz:
Dockerfile: use k8s.gcr.io/debian-base as the base for production image

Aligns the production base image with Kubernetes upstream. Slightly
reduces the size of the production container image from ca. 86MB to ca.
73MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,8 +19,14 @@ RUN make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
 RUN make test
 
 
+# Use debian-base for the extra files we need
+FROM k8s.gcr.io/build-image/debian-base:v2.1.3 as extra
+
+RUN apt-get update -y && apt-get -q -yy install --no-install-recommends --no-install-suggests --fix-missing busybox-static
+
+
 # Create production image for running node feature discovery
-FROM k8s.gcr.io/build-image/debian-base:v2.1.3
+FROM gcr.io/distroless/static
 
 # Run as unprivileged user
 USER 65534:65534
@@ -30,3 +36,4 @@ ENV GRPC_GO_LOG_SEVERITY_LEVEL="INFO"
 
 COPY --from=builder /go/node-feature-discovery/nfd-worker.conf.example /etc/kubernetes/node-feature-discovery/nfd-worker.conf
 COPY --from=builder /go/bin/* /usr/bin/
+COPY --from=extra /bin/busybox /bin/sh
--- a/Dockerfile
+++ b/Dockerfile
@@ -20,7 +20,7 @@ RUN make test
 
 
 # Create production image for running node feature discovery
-FROM debian:buster-slim
+FROM k8s.gcr.io/build-image/debian-base:v2.1.3
 
 # Run as unprivileged user
 USER 65534:65534
--- a/cmd/nfd-master/Dockerfile
+++ b/cmd/nfd-master/Dockerfile
@@ -22,7 +22,6 @@ RUN make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
 
 RUN make test
 
-
 # Create production image for nfd-master
 FROM gcr.io/distroless/static
 
--- a/cmd/nfd-master/Dockerfile
+++ b/cmd/nfd-master/Dockerfile
@@ -1,0 +1,34 @@
+# Build node feature discovery
+#
+# NOTE: we keep the builder stage of nfd-master and nfd-worker Dockerfiles in
+# sync in order to re-use docker layers
+#
+FROM golang:1.15.5-buster as builder
+
+# Get (cache) deps in a separate layer
+COPY go.mod go.sum /go/node-feature-discovery/
+
+WORKDIR /go/node-feature-discovery
+
+RUN go mod download
+
+# Do actual build
+COPY . /go/node-feature-discovery
+
+ARG VERSION
+ARG HOSTMOUNT_PREFIX
+
+RUN make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
+
+RUN make test
+
+
+# Create production image for nfd-master
+FROM gcr.io/distroless/static
+
+# Use more verbose logging of gRPC
+ENV GRPC_GO_LOG_SEVERITY_LEVEL="INFO"
+
+COPY --from=builder /go/bin/nfd-master /usr/bin/
+
+ENTRYPOINT ["/usr/bin/nfd-master"]
--- a/cmd/nfd-worker/Dockerfile
+++ b/cmd/nfd-worker/Dockerfile
@@ -1,4 +1,8 @@
 # Build node feature discovery
+#
+# NOTE: we keep the builder stage of nfd-master and nfd-worker Dockerfiles in
+# sync in order to re-use docker layers
+#
 FROM golang:1.15.5-buster as builder
 
 # Get (cache) deps in a separate layer
@@ -18,15 +22,8 @@ RUN make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
 
 RUN make test
 
-
-# Use debian-base for the extra files we need
-FROM k8s.gcr.io/build-image/debian-base:v2.1.3 as extra
-
-RUN apt-get update -y && apt-get -q -yy install --no-install-recommends --no-install-suggests --fix-missing busybox-static
-
-
-# Create production image for running node feature discovery
-FROM gcr.io/distroless/static
+# Create production image for nfd-worker
+FROM k8s.gcr.io/build-image/debian-base:v2.1.3
 
 # Run as unprivileged user
 USER 65534:65534
@@ -35,5 +32,6 @@ USER 65534:65534
 ENV GRPC_GO_LOG_SEVERITY_LEVEL="INFO"
 
 COPY --from=builder /go/node-feature-discovery/nfd-worker.conf.example /etc/kubernetes/node-feature-discovery/nfd-worker.conf
-COPY --from=builder /go/bin/* /usr/bin/
-COPY --from=extra /bin/busybox /bin/sh
+COPY --from=builder /go/bin/nfd-worker /usr/bin/
+
+ENTRYPOINT ["/usr/bin/nfd-worker"]
--- a/nfd-daemonset-combined.yaml.template
+++ b/nfd-daemonset-combined.yaml.template
@@ -67,7 +67,7 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
+          image: gcr.io/k8s-staging-nfd/node-feature-discovery-master:master
           imagePullPolicy: Always
           name: nfd-master
           securityContext:
@@ -76,14 +76,12 @@ spec:
               drop: ["ALL"]
             readOnlyRootFilesystem: true
             runAsNonRoot: true
-          command:
-            - "nfd-master"
         - env:
           - name: NODE_NAME
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
+          image: gcr.io/k8s-staging-nfd/node-feature-discovery-worker:master
           imagePullPolicy: Always
           name: nfd-worker
           securityContext:
@@ -92,8 +90,6 @@ spec:
               drop: ["ALL"]
             readOnlyRootFilesystem: true
             runAsNonRoot: true
-          command:
-            - "nfd-worker"
           args:
             - "--sleep-interval=60s"
           volumeMounts:
--- a/nfd-master.yaml.template
+++ b/nfd-master.yaml.template
@@ -81,7 +81,7 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
+          image: gcr.io/k8s-staging-nfd/node-feature-discovery-master:master
           imagePullPolicy: Always
           name: nfd-master
           securityContext:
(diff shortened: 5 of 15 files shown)


Dockerfile
```diff
@@ -1,5 +1,7 @@
 # Build node feature discovery
-FROM golang:1.8 as builder
+FROM golang:1.11-alpine as builder
+
+RUN apk add --no-cache gcc git libc-dev
```
Ethyling: As we do `RUN make test` instead of `RUN go test` since #209, we must install `make` too. Otherwise it seems to be working as expected.

Ethyling:
I think we only need to rebase this MR and add `make` in the `builder` stage and we can merge it

marquiz:
Thanks for taking a peek at this. I thing the *correct* base image to use is a good subject for discussion, alpine might not be the one 😄  I prolly should open an issue about this

Ethyling:
> I thing the _correct_ base image to use is a good subject for discussion, alpine might not be the one :smile: I prolly should open an issue about this

Do so, I really don't know what makes a good base image to be honest

kad:
I think main kubernetes repository moved away from alpine not that long time ago and mostly using `k8s.gcr.io/debian-base:v1.0.0` as a base image. Maybe it is better to follow example of k/k ?

marquiz:
> I think main kubernetes repository moved away from alpine not that long time ago and mostly using `k8s.gcr.io/debian-base:v1.0.0` as a base image. Maybe it is better to follow example of k/k ?

I think this might be a good alternative. It slightly reduces the size of the image, and, aligns with k8s upstream. I pushed a new version and updated the PR subject and description

marquiz:
Using `gcr.io/distroless/base` would further reduce the size of the image to about 48MB. But, it's very minimal image and I think `k8s.gcr.io/debian-base` is a better choice. That's mainly because distroless limits the usage/flexibility of local hooks as it has no shell, for example.

zvonkok:
Do we have a list of supported language/runtimes for the hooks, that are actually limited to the base image? The doc only states that the hooks have to be executables (python, bash, c++, go). 

For "complicated discovery" and extra dependencies, the current pattern (I have seen so far) is to use a side-car container to inject feature labels into NFD. Should we add some documentation here?

marquiz:
> Do we have a list of supported language/runtimes for the hooks, that are actually limited to the base image? The doc only states that the hooks have to be executables (python, bash, c++, go).
> 
> For "complicated discovery" and extra dependencies, the current pattern (I have seen so far) is to use a side-car container to inject feature labels into NFD. Should we add some documentation here?

I think these both good suggestions. I opened a new issue #247 to track this

marquiz:
No progress here in a long time. Any comments using ubi-minimal as a base image?

/remove-lifecycle stale

marquiz:
Back to the almost original. What do you think about using the "official" debian-base from Kubernetes as a base image (@zvonkok @Ethyling @kad ) like @kad suggested back in the days? 

I think it would make sense by aligning us with the kubernetes/kubernetes upstream. In addition, it would provide a marginal reduction on the container image size (90.9 MB vs. 93.9 MB on my latest test)

mythi:
> I think it would make sense by aligning us with the kubernetes/kubernetes upstream

that's gcr.io/distroless/base isn't it

marquiz:
> that's `gcr.io/distroless/base` isn't it

You mean distroless/base is same as debian-base? Or using distroless/base would be aligning with k/k? Anyway, I think the answer to both of those questions is negative 😄

mythi:
k/k uses `gcr.io/distroless/base`

marquiz:
> k/k uses `gcr.io/distroless/base`

Nope
https://github.com/kubernetes/kubernetes/tree/master/build/debian-base

mythi:
@marquiz where do you see `debian-base` being used? we evaluated the `gcr.io/distroless` flavors when k8s moved to use them about a year ago. to me it looks all images are now based on `go-runner` which wraps `gcr.io/distroless`: https://github.com/kubernetes/kubernetes/blob/master/build/common.sh#L90-L107

marquiz:
Hmm, looks like you're right. Looks like I missed the switchover as it's very recent like two weeks ago or so. But I think debian-base still makes sense us (and is still aligned with k/k) as we probably want to support hooks other than just plain go binaries.

mythi:
@marquiz this KEP discusses the base image policy and rationale behind all the distroless work: https://github.com/kubernetes/enhancements/tree/master/keps/sig-release/1729-rebase-images-to-distroless

marquiz:
Inspired by the recent k/k switch to distroless (note that e.g. kube-proxy is still based on debian) I did some experimentation on NFD. I now pushed a version which builds statically linked nfd binaries and uses distroless/static as the container base image. Hooks are the problem for us and this version now has busybox to provide shell and coreutils. Dunno if this would be a acceptable(?)

In any case, nfd-master does not have any dependencies on other utils so it could be separated into its own image based on distroless/static

marquiz:
Still valid. Any thoughts or comments on this?

/remove-lifecycle stale

ArangoGutierrez (approved):
/lgtm

ArangoGutierrez:
```bash
[eduardo@fedora-32-ws node-feature-discovery]$ podman run --rm --name=nfd-test k8s.gcr.io/nfd/node-feature-discovery:v0.6.0-32-g900c20e-dirty nfd-master --no-publish
2020/09/06 21:30:40 Node Feature Discovery Master v0.6.0-32-g900c20e-dirty
2020/09/06 21:30:40 NodeName: ''
2020/09/06 21:30:40 gRPC server serving on port: 8080
```

ArangoGutierrez:
@marquiz I tested this PR, the image builds and NFD works, but this PR is old enough to ask, should we take an Updated look to the images being used upstream?

marquiz:
> I tested this PR, the image builds and NFD works, but this PR is old enough to ask, should we take an Updated look to the images being used upstream?

Yeah, certainly. Would be nice to hear comments from others, too.

As I stated above we should be able to build a separate image for nfd-master, using distroless. What do you guys think?

marquiz:
Rebased the PR, squashed the patchset a bit to ditch the oldest experiments, and, added a patch to demonstrate separate master and worker images. Comments?

Still a WIP, I haven't really tested this thoroughly.
/hold

ArangoGutierrez:
@marquiz dis PR didn't make it on the next release cut, do you think it still holds hope to be merged?  I am happy to collab with you on reviews so we can move this patch forward

cmd/nfd-master/Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Build node feature discovery
+#
+# NOTE: we keep the builder stage of nfd-master and nfd-worker Dockerfiles in
+# sync in order to re-use docker layers
+#
+FROM golang:1.14.7 as builder
```
ArangoGutierrez, line 6: we are now on 1.15 soon to be  1.16

cmd/nfd-master/Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY . /go/node-feature-discovery
+
+ARG VERSION
+ARG HOSTMOUNT_PREFIX
+
+RUN make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
+
+RUN make test
+
+
```
ArangoGutierrez, line 25:
```suggestion
```

nfd-daemonset-combined.yaml.template
```diff
@@ -62,31 +62,27 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          image: k8s.gcr.io/nfd/node-feature-discovery:v0.6.0
+          image: k8s.gcr.io/nfd/node-feature-discovery-master:v0.6.0
```
ArangoGutierrez, line 65:
```suggestion
          image: k8s.gcr.io/nfd/node-feature-discovery-master:v0.7.0
```
or v0.8.0 for the future case

nfd-daemonset-combined.yaml.template
```diff
@@ -62,31 +62,27 @@ spec:
...
             runAsNonRoot: true
-          command:
-            - "nfd-master"
         - env:
           - name: NODE_NAME
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          image: k8s.gcr.io/nfd/node-feature-discovery:v0.6.0
+          image: k8s.gcr.io/nfd/node-feature-discovery-worker:v0.6.0
```
ArangoGutierrez, line 78:
```suggestion
          image: k8s.gcr.io/nfd/node-feature-discovery-worker:v0.7.0
```

nfd-master.yaml.template
```diff
@@ -77,16 +77,14 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          image: k8s.gcr.io/nfd/node-feature-discovery:v0.6.0
+          image: k8s.gcr.io/nfd/node-feature-discovery-master:v0.6.0
```
ArangoGutierrez, line 80:
```suggestion
          image: k8s.gcr.io/nfd/node-feature-discovery-master:v0.7.0
```

nfd-worker-daemonset.yaml.template
```diff
@@ -21,16 +21,14 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          image: k8s.gcr.io/nfd/node-feature-discovery:v0.6.0
+          image: k8s.gcr.io/nfd/node-feature-discovery-worker:v0.6.0
```
ArangoGutierrez, line 24:
```suggestion
          image: k8s.gcr.io/nfd/node-feature-discovery-worker:v0.7.0
```

nfd-worker-job.yaml.template
```diff
@@ -30,16 +30,14 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
-          image: k8s.gcr.io/nfd/node-feature-discovery:v0.6.0
+          image: k8s.gcr.io/nfd/node-feature-discovery-worker:v0.6.0
```
ArangoGutierrez, line 33:
```suggestion
          image: k8s.gcr.io/nfd/node-feature-discovery-worker:v0.7.0
```

ArangoGutierrez (changes requested):

cmd/nfd-worker/Dockerfile
```diff
@@ -1,4 +1,8 @@
 # Build node feature discovery
+#
+# NOTE: we keep the builder stage of nfd-master and nfd-worker Dockerfiles in
+# sync in order to re-use docker layers
+#
```
ArangoGutierrez, line 5: now that we are here, change also the version of the base here

marquiz:
> @marquiz dis PR didn't make it on the next release cut, do you think it still holds hope to be merged? I am happy to collab with you on reviews so we can move this patch forward

Thanks for bringing this up @ArangoGutierrez . I think we could squeeze this in the next release. However, I'd like to hear comments what people think about this 😄  There are two main questions:
1. What do we want to use as the base image (for nfd-worker)
2. Are separate images for nfd-master and worker a good or a bad thing

ArangoGutierrez:
> * What do we want to use as the base image (for nfd-worker)
> 
> * Are separate images for nfd-master and worker a good or a bad thing

Well If you ask me No.1 I will respond UBI8 image ;P , but in ref to No.2, I think for consistency, both images should be the same.
Maybe at some point we could add `kustomize` and let the user set his own images via `make`. but for No.2 I vote for having the same. 

@Ethyling @zvonkok @balajismaniam @kad thoughts?

kad:
IMHO, it is better to use base images similar to other base images in main kubernetes repo.
UBI8 images might be ok, but I'm a bit doubtful about licenses on those, maybe there are some underwater stones in using those.
Other option of course can be to build several variants and allow users to select via kustomize which are more appropriate for their usage. However, that means that it should be somehow ensured that worker and master images are from same variant.

cmd/nfd-master/Dockerfile
marquiz: Fixed on the latest rebase

nfd-daemonset-combined.yaml.template
marquiz: Synced with latest master, using staging:master image

nfd-daemonset-combined.yaml.template
marquiz: Using staging:master

nfd-master.yaml.template
marquiz: Using staging:master

nfd-worker-daemonset.yaml.template
marquiz: Using staging:master

nfd-worker-job.yaml.template
marquiz: Using staging:master

cmd/nfd-worker/Dockerfile
marquiz: Done

marquiz:
> IMHO, it is better to use base images similar to other base images in main kubernetes repo.

k/k only uses distroless nowadays, in all of the containers. That's not optimal because we have the hooks support in the worker, and have supported bash as the "runtime".
[ A bit of an off-topic, but, I'm not sure how many users for the hooks really exist, and, should the support be deprecated and eventually removed 🤔 Feature files and sidecar containers could be enough. ]


> UBI8 images might be ok, but I'm a bit doubtful about licenses on those, maybe there are some underwater stones in using those.

This might be an option, yes.

> Other option of course can be to build several variants and allow users to select via kustomize which are more appropriate for their usage. However, that means that it should be somehow ensured that worker and master images are from same variant.

This might actually be worthwhile considering. We could have distroless image as a (default?) option with only support for statically linked hook binaries. In addition, we could have an "extended" image that has e.g. bash (and possibly other runtimes). Having master and worker from the different variant shouldn't be a problem, as long as they are of the same version.

ArangoGutierrez:
I have tested this PR in a kubernetes 1.20 and works.
Should be ready to go, we need OWNERS consensus on this patch that is getting quite old
OWNERS: @Ethyling @zvonkok  @kad

ArangoGutierrez:
If someone wants to give this PR a try here are the built images (x86), on my quay repo
 - https://quay.io/repository/eduardoarango/node-feature-discovery-master
 - https://quay.io/repository/eduardoarango/node-feature-discovery-worker

ArangoGutierrez:
Needs rebase to catch up with the markdown linter

marquiz:
Exploring @kad's idea on image variants, I submitted #469

marquiz:
Closing this PR as not valid anymore.

All images in k/k are based on distroless as is our new **minimal** image. Moreover, our full image is based on Debian buster/10 which is the underlying base for distroless, too. So that I'd say our **full** and **minimal** are aligned as well.

PR #223, opened 2019-03-05
Makefile: reorganise container image variables

marquiz:
Change the structure and naming of the make variables that control the
container image name/tag that gets created. Default values and behavior
stay the same, but, this change tries to make it easier to customize the
build from command line.
    
Also, document all the relevant make variables in readme.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,19 +2,20 @@
 
 IMAGE_BUILD_CMD := docker build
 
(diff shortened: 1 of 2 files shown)


marquiz:
ping @kad, @zvonkok

zvonkok:
Works for me :) 
/lgtm

Issue #130, opened 2018-05-28
Improve SGX labels


mythi:
Currently, the SGX label is added as true if the ```cpuid``` instruction says SGX is supported. However, it may be that the SGX leaf instructions are disabled from the feature control register making the label void (SXG=true but you cannot  execute SGX enclaves).

Additionally, SGX Launch control feature bit is necessary as the label.

okartau:
cpuid-SGX label is correct (note the cpuid part in label name).
"SGX-capable CPU" is not equal to "SGX can be used by SW". 
This additional detection complexity is described in docs like https://en.wikipedia.org/wiki/Software_Guard_Extensions
(perhaps we can add a warning note for SGX case in NFD doc?).

The fact that use of SGX requires more (BIOS support, opt-in enabling), does not make cpuid-SGX label void or false.

If we want to see label "SGX can be used" then we need additional detection code somewhere.

Extending related CPUID-based flags:
Current klauspost/cpuid that is used by NFD, does not look further than SGX bit,
so if we want to see more on (Launch control, SGX1/SGX2 capability), we need to either push for additions on cpuid package side, or add own detection code (like we have on NFD/RDT side).

End goal, i.e. to get label "SGX can be used":
Can we rely on kernel cpu flag?
This flag gets added by SGX kernel driver, right?
(As I don't see it on a plain system (Xeon v6) where CPUID shows SGX capability).

mythi:
IMO, the title of this issue is valid and the improvement is justified: ```cpuid-SGX``` alone cannot be use to determine the node can run enclaves.

marquiz:
Probably so. I just tried to point out that the CPUID SGX label is correct and should not be changed (possibly only masked out), and, we need a new feature source (plugin) if we want to have SGX detection.

marquiz:
@mythi: Is this feature still needed? What would be the use case for this?

mythi:
/reopen
/remove-lifecycle rotten

I'll be working on addressing this.

PR #224, opened 2019-03-05
source/cpu: make cpuid configurable

marquiz:
Add 'cpuid/attributeBlacklist' and 'cpuid/attributeWhitelist' config
options for the cpu feature source. These can be used to filter the set
of cpuid capabilities that get published. The intention is to reduce
clutter in the NFD label space, getting rid of "obvious" or misleading
cpuid labels. Whitelisting has higher priority, i.e.  only whitelist
takes effect if both attributeWhitelist and attributeBlacklist are
specified.

### DIFF
--- a/README.md
+++ b/README.md
@@ -215,6 +215,16 @@ such as restricting discovered features with the --label-whitelist option._
 | <br>                    | RDTL2CA            | Intel L2 Cache Allocation Technology
 | <br>                    | RDTMBA             | Intel Memory Bandwidth Allocation (MBA) Technology
 
+The (sub-)set of CPUID attributes to publish is configurable via the
+`attributeBlacklist` and `attributeWhitelist` cpuid options of the cpu source.
+If whitelist is specified, only whitelisted attributes will be published. With
+blacklist, only blacklisted attributes are filtered out. `attributeWhitelist`
+has priority over `attributeBlacklist`.  For examples and more information
+about configurability, see [Configuration Options](#configuration-options).
+By default, the following CPUID flags have been blacklisted:
+BMI1, BMI2, CLMUL, CMOV, CX16, ERMS, F16C, HTT, LZCNT, MMX, MMXEXT, NX, POPCNT,
+RDRAND, RDSEED, RDTSCP, SGX, SSE, SSE2, SSE3, SSE4.1, SSE4.2 and SSSE3.
+
 **NOTE** The cpuid features advertise *supported* CPU capabilities, that is, a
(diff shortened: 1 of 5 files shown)


source/cpuid/cpuid.go
```diff
@@ -16,8 +16,100 @@ limitations under the License.
...
+		for _, k := range Config.AttributeBlacklist {
+			newFilter.keys[k] = true
+		}
+		newFilter.whitelist = false
+	}
+	filter = &newFilter
+}
+
+func (f keyFilter) unmask(k string) bool {
+	if f.whitelist {
```
kad: this can be simplified to one if. you need to return true when whitelist and ok is true or when whitelist and ok are false. Thus,
```go
if _, ok := f.keys[k]; f.whitelist == ok {
...
}
```

marquiz: True  (I  didn't even think about that), but, I think think that this optimization clearly worsens the readability of the code. In the future, it  may be  hard to understand the logic  behind `f.whitelist == ok`. So, I'd leave it as it is (and let the compiler do the optimization)

marquiz: @kad ping

marquiz:
ping @kad @zvonkok

README.md
```diff
@@ -218,21 +218,29 @@ not enabled) as reported by the `cpuid` instruction.
...
+`attributeBlacklist` and `attributeWhitelist` options of the cpuid source. If
+whitelist is specified, only whitelisted attributes will be published. With
+blacklist, only blacklisted attributes are filtered out. `attributeWhitelist`
+has priority over `attributeBlacklist`.
+For examples and more information about configurability, see
+[Configuration Options](#configuration-options).
+
+By default, the following CPUID flags have been blacklisted:
+BMI1, BMI2, CLMUL, CMOV, CX16, ERMS, F16C, HTT, LZCNT, MMX, MMXEXT, NX, POPCNT,
+RDRAND, RDSEED, RDTSCP, SGX, SSE, SSE2, SSE3, SSE4.1, SSE4.2 and SSSE3.
```
kad: @mythi what do you think about by default blacklisted SGX?

marquiz: SGX cpuid label is pretty worthless, if not even misleading as is discussed in #130

mythi: +1, disabling it makes sense for now

marquiz:
Rebased.

ping @kad @zvonkok

marquiz:
Rebased and adapted to #217

PR #230, opened 2019-04-05
Extend unit test coverage

marquiz:
First there is a big refactoring, moving most of functionality from `cmd/nfd-{master|worker}/main.go` to `pkg/nfd-{master|worker}/`.

Next, we extend unit tests to test the moved functionality. The tests validate the basic labeling functionality with the built-in "fake" source. They also test the communication between nfd-master and nfd-worker, with and without TLS authentication enabled.


pkg/nfd-master/nfd-master.go
```diff
@@ -0,0 +1,337 @@
...
+
+	// Create server listening for TCP connections
+	lis, err := net.Listen("tcp", fmt.Sprintf(":%d", m.args.Port))
+	if err != nil {
+		return fmt.Errorf("failed to listen: %v", err)
+	}
+	// Notify that we're ready to accept connections
+	m.ready <- true
+	close(m.ready)
+	time.Sleep(time.Second)
```
Ethyling: Why do we need to sleep here ?

pkg/nfd-master/nfd-master.go
```diff
@@ -0,0 +1,337 @@
...
+	if err != nil {
+		stderrLogger.Printf("can't update node: %s", err.Error())
+		return err
+	}
+
+	return nil
+}
+
+// Filter labels if whitelist has been defined
+func filterFeatureLabels(labels *Labels, labelWhiteList *regexp.Regexp) {
```
Ethyling: As said in another PR, this function is not used. I'm not sure we may fix it in this PR, but we should open an issue or something for this I think.

marquiz: Argh, yes. Let's handle this separately. I opened #237 where we can discuss it

pkg/nfd-master/nfd-master.go
marquiz: Well spotted! To be honest, I don't know (anymore). Must be some accidental leftover from early development/testing phase of this patchset. Removed this in the new version of the PR

Ethyling:
Looks good to me!

marquiz:
@zvonkok, @kad: any comments? I'd be inclined to merge this soon

marquiz:
@zvonkok, @kad, anything to add?

I'm planning to merge this by the end of the week (i.e., Sunday evening Finnish time) if I do not see any objection before that.

PR #241, opened 2019-05-07
nfd-master: fix --label-whitelist

marquiz:
Make the --label-whitelist effective. Previously, it was unused and had
no effect. Also, add simple unit test for that.

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -203,14 +203,15 @@ func updateMasterNode(helper apihelper.APIHelpers) error {
 }
 
 // Filter labels if whitelist has been defined
-func filterFeatureLabels(labels *Labels, labelWhiteList *regexp.Regexp) {
-	for name := range *labels {
+func filterFeatureLabels(labels Labels, labelWhiteList *regexp.Regexp) Labels {
+	for name := range labels {
 		// Skip if label doesn't match labelWhiteList
 		if !labelWhiteList.MatchString(name) {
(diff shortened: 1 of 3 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -290,13 +290,14 @@ func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Lab
 	removeLabelsWithPrefix(node, "node.alpha.kubernetes-incubator.io/node-feature-discovery")
 
 	// Add labels to the node object.
+	filterFeatureLabels(&labels, s.args.LabelWhiteList)
```
Ethyling: I think we must filter the labels before the annotation is generated no? Here we are adding all the labels in the annotation and then filter the labels using the regex.

marquiz: Yup, you're correct 🤭 I obviously wrote this patch a bit too hastily.

This is now fixed and the unit test is updated to check this. I also added another patch fixing a former issue in unit tests

marquiz:
I pushed a new version where whitelist filter is applied even if --no-publish is specified

Ethyling:
It seems like the regexp in the function `filterFeatureLabels` is `NULL` when we are running the tests `TestRun` and `TestRunTls` from `pkg/nfd-worker/nfd-worker_test.go`. I don't know if the fix is to pass a regexp in the master args?

marquiz:
*Arrgh*, I hope I got it right, this time 😊

Issue #207, opened 2018-12-20
New hierarchy of feature sources


marquiz:
The current set of feature sources (and labels), is somewhat arbitrary, reflecting the (commit) history of the project – typically new features have been implemented as a completely new feature source. However, the feature sources and their associated labels could be organized in a more systematic way, providing a bit more standardized and logical hierarchy of the feature labels.

For example, we could have one source per type of hw (plus one for kernel sw and one for userspace sw). An example, how the current set of feature sources and labels could be transformed:

* cpu
  * hardware_multithreading
  * cpuid.&lt;cpuid flag&gt;
  * pstate.turbo
  * rdt.&lt;rdt feature&gt;
* iommu
  * enabled
* kernel
  * config.&lt;kconfig option&gt;
  * selinux.enabled
  * version.&lt;version component&gt;
* memory
  * numa
* pci
  * &lt;device&gt;.present
* storage
  * nonrotationaldisk
* system
  * os_release.&lt;version field&gt;

This issue was split out from #200. Similar to that, this would be a good time to go for the change as all the feature labels have changed after the repo migration, anyway.

jeremyeder:
Not sure we have a lot to say about this since we're not depending on it yet, and thus can adapt quickly to any changes. /cc @zvonkok

marquiz:
Tried to consult [SIG-node mailing list](https://groups.google.com/forum/#!topic/kubernetes-sig-node/R-Q6XTAs-EI) on this

marquiz:
#217 has an implementation re-organizing CPU related features

Ethyling:
I think #217 solved this issue

marquiz:
Yes, it did – I don't have any other changes in mind

Issue #202, opened 2018-12-12
Release v0.4.0


marquiz:
## Release checklist:
- [x] All [OWNERS](OWNERS) must LGTM the release proposal.
- [x] Update the [daemonset template](node-feature-discovery-daemonset.yaml.template) and the [job template](node-feature-discovery-job.yaml.template) to use the new tagged container image
- [x] An OWNER runs `git tag -s $VERSION` and insert the changelog into the tag description.
- [x] [Build and push](https://github.com/kubernetes-sigs/node-feature-discovery#building-from-source) a container image with the same tag to [quay.io](https://quay.io/kubernetes_incubator).
- [x] Update the `:latest` virtual tag in quay.io to track the last stable (this) release.
- [x] An OWNER pushes the tag with `git push $VERSION`.
- [x] Write the change log into the [Github release info](https://github.com/kubernetes-sigs/node-feature-discovery/releases).
- [x] Add a link to the tagged release in this issue.
- [x] An announcement email is sent to `kubernetes-dev@googlegroups.com` with the
   subject `[ANNOUNCE] node-feature-discovery $VERSION is released`
- [x] Close this issue.

## Changelog
Node-feature-discovery was migrated into a [new repository](https://github.com/kubernetes-sigs/node-feature-discovery) under the [kubernetes-sigs](https://github.com/kubernetes-sigs) organization in Github (#175). Related to the migration, the final container image registry/repo hasn't been dediced yet (#177) – for this release we still use the old repo.

### Major changes
- Split NFD into client and server (#209)
- Changes in labels:
  - NFD label namespace was changed to `feature.node.kubernetes.io/` (#176)
    - `nfd-` prefix was dropped from all feature labels
    - NFD version label (`feature.node.kubernetes.io/node-feature-discovery.version`) was replaced by an annotation (`nfd.node.kubernetes.io/version`)
  - network SRIOV labels were changed (#173):
    - `network-sriov` -> `network-sriov.capable`
    - `network-sriov-configured` -> `network-sriov.configured`
  - selinux detection was moved to kernel feature source
    - `selinux` -> `kernel-selinux.enabled`
  - cpuid, pstate and RDT labels moved under cpu feature source (#217)
    - `cpuid-<cpuid flag>` -> `cpu-cpuid.<cpuid flag>`
    - `pstate-turbo` -> `cpu-pstate.turbo`
    - `rdt-<rdt feature>` -> `cpu-rdt.<rdt feature>`
[...]

PR #243, opened 2019-05-08
Better document the difference between latest release and master branch

marquiz:
Point to the latest release in the README, and, point out that a
user-built custom image is required to run the latest development
version. Update the deployment instructions to reflect the need to
specify the container image when using the deployment spec template(s).
Also, update the Job deployment script to set a user-defined container
image.

Use a placeholder for the container image in the template specs as they were not compatible with the image they were pointing to (v0.3.0).

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,6 +421,13 @@ directory. It is the user's responsibility to review the hooks for e.g.
 possible security implications.
 
 ## Getting started
+
+For a stable version with ready-built images see the
+[latest released version](https://github.com/kubernetes-sigs/node-feature-discovery/tree/v0.3.0) ([release notes](https://github.com/kubernetes-sigs/node-feature-discovery/releases/latest)).
+
+If you want to use the latest development version (master branch) you need to
+[build your own custom image](#building-from-source).
+
 ### System requirements
 
 1. Linux (x86_64/Arm64)
@@ -433,9 +440,11 @@ possible security implications.
 #### nfd-master
 
 Nfd-master runs as a DaemonSet, by default in the master node(s) only. You can
-use the template spec provided to deploy nfd-master:
+use the template spec provided to deploy nfd-master. You only need to update the
+template to use the correct image:
 ```
-kubectl create -f nfd-master.yaml.template
+sed -E s'/^(\s*)image:.+$/\1image: <YOUR_IMAGE_REPO>:<YOUR_IMAGE_TAG>/' nfd-master.yaml.template > nfd-master.yaml
+kubectl create -f nfd-master.yaml
 ```
 Nfd-master listens for connections from nfd-worker(s) and connects to the
 Kubernetes API server to adds node labels advertised by them.
@@ -450,9 +459,11 @@ labels. The provided template will configure these for you.
 
 Nfd-worker is preferably run as a Kubernetes DaemonSet. There is an
 example spec that can be used as a template, or, as is when just trying out the
-service:
+service. Similarly to nfd-master above, you need to update the
+template with the correct image:
 ```
-kubectl create -f nfd-worker-daemonset.yaml.template
+sed -E s'/^(\s*)image:.+$/\1image: <YOUR_IMAGE_REPO>:<YOUR_IMAGE_TAG>/' nfd-worker-daemonset.yaml.template > nfd-worker-daemonset.yaml
+kubectl create -f nfd-worker-daemonset.yaml
 ```
 
 Nfd-worker connects to the nfd-master service to advertise hardware features.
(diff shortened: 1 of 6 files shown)


Ethyling:
I think this is a good beginning, but what append when we are doing a new release? I mean, on a new release, we will need to push the new YAMLs with the `image` field pointing to the released image and then commit again to replace it with `<IMAGE_REPO>:<IMAGE_TAG>`? May we just keep the `image` field pointing to the old released image and let the sed part in the `README.md`. What do you think about it?

marquiz:
> I think this is a good beginning, but what happen when we are doing a new release? I mean, on a new release, we will need to push the new YAMLs with the `image` field pointing to the released image and then commit again to replace it with `<IMAGE_REPO>:<IMAGE_TAG>`? May we just keep the `image` field pointing to the old released image and let the sed part in the `README.md`. What do you think about it?

Yup, that's what I was thinking. For release we could make the templates point to the correct release. But, after the release, the `image:` fields  would be changed to a placeholder – probable only when the template specs are modified otherwise (making them incompatible/non-aligned with the release). I.e., when the templates(s) are modified in devel cycle, the image field must  be modified. How do you see this?

Of course, a big problem is that there has been a very long gap between releases, now. I'd like to solve #207/#217 and cut a release after that.

Ethyling:
> > I think this is a good beginning, but what happen when we are doing a new release? I mean, on a new release, we will need to push the new YAMLs with the `image` field pointing to the released image and then commit again to replace it with `<IMAGE_REPO>:<IMAGE_TAG>`? May we just keep the `image` field pointing to the old released image and let the sed part in the `README.md`. What do you think about it?
> 
> Yup, that's what I was thinking. For release we could make the templates point to the correct release. But, after the release, the `image:` fields would be changed to a placeholder – probable only when the template specs are modified otherwise (making them incompatible/non-aligned with the release). I.e., when the templates(s) are modified in devel cycle, the image field must be modified. How do you see this?

I agree with you on the part that we need to set a placeholder when the images and the templates are incompatible.

> Of course, a big problem is that there has been a very long gap between releases, now. I'd like to solve #207/#217 and cut a release after that.

Yes, I think we need to do a release soon with all the new features. We may need to update https://github.com/kubernetes-sigs/node-feature-discovery/issues/202 to match the features we want to include in this release. I will take a look at #217

marquiz:
> I think we need to do a release soon with all the new features. We may need to update #202 to match the features we want to include in this release.

Yes, definitely, that needs to be updated

label-nodes.sh
```diff
@@ -1,10 +1,19 @@
...
 # We set the .spec.completions and .spec.parallelism to the node count
 # We request a specific hostPort in the job spec to limit the number of pods
 # that run on a node to one. As a result, one pod runs on each node in parallel
 # We set the NODE_NAME environemnt variable to get the Kubernetes node object.
-sed -e "s/COMPLETION_COUNT/$NumNodes/" -e "s/PARALLELISM_COUNT/$NumNodes/" nfd-worker-job.yaml.template > node-feature-discovery-job.yaml
+sed -E -e "s/COMPLETION_COUNT/$NumNodes/" \
+    -e "s/PARALLELISM_COUNT/$NumNodes/" \
+    -e "s/^(\s*)image:.+$/\1image: $1/" \
+    nfd-worker-job.yaml.template > nfd-worker-job.yaml
 kubectl create -f node-feature-discovery-job.yaml
```
Ethyling: Should be `kubectl create -f nfd-worker-job.yaml`

README.md
```diff
@@ -421,6 +421,14 @@ directory. It is the user's responsibility to review the hooks for e.g.
 possible security implications.
 
 ## Getting started
+
+If you want to use a stable version with ready-built images see the
+documentation of the
+[latest released version](https://github.com/kubernetes-sigs/node-feature-discovery/tree/v0.3.0).
```
Ethyling: I would like to have something that tells to use all the repository of the latest release instead of the documentation. I think it could be less confusing for users. What do you think?

label-nodes.sh
marquiz: Hmm, I actually fixed this already yesterday after testing the script myself, but, obviously failed to push the updated branch. Fixed, now

Thanks for spotting this.

README.md
marquiz: Yeah, I think that's a good suggestion. Did you have some specific wording in mind? If not, I'll make up something 😄

marquiz:
> Yes, I think we need to do a release soon with all the new features. We may need to update #202 to match the features we want to include in this release.

I did some updates on #202. Feel free to act on #217

README.md
Ethyling: No, nothing really nice

marquiz: I made a small adjustment

PR #250, opened 2019-05-17
Use v0.4.0 release in README and deployment templates and script

marquiz:
Also, slightly adjust the deployment instructions in README to point out
that the templates should now be usable as is to run the latest released
version of NFD.

### DIFF
--- a/label-nodes.sh
+++ b/label-nodes.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 this=`basename $0`
-if [ $# -ne 1 ]; then
-    echo Usage: $this IMAGE[:TAG]
+if [ $# -gt 1 ]; then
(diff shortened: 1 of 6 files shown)


marquiz:
Need to update/remove the ascii demo from readme as it's totally outdated

/hold

marquiz:
Demo was updated in #251 

/hold cancel

label-nodes.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 this=`basename $0`
-if [ $# -ne 1 ]; then
+if [ $# -gt 1 ]; then
     echo Usage: $this IMAGE[:TAG]
```
Ethyling: You need to update the usage message

marquiz: Ha, thanks for pointing this out! Fixed now

PR #252, opened 2019-05-21
Add label for network bridges

mmirecki:
This patch will add labeling for linux network bridges.
If a bridge named 'br1' exists on a node, the following label
will be added:
    feature.node.kubernetes.io/network-bridge.br1: "true"

This feature is need by the kubevirt project in order to schedule
pods with custom networks.

### DIFF
--- a/source/network/network.go
+++ b/source/network/network.go
@@ -22,6 +22,7 @@ import (
 	"io/ioutil"
 	"log"
 	"net"
+	"os"
 	"strconv"
 	"strings"
 
@@ -81,6 +82,11 @@ func (s Source) Discover() (source.Features, error) {
 				}
 			}
 		}
+
+		if bridgeDir, _ := os.Stat("/sys/class/net/" + netInterface.Name + "/bridge"); bridgeDir != nil {
+			features["bridge."+netInterface.Name] = true
+		}
+
 	}
 	return features, nil
 }
--- a/README.md
+++ b/README.md
@@ -281,10 +281,11 @@ See [configuration options](#configuration-options) for more information.
 
 ### Network Features
 
-| Feature | Attribute  | Description                                           |
-| ------- | ---------- | ----------------------------------------------------- |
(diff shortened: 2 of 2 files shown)


zvonkok:
@mmirecki is this the only network feature you need for kubevirt? 
@marquiz If more and more network things arise and are needed do we want a logical grouping for those features as well,  `network-<iface>-sriov-xxx` `network-<iface>-br1` ?

marquiz:
Sorry, I haven't had the time to look this, yet, because of KubeCon

> If more and more network things arise and are needed do we want a logical grouping for those features as well, `network-<iface>-sriov-xxx` `network-<iface>-br1`

Yes, certainly. One (possible) problem I see is that labels do not provide good hierarchy, and, we  might end up cluttering the label space with a lot of labels.

As a quick thought it could be something like `network-if_<iface_name>.<attribute>=<value>` e.g. `network-if_br1.present=true` etc. And, I think this needs to be configurable

mmirecki:
@zvonkok No, this is the the first one.

mmirecki:
Please review this PR before looking at this one: [https://github.com/kubernetes-sigs/node-feature-discovery/pull/253](https://github.com/kubernetes-sigs/node-feature-discovery/pull/253)
That checks bridge capacity on a node, and reports it as node/status/capacity

source/network/network.go
```diff
@@ -81,6 +82,11 @@ func (s Source) Discover() (source.Features, error) {
 				}
 			}
 		}
+
+		if bridgeDir, _ := os.Stat("/sys/class/net/" + netInterface.Name + "/bridge"); bridgeDir != nil {
```
Ethyling: Not sure about this, but what do you think about something like:
```golang
if _, err := os.Stat("/sys/class/net/" + netInterface.Name + "/bridge"); err == nil {
        // Set label
} else if !os.IsNotExist(err) {
        // Add some error logging
}
```

marquiz:
I still fail to see where you would need the label if we report capacity (implemented in #253)

/remove-lifecycle stale

marquiz:
Any comments @mmirecki ?

/remove-lifecycle rotten

PR #253, opened 2019-05-27
Report feature capacity

mmirecki:
This is a POC PR, demonstrating a proposed feature to report node resource capacity.
The PR checks the nodes for linux bridges, and reports each linux bridge as available resource (with a capacity of 1000 per bridge). These resources are then reported to the master, which updates the /status/capacity section of the node.
For now, the PR only reports linux bridge resources. If proposal is accepted this will be extended by sriov, ovs bridges, etc..
The master will only update the node capacity the first time a worker reports it, and will remove it when the worker stops reporting it. This is done in order to not overwrite changes to the reported capacity when pods utilizing this resource are created.
The capacity reporting feature will be disabled on a node by default. It can be enabled by adding the following parameter: enable-capacity-reporting

This functionality is needed by the kubevirt project. The functionality provided by the the PR is currently provided by: https://github.com/kubevirt/bridge-marker
Measuring other resources is done by other, similar projects. With the changes in the PR we will hopefully be able to replace all these with nfd.


pkg/nfd-master/nfd-master.go
```diff
@@ -282,6 +289,14 @@ func (s *labelerServer) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*p
 			stderrLogger.Printf("failed to advertise labels: %s", err.Error())
 			return &pb.SetLabelsReply{}, err
 		}
+
+		if len(capacities) > 0 {
+			err := updateNodeCapacity(s.apiHelper, r.NodeName, capacities)
+			if err != nil {
+				stderrLogger.Printf("failed to update node capcity: %s", err.Error())
```
zvonkok:
```suggestion
				stderrLogger.Printf("failed to update node capacity: %s", err.Error())
```

pkg/nfd-master/nfd-master.go
```diff
@@ -363,3 +378,77 @@ func addAnnotations(n *api.Node, annotations map[string]string) {
...
+	Op    string      `json:"op"`
+	Path  string      `json:"path"`
+	Value interface{} `json:"value,omitempty"`
+}
+
+func updateNodeCapacity(helper apihelper.APIHelpers, nodeName string, capacities map[string]string) error {
+	patchOperations := make([]patchOperation, 0)
+	reportedCapacities, err := getReportedCapacity(helper, nodeName)
+	if err != nil {
+		stderrLogger.Printf("Failed to retrieve exising node capacity: %s", nodeName, err.Error())
```
zvonkok:
```suggestion
		stderrLogger.Printf("Failed to retrieve existing node capacity: %s", nodeName, err.Error())
```

marquiz:
I took a quick look at this, some comments and questions below.

How does this relate to #252? I somehow like the extended resources more, in this case. What I don't like is the coupling with interface names, as they are totally user-specific. But, in the case of extended resources I don't think it may be ok.

Do you need the "bookkeeping" of ERs for something? How did you pick up 1000 as the capacity of bridge interfaces, round from Linux `BR_MAX_PORTS`? Do you have use case where you are hitting (close to) this limit?

Some other random comments and questions:
- I think the capacity update logic in nfd-master is too simple, e.g. in case of kubelet restart the capacity is re-set to 0 and we're stuck with that. Need to check if the old cap is lower than the new cap or smth
- We possibly would like to have some switch for enabling/disabling node status updates in nfd-master, too. Not just on nfd-worker
- Some unit tests would be needed, e.g. added to the `fake` sources
- RBAC rules (clusterrole definition) in nfd-master template should be updated to enable node status update
- `capacityNs` would need to be changed to something similar to `labelNs` or `annotationNs` (I'm not sure what exactly, yet :smile:)
- could/should the capacities in the gRPC interface passed as a `map[string]int64` instead of `map[string]string`?

In any case, I think we should discuss this in SIG Node in order to get some wider coverage

zvonkok:
I have a couple of use-cases in my mind that ER could solve.
My concern right now is, 
- how should one expose infinite resources as an extended resource? (count to -1?)
- should we even expose an infinite resource as an ER? 

The case for a countable resource is easy, what are we going to do with flags, enablements etc.

marquiz:
> The case for a countable resource is easy, what are we going to do with flags, enablements etc.

Wouldn't labels be correct solution for infinite resources, with the current APIs? Is there some use case for reference counting of infinite resources, or something else in your mind that ERs would provide but labels couldn't?

This might change in the future when/if the k8s resource model is updated/extended.

mmirecki:
> I took a quick look at this, some comments and questions below.
> 
> How does this relate to #252? 
#252 just reports the bridges as labels, this reports them as resources

> I somehow like the extended resources more, in this case. What I don't like is the coupling with interface names, as they are totally user-specific. But, in the case of extended resources I don't think it may be ok.

We need the coulping with bridge name. Kubevirt vm networks depend need the underlying bridges (in case of bridge-cni). Kubevirt sets the pod resources to one with a specific bridge, and then depends on k8s to find an appropriate node.

> 
> Do you need the "bookkeeping" of ERs for something? How did you pick up 1000 as the capacity of bridge interfaces, round from Linux `BR_MAX_PORTS`? Do you have use case where you are hitting (close to) this limit?

In case of bridges we care more about the bridge just being present (running out of resources is not really an issue). 1000 was taken as the os max value.

> 
> Some other random comments and questions:
> 
> * I think the capacity update logic in nfd-master is too simple, e.g. in case of kubelet restart the capacity is re-set to 0 and we're stuck with that. Need to check if the old cap is lower than the new cap or smth

I copied the code from what we have in kubevirt. I think in this case we can just overwrite whenever a new value is delivered.

> * We possibly would like to have some switch for enabling/disabling node status updates in nfd-master, too. Not just on nfd-worker
> * Some unit tests would be needed, e.g. added to the `fake` sources

Sure, will add tests once the general idea is accepted

> * RBAC rules (clusterrole definition) in nfd-master template should be updated to enable node status update
> * `capacityNs` would need to be changed to something similar to `labelNs` or `annotationNs` (I'm not sure what exactly, yet smile)
> * could/should the capacities in the gRPC interface passed as a `map[string]int64` instead of `map[string]string`?
> 
> In any case, I think we should discuss this in SIG Node in order to get some wider coverage

Will push an update to the PR asap.

marquiz:
> > I took a quick look at this, some comments and questions below.
> > How does this relate to #252?
> > #252 just reports the bridges as labels, this reports them as resources
> 
> > I somehow like the extended resources more, in this case. What I don't like is the coupling with interface names, as they are totally user-specific. But, in the case of extended resources I don't think it may be ok.

Accidental negation: what I meant to say above is that for extended resources the interface names are probably more acceptable IMO.


> We need the coulping with bridge name. Kubevirt vm networks depend need the underlying bridges (in case of bridge-cni). Kubevirt sets the pod resources to one with a specific bridge, and then depends on k8s to find an appropriate node.

So, if I understand this correctly, either labels or ERs should be enough.


> > Do you need the "bookkeeping" of ERs for something? How did you pick up 1000 as the capacity of bridge interfaces, round from Linux `BR_MAX_PORTS`? Do you have use case where you are hitting (close to) this limit?
> 
> In case of bridges we care more about the bridge just being present (running out of resources is not really an issue). 1000 was taken as the os max value.

Should it then be BR_MAX_PORTS, i.e. 1024? It would be good to document in the code from where this number is originated.

mmirecki:
> > > I took a quick look at this, some comments and questions below.
> > > How does this relate to #252?
> > > #252 just reports the bridges as labels, this reports them as resources
> > 
> > 
> > > I somehow like the extended resources more, in this case. What I don't like is the coupling with interface names, as they are totally user-specific. But, in the case of extended resources I don't think it may be ok.
> 
> Accidental negation: what I meant to say above is that for extended resources the interface names are probably more acceptable IMO.
> 
> > We need the coulping with bridge name. Kubevirt vm networks depend need the underlying bridges (in case of bridge-cni). Kubevirt sets the pod resources to one with a specific bridge, and then depends on k8s to find an appropriate node.
> 
> So, if I understand this correctly, either labels or ERs should be enough.

To an extend yes. The downside of labels is that you can not nicely track the usage, which you get out of the box for resources. Also, if we used labels for infinite resources, we would have to use two different mechanism (labels & resources) for node scheduling.

> 
> > > Do you need the "bookkeeping" of ERs for something? How did you pick up 1000 as the capacity of bridge interfaces, round from Linux `BR_MAX_PORTS`? Do you have use case where you are hitting (close to) this limit?
> > 
> > 
> > In case of bridges we care more about the bridge just being present (running out of resources is not really an issue). 1000 was taken as the os max value.
> 
> Should it then be BR_MAX_PORTS, i.e. 1024? It would be good to document in the code from where this number is originated.
Done

mmirecki:
Unit tests were added.

PR #254, opened 2019-05-29
source/memory: detect NVDIMM DAX mode

marquiz:
Extend NVDIMM (non-volatile DIMM) discovery by adding detection of DAX
mode, i.e. detection of regions in DAX/AppDirect mode.
The new label is:
    feature.node.kubernetes.io/memory-nv.dax: true

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,7 @@ See [configuration options](#configuration-options) for more information.
 | ------- | --------- | ------------------------------------------------------ |
 | numa    | <br>      | Multiple memory nodes i.e. NUMA architecture detected
 | nv      | present   | NVDIMM device(s) are present
+| nv      | dax       | NVDIMM region(s) configured in DAX mode are present
 
 ### Network Features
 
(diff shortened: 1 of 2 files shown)


marquiz:
Not fully verified

/hold

marquiz:
@pohly, @okartau, are you able to verify if this works for you?

okartau:
> are you able to verify if this works for you?

I tried running code this PR in a cluster where one node has Optane DC memory which is already configured in AppDirect mode. There is another node in cluster without Optane DC memory.
In both cases the code works as expected:
- on Optane-equipped node there is label added:
feature.node.kubernetes.io/memory-nv.dax=true

- on node(s) without Optane DC dimms there is no such label.

okartau:
I also tried the case where Optane DC device is present but set to MemoryMode=100% i.e. no DAX-capable storage space is configurable.
In that case the node in question gets memory-nv presence label:
feature.node.kubernetes.io/memory-nv.present=true

and no dax label (as there is no entry named `*dax*` under /sys/bus/nd/devices/)

marquiz:
So this is now verified. Thanks @okartau

/hold cancel

PR #256, opened 2019-05-31
Makefile: create default yamls, configurable namespace, docker-push

askervin:
- Create default yamls for deploying nfd-master and
  nfd-worker-daemonset.
- Use kube-system namespace by default.
- Configurable namespace:
  make IMAGE_REGISTRY=myhost:5000 K8S_NAMESPACE=my-nfd-devel
- Add standard rule for pushing docker image (make docker-push).


Makefile
```diff
@@ -9,18 +9,32 @@ IMAGE_NAME := node-feature-discovery
 IMAGE_TAG_NAME := $(VERSION)
 IMAGE_REPO := $(IMAGE_REGISTRY)/$(IMAGE_NAME)
 IMAGE_TAG := $(IMAGE_REPO):$(IMAGE_TAG_NAME)
-
+K8S_NAMESPACE := kube-system
 
 all: image
 
-image:
+image: nfd-master.yaml nfd-worker-daemonset.yaml
```
marquiz: We have other yaml, too, namely `nfd-worker-job.yaml.template` and `nfd-daemonset-combined.yaml.template`

Makefile
```diff
@@ -9,18 +9,32 @@ IMAGE_NAME := node-feature-discovery
...
 	$(IMAGE_BUILD_CMD) --build-arg NFD_VERSION=$(VERSION) \
 		-t $(IMAGE_TAG) ./
 
+%.yaml: %.yaml.template FORCE
+	@echo "$@: namespace: ${K8S_NAMESPACE}"
+	@echo "$@: image: ${IMAGE_TAG}"
+	@sed -E \
+	     -e s',^(\s*)name: node-feature-discovery,\1name: ${K8S_NAMESPACE},' \
```
marquiz: Could it be made even a bit more obvious what line in the template(s) this expression targets? E.g. having something like this in the template and the sed expression:
```
name: node-feature-discovery  # NFD namespace
```

Makefile
```diff
@@ -9,18 +9,32 @@ IMAGE_NAME := node-feature-discovery
...
+	     -e s',^(\s*)image:.+$$,\1image: ${IMAGE_TAG},' \
+	     -e s',^(\s*)namespace:.+$$,\1namespace: ${K8S_NAMESPACE},' \
+	     $< > $@
+
+docker-push: image
+	docker push "${IMAGE_TAG}"
```
marquiz: I have a corresponding addition in #160. We support other image builders than docker, too, so I think the make target should be `image-push`. Also, the push command should be configurable, e.g. `$(IMAGE_PUSH_CMD)`

Makefile
```diff
@@ -9,18 +9,32 @@ IMAGE_NAME := node-feature-discovery
...
+K8S_NAMESPACE := kube-system
 
 all: image
 
-image:
+image: nfd-master.yaml nfd-worker-daemonset.yaml
 	$(IMAGE_BUILD_CMD) --build-arg NFD_VERSION=$(VERSION) \
 		-t $(IMAGE_TAG) ./
 
+%.yaml: %.yaml.template FORCE
```
marquiz: How about just putting the names of the YAMLs in a variable and creating a phony target for generating them? Would get rid of the `FORCE` target. WDYT?

Makefile
askervin: Good point. Added `yaml_templates` variable to Makefile to find them all.

Makefile
askervin: Variables added.

Phony target (yamls) added. (Good idea!)

We won't get rid of `FORCE`, however, because yaml instances depend on both templates and configuration variable values (`IMAGE_TAG` and `K8S_NAMESPACE`). Dropping `FORCE` would easily lead situations like:
- yamls would use older image than the latest build (example: user changed `IMAGE_REGISTRY` for the latest build)
- `make K8S_NAMESPACE=nfd-devel nfd-master.yaml` would report `...is already up to date` yet the yaml would actually contain different namespace.

Makefile
askervin: Thanks! This fixes the thing that bothered me a bit. :)

Makefile
askervin: As #160 seems to be going in, I dropped `docker-push` completely from this patch.
(I copied `docker-push:` practice from projects created with `kubebuilder`.)
Changed `README.md` examples to use `make push`, that is implemented in #160.

README.md
```diff
@@ -436,16 +436,17 @@ If you want to use the latest development version (master branch) you need to
 
 #### nfd-master
 
-Nfd-master runs as a DaemonSet, by default in the master node(s) only. You can
-use the template spec provided to deploy nfd-master. If deploying a custom
-image, you need to update the template to use the correct image (for the latest
-released version, omit the `sed` part):
+Nfd-master runs as a DaemonSet, by default in the master node(s)
+only. You can use the template spec provided to deploy nfd-master, or
+use ``nfd-master.yaml`` generated by ``Makefile``. The latter includes
```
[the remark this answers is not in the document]
askervin: They definitely seem to be enough. `Brains fixed.`

Makefile
marquiz: > We won't get rid of `FORCE`

True, my bad. I thought you could nicely generate all yamls with one _simple_ command (like `sed -i`), but, that is not possible because you need to copy files

README.md
```diff
@@ -592,7 +594,7 @@ make
 Optional, this example with Docker.
 
 ```
-docker push <image registry>/<image-name>:<version>
+make push
```
marquiz: Remove this change as #160 is not in, yet.

Makefile
```diff
@@ -9,18 +9,34 @@ IMAGE_NAME := node-feature-discovery
...
 IMAGE_TAG := $(IMAGE_REPO):$(IMAGE_TAG_NAME)
+K8S_NAMESPACE := kube-system
 
+yaml_templates := $(wildcard *.yaml.template)
+yaml_instances := $(patsubst %.yaml.template,%.yaml,$(yaml_templates))
 
 all: image
 
-image:
+image: $(yaml_instances)
```
marquiz: Couldn't/shouldn't `image` depend simply on the `yamls` target?

Makefile
```diff
@@ -9,18 +9,34 @@ IMAGE_NAME := node-feature-discovery
...
+
 mock:
 	mockery --name=FeatureSource --dir=source --inpkg --note="Re-generate by running 'make mock'"
 	mockery --name=APIHelpers --dir=pkg/apihelper --inpkg --note="Re-generate by running 'make mock'"
 	mockery --name=LabelerClient --dir=pkg/labeler --inpkg --note="Re-generate by running 'make mock'"
 
 test:
 	go test ./cmd/... ./pkg/...
+
+FORCE:
```
marquiz: Very subtle detail, but I think `.FORCE` (with a dot) would be a better name as it's not a "real" target, similar to `.PHONY`. You could also move the definition to the top of the file. What do you think?

Makefile
askervin: Yup, it should. Thanks!

Makefile
askervin: Good one, done.

README.md
askervin: Ok

askervin: Replaced all `make ... push` lines with `docker push ...`.

askervin:
`label-nodes.sh` migrated to support new workflow.

PR #258, opened 2019-06-10
kernel: fix undefined symbol error

marquiz:
Fix a build failure that slipped through when adding support for
non-binary kconfig options in #197,


Ethyling:
Oh sorry, missed this one... Was testing with the wrong import...
Anyway, lgtm

PR #260, opened 2019-07-02
nfd-worker: try connecting to nfd-master for 60s

marquiz:
Instead of erroring out right away, try to connect to nfd-master for 60
seconds until giving up.


zvonkok:
This is helpful, most deployments of the nfd-worker CrashLoopBack right away on my clusters.

PR #261, opened 2019-07-17
Add RDMA capable feature to PCI Features

adrianchiris:
This commit adds a new PCI feature : rdma.capable
If present it means that the node has RDMA capable PCI devices.

This can be used by other entities (e.g operators) to trigger loading
of RDMA modules on the node.

Currently we determine that by comparing the PCI vendor ID.
This can later be adapted to accomodate other vendors/devices.

### DIFF
--- a/source/pci/pci.go
+++ b/source/pci/pci.go
@@ -40,6 +40,10 @@ var Config = NFDConfig{
 
 var devLabelAttrs = []string{"class", "vendor", "device", "subsystem_vendor", "subsystem_device"}
 
+var rdmaCapablePCIVendorIds = []string{"15b3"}
+
+const rdmaFeatureCapable = "rdma.capable"
+
 // Implement FeatureSource interface
 type Source struct{}
 
@@ -99,6 +103,10 @@ func (s Source) Discover() (source.Features, error) {
 		}
 	}
 
+	if hasRdmaCapableDevice(devs) {
+		features[rdmaFeatureCapable] = true
+	}
+
 	return features, nil
 }
 
@@ -145,4 +153,26 @@ func detectPci() (map[string][]pciDeviceInfo, error) {
 	}
 
 	return devInfo, nil
+}
+
+// Check if the system has a remote DMA capable device.
+// Use PCI vendor ID for now.
+func hasRdmaCapableDevice(devs map[string][]pciDeviceInfo) bool {
+	for _, classDevs := range devs {
+		for _, dev := range classDevs {
+			if contains(rdmaCapablePCIVendorIds, dev["vendor"]) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func contains(slice []string, elem string) bool {
+	for _, e := range slice {
+		if e == elem {
+			return true
+		}
+	}
+	return false
 }
--- a/source/pci/pci.go
+++ b/source/pci/pci.go
@@ -21,6 +21,7 @@ import (
 	"io/ioutil"
 	"log"
 	"path"
+	"regexp"
 	"strings"
 
 	"sigs.k8s.io/node-feature-discovery/source"
@@ -31,16 +32,30 @@ type pciDeviceInfo map[string]string
 type NFDConfig struct {
 	DeviceClassWhitelist []string `json:"deviceClassWhitelist,omitempty"`
 	DeviceLabelFields    []string `json:"deviceLabelFields,omitempty"`
+	RdmaCapableDevices   []string `json:"rdmaCapableDevices,omitempty"`
 }
 
 var Config = NFDConfig{
 	DeviceClassWhitelist: []string{"03", "0b40", "12"},
 	DeviceLabelFields:    []string{"class", "vendor"},
+	RdmaCapableDevices:   []string{"15b3:.*"},
 }
 
 var devLabelAttrs = []string{"class", "vendor", "device", "subsystem_vendor", "subsystem_device"}
 
-var rdmaCapablePCIVendorIds = []string{"15b3"}
+type rdmaCapableDev struct {
+	vendorID      string
+	deviceIDRegex string
+}
+
+func (r rdmaCapableDev) match(vendor, device string) bool {
+	match, err := regexp.MatchString(r.deviceIDRegex, device)
+	if err != nil {
+		log.Printf("WARNING: Failed to match on deviceIDRegex: %s", err.Error())
+		return false
+	}
+	return vendor == r.vendorID && match
+}
 
 const rdmaFeatureCapable = "rdma.capable"
 
@@ -155,23 +170,30 @@ func detectPci() (map[string][]pciDeviceInfo, error) {
 	return devInfo, nil
 }
 
-// Check if the system has a remote DMA capable device.
-// Use PCI vendor ID for now.
+// Check if the system has a remote DMA capable PCI device.
 func hasRdmaCapableDevice(devs map[string][]pciDeviceInfo) bool {
+
+	var rdmaCapableDevs []rdmaCapableDev
+
+	// Parse configurations and create a list of RDMA capable devices
+	for _, identifier := range Config.RdmaCapableDevices {
+		// Identifier is in <vendorID:deviceIDRegex> format.
+		out := strings.Split(identifier, ":")
+		if len(out) != 2 {
+			log.Printf("WARRNING: invalid RDMA Capable device identifier: %s", identifier)
+			continue
+		}
+		rdmaCapableDevs = append(rdmaCapableDevs, rdmaCapableDev{out[0], out[1]})
+	}
+
+	// Search for a match on the system's PCI devices
 	for _, classDevs := range devs {
 		for _, dev := range classDevs {
-			if contains(rdmaCapablePCIVendorIds, dev["vendor"]) {
-				return true
+			for _, rdma := range rdmaCapableDevs {
+				if rdma.match(dev["vendor"], dev["device"]) {
+					return true
+				}
 			}
-		}
-	}
-	return false
-}
-
-func contains(slice []string, elem string) bool {
-	for _, e := range slice {
-		if e == elem {
-			return true
 		}
 	}
 	return false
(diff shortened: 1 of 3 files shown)


source/pci/pci.go
```diff
@@ -40,6 +40,10 @@ var Config = NFDConfig{
 
 var devLabelAttrs = []string{"class", "vendor", "device", "subsystem_vendor", "subsystem_device"}
 
+var rdmaCapablePCIVendorIds = []string{"15b3"}
```
zvonkok: Are all Mellaonx cards RDMA capable?

source/pci/pci.go
```diff
@@ -146,3 +154,25 @@ func detectPci() (map[string][]pciDeviceInfo, error) {
...
 }
+
+// Check if the system has a remote DMA capable device.
+// Use PCI vendor ID for now.
+func hasRdmaCapableDevice(devs map[string][]pciDeviceInfo) bool {
+	for _, classDevs := range devs {
+		for _, dev := range classDevs {
+			if contains(rdmaCapablePCIVendorIds, dev["vendor"]) {
+				return true
+			}
```
zvonkok: For other vendors, we might need to filter by more PCI attributes, e.g. Intel has a variety of different cards.

adrianchiris: I can filter by  both vendor id and device ID as they uniquely identify a PCI device.
allowing "wild card"/regex for the device ID.

source/pci/pci.go
adrianchiris: Yes, they are.

moshe010:
any update on this PR?

kad:
In theory, this kind of detection based solely on PCI device IDs is not really reliable. Neither it shows real state of the kernel support. Do we have any flag in e.g. sysfs, which can say that device is rdma capable and kernel is ready to support it?

adrianchiris:
So, all RDMA related sysfs files are located under /dev/infiniband/
these files (and folder) are created when RDMA modules are loaded if there are RDMA capable devices in the system (assuming low level vendor drivers implemented required kernel APIs)

we would say that a node is rdma capable if it has a PCI device that is RDMA capable
we would say that a node is rdma enabled/configured if those RDMA modules were loaded
this PR deals with the former.

marquiz:
I must apologise for slow response on this. My excuse is long summer holidays...

Anyway, I'm a bit hesitant to merge this patch. The reason is that I see the current approach as effectively starting to maintain a device database in NFD. That is, a list of vendors/devices that support RDMA. And what if Mellanox would produce a card without RDMA capability? I'd very much like to see RDMA capability detected from something that the kernel directly reports, (e.g. under `/sys/class/net/` or somewhere).

If we're just looking at Mellanox cards, would it be enough to just use the pci label for now. Until we find a better/more general way to detect it. E.g. enable device class `0200` in config and look at the `
feature.node.k8s.io/pci-0200_15b3.present` label?

What do you think?

adrianchiris:
Hi,
The devices that their presence will cause NFD to mark the node as rdma.capable are specified in configuration file(nfd-worker.conf), so no database will be maintained in NFD.

Unfortunately, today, there is no kernel sysfs to indicate rdma capability.
Indication from kernel is only available after the relevant RDMA modules are loaded, which is what we are trying to do (load RMDA modules on nodes marked with rdma.capable)

Mellanox sees RDMA as one of its core technologies. It is unlikely that its going to be abandoned in future products (for obvious reasons i cannot guarantee it).

The in-code defaults can be removed if that's a concern.

kad:
> Indication from kernel is only available after the relevant RDMA modules are loaded, which is what we are trying to do (load RMDA modules on nodes marked with rdma.capable)
Would it be so that set of modules will be different for each vendor of RDMA capable HW ?

marquiz:
> The devices that their presence will cause NFD to mark the node as rdma.capable are specified in configuration file(nfd-worker.conf), so no database will be maintained in NFD.
> 
> Unfortunately, today, there is no kernel sysfs to indicate rdma capability.
> Indication from kernel is only available after the relevant RDMA modules are loaded, which is what we are trying to do (load RMDA modules on nodes marked with rdma.capable)

OK. Let's find a way to satisfy this use case.

What do you think about using the "raw" pci.present label (`feature.node.k8s.io/pci-0200_15b3.present`) I suggested above?


> Mellanox sees RDMA as one of its core technologies. It is unlikely that its going to be abandoned in future products (for obvious reasons i cannot guarantee it).
> 
> The in-code defaults can be removed if that's a concern.

This looks like a vendor-specific label to me, i.e. different driver modules would need to be loaded depending on the vendor. Am I right? In this case `rdma.present` would not necessarily be enough in the general case(?)

One thing that came into my mind would be extending the configurability in your patch a bit further. Allow any custom feature labels in the pci source based on similar regex or smth. That is, the configuration (of pci source) would contain something like
```
    customFeatures:
        - name: "rdma.mellanox.present"
        - deviceIdRegex: "15b3:.*"
```

This same mechanism could then be relatively easily used for GPU labels, too – for people that have been after more human readable labels. WDYT?

adrianchiris:
> This looks like a vendor-specific label to me, i.e. different driver modules would need to be loaded depending on the vendor. Am I right? In this case rdma.present would not necessarily be enough in the general case(?)

So, to my understanding, loading (the general) rdma kernel modules will trigger loading of vendor low level modules according to the devices in the system. (Will double check that)

Doing a generalization and extending this to custom features sounds good to me.

marquiz:
> Doing a generalization and extending this to custom features sounds good to me.

OK, I think this would be a useful feature


> The in-code defaults can be removed if that's a concern.

I think it would be best to leave the in-code defaults empty. Examples in the sample config would be very good, though.

I thought this a little further. Maybe we should allow multiple rules per label(?) Something like
```
    customFeatures:
        - name: "rdma.mellanox.present"
          rules:
            - deviceIdRegex: "0200:15b3:.*"
```
In this scheme there could be multiple rules (ORed) per label. And later, other rules than plain regexp could be implemented. Looks like we need to iterate a bit on the design, aiming for flexible enough and future proof "api". Any thoughts?

adrianchiris:
@marquiz We can have something like:

```
customFeatures:
  - name: "rdma.capable"
    matchOn:
      - pciIdRegex:
          - "[<class>:]<vendorID>:<deviceID>"  // class optional
          - "15b3:.*"
          - "1010:41\d+"
  - name: "rdma.available"
    matchOn:
      - kModLoaded :
          - "ib_core"
          - "rdma_cm"
```

customFeatures is a list of features.
a features specifies its name and on what it should match (matchOn)
`matchOn` is a list of "matchers" or "rules"

the `pciIdRegex` matcher/rule performs the matching based on a list of PCI ID regex.
you can have multiple matchers per feature. (Ored as you suggested)

In the future it will be possible to add new matchers if deemed worthy.
(e.g kModLoaded matcher,  filePath matcher)

Reading again now, its basically as you suggested only the PCI regex is a list :)

marquiz:
Thinking this further, it seems like this functionality should be moved out of `pci` into the `local` source (or entirely in its own source).

adrianchiris:
For CustomFeatures it would be best to have its own source as future matchers may not be pci related. 
For the PCI bits its possible to reuse code from `pci `package.

Thinking about RDMA again, as its a (standard) technology, I would like RDMA capability label (`rdma.capable`) to be a "first class citizen" like` sriov.capable` (and not depend on how a user chose to configure the label).
Acting on a well known label is much easier than relying on the user to configure its name properly.
The internal logic can go through the same path as CustomFeatures.
Does this make sense ?

P.S
Have doubled checked the loaded modules question, and loading RDMA upper layer modules will automatically trigger loading the low level vendor specific modules.

marquiz:
> @marquiz For CustomFeatures it would be best to have its own source as future matchers may not be pci related.

Yup.

> For the PCI bits its possible to reuse code from `pci `package.

Yes, that would be preferable


> Thinking about RDMA again, as its a standard technology, I would like RDMA capability label (`rdma.capable`) to be a "first class citizen" like` sriov.capable` (and not depend on how a user chose to configure the label).
> Acting on a well known label is much easier than relying on the user to configure its name properly.
> The internal logic can go through the same path as CustomFeatures.
> Does this make sense ?

I understand you point of view. We could include rdma rule(s) in the default configuration of the new feature source. Or, possibly create a list of const/fixed rules that are always run and cannot be overridden by the user (not by accident – at least we could have a config flag to disable these). WDYT?

> P.S
> Have doubled checked the loaded modules question, and loading RDMA upper layer modules will automatically trigger loading the low level vendor specific modules. @kad

OK, thanks for checking this

zvonkok:
@adrianchiris Any update from your side?

adrianchiris:
> I understand you point of view. We could include rdma rule(s) in the default configuration of the new feature source. Or, possibly create a list of const/fixed rules that are always run and cannot be overridden by the user (not by accident – at least we could have a config flag to disable these). WDYT?

@marquiz @zvonkok 
I lean towards the second option of having it defined as a part of a fixed/const list, blocking any attempt to override it.

Do you see a valid use-case of overriding the logic by the user ? if not, then allowing the user to override the rule could be another feature on top of it when the need arise.

marquiz:
> I lean towards the second option of having it defined as a part of a fixed/const list, blocking any attempt to override it.

K, I think this is justified



> Do you see a valid use-case of overriding the logic by the user ? if not, then allowing the user to override the rule could be another feature on top of it when the need arise.

I necessarily don't. I see a valid use case of adding new logic (for this label), and, that would be covered by the user-specific/custom rules. In other words there would be two sets of rules (*fixed* and *user-defined*), both of which are run, and the set of labels from those would be joined (an union).

If there really would be a need to disable the execution of the fixed ruleset we could add a flag for that later on.

adrianchiris:
So, i think we are in agreement of the overall requirement and high level design.
I will check internally when i can work on this Feature, then we can have a more in-detail discussion over the code itself.

marquiz:
Yes, nice 👍

marquiz:
@adrianchiris have you looked at this lately?

adrianchiris:
@marquiz, Hi.
I was caught up in other tasks. Hopefully i will have something ready for reviews by Feb end.

PR #266, opened 2019-08-29
source/cpu: support 'false' status of cpu-pstate.turbo

marquiz:
Some workloads may benefit from Intel Turbo Boost technology being
disabled. This patch sets the
'feature.node.kubernetes.io/cpu-pstate.turbo' label to 'false' if we can
detect that it has been disabled. If detection fails no label is
published.

### DIFF
--- a/source/cpu/pstate.go
+++ b/source/cpu/pstate.go
@@ -22,22 +22,23 @@ import (
 	"runtime"
 )
 
-// Discover returns feature names for p-state related features such as turbo boost.
-func turboEnabled() (bool, error) {
+// Discover p-state related features such as turbo boost.
+func detectPstate() (map[string]string, error) {
 	// On other platforms, the frequency boost mechanism is software-based.
 	// So skip pstate detection on other architectures.
 	if runtime.GOARCH != "amd64" && runtime.GOARCH != "386" {
-		return false, nil
+		return nil, nil
 	}
 
 	// Only looking for turbo boost for now...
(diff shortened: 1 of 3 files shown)


source/cpu/pstate.go
```diff
@@ -22,22 +22,23 @@ import (
...
+func detectPstate() (map[string]string, error) {
 	// On other platforms, the frequency boost mechanism is software-based.
 	// So skip pstate detection on other architectures.
 	if runtime.GOARCH != "amd64" && runtime.GOARCH != "386" {
-		return false, nil
+		return nil, nil
 	}
 
 	// Only looking for turbo boost for now...
 	bytes, err := ioutil.ReadFile("/sys/devices/system/cpu/intel_pstate/no_turbo")
```
Ethyling: Just to be sure, is this normal to use `/sys` and not `/host-sys`? We are only using `/host-sys` for the `selinux` label. Why this case is different?

marquiz: For `selinux` we need host-sys as `/sys/fs/selinux` is not mounted inside the nfd container, by default. For pstate we can rely on "default" sys which is better as the detection works even if `/host-sys` is not mounted.

Of course. for selinux we wouldn't need the whole host sysfs (selinuxfs would be enough) but that's another story...

Ethyling: Ok, thank you for the details!

PR #267, opened 2019-08-30
Added verify target to Makefile

zvonkok:
Adding verify target to Makefile to automatically format go code with `gofmt`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,8 @@ IMAGE_TAG := $(IMAGE_REPO):$(IMAGE_TAG_NAME)
 K8S_NAMESPACE := kube-system
 KUBECONFIG :=
 
+GOFMT_CHECK=$(shell find . -not \( \( -wholename './.*' -o -wholename '*/vendor/*' \) -prune \) -name '*.go' | sort -u | xargs gofmt -s -l)
+
 yaml_templates := $(wildcard *.yaml.template)
 yaml_instances := $(patsubst %.yaml.template,%.yaml,$(yaml_templates))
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -48,5 +48,19 @@ e2e-test:
 	dep ensure -v
 	go test -v ./test/e2e/ -args -nfd.repo=$(IMAGE_REPO) -nfd.tag=$(IMAGE_TAG_NAME) -kubeconfig=$(KUBECONFIG)
 
+verify:	verify-gofmt
```
Ethyling: Some details, but I think you should replace this tab by a space

Makefile
```diff
@@ -48,5 +50,19 @@ e2e-test:
...
```
zvonkok:
```suggestion
verify: verify-gofmt
```

Makefile
zvonkok: Done.

Makefile
```diff
@@ -15,6 +15,8 @@ IMAGE_TAG := $(IMAGE_REPO):$(IMAGE_TAG_NAME)
 K8S_NAMESPACE := kube-system
 KUBECONFIG :=
 
+GOFMT_CHECK=$(shell find . -not \( \( -wholename './.*' -o -wholename '*/vendor/*' \) -prune \) -name '*.go' | sort -u | xargs gofmt -s -l)
```
marquiz: You're already doing the gofmt check here. Could that be moved to the actual make target? I.e. have something like `GO_FILES` here and then running `gofmt -s -l $(GO_FILES)` in the make target

Makefile
```diff
@@ -48,5 +50,19 @@ e2e-test:
 	dep ensure -v
 	go test -v ./test/e2e/ -args -nfd.repo=$(IMAGE_REPO) -nfd.tag=$(IMAGE_TAG_NAME) -kubeconfig=$(KUBECONFIG)
 
+verify:	verify-gofmt
+
+verify-gofmt:
```
marquiz: .PHONY target as well(?)

Makefile
marquiz: I think this should be a .PHONY target

PR #279, opened 2020-01-27
vendor: update klauspost/cpuid

askervin:
Update cpuid from v1.2.2 to v1.2.3. Brings in SGX improvements and
CPUID leaf 7 feature detection (VBMI2, VPOPCNTDQ, GFNI, VAES,
AVX512BITALG, VPCLMULQDQ, AVX512BF16, AVX512VP2INTERSECT).

Signed-off-by: Antti Kervinen <antti.kervinen@intel.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,7 +37,7 @@ require (
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/json-iterator/go v1.1.7 // indirect
 	github.com/jtolds/gls v4.20.0+incompatible // indirect
-	github.com/klauspost/cpuid v1.2.2-0.20191106130511-5a626f7029c9
+	github.com/klauspost/cpuid v1.2.3
 	github.com/munnerz/goautoneg v0.0.0-20190414153302-2ae31c8b6b30 // indirect
 	github.com/mxk/go-flowrate v0.0.0-20140419014527-cca7078d478f // indirect
 	github.com/onsi/ginkgo v1.10.1
--- a/go.sum
+++ b/go.sum
@@ -198,6 +198,8 @@ github.com/klauspost/cpuid v1.2.0 h1:NMpwD2G9JSFOE1/TJjGSo5zG7Yb2bTe7eq1jH+irmeE
 github.com/klauspost/cpuid v1.2.0/go.mod h1:Pj4uuM528wm8OyEC2QMXAi2YiTZ96dNQPGgoMS4s3ek=
 github.com/klauspost/cpuid v1.2.2-0.20191106130511-5a626f7029c9 h1:bBK4GTyzBQHZWkqWwl5j2Wh2GECowFzo0om/r98FCII=
 github.com/klauspost/cpuid v1.2.2-0.20191106130511-5a626f7029c9/go.mod h1:Pj4uuM528wm8OyEC2QMXAi2YiTZ96dNQPGgoMS4s3ek=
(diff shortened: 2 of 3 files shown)


askervin:
Tested builds/platforms:
- Separate nfd-worker build: `go build -ldflags...` under `cmd/nfd-worker`. Ran the binary on a legacy and a future CPU.
- Full image build with `make ... push`, deployed nfd master and worker daemonsets to a Qemu cluster. Checked that the worker binary includes the latest cpuid code.

mythi:
I'd propose we drop `cpuid.SGX` from the default blacklist.

mythi:
> I'd propose we drop cpuid.SGX from the default blacklist.

my arguments for this is that: 1) this update brings new SGX cpuid bits so it's better either accept them all or have them all blacklisted, 2) in we #130 we discussed whether this bit alone adds any value but after all, it's `cpuid.SGX` so it servers the purpose it promises.

marquiz:
> my arguments for this is that: 1) this update brings new SGX cpuid bits so it's better either accept them all or have them all blacklisted, 2) in we #130 we discussed whether this bit alone adds any value but after all, it's `cpuid.SGX` so it servers the purpose it promises.

Maybe we should just blacklist the new sgx labels, too? Maybe it delivers what it promises but if the promise is "too weak" the information is rather indifferent for the end use. I still think that it's better to avoid cluttering the label space with misleading('ish) labels.

Is there a use for the cpuid.SGX label in a scenario where SGX would be disabled from the system? Or maybe we should re-open #130 and implement something more useful?

mythi:
> I still think that it's better to avoid cluttering the label space with misleading('ish) labels.

OTOH, this is not very common bit these days so it's unlikely to show up to many.

> Is there a use for the cpuid.SGX label in a scenario where SGX would be disabled from the system? 

Taking a closer look in to the SDM, it says `cpuid.SGX` is a pre-condition for users to check whether SGX is properly configured: if `cpuid.SGX`, cpuid leaf 12h is available which gives the final result.

Therefore, `cpuid.SGX` alone is already valid.

> Or maybe we should re-open #130 and implement something more useful?

we should re-open #130 and maybe implement another source plugin that returns detailed SGX info from klauspost/cpuid `hasSGX()`.

marquiz:
> Taking a closer look in to the SDM, it says `cpuid.SGX` is a pre-condition for users to check whether SGX is properly configured: if `cpuid.SGX`, cpuid leaf 12h is available which gives the final result.
> 
> Therefore, `cpuid.SGX` alone is already valid.

Hmm, valid in which sense? SGX may still be disabled (and it is, by default?) even if cpuid.SGX is true.

I'm just trying to figure out if there's a use case or some new angle that would invalidate conclusion made in the discussions under #130 😄

mythi:
@marquiz valid in a sense that someone who sees the bit could then check further details using `cpuid.12h`

I'm OK we digest this further, blacklist `cpuid.SGX_LE` and get the PR merged. WDYT?

marquiz:
> I'm OK we digest this further, blacklist `cpuid.SGX_LE` and get the PR merged. WDYT?

Yeah, let's do this.

We can re-open #130 and continue the discussion there.

askervin:
force-pushed change: SGXLC blacklisted as discussed in
https://github.com/kubernetes-sigs/node-feature-discovery/pull/279#issuecomment-579217600

mythi:
+1

PR #280, opened 2020-02-04
Update go version in Dockerfile

ArangoGutierrez:
Update go version in Docker build image to v1.13.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - "1.12"
+  - "1.13.5"
 
 env:
   global:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build node feature discovery
-FROM golang:1.12 as builder
+FROM golang:1.13.5 as builder
 
 # Get (cache) deps in a separate layer
 COPY go.mod go.sum /go/node-feature-discovery/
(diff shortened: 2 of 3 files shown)


marquiz:
Thank you for the patch. I think we should keep travis.yml in sync with the Dockerfile so that would need to be updated, too.

ArangoGutierrez:
sure, will patch the `.travis.yml` file as well

ArangoGutierrez:
@marquiz  you think we should do 
```yaml
language: go
go:
  - "1.13"
  - "1.13.x"
```
or just 1.13?

marquiz:
I think one version should be enough. We could stay in full sync with the dockerfile and use `1.13.5`(?)

ArangoGutierrez:
done

marquiz:
Hmm, end-to-end tests fail with go 1.13 🤔 

Looks like the kubernetes version we're using as a dep (the e2e test framework from it) is incompatible with go 1.13. We should update the dep(s) to k8s v1.17 and try with that.

marquiz:
Updating the deps wasn't quite that simple. I just submitted #282 to do that. We should merge that first and then try your PR on top of it.

ArangoGutierrez:
thanks for #282 , rebased and updated

Issue #281, opened 2020-02-05
unable to run e2e test


ArangoGutierrez:
When trying `make e2e-test` cmd I run into a flag error

```bash
[eduardo@fedora-ws node-feature-discovery]$ make e2e-test -e KUBECONFIG=${KUBECONFIG}
go test -v ./test/e2e/ -args -nfd.repo=quay.io/eduardoarango/node-feature-discovery -nfd.tag=v0.5.0-11-g8fcd7c1-dirty -kubeconfig=[...]
flag provided but not defined: -test.timeout
Usage of /tmp/go-build485178007/b001/e2e.test:
 [...]
FAIL	sigs.k8s.io/node-feature-discovery/test/e2e	0.041s
FAIL
make: *** [Makefile:50: e2e-test] Error 1
```

I am running a `go version go1.13.5 linux/amd64` over 8fcd7c1abd2b990ade7777b9b48fc2ba5059cf37

ArangoGutierrez:
ref: #280

sjug:
Seeing this as well on `master` with `go version go1.13.7 linux/amd64`

marquiz:
Older Kubernetes versions (specifically the e2e test framework) was incompatible with golang 1.13. Updating to k8s v1.17.x fixes this, see #282

ArangoGutierrez:
can confirm  #282 patch works 
```bash
Ran 1 of 1 Specs in 29.401 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestE2E (29.40s)
PASS
ok  	sigs.k8s.io/node-feature-discovery/test/e2e	29.415s
```

PR #282, opened 2020-02-05
go.mod: update kubernetes and its deps to v1.17.2

marquiz:
Also syncs e2e tests with the Kubernetes e2e test framework.


zvonkok:
@marquiz The backward compatibility for the API resources would be >=1.15, right? 
Do we have somewhere stated which version of NFD works with which version of Kubernetes? 
Going forward we might also update `apiVersion` of all resources used to deploy NFD.

go.mod
```diff
@@ -3,96 +3,55 @@ module sigs.k8s.io/node-feature-discovery
 go 1.12
```
ArangoGutierrez:
```suggestion
go 1.13
```

ArangoGutierrez:
closes #281

zvonkok:
@ArangoGutierrez thanks for testing, see comment https://github.com/kubernetes-sigs/node-feature-discovery/issues/281#issuecomment-582548851 

/lgtm

PR #283, opened 2020-02-10
pci: include device information to co-processor class device labels

askervin:
Deploying [the QAT plugin](https://github.com/intel/intel-device-plugins-for-kubernetes) *only* on nodes with the QuickAssist Technology HW needs more information on node labels.  Currently available PCI device `class` and `vendor` are not sufficient.

This patch adds PCI `device` information to labels in case a PCI device is a co-processor, that is, devices in class `0b40` .

Tested on a node with a QuickAssist Technology coprocessor and NVidia graphics card:
```
2020/02/10 14:22:46 pci-37c8_0b40_8086.present = true
2020/02/10 14:22:46 pci-0300_10de.present = true
```

### DIFF
--- a/source/pci/pci.go
+++ b/source/pci/pci.go
@@ -29,19 +29,41 @@ import (
 type pciDeviceInfo map[string]string
 
 type NFDConfig struct {
-	DeviceClassWhitelist []string `json:"deviceClassWhitelist,omitempty"`
-	DeviceLabelFields    []string `json:"deviceLabelFields,omitempty"`
+	DeviceClassWhitelist      []string            `json:"deviceClassWhitelist,omitempty"`
+	DeviceLabelFieldsForClass map[string][]string `json:"deviceLabelFieldsForClass"`
+	DeviceLabelFields         []string            `json:"deviceLabelFields,omitempty"`
 }
 
 var Config = NFDConfig{
 	DeviceClassWhitelist: []string{"03", "0b40", "12"},
-	DeviceLabelFields:    []string{"class", "vendor"},
+	// Custom fields for selected classes
+	DeviceLabelFieldsForClass: map[string][]string{
+		"0b40": {"device", "class", "vendor"}, // Class 0b40: Co-processor
+	},
+	// Default fields for the rest of the classes
+	DeviceLabelFields: []string{"class", "vendor"},
 }
 
-var devLabelAttrs = []string{"class", "vendor", "device", "subsystem_vendor", "subsystem_device"}
+var devLabelAttrs = []string{} // All device fields mentioned in NFDConfig
 
 // Implement FeatureSource interface
 type Source struct{}
+
+func init() {
+	// Initialize devLabelAttrs to include all fields in NFDConfig
+	seenFields := map[string]bool{}
+	for _, fieldList := range Config.DeviceLabelFieldsForClass {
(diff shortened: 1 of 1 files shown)


marquiz:
Hmm, I'm not quite sure about this. I would not change the default configuration that we currently have lightly.

Is there a problem with deploying the device plugin on all nodes with *any* Intel co-processor? I.e.doing some unnecessary deployments with some other Intel co-processor than QAT. One can always change the NFD configuration if absolutely optimal device plugin deployment is required

mythi:
> Is there a problem with deploying the device plugin on all nodes with any Intel co-processor?

That's why we have this PR. With QAT we want to be able to "pin" to a specific PCI device ID and give the plugin the device specific parameters. This would be used, for instance, to guarantee that `qat.intel.com/generic` resource is always a specific device...

That's not a problem with NFD as such because we could change the config to `device_vendor`. The "problem" triggers with GPU + QAT co-existence: if we want to deploy QAT using `device_vendor`, we cannot deploy GPU with `class_vendor` anymore.

This PR allows to add labels for all devices in a particular device class.

marquiz:
> That's why we have this PR. With QAT we want to be able to "pin" to a specific PCI device ID and give the plugin the device specific parameters. This would be used, for instance, to guarantee that `qat.intel.com/generic` resource is always a specific device

Shouldn't the device plugin be able to handle the device detection by itself?


> That's not a problem with NFD as such because we could change the config to `device_vendor`. The "problem" triggers with GPU + QAT co-existence: if we want to deploy QAT using `device_vendor`, we cannot deploy GPU with `class_vendor` anymore.

If we go this route I would possibly add this on the side as a "parallel" feature with even a bit more flexible configuration. E.g. something like this
```
pci:
  deviceClassWhitelist: ["03", "0b40"]
  deviceLabelFields: ["class", "vendor"]
  extraDeviceLabels:
    - match:
        class: "0b40"
        vendor: "8086"
      fields: ["class", "vendor", "device"]
```
...or then this might be a bad idea, don't know yet, have to think about this 😉

mythi:
> Shouldn't the device plugin be able to handle the device detection by itself?

The user needs to configure whether he wants to detect QAT version X instead of QAT version Y. We can have a YAML that deploys the plugin with that setting but to make the plugin run only on the nodes that have the right version we want to use NFD labels.

marquiz:
> The user needs to configure whether he wants to detect QAT version X instead of QAT version Y. We can have a YAML that deploys the plugin with that setting but to make the plugin run only on the nodes that have the right version 

Is it a problem if the plugin runs also on nodes having version Y? I understand the waste of one unnecessary pod per node but is there something else?

However, I also understand the "coarse vs. fine-grained label co-existence" problem you described above. Thus, I think that a solution might be needed. I wouldn't necessarily include that in the default configuration, though.

askervin:
In my perspective, this PR is obsolete after merging PR #290. _Custom features_ offer PCI matching that enables matching all class/device/vendor PCI IDs, which was the original driving factor for this PR.

However, yet there is overlap, the configuration suggested by @marquiz in https://github.com/kubernetes-sigs/node-feature-discovery/pull/283#issuecomment-584592738 would be more general and possibly useful pattern for many sources to configure the level of detail in exposed data.

Closing this PR for now.

PR #285, opened 2020-02-11
README: better document the usage of hooks and feature files

marquiz:
Mention supported runtimes that the hooks can use and describe how hooks
and feature files can be used for injecting labels from other pods.

### DIFF
--- a/README.md
+++ b/README.md
@@ -335,8 +335,11 @@ feature sources.
 
 The *local* feature source gets its labels by two different ways:
 * It tries to execute files found under `/etc/kubernetes/node-feature-discovery/source.d/`
-directory. The hook files must be executable. When executed, the hooks are
-supposed to print all discovered features in `stdout`, one per line.
+directory. The hook files must be executable and they are supposed to print all
+discovered features in `stdout`, one per line. With ELF binaries static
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -336,7 +336,9 @@ feature sources.
 The *local* feature source gets its labels by two different ways:
 * It tries to execute files found under `/etc/kubernetes/node-feature-discovery/source.d/`
 directory. The hook files must be executable. When executed, the hooks are
-supposed to print all discovered features in `stdout`, one per line.
+supposed to print all discovered features in `stdout`, one per line. In
+addition to ELF binaries, runtimes currently supported by the NFD release image
```
Ethyling: Should we mention `statically linked ELF binaries` to prevent any problem with dynamic libraries?

marquiz: Thanks for thinking about this. I was also actually thinking about this, too, but wrote it in a hurry and did not mention this. But, now that you brought it up might be good to add a note. Possibly recommend statically linked binaries and mention that there is a very limited number of libraries available in the image (basically "try and see if it works")

marquiz: Pushed a new version of the PR. Static binaries are now mentioned as the recommended ELF format

Issue #286, opened 2020-02-12
pci: add sriov detection


mythi:
SR-IOV is a PCI attribute. I'd suggest we add new attributes to PCI features:

```
feature.node.kubernetes.io/pci-0200_8086.sriov.capable=true
feature.node.kubernetes.io/pci-0200_8086.sriov.configured=true
```
Benefits:
* all SRIOV PCI devices can be detected
* For NIC class devices it is possible get to know which NICs have that capability (right now it's just "there's an SRIOV NIC on this node")

marquiz:
This might be useful. Do you have a specific use case in mind that could be documented here? Do you think that only one label would be sufficient, e.g. `sriov.configured` (thinking about limiting the number of labels)?

mythi:
QAT is an SR-IOV device and the QAT device plugin uses SR-IOV VF devices. My preference would be `sriov.capable` because the configuration could also be done in an initContainer before the actual device plugin runs... Just a quick thought.

marquiz:
K, sounds reasonable. I'd prolly only add `sriov.capable` for now (in order to limit the number of labels). We can add `sriov.configured` later if required. WDYT?

mythi:
+1 to `sriov.capable` only.

PR #288, opened 2020-02-14
pci: add sriov.capable attribute

mythi:
SR-IOV is a PCI attribute and also non-NIC PCI devices can have it. Therefore,
it is useful to label all PCI devices with that capability.

After this commit the following labels for Intel NICs are overlapping:
```
feature.node.kubernetes.io/pci-0200_8086.sriov.capable=true
feature.node.kubernetes.io/network-sriov.capable=true
```
Fixes: #286

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>

### DIFF
--- a/source/pci/pci.go
+++ b/source/pci/pci.go
@@ -39,6 +39,7 @@ var Config = NFDConfig{
 }
 
 var devLabelAttrs = []string{"class", "vendor", "device", "subsystem_vendor", "subsystem_device"}
+var extraDevAttrs = []string{"sriov_totalvfs"}
 
 // Implement FeatureSource interface
 type Source struct{}
@@ -49,11 +50,6 @@ func (s Source) Name() string { return "pci" }
 // Discover features
 func (s Source) Discover() (source.Features, error) {
 	features := source.Features{}
-
-	devs, err := detectPci()
-	if err != nil {
-		return nil, fmt.Errorf("Failed to detect PCI devices: %s", err.Error())
-	}
 
 	// Construct a device label format, a sorted list of valid attributes
 	deviceLabelFields := []string{}
@@ -80,6 +76,21 @@ func (s Source) Discover() (source.Features, error) {
 		deviceLabelFields = []string{"class", "vendor"}
 	}
 
+	// Read extraDevAttrs + configured or default labels. Attributes
+	// set to 'true' are considered must-have.
+	deviceAttrs := map[string]bool{}
+	for _, label := range extraDevAttrs {
+		deviceAttrs[label] = false
+	}
+	for _, label := range deviceLabelFields {
+		deviceAttrs[label] = true
+	}
+
+	devs, err := detectPci(deviceAttrs)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to detect PCI devices: %s", err.Error())
+	}
+
 	// Iterate over all device classes
 	for class, classDevs := range devs {
 		for _, white := range Config.DeviceClassWhitelist {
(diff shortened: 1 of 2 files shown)


mythi:
What makes the commit message invalid?

marquiz:
> What makes the commit message invalid?

You should drop the `Fixes:` meta tag from the commit message

source/pci/pci.go
```diff
@@ -38,7 +39,7 @@ var Config = NFDConfig{
 	DeviceLabelFields:    []string{"class", "vendor"},
 }
 
-var devLabelAttrs = []string{"class", "vendor", "device", "subsystem_vendor", "subsystem_device"}
+var devLabelAttrs = []string{"class", "vendor", "device", "subsystem_vendor", "subsystem_device", "sriov_totalvfs"}
```
marquiz: This will effectively make `sriov_totalvfs` a valid field in the `pci.deviceLabelFields` configuration option (i.e. `sriov_totalvfs` could be made part of the label name). I don't think that's desired so we should have a separate list for the extra fields that will be scanned by `readDevInfo()` but not available as fields in the label name.

mythi: ah, of course! I split this to `extraDevAttrs`.

source/pci/pci.go
```diff
@@ -106,7 +111,10 @@ func (s Source) Discover() (source.Features, error) {
 func readDevInfo(devPath string) (pciDeviceInfo, error) {
 	info := pciDeviceInfo{}
 
-	for _, attr := range devLabelAttrs {
+	for _, attr := range append(devLabelAttrs, extraDevAttrs...) {
+		if _, err := os.Stat(path.Join(devPath, attr)); os.IsNotExist(err) {
+			continue
+		}
```
marquiz: nit: I started to think if we should check whether the failing attribute belongs to extraDevAttrs or not. If one of the "mandatory" attributes is not found, we should prolly return an error. WDYT?

mythi: In case of `sriov_totalvfs`, all PCI devices don't have it so I thought we could silently ignore if it's not there. `network` source is a good example of noisy logs:

`2020/02/17 14:50:16 SR-IOV not supported for network interface: eth0: open /sys/class/net/eth0/device/sriov_totalvfs: no such file or directory`

marquiz: > n case of `sriov_totalvfs`, all PCI devices don't have it so I thought we could silently ignore if it's not there

Yeah, sure, we can ignore that. But i was thinking about the others listed in `devLabelAttrs` that *should* not fail

mythi: I misread your comment earlier. Yes, an error with mandatory attrs certainly makes sense.

marquiz: K, as I said they *should* not fail but prolly better be safe 😉

mythi: I've updated the error handling. While debugging I realized that `ioutil.ReadFile` returns no error if the file is not available so the `if err != nil { ... }` has been *broken* all the time.

marquiz: > `ioutil.ReadFile` returns no error if the file is not available

Umm, waat?? Yes it does return an error, at least for me

mythi: some glitch. nevermind

source/pci/pci.go
```diff
@@ -106,10 +110,16 @@ func (s Source) Discover() (source.Features, error) {
...
+	for _, attr := range append(devLabelAttrs, extraDevAttrs...) {
 		data, err := ioutil.ReadFile(path.Join(devPath, attr))
 		if err != nil {
-			return info, fmt.Errorf("Failed to read device %s: %s", attr, err)
+			// extraDevAttrs are considered optional so we ignore errors and continue
+			for _, optional := range extraDevAttrs {
+				if strings.Compare(attr, optional) != 0 {
+					return info, fmt.Errorf("Failed to read device %s: %s", attr, err)
+				}
+			}
```
marquiz: This logic looks like it breaks if there are more than one item in `extraDevAttrs`.

How about creating a `map[string]struct{}` of the optional keys and then doing a lookup in the for loop? Another (better?) solution would be to change give the list of attrs (plus a bool flag if to fail on error) as an argument to `readDevInfo()` and calling it twice, first with the mandatory and then with the extra attrs.

PR #289, opened 2020-02-17
source/cpu: detect SST-CP

phillyb7:
Detection for Intel SST-CP ( Speed Select Technology - Core Power )

Adds the following feature label:

> feature.node.kubernetes.io/cpu-power.sst_cp.enabled=true

### DIFF
--- a/source/cpu/power_amd64.go
+++ b/source/cpu/power_amd64.go
@@ -74,3 +74,57 @@ func discoverSSTBF() (bool, error) {
 
 	return false, nil
 }
+
+func discoverSSTCP() (bool, error) {
+	/*
+	/ There are two requirements for SST-CP being enabled on a machine:
+	/ 1) Turbo is enabled
+	/ 2) There is at least two different values for EPP (Energy Performance Preference) among the cores of the system
+	*/
+
+	epp_file := "cpufreq/energy_performance_preference"
+	epp_values := []string{}
+
+	// 1) Check Turbo is enabled
+	features, _ := detectPstate()
+	if features["turbo"] == "false" {
+		return false, nil
+	}
+
+	// Function to check if an EPP value has already been seen
+	value_seen := func(value string) bool {
+		for _, existing_value := range epp_values {
+			if existing_value == value {
+				return true
+			}
+		}
+		return false
+	}
+
+	// 2) Loop over all CPUs in the system to check for multiple EPP values
+	files, err := ioutil.ReadDir(cpuDevicesBaseDir)
+	if err != nil {
+		return false, err
+	}
+
+	// If the first logical CPU doesn't have EPP file, neither do the others, so no need to continue
+	if _, err = os.Stat(path.Join(cpuDevicesBaseDir, files[0].Name(), epp_file)); os.IsNotExist(err) {
+		return false, nil
+	}
+
+	for _, file := range files {
+		cpu_epp_value, err := ioutil.ReadFile(path.Join(cpuDevicesBaseDir, file.Name(), epp_file))
+		if err != nil {
+			return false, err
+		}
+		epp_value_str := strings.TrimSuffix(string(cpu_epp_value), "\n")
+		if !value_seen(epp_value_str) {
+			epp_values = append(epp_values, epp_value_str)
+			if len(epp_values) > 1 {
+				return true, nil
+			}
+		}
+	}
+
+	return false, nil
+}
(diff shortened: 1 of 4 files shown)


source/cpu/power_amd64.go
```diff
@@ -74,3 +74,57 @@ func discoverSSTBF() (bool, error) {
...
+	*/
+
+	epp_file := "cpufreq/energy_performance_preference"
+	epp_values := []string{}
+
+	// 1) Check Turbo is enabled
+	features, _ := detectPstate()
+	if features["turbo"] == "false" {
+		return false, nil
+	}
```
marquiz: I started to think that maybe we could/should cache the features somewhere, e.g. in the `Source` struct itself. But maybe that's out of the scope of this patch and should be done separately.

source/cpu/power_amd64.go
```diff
@@ -74,3 +74,57 @@ func discoverSSTBF() (bool, error) {
...
+		if err != nil {
+			return false, err
+		}
+		epp_value_str := strings.TrimSuffix(string(cpu_epp_value), "\n")
+		if !value_seen(epp_value_str) {
+			epp_values = append(epp_values, epp_value_str)
+			if len(epp_values) > 1 {
+				return true, nil
+			}
+		}
```
marquiz: In order to simplify this (drop the `value_seen` lookup function) you could simply store the `epp_value_str`s in a `map[string]struct{}` and return `true` if length of the map is greater than 1.

source/cpu/power_amd64.go
```diff
@@ -74,3 +74,57 @@ func discoverSSTBF() (bool, error) {
...
+	// 2) Loop over all CPUs in the system to check for multiple EPP values
+	files, err := ioutil.ReadDir(cpuDevicesBaseDir)
+	if err != nil {
+		return false, err
+	}
+
+	// If the first logical CPU doesn't have EPP file, neither do the others, so no need to continue
+	if _, err = os.Stat(path.Join(cpuDevicesBaseDir, files[0].Name(), epp_file)); os.IsNotExist(err) {
+		return false, nil
+	}
```
Ethyling: I think we should also check other error types here

marquiz:
@phillyb7: is it really sufficient to examine the *epp* parameter only? Looking at the SST admin guide at https://lore.kernel.org/patchwork/patch/1156125/ there are other parameters. too (*frequency weight*, *min*, *max* and *desired*). Also, what if SST-CP is enabled in the *Ordered* mode mentioned in the guide?

phillyb7:
Hi @marquiz, from my understanding these extra parameters are associated with the CLOS groups themselves. So to have a core utilize a different set of parameters it would have to be associated with a given CLOS group, denoting its priority, which is done via the epp values - each CLOS group is associated with an epp value, so for example the value 'performance' is priority 0.

The ordered throttling aspect in the Ordered mode would follow along the same lines - the order is based off of the epp value.

Because of this I would have thought that having different epp values alone (based on the sysadmin configuring it beforehand) would be enough to determine that SST-CP was enabled.

marquiz:
OK, you know the feature better than me 😄 So it's not possible/feasible to configure multiple CLOSes with the same epp value but different frequency limits, for example?

phillyb7:
You've actually raised some interesting questions and I may be wrong about the Ordered mode, I'll need to make sure. Do you mind if I come back to you with more information to determine if basing it off the epp value is enough?

marquiz:
> Do you mind if I come back to you with more information to determine if basing it off the epp value is enough?

No problem, take your time. Thanks for taking the task of investigating this further, it's interesting to hear the answers 😄

marquiz:
ping @phillyb7 any update on this?

/remove-lifecycle stale

phillyb7:
> ping @phillyb7 any update on this?

Hey @marquiz, sorry for the extended delay - I changed around a few things with how the detection works and forgot about replying.

After thinking about it, the detection here is very specific to what we define as SST-CP being enabled. The detection is for a feature available in Intel's CMK (https://github.com/intel/CPU-Manager-for-Kubernetes/commit/05886988e3f1736e0d6d07d1dcff309d5d9ed2fb) that supports the use of CMK with a node that has SST-CP enabled. With CMK though, it will consider a node to have SST-CP enabled if there are multiple EPP values among the cores (as you've mentioned above), each signifying a different priority level. The concerns you brought up about the sufficiency of only looking at these values to say whether SST-CP is enabled or not got me thinking about how this way of doing it is only specific to CMK, and not very fair to other applications in the future that may depend on other aspects as well.

Because of that I've extracted the code changes here and put it into a script that NFD will look for - so it's using the feature detector hooks that are available with NFD instead of changing the source code. So this PR is actually unnecessary and can be removed, and maybe it's better to wait until SST-CP is more clearly defined before adding it as a detectable feature

marquiz:
Np, thank you for the prompt and detailed reply 😄 Do you think we should close this PR for now?

phillyb7:
Yeah I'll close this PR now, thanks! 😄

Issue #287, opened 2020-02-12
Add support for Custom features


adrianchiris:
Add a new feature source, `custom`, that will enable node labeling of custom features by defining a set of rules to match on per feature.

The custom features shall have two sources:
1. predefined list of node features in repository that are vendor agnostic (e.g RDMA)
2. configuration file

More info in the discussion in [PR#261](https://github.com/kubernetes-sigs/node-feature-discovery/pull/261)

adrianchiris:
I will propose a PR for this based on the discussion in #261

marquiz:
Implemented in #290.

PR #290, opened 2020-02-18
Support custom features

adrianchiris:
This PR implements #287

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,40 +255,80 @@ capability might be supported but not enabled.
 
 ### Custom Features
 The Custom feature source allows the user to define features based on a mix of predefined rules.
+A rule is provided input witch affects its process of matching for a defined feature.
 
-#### Rule Types
+To aid in making Custom Features clearer, we define a general and a per rule nomenclature, keeping things as
+consistent as possible.
 
+#### General Nomenclature & Definitions
+```
+Rule        :Represents a matching logic that is used to match on a feature.
+Rule Input  :The input a Rule is provided. This determines how a Rule performs the match operation.
+Matcher     :A composition of Rules, each Matcher may be composed of at most one instance of each Rule.
+```
+
+#### Custom Features Format (using the Nomenclature defined above)
+```yaml
+- name: <feature name>
+  matchOn:
+  - <Rule-1>: <Rule-1 Input>
+    [<Rule-2>: <Rule-2 Input>]
+  - <Matcher-2>
+  - ...
+  - ...
+  - <Matcher-N>
+- <custom feature 2>
+- ...
+- ...
+- <custom feature M>
+```
+
+#### Matching process
+Specifying Rules to match on a feature is done by providing a list of Matchers.
+Each Matcher contains one or more Rules.
+
+Logical _OR_ is performed between Matchers and logical _AND_ is performed between Rules
+of a given Matcher.
+
+#### Rules
 ##### PciId Rule
-The PciId rule allows matching the PCI devices in the system on the following PCI attributes: `class`,`vendor` and
-`device` identifiers.
+###### Nomenclature
+```
+Attribute   :A PCI attribute.
+Element     :An identifier of the PCI attribute.
+```
 
+The PciId Rule allows matching the PCI devices in the system on the following Attributes: `class`,`vendor` and
+`device`. A list of Elements is provided for each Attribute.
+
+###### Format
 ```yaml
 pciId :
   class: [<class id>, ...]
   vendor: [<vendor id>,  ...]
   device: [<device id>, ...]
 ```
-Matching is done by performing a logical OR for elements of an attribute and logical AND between specified attributes for
+
+Matching is done by performing a logical _OR_ between Elements of an Attribute and logical _AND_ between the specified Attributes for
 each PCI device in the system.
-At least one attribute must be specificed, missing attributes will not partake in the matching process.
+At least one Attribute must be specified. Missing attributes will not partake in the matching process.
 
 ##### LoadedKMod Rule
-The LoadedKMod rule allows matching the loaded kernel modules in the system against a provided list of kernel modules.
+###### Nomenclature
+```
+Element     :A kernel module
+```
 
+The LoadedKMod Rule allows matching the loaded kernel modules in the system against a provided list of Elements.
+
+###### Format
 ```yaml
 loadedKMod : [<kernel module>, ...]
 ```
- Matching is done by performing a logical AND for each provided kernel module, i.e the rule will match if all provided kernel modules are loaded
+ Matching is done by performing logical _AND_ for each provided Element, i.e the Rule will match if all provided Elements (kernel modules) are loaded
  in the system.
 
-#### Multiple Rules Matching process
-Specifying Rules to match on a feature is done by a list of rule elements.
-Each rule element contains one or more rule types.
-
-Logical OR is performed between rule elements and logical AND is performed between rule types
-of a given rule element.
-
-##### Example
+#### Example
 ```yaml
 custom:
   - name: "my.kernel.feature"
@@ -327,7 +367,7 @@ with a PCI vendor ID of `15b3` _AND_ PCI device ID of `1014` _OR_ `1017`.
 
 #### Statically defined features
 Some feature labels which are common and generic are defined statically in the `custom` feature source.
-A user may add additional rules to these feature labels by defining them in the `nfd-worker` configuration file
+A user may add additional Matchers to these feature labels by defining them in the `nfd-worker` configuration file.
 
 | Feature | Attribute | Description |
 | ------- | --------- | -----------|
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ for up-to-date information about the required volume mounts.
 The current set of feature sources are the following:
 
 - CPU
+- Custom
 - IOMMU
 - Kernel
 - Memory
@@ -181,6 +182,7 @@ feature logically has sub-hierarchy, e.g. `sriov.capable` and
 ```json
 {
   "feature.node.kubernetes.io/cpu-<feature-name>": "true",
+  "feature.node.kubernetes.io/custom-<feature-name>": "true",
   "feature.node.kubernetes.io/iommu-<feature-name>": "true",
   "feature.node.kubernetes.io/kernel-<feature name>": "<feature value>",
   "feature.node.kubernetes.io/memory-<feature-name>": "true",
@@ -250,6 +252,87 @@ capability might be supported but not enabled.
 | PMULL     | Optional Cryptographic and CRC32 Instructions
 | JSCVT     | Perform Conversion to Match Javascript
 | DCPOP     | Persistent Memory Support
+
+### Custom Features
+The Custom feature source allows the user to define features based on a mix of predefined rules.
+
+#### Rule Types
+
+##### PciId Rule
+The PciId rule allows matching the PCI devices in the system on the following PCI attributes: `class`,`vendor` and
+`device` identifiers.
+
+```yaml
+pciId :
+  class: [<class id>, ...]
+  vendor: [<vendor id>,  ...]
+  device: [<device id>, ...]
+```
+Matching is done by performing a logical OR for elements of an attribute and logical AND between specified attributes for
+each PCI device in the system.
+At least one attribute must be specificed, missing attributes will not partake in the matching process.
+
+##### LoadedKMod Rule
+The LoadedKMod rule allows matching the loaded kernel modules in the system against a provided list of kernel modules.
+
+```yaml
+loadedKMod : [<kernel module>, ...]
+```
+ Matching is done by performing a logical AND for each provided kernel module, i.e the rule will match if all provided kernel modules are loaded
+ in the system.
+
+#### Multiple Rules Matching process
+Specifying Rules to match on a feature is done by a list of rule elements.
+Each rule element contains one or more rule types.
+
+Logical OR is performed between rule elements and logical AND is performed between rule types
+of a given rule element.
+
+##### Example
+```yaml
+custom:
+  - name: "my.kernel.feature"
+    matchOn:
+      - loadedKMod: ["kmod1", "kmod2"]
+  - name: "my.pci.feature"
+    matchOn:
+      - pciId:
+          vendor: ["15b3"]
+          device: ["1014", "1017"]
+  - name: "my.combined.feature"
+    matchOn:
+      - loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
+        pciId:
+          vendor: ["15b3"]
+          device: ["1014", "1017"] 
+  - name: "my.accumulated.feature"
+    matchOn:
+      - loadedKMod : ["some_kmod1", "some_kmod2"]
+      - pciId:
+          vendor: ["15b3"]
+          device: ["1014", "1017"]
+```
+
+__In the example above:__
+- A node would contain the label: `feature.node.kubernetes.io/custom-my.kernel.feature=true`
+if the node has `kmod1` _AND_ `kmod2` kernel modules loaded. 
+- A node would contain the label: `feature.node.kubernetes.io/custom-my.pci.feature=true`
+if the node contains a PCI device with a PCI vendor ID of `15b3` _AND_ PCI device ID of `1014` _OR_ `1017`.
+- A node would contain the label: `feature.node.kubernetes.io/custom-my.combined.feature=true`
+if `vendor_kmod1` _AND_ `vendor_kmod2` kernel modules are loaded __AND__ the node contains a PCI device
+with a PCI vendor ID of `15b3` _AND_ PCI device ID of `1014` _or_ `1017`.
+- A node would contain the label: `feature.node.kubernetes.io/custom-my.accumulated.feature=true`
+if `some_kmod1` _AND_ `some_kmod2` kernel modules are loaded __OR__ the node contains a PCI device
+with a PCI vendor ID of `15b3` _AND_ PCI device ID of `1014` _OR_ `1017`.
+
+#### Statically defined features
+Some feature labels which are common and generic are defined statically in the `custom` feature source.
+A user may add additional rules to these feature labels by defining them in the `nfd-worker` configuration file
+
+| Feature | Attribute | Description |
+| ------- | --------- | -----------|
+| rdma  | capable | The node has an RDMA capable Network adapter |
+| rdma | enabled | The node has the needed RDMA modules loaded to run RDMA traffic |
 
 ### IOMMU Features
 
--- a/cmd/nfd-worker/main.go
+++ b/cmd/nfd-worker/main.go
@@ -91,7 +91,7 @@ func argsParse(argv []string) (worker.Args, error) {
                               in testing
                               [Default: ]
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpu,iommu,kernel,local,memory,network,pci,storage,system]
+                              [Default: cpu,custom,iommu,kernel,local,memory,network,pci,storage,system]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
--- a/source/custom/custom.go
+++ b/source/custom/custom.go
@@ -17,11 +17,17 @@ limitations under the License.
 package custom
 
 import (
+	"fmt"
+	"log"
+
 	"sigs.k8s.io/node-feature-discovery/source"
+	"sigs.k8s.io/node-feature-discovery/source/custom/rules"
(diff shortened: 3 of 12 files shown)


adrianchiris:
@marquiz

source/custom/static_features.go
```diff
@@ -0,0 +1,36 @@
...
+		CustomFeature{
+			Name: "rdma.capable",
+			MatchOn: MatchRules{
+				PciIdRegex: []string{"15b3:.*"},
+			},
+		},
+		CustomFeature{
+			Name: "rdma.available",
+			MatchOn: MatchRules{
+				LoadedKMod: []string{"ib_core", "rdma_cm"},
```
adrianchiris: this will need to be updated.
there are two approaches here IMO:
1. define the bare-minimum kernel modules for RDMA applications to be loaded then allow user to add more (e.g rdma_cm, ib_iser)
2.  ensure the same modules as in [rdma-core ](https://github.com/linux-rdma/rdma-core/tree/master/kernel-boot/modules)are loaded which is what majors distros are using

frankly i prefer 2.

source/custom/rules/loaded_kmod_rule.go
```diff
@@ -0,0 +1,67 @@
...
+		}
+		if !kmodFound {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func (r LoadedKModRule) getLoadedModules() ([]string, error) {
+	out, err := exec.Command("lsmod").Output()
```
marquiz: How about reading directly from `/proc/modules`

source/custom/rules/loaded_kmod_rule.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+		}
+		if !kmodFound {
+			return false, nil
+		}
+	}
+	return true, nil
+}
+
+func (r LoadedKModRule) getLoadedModules() ([]string, error) {
```
marquiz: I'd suggest returning a `map[string]struct{}` for nicer lookup.

cmd/nfd-worker/main.go
```diff
@@ -91,7 +91,7 @@ func argsParse(argv []string) (worker.Args, error) {
                               in testing
                               [Default: ]
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpu,iommu,kernel,local,memory,network,pci,storage,system]
+                              [Default: cpu,iommu,kernel,local,memory,network,pci,storage,system,custom]
```
marquiz: nit: the sources are in alphabetical order here

source/custom/rules/loaded_kmod_rule.go
adrianchiris: yes, the implementation was changed to exactly that, unfortunately i amended the wrong commit.

adrianchiris:
> Put the common pci discovery library under pkg/

@marquiz the intention was to utilize golang [internal packages ](https://golang.org/doc/go1.4#internalpackages ) which will prevent other projects from importing the package.

/pkg usually is intended (i think...) for exposing stable API for other projects to consume.

anyway, if that's not a concern i can move no problem.

marquiz:
> the intention was to utilize golang [internal packages ](https://golang.org/doc/go1.4#internalpackages) which will prevent other projects from importing the package.

@adrianchiris thanks for educating me :mortar_board: I wasn't even aware of this golang "kludge". In this  light the path is fine. We might consider moving some other stuff there, too...

marquiz:
@adrianchiris : needs rebase after #288 was merged

zvonkok:
> Thanks @adrianchiris for the patchset!
> 
> I started to think that is _PciIdRegex_ in this form the sort of matching we want? I was thinking about #288 which adds detection of one new PCI device attribute, and, having a bit more "k8s-style" matcher something like:
> 
> ```
> PciDevAttrs:
> - vendor: ['15b3']
> - vendor: ['1234']
>   device: ['1000', '1010', '1020']
> ```
> 
> don't know if that's a good or bad idea, though. @zvonkok ?

@marquiz  @adrianchiris  I do not think we need a regex to match against specific PCI device. It looks cumbersome and people who do not understand regex, beware we are not all developers, may have trouble understanding what is this matched against. 

The `vendor: ['15b3']` has the same information content as `.*:15b3:.*`  is easier to read and one immediately sees what we're matching against. 

To understand  `.*:15b3:.*` one would have to know that 15b3 is the vendor id, and what `.*:`, `:.*` means in this context

We have the functions to dissect the PCI attributes, we can leverage that. 

The chaining of attributes to be more specific looks IMHO nicer when used e.g.:

```
PciDevAttrs:
- vendor: ['15b3']
  device:  ['1000', '1010', '1020']
```

Without thinking about we can see what we're matching against.

adrianchiris:
@zvonkok @marquiz  thank you for providing feedback,  i see your point.
ill respin the PR and address the comments

Instead of a PciIdRegex rule we shall have a PciDevAttrs rule which will match on a list of PCI device attributes. each device attribute will be composed of list elements. 

Match operation: OR will be performed per element and AND will be performed per attribute.
An empty attribute will not be included in the matching process.

`
PciDevAttrs:
  - class : ["0200"]
    vendor : ["15b3"]
    device : ["1014", "1016"]
  - vendor : ["14e4"]
    device : ["165f"]
`
in the example above we will match on:
PciDevAttrs[0] : class 0200 AND vendor 15b3 AND device (1014 OR 1016)
PciDevAttrs[1] : vendor 14e4 AND device 165f

Sounds good ?

marquiz:
> Sounds good ?
@adrianchiris thanks for excplicitly writing out the logic. Yes, that is how I envisioned it would work. @zvonkok ?

marquiz:
But, of course, there are other alternatives for the interface, too. For example, a more flexible (and clumsy) would be to have something similar to `matchExpressions` used in k8s selectors
```
- vendor: {operator: In, values: ["15b3"]}
  device: {operator: In, values: ["1014", "1016"]}
```
That might be just overcomplication, though, because the only meaningful operators I can currently think of are `In` and `NotIn`. OTOH, also other operators (like `Lt`, `Gt`, `Exists`, `DoesNotExist`) could make sense in the future if more device attributes for matching are enabled in the future

adrianchiris:
I think we should avoid over complication (thinking from the user perspective here)
Perhaps name the match rule to PciID to limit the scope ?

Then if/when there are more use cases have one that matches on other attributes.

marquiz:
> I think we should avoid over complication (thinking from the user perspective here)

Agree

> Perhaps name the match rule to PciID to limit the scope ?

Mm, that might be better



> Then if/when there are more use cases have one that matches on other attributes.

We can do that

mythi:
I've looked into `source/pci/` recently and found this PR interesting.

One question though: what are 'custom features'. 

My observations are: 1) the functionality does not bring sources as such (it re-uses what pci source does and adds a bit on top of `source/kernel`), 2) a new match needs new rules (that are built-in).

Should this feature be turned to a configurable label-filter where user specified filters are run for each source instead?

adrianchiris:
```
My observations are: 1) the functionality does not bring sources as such (it re-uses what pci source does and adds a bit on top of source/kernel), 2) a new match needs new rules (that are built-in).

Should this feature be turned to a configurable label-filter where user specified filters are run for each source instead?
```

The intention is not to limit the functionality to other sources.
There are bits of code that's going to be reused across sources (hence i also move some code to a common place).

But rather to define framework and an expandable set of match rules which users can (simply) compose to trigger NFD to label a node with a label of their choosing given the rules match.

marquiz:
Yes, in the first phase we would have pci and kmod matchesrs, butm the idea is that we would be able to add new matchers when needed.

And, there would be a set of built-in rules (to satisfy the `rdma.capable` use case that @adrianchiris  presented in #261. In addition, there would be a set of configurable user-defined rules.

mythi:
> Yes, in the first phase we would have pci and kmod matchesrs, butm the idea is that we would be able to add new matchers when needed.

My comment was that why not to have a generic matcher that filters source plugins' output.

marquiz:
Matchers might not have any "source plugin". An example is the mkod matcher

source/custom/custom.go
```diff
@@ -48,7 +48,7 @@ func (s Source) Name() string { return "custom" }
 // Discover features
 func (s Source) Discover() (source.Features, error) {
 	features := source.Features{}
-	allFeatureConfig := s.appendFeatures(getStaticFeatureConfig(), Config.CustomFeatures)
+	allFeatureConfig := s.appendFeatures(getStaticFeatureConfig(), Config)
```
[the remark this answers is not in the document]
marquiz: Agree, I feel that this would be the more logical thing to do

zvonkok:
> Sounds good ?

Yes, sounds good. This makes it fairly easy to add some "simple" rules to detect a specific device. 

Thinking of the operators I might have a use-case where a `Gt` could help but than we would need to rely on the vendor to populate the PCI class ID with ascending numbers, which is not a guarantee, there is no PCI class ID naming spec. e.g NVIDIAs M60 is a 0x300 class id and the V100 and T4 are 0x302. To cover only the "latest and greatest" detection of cards one could write

```
- vendor: {operator: In, values: ["10de"]}
  class: {operator: Gt, values: ["300"]}
```

But this can be also easily be done by the current approach. Lets start with the simple config and extend if we need it. 

Configs from customers and internal POCs will tell us what we need and which pattern we would need to look for.

mythi:
> Matchers might not have any "source plugin". An example is the mkod matcher

My concern is that conceptually the matchers become data sources too.

marquiz:
> My concern is that conceptually the matchers become data sources too.

Well yes, they are. But not feature sources (as specified in the nfd internal api)

adrianchiris:
@marquiz @zvonkok inputs appreciated :)

marquiz:
> @marquiz @zvonkok inputs appreciated :)

Oops, sorry. I missed the new version of the patchset, I'll take a closer look soon

zvonkok:
@marquiz Any additional comments?

source/custom/rules/pci_id_rule.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (r PciIdRule) matchDevOnRule(dev pciutils.PciDeviceInfo, rule pciIdRuleInput) bool {
+	if len(rule.Class) == 0 && len(rule.Vendor) == 0 && len(rule.Device) == 0 {
+		return false
+	}
+
+	if len(rule.Class) == 0 || in(dev["class"], rule.Class) == true {
+		if len(rule.Vendor) == 0 || in(dev["vendor"], rule.Vendor) == true {
+			if len(rule.Device) == 0 || in(dev["device"], rule.Device) == true {
```
marquiz: nit: you can drop the `== true` from the conditionals

suggestion: to make this more readable (especially in the case of future additions) we could drop the nested `if`s by doing smth like:
```
if len(rule.Class) > 0 && !in(dev["class"], rule.Class) {
    return false
}
```

source/custom/rules/pci_id_rule.go
```diff
@@ -0,0 +1,92 @@
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func in(item string, arr []string) bool {
+	for i := 0; i < len(arr); i++ {
```
marquiz: nit: why not use for `for _, val := range arr`?

source/custom/custom.go
```diff
@@ -46,5 +49,64 @@ func (s Source) Name() string { return "custom" }
 // Discover features
 func (s Source) Discover() (source.Features, error) {
 	features := source.Features{}
+	allFeatureConfig := s.mergeFeatures(getStaticFeatureConfig(), Config)
```
marquiz: Do we need to merge? Would just a simple append work? First run the static and the user-defined ones. It might even be desirable to have multiple rules for the same label. WDYT?

marquiz:
> @marquiz @zvonkok inputs appreciated :)

Thanks for the update and apologies for the delay.

Some comments above, but, overall, I start to like this 😉

source/custom/custom.go
adrianchiris: I have no special Preference, We can do this as you suggest.
The current logic allows a user to override a specific feature that was defined statically.

marquiz: @zvonkok do you have any preference? I think, later on, we could add a config option to disable the static rule(s) if there is need for that. WDYT?

source/custom/rules/loaded_kmod_rule.go
```diff
@@ -0,0 +1,73 @@
...
+type loadedKModRuleInput []string
+
+const kmodProcfsPath = "/proc/modules"
+
+func (r LoadedKModRule) Match(input []byte) (bool, error) {
+	kmods := loadedKModRuleInput{}
+	err := json.Unmarshal(input, &kmods)
+	if err != nil {
+		return false, fmt.Errorf("failed to convert raw JSON input: %s", err.Error())
+	}
```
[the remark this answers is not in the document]
adrianchiris: yes, it will simplify things. will update the PR soon

adrianchiris:
Updated code according to comments

source/custom/rules/pci_id_rule.go
```diff
@@ -0,0 +1,87 @@
...
+// Match operation: OR will be performed per element and AND will be performed per attribute.
+// An empty attribute will not be included in the matching process.
+type PciIdRuleInput struct {
+	Class  []string `json:"class,omitempty"`
+	Vendor []string `json:"vendor,omitempty"`
+	Device []string `json:"device,omitempty"`
+}
+
+type PciIdRule struct {
+	PciIdRuleInput
```
marquiz: Is there a reason to have the intermediate `PciIdRuleInput` type? Could it be dropped and just define the members (Class, Vendor, Device) here?

source/custom/custom.go
```diff
@@ -0,0 +1,112 @@
...
+		// Loaded kernel module rule
+		if rule.LoadedKMod != nil {
+			match, err := rule.LoadedKMod.Match()
+			if err != nil {
+				return false, err
+			}
+			if match {
+				return true, nil
+			}
+		}
```
marquiz: What should we do if both `PciId` and `LoadedKMod` rules are specified? Now, we essentially perform a logical OR, but, should we do logical AND insteadl, WDYT?

source/custom/custom.go
```diff
@@ -0,0 +1,112 @@
...
+
+// Custom Features Configurations
+type MatchRule struct {
+	PciId *rules.PciIdRule `json:"pciId,omitempty""`
+	LoadedKMod *rules.LoadedKModRule `json:"loadedKMod,omitempty""`
+}
+
+type CustomFeature struct {
+	Name    string      `json:"name"`
+	MatchOn []MatchRule `json:"matchOn"`
```
marquiz: Just a thought: should we have `Value`, too? That can be added later on, so no need to change that unless you want to...

source/custom/rules/pci_id_rule.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// Match PCI devices on the provided list of PCI regular expressions
```
marquiz: nit: the comment is outdated, we do not use regexs, anymore 😉

source/custom/rules/pci_id_rule.go
adrianchiris: its just so we have clear boundary between whats being provided by the user via config and and possibly additional (internal) attributes added to the rule .

I wanted to add that for the LoadedKMod rule as well but as the rule is simple it would make the user config more complex.

source/custom/rules/pci_id_rule.go
adrianchiris: Right! remnants... ill modify

source/custom/rules/pci_id_rule.go
marquiz: > its just so we have clear boundary between whats being provided by the user via config and and possibly additional (internal) attributes added to the rule .

OK. You could do a similar separation with public/non-public members(?) But, let's keep this if you prefer the really tight separation.

source/custom/custom.go
adrianchiris: lets keep this as is, later on if needed, we can extend with Value.

source/custom/custom.go
adrianchiris: we can do that,  the OR can be expressed by adding more MatchRules.

adrianchiris:
> Getting closer. What is still missing is documentation (README.md).
>
>I'd suggest dropping mergeFeatures and use simpler appending. WDYT?

As we are converging, ill add documentation.
ill drop mergeFeatures and do a simple append, originally i thought i'll start merging rules to optimize i dont think its needed for now (also the rule and its input are the same now).

source/custom/custom.go
marquiz: fair enough

source/custom/custom.go
zvonkok: Let's go on and do append like how it is now in the code. Implicit behaviour is sometimes misleading and can lead to error. I am in favour of explicitly disabling the static options via config.

zvonkok:
When we have the documentation in place, looks good to me.

source/custom/custom.go
```diff
@@ -0,0 +1,114 @@
...
+		} else {
+			customFeatureMap[allFeatures[idx].Name] = &allFeatures[idx]
+		}
+	}
+	// convert back to list of features
+	appendedFeatures := make([]CustomFeature, 0, len(customFeatureMap))
+	for _, feature := range customFeatureMap {
+		appendedFeatures = append(appendedFeatures, *feature)
+	}
+	return appendedFeatures
```
marquiz: I think you could/should drop this method(?)

README.md
```diff
@@ -251,6 +253,86 @@ capability might be supported but not enabled.
...
+
+##### PciId Rule
+The PciId rule allows matching the PCI devices in the system on the following PCI attributes: `class`,`vendor` and
+`device` identifiers.
+
+```yaml
+pciId :
+  class: ["<PCI_CLASS_ID>", ...]
+  vendor: ["<PCI_VENDOR_ID>", ...]
+  device: ["<PCI_DEVICE_ID>", ...]
```
marquiz: Possibly a personal preference, but, I would re-format this a bit, at least drop the quotes, for better readability.

I'd possibly also use lower case as a similar format is used with the other feature sources:
```yaml
  class: [<class id>, ...]
  vendor: [<vendor id>, ...]
  device: [<device id>, ...]
```

README.md
```diff
@@ -251,6 +253,86 @@ capability might be supported but not enabled.
...
+```
+Matching is done by performing a logical OR for each attribute and logical AND between the specified attributes for
+each PCI device in the system.
```
marquiz: > logical OR for each attribute and logical AND between the specified attributes

A bit cryptic. One suggestion (minimal change): "logical OR for elements of an attribute and logical AND between the specified attributes"

README.md
```diff
@@ -251,6 +253,86 @@ capability might be supported but not enabled.
...
+
+##### LoadedKMod Rule
+The LoadedKMod rule allows matching the loaded kernel modules in the system against a provided list of kernel modules.
+
+```yaml
+loadedKMod : ["<KERNEL_MODULE>", ...]
```
marquiz: I'd re-format this, too, at least drop the quotes

Corresponding the suggestion for PciId rule above:
```yaml
loadedKMod : [<module name>, ...]
```

source/custom/custom.go
adrianchiris: ofc, thanks.

README.md
adrianchiris: will modify

README.md
```diff
@@ -251,6 +253,86 @@ capability might be supported but not enabled.
...
+`device` identifiers.
+
+```yaml
+pciId :
+  class: [<class id>, ...]
+  vendor: [<vendor id>,  ...]
+  device: [<device id>, ...]
+```
+Matching is done by performing a logical OR for elements of an attribute and logical AND between specified attributes for
+each PCI device in the system.
```
marquiz: One more thing (just came to my mind) 😉 Should we explicitly state that all attributes are optional (not all of them need to be specified in a rule), the only (quite logical) requirement is that the user should specify at least one attribute? Or is it obvious from the documentation WDYT?

adrianchiris: ill explicitly state that the attributes are optional and missing attributes will not partake in the matching process.

there is an example below that depicts that, but i think its worth specifying it here

zvonkok:
I am good with it, the only nitpick I would have is, reading the documentation it may be no clear from the beginning which words (rule, attribute, element) correspond to which yaml tags, 

Maybe add a nomenclature? 
```
rule:
  attribute: [element0, element1] 
```
And try to always use the same words for the same yaml tags throughout the documentation.

marquiz:
> Maybe add a nomenclature?

+1 for adding something like from me. I was also struggling with the terminology a bit 😄

Issue #229, opened 2019-04-03
Supporting extended resources


fabiand:
Similar to #45 this RFE is about enhancing NFD to also permit exposing extended resources.

Just like labels, extended resources can be used to expose compute related features of nodes, which do have some kind of limit.

The mechanics and permissions of how they get exposed are close to equal to labels.

fabiand:
This was also raised here https://groups.google.com/forum/#!topic/kubernetes-sig-node/PDJ0wTTEdv4

marquiz:
As I said in my reply to the email above, I'm favourable to the idea. This could fly,  expecting that there is a well justified use case, and, SIG-node is ok with this.

zvonkok:
What specific extended resources is it all about? I do not think that NFD should advertise extended resources where one needs a driver deployed to detect the extended resource. 

I do not want to deploy a complete GPU stack to all nodes just to find out that 90% of my nodes are not GPU nodes. The pci label is for me the anchor to deploy all needed parts to have the card + stack activated and advertised to the cluster, which is of course not part of NFD.

Or is it here more about exposing the now labels as extended resources? 

Can someone shed some light, with a real-world example?

mmirecki:
A real-world example:
Kubevirt needs to verify resources on a node to verify that a vm (wrapped in a pod) can be created on it.
This could for example be a bridge, ovs-bridge or vf to which a vm nic would be connected. For that we need to verify what resources are available on node.

marquiz:
Do you need extended resources, more specifically, bookkeeping done by the master?

mmirecki:
We need extended resources to track which nodes we can schedule our
workloads to.
Kubevirt provisions vm's inside pods. As an example, it's quite common for
a vm to have
more than one network, for which resources such as bridges, ovs-bridges or
virtual functions
are required. We need the extended resources to be able to schedule our
pod/vm's on
appropriate nodes.

fabiand:
@zvonkok in our case no heavy logic is needed to gather the informationsin order to do the "labeling", we just need to express something else.

@marquiz the accounting is useful, as we don't consider some networking resources to be "indefinitely consumable", but are limited in fact. We'd like to use extended resources to reflect this fact.

fabiand:
@mmirecki doy ou mind opening a PR to demonstrate how it could look?

mmirecki:
Sure (have this waiting on my task queue for quite some time now).

mmirecki:
[...]

PR #291, opened 2020-02-19
master: add extended resource support

uniemimu:
@marquiz 
#229 

This adds support for making selected labels extended resources.

Signed-off-by: Ukri Niemimuukko <ukri.niemimuukko@intel.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,7 @@
 - [Feature discovery](#feature-discovery)
   - [Feature sources](#feature-sources)
   - [Feature labels](#feature-labels)
+- [Extended resources (experimental)](#extended-resources-experimental)
 - [Getting started](#getting-started)
   - [System requirements](#system-requirements)
   - [Usage](#usage)
@@ -53,7 +54,7 @@ nfd-master.
   Usage:
   nfd-master [--no-publish] [--label-whitelist=<pattern>] [--port=<port>]
      [--ca-file=<path>] [--cert-file=<path>] [--key-file=<path>]
-     [--verify-node-name] [--extra-label-ns=<list>]
+     [--verify-node-name] [--extra-label-ns=<list>] [--resource-labels=<list>]
   nfd-master -h | --help
   nfd-master --version
 
@@ -75,6 +76,8 @@ nfd-master.
   --label-whitelist=<pattern>     Regular expression to filter label names to
                                   publish to the Kubernetes API server. [Default: ]
   --extra-label-ns=<list>         Comma separated list of allowed extra label namespaces
+                                  [Default: ]
+  --resource-labels=<list>        Comma separated list of labels to be exposed as extended resources.
                                   [Default: ]
 ```
 
@@ -440,6 +443,37 @@ NFD is running. In order to avoid race conditions you should write into a
 temporary file (outside the `source.d` and `features.d` directories), and,
 atomically create/update the original file by doing a filesystem move
 operation.
+
+## Extended resources (experimental)
+
+This feature is experimental and by no means a replacement for the usage of
+device plugins.
+
+Labels which have integer values, can be promoted to Kubernetes extended
+resources by listing them to the master `--resource-labels` command line flag.
+These labels won't then show in the node label section, they will appear only
+as extended resources.
+
+An example use-case for the extended resources could be based on a hook which
+creates a label for the node SGX EPC memory section size. By giving the name of
+that label in the `--resource-labels` flag, that value will then turn into an
+extended resource of the node, allowing PODs to request that resource and the
+Kubernetes scheduler to schedule such PODs to only those nodes which have a
+sufficient capacity of said resource left.
+
+Similar to labels, the default namespace `feature.node.kubernetes.io` is
+automatically prefixed to the extended resource, if the promoted label doesn't
+have a namespace.
+
+Example usage of the command line arguments, using a new namespace:
+`nfd-master --resource-labels=my_source-my.feature,sgx.some.ns/epc --extra-label-ns=sgx.some.ns`
+
+The above would result in following extended resources provided that related
+labels exist:
+```
+  sgx.some.ns/epc: <label value>
+  feature.node.kubernetes.io/my_source-my.feature: <label value>
+```
 
 ## Getting started
 
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -65,7 +65,7 @@ func argsParse(argv []string) (master.Args, error) {
   Usage:
   %s [--no-publish] [--label-whitelist=<pattern>] [--port=<port>]
      [--ca-file=<path>] [--cert-file=<path>] [--key-file=<path>]
-     [--verify-node-name] [--extra-label-ns=<list>]
+     [--verify-node-name] [--extra-label-ns=<list>] [--resource-labels=<list>]
   %s -h | --help
   %s --version
 
@@ -87,6 +87,8 @@ func argsParse(argv []string) (master.Args, error) {
   --label-whitelist=<pattern>     Regular expression to filter label names to
                                   publish to the Kubernetes API server. [Default: ]
   --extra-label-ns=<list>         Comma separated list of allowed extra label namespaces
+                                  [Default: ]
+  --resource-labels=<list>        Comma separated list of labels to be exposed as extended resources.
                                   [Default: ]`,
 		ProgramName,
 		ProgramName,
@@ -113,6 +115,7 @@ func argsParse(argv []string) (master.Args, error) {
 	}
 	args.VerifyNodeName = arguments["--verify-node-name"].(bool)
 	args.ExtraLabelNs = strings.Split(arguments["--extra-label-ns"].(string), ",")
+	args.ResourceLabels = strings.Split(arguments["--resource-labels"].(string), ",")
 
 	return args, nil
 }
--- a/pkg/apihelper/k8shelpers.go
+++ b/pkg/apihelper/k8shelpers.go
@@ -17,8 +17,11 @@ limitations under the License.
 package apihelper
 
 import (
+	"encoding/json"
+
 	api "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	k8sclient "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 )
@@ -58,4 +61,14 @@ func (h K8sHelpers) UpdateNode(c *k8sclient.Clientset, n *api.Node) error {
 	}
 
 	return nil
+}
+
+func (h K8sHelpers) PatchStatus(c *k8sclient.Clientset, nodeName string, marshalable interface{}) error {
+	// Send the updated node to the apiserver.
+	patch, err := json.Marshal(marshalable)
+	if err == nil {
+		_, err = c.CoreV1().Nodes().Patch(nodeName, types.JSONPatchType, patch, "status")
+	}
+
+	return err
 }
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -26,6 +26,7 @@ import (
 	"os"
 	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 	"time"
 
@@ -57,6 +58,9 @@ var (
 // Labels are a Kubernetes representation of discovered features.
 type Labels map[string]string
 
+// ExtendedResources are k8s extended resources which are created from discovered features.
+type ExtendedResources map[string]string
+
 // Annotations are used for NFD-related node metadata
 type Annotations map[string]string
 
@@ -70,6 +74,7 @@ type Args struct {
 	NoPublish      bool
 	Port           int
 	VerifyNodeName bool
+	ResourceLabels []string
 }
 
 type NfdMaster interface {
@@ -82,6 +87,21 @@ type nfdMaster struct {
 	args   Args
 	server *grpc.Server
 	ready  chan bool
+}
+
+// statusOp is a json marshaling helper used for patching node status
+type statusOp struct {
+	Op    string `json:"op"`
+	Path  string `json:"path"`
+	Value string `json:"value,omitempty"`
+}
+
+func createStatusOp(verb string, resource string, path string, value string) statusOp {
+	if !strings.Contains(resource, "/") {
+		resource = LabelNs + resource
+	}
+	res := strings.ReplaceAll(resource, "/", "~1")
+	return statusOp{verb, "/status/" + path + "/" + res, value}
 }
 
 // Create new NfdMaster server instance.
@@ -204,7 +224,7 @@ func updateMasterNode(helper apihelper.APIHelpers) error {
 }
 
 // Filter labels by namespace and name whitelist
-func filterFeatureLabels(labels Labels, extraLabelNs []string, labelWhiteList *regexp.Regexp) Labels {
+func filterFeatureLabels(labels Labels, extraLabelNs []string, labelWhiteList *regexp.Regexp, extendedResourceNames []string) (Labels, ExtendedResources) {
 	for label := range labels {
 		split := strings.SplitN(label, "/", 2)
 		name := split[0]
@@ -229,7 +249,24 @@ func filterFeatureLabels(labels Labels, extraLabelNs []string, labelWhiteList *r
 			delete(labels, label)
 		}
 	}
-	return labels
+
+	// Remove labels which are intended to be extended resources
+	extendedResources := ExtendedResources{}
+	for _, extendedResourceName := range extendedResourceNames {
+		// remove possibly given default LabelNs to keep annotations shorter
+		extendedResourceName = strings.TrimPrefix(extendedResourceName, LabelNs)
+		if _, ok := labels[extendedResourceName]; ok {
+			if _, err := strconv.Atoi(labels[extendedResourceName]); err != nil {
+				stderrLogger.Printf("bad label value encountered for extended resource: %s", err.Error())
+				continue // non-numeric label can't be used
+			}
+
+			extendedResources[extendedResourceName] = labels[extendedResourceName]
+			delete(labels, extendedResourceName)
+		}
+	}
+
+	return labels, extendedResources
 }
 
 // Implement LabelerServer
@@ -265,19 +302,28 @@ func (s *labelerServer) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*p
 	}
 	stdoutLogger.Printf("REQUEST Node: %s NFD-version: %s Labels: %s", r.NodeName, r.NfdVersion, r.Labels)
 
-	labels := filterFeatureLabels(r.Labels, s.args.ExtraLabelNs, s.args.LabelWhiteList)
+	labels, extendedResources := filterFeatureLabels(r.Labels, s.args.ExtraLabelNs, s.args.LabelWhiteList, s.args.ResourceLabels)
 
 	if !s.args.NoPublish {
-		// Advertise NFD worker version and label names as annotations
-		keys := make([]string, 0, len(labels))
-		for k, _ := range labels {
-			keys = append(keys, k)
+		// Advertise NFD worker version, label names and extended resources as annotations
+		labelKeys := make([]string, 0, len(labels))
+		for k := range labels {
+			labelKeys = append(labelKeys, k)
 		}
-		sort.Strings(keys)
-		annotations := Annotations{"worker.version": r.NfdVersion,
-			"feature-labels": strings.Join(keys, ",")}
+		sort.Strings(labelKeys)
 
-		err := updateNodeFeatures(s.apiHelper, r.NodeName, labels, annotations)
+		extendedResourceKeys := make([]string, 0, len(extendedResources))
+		for key := range extendedResources {
+			extendedResourceKeys = append(extendedResourceKeys, key)
+		}
+		sort.Strings(extendedResourceKeys)
+
+		annotations := Annotations{"worker.version": r.NfdVersion,
+			"feature-labels":     strings.Join(labelKeys, ","),
+			"extended-resources": strings.Join(extendedResourceKeys, ","),
+		}
+
+		err := updateNodeFeatures(s.apiHelper, r.NodeName, labels, annotations, extendedResources)
 		if err != nil {
 			stderrLogger.Printf("failed to advertise labels: %s", err.Error())
 			return &pb.SetLabelsReply{}, err
@@ -288,7 +334,7 @@ func (s *labelerServer) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*p
 
 // advertiseFeatureLabels advertises the feature labels to a Kubernetes node
 // via the API server.
-func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Labels, annotations Annotations) error {
+func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Labels, annotations Annotations, extendedResources ExtendedResources) error {
 	cli, err := helper.GetClient()
 	if err != nil {
 		return err
@@ -299,6 +345,9 @@ func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Lab
 	if err != nil {
 		return err
 	}
+
+	// Resolve publishable extended resources before node is modified
+	statusOps := getExtendedResourceOps(node, extendedResources)
 
 	// Remove old labels
 	if l, ok := node.Annotations[AnnotationNs+"feature-labels"]; ok {
@@ -323,7 +372,16 @@ func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Lab
 		return err
 	}
 
-	return nil
+	// patch node status with extended resource changes
+	if len(statusOps) > 0 {
+		err = helper.PatchStatus(cli, node.Name, statusOps)
+		if err != nil {
+			stderrLogger.Printf("error while patching extended resources: %s", err.Error())
+			return err
+		}
+	}
+
+	return err
 }
 
 // Remove any labels having the given prefix
@@ -346,6 +404,42 @@ func removeLabels(n *api.Node, labelNames []string) {
 	}
 }
 
+// getExtendedResourceOps returns a slice of operations to perform on the node status
+func getExtendedResourceOps(n *api.Node, extendedResources ExtendedResources) []statusOp {
+	var statusOps []statusOp
+
+	oldResources := strings.Split(n.Annotations[AnnotationNs+"extended-resources"], ",")
+
+	// figure out which resources to remove
+	for _, resource := range oldResources {
+		if _, ok := n.Status.Capacity[api.ResourceName(addNs(resource, LabelNs))]; ok {
+			// check if the ext resource is still needed
+			_, extResNeeded := extendedResources[resource]
+			if !extResNeeded {
+				statusOps = append(statusOps, createStatusOp("remove", resource, "capacity", ""))
+				statusOps = append(statusOps, createStatusOp("remove", resource, "allocatable", ""))
+			}
+		}
+	}
+
+	// figure out which resources to replace and which to add
+	for resource, value := range extendedResources {
+		// check if the extended resource already exists with the same capacity in the node
+		if quantity, ok := n.Status.Capacity[api.ResourceName(addNs(resource, LabelNs))]; ok {
+			val, _ := quantity.AsInt64()
+			if strconv.FormatInt(val, 10) != value {
+				statusOps = append(statusOps, createStatusOp("replace", resource, "capacity", value))
+				statusOps = append(statusOps, createStatusOp("replace", resource, "allocatable", value))
+			}
+		} else {
+			statusOps = append(statusOps, createStatusOp("add", resource, "capacity", value))
+			// "allocatable" gets added implicitly after adding to capacity
+		}
+	}
+
+	return statusOps
+}
+
 // Add NFD labels to a Node object.
 func addLabels(n *api.Node, labels map[string]string) {
 	for k, v := range labels {
@@ -362,4 +456,12 @@ func addAnnotations(n *api.Node, annotations map[string]string) {
 	for k, v := range annotations {
 		n.Annotations[AnnotationNs+k] = v
 	}
+}
+
+// addNs adds a namespace if one isn't already found from src string
+func addNs(src string, nsToAdd string) string {
+	if strings.Contains(src, "/") {
+		return src
+	}
+	return nsToAdd + src
 }
--- a/nfd-master.yaml.template
+++ b/nfd-master.yaml.template
@@ -18,6 +18,9 @@ rules:
   - ""
   resources:
   - nodes
+# when using command line flag --resource-labels to create extended resources
+# you will need to uncomment "- nodes/status"
+# - nodes/status
   verbs:
   - get
   - patch
--- a/pkg/apihelper/apihelpers.go
+++ b/pkg/apihelper/apihelpers.go
@@ -31,4 +31,7 @@ type APIHelpers interface {
 
 	// UpdateNode updates the node via the API server using a client.
 	UpdateNode(*k8sclient.Clientset, *api.Node) error
+
+	// PatchStatus updates the node status via the API server using a client.
+	PatchStatus(*k8sclient.Clientset, string, interface{}) error
 }
--- a/pkg/apihelper/mock_APIHelpers.go
+++ b/pkg/apihelper/mock_APIHelpers.go
@@ -59,6 +59,20 @@ func (_m *MockAPIHelpers) GetNode(_a0 *kubernetes.Clientset, _a1 string) (*v1.No
 	return r0, r1
 }
 
+// PatchStatus provides a mock function with given fields: _a0, _a1, _a2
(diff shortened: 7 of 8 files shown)


cmd/nfd-master/main.go
```diff
@@ -65,7 +65,7 @@ func argsParse(argv []string) (master.Args, error) {
   Usage:
   %s [--no-publish] [--label-whitelist=<pattern>] [--port=<port>]
      [--ca-file=<path>] [--cert-file=<path>] [--key-file=<path>]
-     [--verify-node-name] [--extra-label-ns=<list>]
+     [--verify-node-name] [--extra-label-ns=<list>] [--ext-res=<resources>]
```
uniemimu: In hindsight, I should maybe replace "resources" by "list"

pkg/nfd-master/nfd-master.go
```diff
@@ -346,6 +363,84 @@ func removeLabels(n *api.Node, labelNames []string) {
...
+			}
+		}
+	}
+	if sep == "," {
+		patch += "}}}"
+		cli, err := helper.GetClient()
+		if err != nil {
+			return false, err
+		}
+		err = helper.PatchStatus(cli, nodeName, []byte(patch))
```
uniemimu: Note: unlike pruning, this does not avoid patching even if the ext resources stay the same all the time. I do not know about the actual implementation of CoreV1().Nodes().PatchStatus, how quick it is figuring out that the diff is zero.

It could be possible to check in this function whether the resource already exists based on annotations, but then the annotation should store also the value of the extended resource capacity. Then with this check the unnecessary calling of PatchStatus could be avoided, when the resource hasn't really changed.

pkg/nfd-master/nfd-master.go
```diff
@@ -323,7 +331,16 @@ func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Lab
 		return err
 	}
 
-	return nil
+	// Prune redundant extended resources (potentially patches the node)
+	_, err = pruneExtResources(nodeName, labels, extResources, oldResources, helper)
+	if err != nil {
+		return err
```
Ethyling: We may need to add some logging in case of error here

pkg/nfd-master/nfd-master.go
```diff
@@ -323,7 +331,16 @@ func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Lab
...
+	}
+
+	// Add new extended resources (potentially patches the node)
+	_, err = addExtResources(nodeName, labels, extResources, helper)
+
+	return err
```
Ethyling: Same as above

cmd/nfd-master/main.go
```diff
@@ -65,7 +65,7 @@ func argsParse(argv []string) (master.Args, error) {
   Usage:
   %s [--no-publish] [--label-whitelist=<pattern>] [--port=<port>]
      [--ca-file=<path>] [--cert-file=<path>] [--key-file=<path>]
-     [--verify-node-name] [--extra-label-ns=<list>]
+     [--verify-node-name] [--extra-label-ns=<list>] [--ext-res=<list>]
```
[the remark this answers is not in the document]
uniemimu: Since the feature in k8s is "extended resource" and adding one has an impact on multiple places inside the node status, not just capacity, I wouldn't go there. But yes, I agree that --ext-res is not perhaps the best either. This may require a bit more time to come up with something good.

cmd/nfd-master/main.go
```diff
@@ -113,6 +115,7 @@ func argsParse(argv []string) (master.Args, error) {
 	}
 	args.VerifyNodeName = arguments["--verify-node-name"].(bool)
 	args.ExtraLabelNs = strings.Split(arguments["--extra-label-ns"].(string), ",")
+	args.ExtResources = strings.Split(arguments["--ext-res"].(string), ",")
```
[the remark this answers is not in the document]
uniemimu: sure

pkg/nfd-master/nfd-master.go
```diff
@@ -288,7 +291,7 @@ func (s *labelerServer) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*p
 
 // advertiseFeatureLabels advertises the feature labels to a Kubernetes node
 // via the API server.
-func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Labels, annotations Annotations) error {
+func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Labels, annotations Annotations, extResources []string) error {
```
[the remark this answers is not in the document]
uniemimu: sure

pkg/apihelper/k8shelpers.go
```diff
@@ -59,3 +59,13 @@ func (h K8sHelpers) UpdateNode(c *k8sclient.Clientset, n *api.Node) error {
 
 	return nil
 }
+
+func (h K8sHelpers) PatchStatus(c *k8sclient.Clientset, nodeName string, patch []byte) error {
+	// Send the updated node to the apiserver.
+	_, err := c.CoreV1().Nodes().PatchStatus(nodeName, patch)
```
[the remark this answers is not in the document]
uniemimu: I'll give it a shot.

Much of the burden in the patch creation is from the parts that try to avoid the patching when it really isn't needed.

uniemimu:
> What I was wondering whether it makes sense to create the label at all if the feature is published as an extended resource. My first thought is that it doesn't as they are sort of overlapping mechanisms. WDYT?

The rationale for the current way of having the label at the same time was:
1) smaller change, no need to filter out anything while adding labels
2) it doesn't hurt, and somebody might come with a use-case where the label is wanted anyway. Suppose you have nodes with an extended resource Foo with capacity value 1024, and some other nodes with the extended resource Foo with capacity value 2048. Also suppose that for some reason, you want some special POD which consumes 1000 Foos to be only deployed to the nodes which have a capacity of 2048 Foos, because those nodes are so much better. Now you could easily use a node selector for that, if the label is there also. If you leave it only for the scheduler, it can pick the wrong nodes and you need to figure out something else like another label. A corner case, but maybe plausible.

cmd/nfd-master/main.go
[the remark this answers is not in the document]
marquiz: I don't have strong vision about the name of the flag. My thinking behind capacity was that our interface is the node capacity attribute, and, we're not doing any resource management, just announcing the capacity.

pkg/apihelper/k8shelpers.go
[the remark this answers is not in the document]
marquiz: K, give it a try. It might be worth to define a simple struct type for jsonpatch, create an array of them and then stringify that in one shot(?)

marquiz:
> 1. smaller change, no need to filter out anything while adding labels

Mm, I understand :)


> 2. it doesn't hurt, and somebody might come with a use-case where the label is wanted anyway.

My motive was limiting unnecessary labels.

> Suppose you have nodes with an extended resource Foo with capacity value 1024, and some other nodes with the extended resource Foo with capacity value 2048. Also suppose that for some reason, you want some special POD which consumes 1000 Foos to be only deployed to the nodes which have a capacity of 2048 Foos, because those nodes are so much better.

I understand the thinking, but I don't know how valid use case this is and what would be the right thing to do. In the case you presented it might be good to have some other label indicating the superiority of the other node(?) OTOH, this new feature (i.e. turning labels into extended resources) needs to be explicitly enabled and probably only used in specialised deployments so having few extra labels doesn't hurt.

marquiz:
@zvonkok @fabiand @mmirecki any thoughts about this PR?

zvonkok:
> The rationale for the current way of having the label at the same time was:
> 
> 1. smaller change, no need to filter out anything while adding labels
> 2. it doesn't hurt, and somebody might come with a use-case where the label is wanted anyway. Suppose you have nodes with an extended resource Foo with capacity value 1024, and some other nodes with the extended resource Foo with capacity value 2048. Also suppose that for some reason, you want some special POD which consumes 1000 Foos to be only deployed to the nodes which have a capacity of 2048 Foos, because those nodes are so much better. Now you could easily use a node selector for that, if the label is there also. If you leave it only for the scheduler, it can pick the wrong nodes and you need to figure out something else like another label. A corner case, but maybe plausible.

I am just wondering if a Foo:2048 is so much better than a Foo:1000, what features make these nodes "better". Is it the device that is exposed or the combination of the device + node features? 

If it is a different Foo device, this is maybe already exposed through PCI vendor,class,device.
e.g. older GPUs with lower CUDA core count have a class id 0x300, 
newer GPUs with higher CUDA core count have a class id 0x302 . 

If it is a feature of the node, what would this be that NFD is currently not exposing? 

Any specific use-case you have in mind? What was the rationale in using extended resources? 
An example would help to understand your point.

cmd/nfd-master/main.go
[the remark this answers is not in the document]
uniemimu: In case longer argument names are not frowned upon, I propose "--resource-labels"

uniemimu:
> I am just wondering if a Foo:2048 is so much better than a Foo:1000, what features make these nodes "better". Is it the device that is exposed or the combination of the device + node features?
> 
> If it is a different Foo device, this is maybe already exposed through PCI vendor,class,device.
> e.g. older GPUs with lower CUDA core count have a class id 0x300,
> newer GPUs with higher CUDA core count have a class id 0x302 .
> 
> If it is a feature of the node, what would this be that NFD is currently not exposing?
> 
> Any specific use-case you have in mind? What was the rationale in using extended resources?
> An example would help to understand your point.

I'm not really claiming bigger to be always better, sometimes you may want just the opposite. That was just an example, one could want to prefer the nodes with less resources for some PODs and keep the nodes with a huge amount of resources for some other PODs that really need huge amount of resources. I'm just trying to describe plausible scenarios. The resource capacity can form a fingerprint for the hardware just like something encoded in the PCI class may do.

Sometimes you may have a difference in the PCI information, sometimes maybe not. Sometimes you might get that fingerprint you need, from e.g. an accelerator memory amount exposed through NFD-hooks and labels.

This mechanism in this PR then is just a scheduling aid for keeping book that the extended resource doesn't run out, but like said, I can imagine someone having a need for the label at the same time to further control where different PODs get deployed.

The rationale for this PR comes from the need of sharing accelerator hardware for multiple PODs in a node with known consumption of the accelerator memory resources per POD and a limited amount of said memory resources in the accelerator at each node.

zvonkok:
@marquiz I am still trying to wrap my head around about the extended resources. If a resource is countable we can use extended resources but isn't that a job for the DevicePlugin? If a resource is infinite we can use a label. 

A feature depending on the amount or count of the specific hardware is as specific asset of this hardware, for base detection, I am in favour of just having the extended resource, and if you need specific labels for this specific hardware to steer scheduling you could create namespaced labels that are created by using the hooks or feature capability of NFD. 

Taking the GPU as an example: 

1. NFD labels the node with the PCI class and vendor id: `pci-0300_10de.present = true`
2. Deploy the NV enablement stack on nodeSelector  `pci-0300_10de.present = true`
   (now we have the driver and can extract sophisticated features about the hw)
3. Deploy GPU feature discovery to expose scheduling features (sidecar container using the feature hook) 
```
   nvidia.com/memory = xxxx
   nvidia.com/driver-version = xxxx
   nvidia.com/compute-capability = xxxx
```

Now you can steer your workload on the best `Foo`. 

The DevicePlugin takes care of exposing the count to the cluster and marking unhealthy cards as not usable. 

I am still thinking about a real-world use-case to use extended resources ...

uniemimu:
>    nvidia.com/memory = xxxx
>    nvidia.com/driver-version = xxxx
>    nvidia.com/compute-capability = xxxx
> ```
> 
> Now you can steer your workload on the best `Foo`.
> 
> The DevicePlugin takes care of exposing the count to the cluster and marking unhealthy cards as not usable.
> 
> I am still thinking about a real-world use-case to use extended resources ...

Taking your own example here as a baseline, you could have a use-case where you know the maximum nvidia.com/memory consumption for your PODs, and you would like to have several PODs being run on a single GPU in order to get the GPU utilization up, limited only by their memory consumption. Now with the extended resource you could do that without risking running out of memory, provided that you have the POD spec written correctly. The standard approach of course is that the GPU is not shared between POD instances like that, but suppose it could be.

You are correct in that the plugin could probably do whatever kind of capability labeling needed, it doesn't have to be the same name as the extended resource. And those could then be used for scheduling purposes.

mythi:
> I am still thinking about a real-world use-case to use extended resources ...

Another example from me is similar to what @uniemimu had: SGX

The workloads need the access to `/dev/sgx/enclave` (a shared device) and they can specify how much SGX EPC memory they need. The amount of memory depends on the workload. 

The existing SGX device plugins out there do the SGX EPC book keeping with 1Mb granularity. A workload needing, let's say 64Mb SGX EPC would result in 64 `Allocate()`  and still get that one shared file.

With this feature, extended resources, and a simple NFD source plugin, such as 

```
package main

import (
        "fmt"
        "github.com/klauspost/cpuid"
)

func main() {
        var total uint64 = 0
        if cpuid.CPU.SGX.Available {
                for _, s := range cpuid.CPU.SGX.EPCSections {
                        total += s.EPCSize
                }
                if total != 0 {
                        fmt.Printf("sgx.intel.com/epc=%d", total)
                }
        }
}
```

We could have

```
sgx.intel.com/device: <some big value>
sgx.intel.com/epc: <extended resources reported by NFD source plugin>
feature.node.kubernetes.io/cpu-cpuid.SGX: "true"
feature.node.kubernetes.io/cpu-cpuid.SGXLE: "true"
```

and EPC would be a simple counter...

marquiz:
> If a resource is countable we can use extended resources but isn't that a job for the DevicePlugin?

@uniemimu please correct me if I'm wrong, but AFAIU, the desire to use NFD instead of device plugins is the problematic device plugin API. With my limited understanding of device plugins I assume that for every resource unit you will get one `allocate()` or similar calls which would be unbearable if you have like `8Gi` units of a resource.


> A feature depending on the amount or count of the specific hardware is as specific asset of this hardware, for base detection, I am in favour of just having the extended resource, and if you need specific labels for this specific hardware to steer scheduling you could create namespaced labels that are created by using the hooks or feature capability of NFD.

I'm inclined to think the same way. Having these numerical labels generally feels sort of non-features and user-unfriendly. Well, at least for the moment, unless proven otherwise :wink: 

> I am still thinking about a real-world use-case to use extended resources

I guess this `nvidia.com/memory` presented above could be one. What I'm still missing in this particular case is how you would handle the case where a node would have multiple GPUs(?) Sharing memory between them might be expensive(?)

pkg/nfd-master/nfd-master.go
```diff
@@ -270,14 +291,16 @@ func (s *labelerServer) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*p
...
 			keys = append(keys, k)
 		}
 		sort.Strings(keys)
 		annotations := Annotations{"worker.version": r.NfdVersion,
-			"feature-labels": strings.Join(keys, ",")}
+			"feature-labels":     strings.Join(keys, ","),
+			"extended-resources": strings.Join(s.args.ResourceLabels, ",")}
 
-		err := updateNodeFeatures(s.apiHelper, r.NodeName, labels, annotations)
+		resourceLabels := resourceLabels(s.args.ResourceLabels)
```
marquiz: I'd construct `resourceLabels` (or `extendedResources`) earlier, before the annotations. Only adding labels that exist (not everything that was listed on command line)

pkg/nfd-master/nfd-master.go
```diff
@@ -288,7 +311,7 @@ func (s *labelerServer) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*p
 
 // advertiseFeatureLabels advertises the feature labels to a Kubernetes node
 // via the API server.
-func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Labels, annotations Annotations) error {
+func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Labels, annotations Annotations, resourceLabels ResourceLabels) error {
```
marquiz: I'd change the last argument to be `ExtendedResources` that will be advertised as is. This would make this function more general and future-proof

pkg/nfd-master/nfd-master.go
```diff
@@ -346,6 +388,85 @@ func removeLabels(n *api.Node, labelNames []string) {
...
+		}
+		err = helper.PatchStatus(cli, nodeName, patch)
+	}
+	return err
+}
+
+// pruneExtResources forms a patch which removes redundant ext resources from
+// node status capacity and allocatable arrays
+// returns true if node was patched, false otherwise (and error)
+func pruneExtResources(n *api.Node, labels Labels, resourceLabels ResourceLabels, oldResources []string, helper apihelper.APIHelpers) (bool, error) {
```
marquiz: I think you could combine `pruneExtResources()` and `addExtResources()` into one `updateExtendedResources()` function. 
```
func updateExendedtResources(n *api.Node, extendedResources ExtendedResources, oldResources []string, helper apihelper.APIHelpers)
```

Create a list of patches, doing "replace" for existing resources that needs to be updated, "add" for completely new resources and "remove" for resources to be removed. Then do a atomic update on one `patchNodeStatus()` call

uniemimu:
> > If a resource is countable we can use extended resources but isn't that a job for the DevicePlugin?
> 
> @uniemimu please correct me if I'm wrong, but AFAIU, the desire to use NFD instead of device plugins is the problematic device plugin API. With my limited understanding of device plugins I assume that for every resource unit you will get one `allocate()` or similar calls which would be unbearable if you have like `8Gi` units of a resource.

Something like that. Each increase of the resource value done via the device plugin API requires a registration, and the increase is +1 for each registration. It is not feasible for a larger amount, and indeed each registration quite likely results in memory allocations of its own.

> 
> > A feature depending on the amount or count of the specific hardware is as specific asset of this hardware, for base detection, I am in favour of just having the extended resource, and if you need specific labels for this specific hardware to steer scheduling you could create namespaced labels that are created by using the hooks or feature capability of NFD.
> 
> I'm inclined to think the same way. Having these numerical labels generally feels sort of non-features and user-unfriendly. Well, at least for the moment, unless proven otherwise wink

Well, I will then filter out the labels. It just increases the patch size.

uniemimu:
> I guess this `nvidia.com/memory` presented above could be one. What I'm still missing in this particular case is how you would handle the case where a node would have multiple GPUs(?) Sharing memory between them might be expensive(?)

The multiple GPU case is interesting but in my opinion outside the scope of this PR. A k8s scheduler extension could maybe be used for handling cases where the total resource amount at a node does not match the largest possible single allocation of said resources, or potentially other more complex cases. In practice, it is not overly difficult for the scheduler extension to filter out a node based on whatever heuristics are needed, even if k8s scheduler would initially think that the node is a potential deployment target based on blind resource amount.

In short the answer is: I would handle that case outside NFD, but I would probably still need the resource amount exposed by NFD.

pkg/nfd-master/nfd-master.go
```diff
@@ -346,6 +394,107 @@ func removeLabels(n *api.Node, labelNames []string) {
...
+		replace:     map[string]string{},
+		remove:      map[string]string{},
+		annotations: []string{},
+	}
+
+	oldResources := []string{}
+	if l, ok := n.Annotations[AnnotationNs+"extended-resources"]; ok {
+		oldResources = strings.Split(l, ",")
+	}
+
```
marquiz: you can drop the if and just do 
```
oldResources := strings.Split(n.Annotations[AnnotationNs+"extended-resources"], ",")
```

pkg/nfd-master/nfd-master.go
```diff
@@ -306,6 +344,10 @@ func updateNodeFeatures(helper apihelper.APIHelpers, nodeName string, labels Lab
 		removeLabels(node, oldLabels)
 	}
 
+	// Resolve publishable extended resources
+	resourceActions := getExtendedResourceActions(node, extendedResources)
+	annotations["extended-resources"] = strings.Join(resourceActions.annotations, ",")
```
marquiz: Move the addition of annotations to `SetLabels()` where "label annotations" are added, too. You should be able to use keys directly from `extendedResources`(?)

uniemimu: The annotations are added here, because the logic for figuring out which extended resources (annotations) there are going to be after patching, needs to be run. That logic requires the node object. The node instance does not exist prior to this function. Hence the annotations for extended resources are added here, after the necessary node instance is created. Now, do you want me to try to move the creation of the node instance elsewhere and pass it around, or is this ok, as it is?

pkg/nfd-master/nfd-master.go
```diff
@@ -265,19 +302,20 @@ func (s *labelerServer) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*p
 	}
 	stdoutLogger.Printf("REQUEST Node: %s NFD-version: %s Labels: %s", r.NodeName, r.NfdVersion, r.Labels)
 
-	labels := filterFeatureLabels(r.Labels, s.args.ExtraLabelNs, s.args.LabelWhiteList)
+	extendedResources := extendedResources(s.args.ResourceLabels)
+	labels := filterFeatureLabels(r.Labels, s.args.ExtraLabelNs, s.args.LabelWhiteList, extendedResources)
```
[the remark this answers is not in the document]
uniemimu: Sure. But the name of the function will no longer be overly descriptive considering it also creating ExtendedResources. Just saying - I'll do it.

pkg/nfd-master/nfd-master.go
```diff
@@ -84,6 +90,22 @@ type nfdMaster struct {
...
+}
+
+// resourceActions stores maps for extended resource related operations
+// and a slice of annotations
+type resourceActions struct {
+	add         map[string]string
+	replace     map[string]string
+	remove      map[string]string
+	annotations []string
+}
```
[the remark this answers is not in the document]
uniemimu: Maybe not. I'll see what I can do while trying to avoid creating a blob-functions that does everything.

pkg/nfd-master/nfd-master.go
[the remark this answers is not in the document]
marquiz: > name of the function will no longer be overly descriptive

Kind of true, but it already "created" extended resources by filling the map so I don't know if it's hugely different. You're free to rename the function, though, if you come up with a better name 😄

pkg/nfd-master/nfd-master.go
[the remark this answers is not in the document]
marquiz: I was thinking `getExtendedResourceActions`. Instead of filling maps in a struct, just append JSON patches to a single array

pkg/nfd-master/nfd-master.go
marquiz: Ah, you mean the filtering out of invalid values. Maybe that should be already earlier, in `filterFeatureLabels()`(?)

uniemimu: filterFeatureLabels doesn't sound any better, the needed node instance doesn't exist yet there.

In order to know what the annotations will be, the whole logic currently in getExtendedResourceActions needs to be run. Surely *that* doesn't belong into a function "filterFeatureLabels" which up till my patch has mostly been nicely and cleanly concentrating in doing what the name says - filtering out feature labels. Now up to this point what I've done in filterFeatureLabels is filtering out the extended resources from the labels (plus setting the values from labels to extended resources, but that is a bit of a kludge) which is pretty well still in line with what the function was probably designed for.

I could move the calling of getExtendedResourceActions (which I'll rename probably since changes were requested) into the function "SetLabels" but that would require moving the creation of the node instance into there also, and the resulting patch array and that node instance will need to be moved as arguments of updateNodeFeatures. Do you want that? If you do, I'll do it. Then you will have the annotations set at the same time at the same place where "label annotations" are added. Extended resource logic and patch forming will run before that, and node instance needs to be created before that so that the logic gets a hold of what the extended resources were previously.

marquiz: But, you wouldn't need the node instance for that(?) Everything that's in ´extendedResources` get's added (minus the invalid values) so you could just keys from the map.

pkg/nfd-master/nfd-master.go
```diff
@@ -228,8 +244,18 @@ func filterFeatureLabels(labels Labels, extraLabelNs []string, labelWhiteList *r
...
+
+		// Skip labels which are intended to be extended resources
+		if _, ok := extendedResources[label]; ok {
+			if _, err := strconv.Atoi(labels[label]); err != nil {
+				stderrLogger.Printf("bad label value encountered for extended resource: %s", err.Error())
+				continue // non-numeric label can't be used
+			}
+			extendedResources[label] = labels[label] // store value
+			delete(labels, label)
+		}
```
marquiz: I think you could do this filtering the other way around. In a separate loop, go through `extendedResourceNames`, delete labels that match and add them to `extendedResources` map.

This would also eliminate the need to initialize `extendedResources` above. And, as `extendedResources` would only have legit values, you could possibly omit some `value != ""` checks in other functions.

pkg/nfd-master/nfd-master.go
```diff
@@ -265,19 +291,30 @@ func (s *labelerServer) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*p
...
+		labelKeys := make([]string, 0, len(labels))
+		for k := range labels {
+			labelKeys = append(labelKeys, k)
 		}
-		sort.Strings(keys)
+		sort.Strings(labelKeys)
+
+		extendedResourceKeys := []string{}
+		for key, value := range extendedResources {
+			if value != "" {
```
marquiz: With the change I proposed in `filterFeatureLabels` you could remove this check, I think

pkg/nfd-master/nfd-master.go
```diff
@@ -346,6 +395,51 @@ func removeLabels(n *api.Node, labelNames []string) {
...
+				addStatusOp("remove", resource, "capacity", "")
+				addStatusOp("remove", resource, "allocatable", "")
+			}
+		}
+	}
+
+	// figure out which resources to replace and which to add
+	for resource, value := range extendedResources {
+		// check that there is such a label (value is non-empty)
+		if value != "" {
```
marquiz: With the change I proposed in `filterFeatureLabels` you could remove this check, I think

pkg/nfd-master/nfd-master.go
```diff
@@ -346,6 +395,51 @@ func removeLabels(n *api.Node, labelNames []string) {
...
 }
 
+// getExtendedResourceOps returns a slice of operations to perform on the node status
+func getExtendedResourceOps(n *api.Node, extendedResources ExtendedResources) []statusOp {
+	var statusOps []statusOp
+	addStatusOp := func(verb string, resource string, path string, value string) {
+		res := strings.ReplaceAll(resource, "/", "~1")
+		op := statusOp{verb, "/status/" + path + "/" + res, value}
+		statusOps = append(statusOps, op)
+	}
```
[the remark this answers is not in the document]
uniemimu: Other than being only needed here in this function, there is no reason.

marquiz:
@zvonkok any comments?

I see this mostly as an experimental feature. The default operation is not affected.

pkg/nfd-master/nfd-master.go
```diff
@@ -346,6 +399,43 @@ func removeLabels(n *api.Node, labelNames []string) {
...
 
+// getExtendedResourceOps returns a slice of operations to perform on the node status
+func getExtendedResourceOps(n *api.Node, extendedResources ExtendedResources) []statusOp {
+	var statusOps []statusOp
+
+	oldResources := strings.Split(n.Annotations[AnnotationNs+"extended-resources"], ",")
+
+	// figure out which resources to remove
+	for _, resource := range oldResources {
+		// provided that the old resource exists in the node
```
marquiz: I just realized that we need to do similar handling of the labels namespace/prefix that we do in `addLabels(). Worker can send labels either in non-namespaced format (in which case the namespace is added automatically), or, in the full namespaced format.

pkg/nfd-master/nfd-master.go
```diff
@@ -229,7 +246,22 @@ func filterFeatureLabels(labels Labels, extraLabelNs []string, labelWhiteList *r
...
+		if _, ok := labels[extendedResourceName]; ok {
+			if _, err := strconv.Atoi(labels[extendedResourceName]); err != nil {
+				stderrLogger.Printf("bad label value encountered for extended resource: %s", err.Error())
+				continue // non-numeric label can't be used
+			}
+
+			extendedResources[extendedResourceName] = labels[extendedResourceName]
+			delete(labels, extendedResourceName)
+		}
+	}
```
marquiz: I think we should do label namespace/prefix handling here, dropping the default namespace if specified (i.e. something like `strings.TrimPrefix(extendedResourceName , LabelNs)`). This way, the use could specify labels in the default ns in two ways, either in the fully namespaced format (`feature.node.kubernetes.io/source-feature.name`) or in a shorter (`source-feature.name`) format. This would comply with the label handling/naming used in the `local` feature source in nfd-worker.

This also needs to be documented in README

uniemimu:
> I think we should add a (commented out) example of the required RBAC rule(s) in `nfd-master.yaml.template`

Indeed, that is a good idea. About the namespace adding and removing magic: I tend to be in the KISS-camp, but since such a mechanism exists for the labels already, I'll try to mimic that for the extended resources.

marquiz:
> I tend to be in the KISS-camp, but since such a mechanism exists for the labels already, I'll try to mimic that for the extended resources.

👍 Yes, the namespacing is a unfortunately a bit convoluted, because of historical reasons. However, before this PR it has mostly been an internal implementation detail in worker-master communication

zvonkok:
@marquiz I would also see it as an experimental feature. My only concern is that we here create a loop-hole for not using device-plugins where the community has agreed upon on how to handle devices with health checks and updating the count of resources being used of free to allocate. 

On the other hand, e.g. on-die features that "just" work, do not need a device-plugin one just wants to expose and use them. To not oversubscribe would be the responsibility of the admin or workload maintainer, pipeline architect or whoever is designing the application. 

Regarding how the slices are allocated by a device plugin, it is how the device plugin was developed. If you e.g. allocate 4 GPUs the device-plugin will not allocate one by one but rather return a list of GPUs [https://github.com/NVIDIA/k8s-device-plugin/blob/f737ebb9aab0d50b24b1ce513236410fcfd0dee1/server.go#L212](https://github.com/NVIDIA/k8s-device-plugin/blob/f737ebb9aab0d50b24b1ce513236410fcfd0dee1/server.go#L212)

Long story short I am good with it as an experimental feature, we might get more cases in the future where this functionality makes sense. 

We should explicitly state somewhere that this is experimental and not a replacement for device-plugins.

mythi:
>  If you e.g. allocate 4 GPUs the device-plugin will not allocate one by one but rather return a list of GPUs

Right, this was a miss from my side. The "list" is still very long if the resources are memory slices and the app is requesting a lot of them. Likewise, kubelet has internal structures to deal with all registered "devices".

> We should explicitly state somewhere that this is experimental and not a replacement for device-plugins.

You'd still need the device-plugin to be able to access the device so I see this as an optimization.

marquiz:
@zvonkok thanks for your thoughts. Very valid points (of view) and concerns. I think we're on the same page. And, we can go forward with this.

@uniemimu please add the disclaimers (experimental and not to be used in place of device plugins) to the README.

README.md
```diff
@@ -441,6 +443,33 @@ temporary file (outside the `source.d` and `features.d` directories), and,
...
+An example use-case for the extended resources could be based on a hook which
+creates a label for the node SGX EPC memory section size. By giving the name of
+that label in the `--resource-labels` flag, that value will then turn into an
+extended resource of the node, allowing PODs to request that resource and the
+Kubernetes scheduler to schedule such PODs to only those nodes which have a
+sufficient capacity of said resource left.
+
+Similar to labels, the default namespace `feature.node.kubernetes.io` is
+automatically prefixed to the extended resource, if the promoted label doesn't
+have a namespace. The default namespace is not stored in node annotations.
```
marquiz: I think the sentence about node annotations could/should be removed as the NFD specific annotations are not mentioned in the document elsewhere.

Just as a reminder to myself we should actually document those. Even if some (like feature labels and extended resources) are sort of "implementation details" of NFD there are the version annotations which might be of more interest to the user. Created #296 to track this

README.md
```diff
@@ -441,6 +443,33 @@ temporary file (outside the `source.d` and `features.d` directories), and,
...
+
+Example usage of the command line arguments, using a new namespace:
+`nfd-master --resource-labels=sgx.some.ns/epc --extra-label-ns=sgx.some.ns`
```
marquiz: One final comment 😉 Should we add an example of a default ns label, in order to cover that use case, too. Smth like
```
nfd-master --resource-labels=my_source-my.feature,sgx.some.ns/epc --extra-label-ns=sgx.some.ns
```

Would produce extended resources
```
sgx.some.ns/epc: <label value>
feature.node.kubernetes.io/my_source-my.feature: <label value>
```

README.md
```diff
@@ -441,6 +443,36 @@ temporary file (outside the `source.d` and `features.d` directories), and,
...
+have a namespace.
+
+Example usage of the command line arguments, using a new namespace:
+`nfd-master --resource-labels=my_source-my.feature,sgx.some.ns/epc --extra-label-ns=sgx.some.ns`
+
+The above would result in following extended resources provided that related
+labels exist:
+
+  sgx.some.ns/epc: <label value>
+  feature.node.kubernetes.io/my_source-my.feature: <label value>
```
marquiz: Add the triple backticks (```) so that this block gets formatted correctly

zvonkok:
Thanks  @uniemimu!  No additional comments. 

/lgtm

PR #292, opened 2020-02-27
test/e2e: more comprehensive per-node testing

marquiz:
Implement an end-to-end test with all feature sources enabled. The new
test runs nfd-worker as a daemonset on all (schedulable) nodes of the
test cluster which makes it possible to cover a wide range features,
assuming the test cluster is heterogenous containing nodes with varying
system configurations.

The features available depends on the node(s) the e2e testa are run on.
Thus, some runtime parameterization of the tests is needed. The patch
adds a new command line test flag 'nfd.e2e-config' that is used to
specify the per-node feature labels and annotations that is expected to
be present in the cluster. An example configuration file is provided
with the patch. The pod spec of nfd-worker deployment is changed to
better correspond the default deployment and thus enable wider feature
discovery. This means using hostnetwork and adding mounts for /sys /boot
and /etc/os-release.

The patch changes node object management so that all nfd-related labels
are removed after each test (not just the ones the test is expected to
add). Also, all nfd-related annotations are now removed.

This PR also contains integration with Travis-CI. My personal repository has a [devel/e2e-node-debug](https://github.com/marquiz/node-feature-discovery/tree/devel/e2e-node-debug) branch demonstrating the new end-to-end test in action: test results can be observed in [Travis-CI](https://travis-ci.org/marquiz/node-feature-discovery/builds/655787115)


.travis.yml
```diff
@@ -21,7 +21,8 @@ script:
       chmod a+x $HOME/bin/aws-iam-authenticator;
       export PATH=$PATH:$HOME/bin;
       echo "$KUBECONFIG_AWS" > kubeconfig_aws;
-      make e2e-test -e KUBECONFIG=`pwd`/kubeconfig_aws;
+      echo "$E2E_TEST_CONFIG_DATA" > e2e-test-config.json;
```
Ethyling: Is the config file supposed to be in YAML or JSON? There is some JSON references but the unmarshall function and the config file example are using YAML

test/e2e/node_feature_discovery.go
```diff
@@ -190,65 +247,182 @@ func nfdMasterPod(ns string, image string, onMasterNode bool) *v1.Pod {
...
-						{
-							Name: "NODE_NAME",
-							ValueFrom: &v1.EnvVarSource{
-								FieldRef: &v1.ObjectFieldSelector{
-									FieldPath: "spec.nodeName",
-								},
+		Spec: nfdWorkerPodSpec(image, extraArgs),
+	}
+
+	p.Spec.RestartPolicy = v1.RestartPolicyNever
```
Ethyling: Why do we set this outside of the `nfdWorkerPodSpec` function?

test/e2e/node_feature_discovery.go
```diff
@@ -260,27 +434,153 @@ var _ = framework.KubeDescribe("Node Feature Discovery", func() {
...
+				By("Creating nfd-worker daemonset")
+				workerDS := nfdWorkerDaemonSet(fmt.Sprintf("%s:%s", *dockerRepo, *dockerTag), []string{})
+				workerDS, err := f.ClientSet.AppsV1().DaemonSets(f.Namespace.Name).Create(workerDS)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("Waiting for daemonset pods to be ready")
+				Expect(e2epod.WaitForPodsReady(f.ClientSet, f.Namespace.Name, workerDS.Spec.Template.Labels["name"], 5)).NotTo(HaveOccurred())
+
+				//By("Sleeping....")
+				//time.Sleep(120 * time.Second)
```
Ethyling: Do we need this or should we remote it?

marquiz: Woops 😬 Good catch, good that you're really reading the code. Fixed

marquiz: I.e. removed

.travis.yml
marquiz: Both json and yaml are supported. But good point, I just added .json suffix because I use JSON format in my travis setup. I removed the .json suffix

test/e2e/node_feature_discovery.go
marquiz: That's because we use the same podspec (returned by `nfdWorkerPodSpec()` for the daemonset and daemonset may not have `RestartPolicyNever`

Ethyling:
Seems to be working on my cluster
/lgtm

Ethyling:
https://travis-ci.org/kubernetes-sigs/node-feature-discovery/builds/657731000
@marquiz The test has been skipped: "no e2e-config was specified". Maybe the env variable is not set as expected?

marquiz:
Thanks @Ethyling for the review !

> https://travis-ci.org/kubernetes-sigs/node-feature-discovery/builds/657731000
> @marquiz The test has been skipped: "no e2e-config was specified". Maybe the env variable is not set as expected?

Woops, thanks for the heads-up! My bad :grimacing:  Exactly as you suspected I had forgotten to set the env variable. Fixed now:
https://travis-ci.org/kubernetes-sigs/node-feature-discovery/builds/657731000

PR #294, opened 2020-03-05
Change nfd-master deployment mechanism from daemonset to deployment

RenaudWasTaken:
On some managed services (e.g: GKE), the node feature discovery master doesn't get deployed because there aren't any master nodes in the cluster.

This PR changes the deployment mechanism of the master from Daemonset to Deployment with a single replica. This replica can then be scaled up if HA is required. Additionally, instead of requiring the master label, we use the ["node-affinity" mechanisms](https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity)

Signed-off-by: Renaud Gaubert <rgaubert@nvidia.com>


Ethyling:
Should we add something in the README about HA setup? This would require to scale the deployment and maybe add pod anti-affinity I guess? WDYT @marquiz @RenaudWasTaken ?

RenaudWasTaken:
Updated the README

PR #298, opened 2020-03-20
version: allow undefined version

marquiz:
Just print a warning instead of exiting with an error if no version has
been specified at build-time. This was pointless and just annoying at
development time when doing builds with go directly.


zvonkok:
Definitely, this was soo annoying :) 

/lgtm

Issue #297, opened 2020-03-19
Default location of /etc/os-release in Fedora is /usr/lib/os-release


zvonkok:
Good and bad news regarding Fedora, if Fedora had `os-release` in the "right" place NFD would just work 
```
2020/03/19 15:56:06 system-os_release.ID = fedora
2020/03/19 15:56:06 system-os_release.VERSION_ID = 31
2020/03/19 15:56:06 system-os_release.VERSION_ID.major = 31
2020/03/19 15:56:06 system-os_release.VERSION_ID.minor = 
```
But unfortunately `/etc/os-release` is a symlink to `.../usr/lib/os-release`

Sooner or later this change may appear in RHEL.

zvonkok:
@ArangoGutierrez PTAL

marquiz:
Hmm, shouldn't this just work, already now 🤔  As `/etc/os-release` is mounted inside the nfd container it should be the symlink target that gets there(?)

zvonkok:
I am running a Fedora 31
```
cat /etc/os-release | grep PRETTY
PRETTY_NAME="Fedora 31 (Workstation Edition)"
```
Debian e.g and Fedora symlink `/etc/os-release` to `../usr/lib/os-release`. 
Running an F29 container and mounting the `os-release` will show the wrong information about the host. 
```
podman run -it --rm    -v /etc/os-release:/host-etc/os-release    fedora:29  cat /etc/os-release | grep PRETTY
PRETTY_NAME="Fedora 29 (Container Image)"
```

Running a Debian container shows me that my host is now a Debian. 

```
podman run -it --rm    -v /etc/os-release:/host-etc/os-release    debian  cat /etc/os-release | grep PRETTY
PRETTY_NAME="Debian GNU/Linux 10 (buster)"
```

The symlink will be followed inside the container. Docker will not resolve the symlink and mounting `/usr/lib` additionall, we get the right information. 

```
podman run -it --rm    -v /etc/os-release:/host-etc/os-release  -v /usr/lib/os-release:/usr/lib/os-release   debian  cat /etc/os-release | grep PRETTY
PRETTY_NAME="Fedora 31 (Workstation Edition)"
```

marquiz:
I didn't quite get that 🤔 We expose os-release of the host OS, not the container. We mount it under `/host-etc` not `/etc`

Works as expected, at least for me:
```
[...]

PR #299, opened 2020-03-20
source/system: do not publish empty version labels

marquiz:
Do not publish non-existing version components as empty labels.

Reading the description of #297 I realised that it would probably be better to not publish the missing version components as labels with empty values.

### DIFF
--- a/source/system/system.go
+++ b/source/system/system.go
@@ -51,7 +51,9 @@ func (s Source) Discover() (source.Features, error) {
 				if key == "VERSION_ID" {
 					versionComponents := splitVersion(value)
 					for subKey, subValue := range versionComponents {
-						features[feature+"."+subKey] = subValue
+						if subValue != "" {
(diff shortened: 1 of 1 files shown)


Ethyling:
@marquiz E2E tests are failing. We need to remove the `feature.node.kubernetes.io/system-os_release.VERSION_ID.minor` in the section `expectedLabelKeys` as the distribution we are using in E2E tests does not have a minor.

marquiz:
> E2E tests are failing. We need to remove the `feature.node.kubernetes.io/system-os_release.VERSION_ID.minor` in the section `expectedLabelKeys` as the distribution we are using in E2E tests does not have a minor

Thanks for the note! That should be fixed, now

marquiz:
> should be fixed, now

OK, now it is 😉 (I messed up this fix, too, initially)

PR #304, opened 2020-04-23
Rework config handling

marquiz:
Extend the FeatureSource interface with new methods for configuration
handling. This enables easier on-the fly reconfiguration of the
feature sources. Further, it simplifies adding config support to feature
sources in the future.

On top of that, implement dumb re-read/re-parse of the configuration file on every round of
discoery. Probably not the most elegant solution to watch for config
file changes, but, it works and doesn't cost much overhead.

The patchset also fixes some (corner) cases with the overrides (--options)
handling, too:
- Overrides were not applied if config file was missing or its parsing
  failed
- Overrides for a certain source did not have effect if an empty config
  for the source was specified in the config file. This was caused by
  the first pass of parsing (config file) setting a nil pointer to the
  source-specific config, effectively detaching it from the main config.
  The second pass would then create a new instance of the source
  specific config, but, this was not visible in the feature source, of
  course.


Ethyling:
We should maybe add something in the README about this.

marquiz:
> We should maybe add something in the README about this.

You're right

/hold

Ethyling:
Otherwise the code looks good to me. I still need to test it

marquiz:
> Otherwise the code looks good to me. I still need to test it

README updated. Have you had time to test this?

/unhold

Ethyling:
Not yet, I will check it soon.

Ethyling:
Tested and working fine. Just a small detail, should we add a log line to specify that we are reloading the config file?

marquiz:
> Tested and working fine.

Rebased, but, didn't have the time to properly re-test yet.

> Just a small detail, should we add a log line to specify that we are reloading the config file?

Added a log print after every successful loading of the config.

Ethyling:
When there is an error in the config file we have this output:
```
2020/05/20 20:18:33 Failed to parse config file: error converting YAML to JSON: yaml: line 30: could not find expected ':'
2020/05/20 20:18:33 Configuration successfully loaded from "/etc/kubernetes/node-feature-discovery/nfd-worker.conf"
```

marquiz:
> When there is an error in the config file we have this output:

😭Fixed. Looks like it's time for me to get to bed 😴

marquiz:
Rebased once more

Issue #306, opened 2020-04-27
How can I prevent from labeling/annotating 


gerardo8a:
I only need to annotate and label custom data. Using empty whitelist should do the trick? if have a simple configmap for example:

```
sources:
      cpu:
        cpuid:
          attributeWhitelist:
```

So I assume all cpu should be skipped? the node still has the annotations made before I added the new config.

gerardo8a:
Looking at the code it seems the whitelist for label is done on themaster via argument. I thought this same thing was achive from the worker so no info is send to the master

marquiz:
@gerardo8a I suppose you had your use case resolved.

Both the master and worker have `--label-whitelist` command line flag with which you can specify a regexp for accepted label names (applied on the name part after the label namespace). An empty regexp (the default) matches everything i.e. does not drop anything. Specifying this for nfd-worker let's you control it on node level whereas with nfd-master you get cluster-level control.

The config option you were looking at (`sources.cpu.cpuid.attributeWhitelist`) only controls the CPUID attributes of the `cpu` feature source (i.e. `cpu-cpuid.*` labels) and does not affect any other labels.

What you probably should use is the `--sources` command line flag of nfd-worker. That is, specify `--sources=local` in case you're using the local hooks / feature files for you custom labels. This way, all the other feature sources than local will be disabled and no labels for those are created.

All this might be a bit confusing, I admit.

> the node still has the annotations made before I added the new config.

NFD will remove old labels when doing re-labeling so you should not have any stale labels after changing the config. Currently, nfd-worker does not detect configuration changes on run-time so you need to restart it after modifying the config (stupid!). I have an open PR (#304) to fix that, though

PR #307, opened 2020-04-28
nfd-worker: fix --label-whitelist

marquiz:
After commenting #306 I looked into the code and was not happy what I found (with my own work).

Unify handling of --label-whitelist in nfd-worker and nfd-master. That is,
in nfd-worker, apply the regexp filter on non-namespaced part of the
label name.

Brief history:
1. Originally the whitelist regexp was applied on the full namespaced
   label name (that would be e.g.
   'feature.node.kubernetes.io/cpu-cpuid.AVX' in the current nfd version)

2. Commit 81752b2d changed the behavior so that the regexp was applied
   on the non-namespaced part (that would be `cpu-cpuid.AVX`)

3. Commit 40918827 added support for custom label namespaces. With this
   change, the label whitelist handling diverged between nfd-worker and
   nfd-master. In nfd-master the whitelist regexp is always applied on
   the non-namespaced label name. However, in nfd-worker the whitelist
   handling is two-fold (and inconsistent): for labels in the standard
   nfd namespace regexp is applied on the non-namespaced part (e.g.
   `cpu-cpuid.AVX`, but, for labels in custom namespaces the regexp is
   applied on the full name (e.g. `example.com/my-feature`).

This patch changes nfd-worker to behave similarly to nfd-master. The
namespace part is now always omitted, which should be easier for the
users to comprehend.

Also, fixes a bug in the label name prefixing so that the name of the
feature source is not prefixed into labels with custom label namespace
(effectively mangling the intended namespace). For example, previously a
'example.com/feature' label from the 'custom' feature source would be
prefixed with the source name, mangling it to
'custom-example.com/feature'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,10 @@ nfd-master.
                                   has been enabled.
   --no-publish                    Do not publish feature labels
   --label-whitelist=<pattern>     Regular expression to filter label names to
-                                  publish to the Kubernetes API server. [Default: ]
+                                  publish to the Kubernetes API server.
+                                  NB: the label namespace is omitted i.e. the filter
+                                  is only applied to the name part after '/'.
+                                  [Default: ]
   --extra-label-ns=<list>         Comma separated list of allowed extra label namespaces
                                   [Default: ]
   --resource-labels=<list>        Comma separated list of labels to be exposed as extended resources.
@@ -129,11 +132,14 @@ nfd-worker.
                               in testing
                               [Default: ]
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpu,iommu,kernel,local,memory,network,pci,storage,system,usb]
+                              [Default: cpu,custom,iommu,kernel,local,memory,network,pci,storage,system,usb]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
-                              publish to the Kubernetes API server. [Default: ]
+                              publish to the Kubernetes API server.
+                              NB: the label namespace is omitted i.e. the filter
+                              is only applied to the name part after '/'.
+                              [Default: ]
   --oneshot                   Label once and exit.
   --sleep-interval=<seconds>  Time to sleep between re-labeling. Non-positive
                               value implies no re-labeling (i.e. infinite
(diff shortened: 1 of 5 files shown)


marquiz:
Forgot to add that this is one possible way to "fix" this. Alternatives include:
1. Change nfd-master to behave similarly to nfd-worker, instead
2. Revert back to the original, i.e. always apply the regexp to the fully namespaced label name

/hold

Ethyling:
I think we should apply the same policy for labels regardless of the namespace (default or not). 

However, I don't know if we should apply the filter on the namespace or not. I would say that applying the whitelisting on namespaces would add the possibility to filter labels as we want on each node. For example, currently we can't filter namespaces on nfd-worker, we can only do this on nfd-master, so it would allow us to limit namespaces per node. But it would add more "complexity" on this feature for the user.

What do you think?

marquiz:
These were the same questions I was pondering.

The reason I ended up the solution implemented here (i.e. whitelisting NOT applied on namespace) is that 
1. nfd-master already has other means of controlling the namespaces (`--extra-label-ns`). It is not exactly the same thing, but, probably enough
2. Labels in non-default namespaces are always result of user-made customization (hooks, feature files or the `custom` feature source). Thus, it's less likely that the user would want to filter out labels that she has has specifically enabled herself.

Ethyling:
> The reason I ended up the solution implemented here (i.e. whitelisting NOT applied on namespace) is that
> 
>     1. nfd-master already has other means of controlling the namespaces (`--extra-label-ns`). It is not exactly the same thing, but, probably enough
> 
>     2. Labels in non-default namespaces are always result of user-made customization (hooks, feature files or the `custom` feature source). Thus, it's less likely that the user would want to filter out labels that she has has specifically enabled herself.

Sounds reasonable to me. I will take a look at the code and test it.

mythi:
> Unify handling of --label-whitelist in nfd-worker and nfd-master. 

Maybe slightly off-topic but what's the use-case for double filtering?

marquiz:
> Maybe slightly off-topic but what's the use-case for double filtering?

In short: historical reasons 😄 It was "cloned" when NFD was split to master and worker.

Ethyling:
LGTM but I think we should update the README to specify that the whitelist is not applied on the namespace part. WDYT?

marquiz:
> I think we should update the README to specify that the whitelist is not applied on the namespace part.

Agree. Added documentation on the command line help and via that also to the README. Do you think that would be sufficient?

Also rebased on top of the latest master.

Ethyling:
> > I think we should update the README to specify that the whitelist is not applied on the namespace part.
> 
> Agree. Added documentation on the command line help and via that also to the README. Do you think that would be sufficient?

Yes I think. Last checks and we can merge this :smile:

Issue #308, opened 2020-05-14
Kernel version full is not compatible with L4T


paroque28:
```
2020/05/14 15:12:57 Ignoring invalid feature value kernel-version.full=4.9.140-l4t-r32.3.1+g47e7e1cb0b49: [a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')]
```

Should NFD replace `+` characters with something else?

marquiz:
I think this makes sense for the version.full label. We should prolly replace all invalid characters, not just `+`

paroque28:
Fixed with mentioned pull request

PR #309, opened 2020-05-14
Proposal kernel.go: Replace forbidden symbols and check format

paroque28:
Some Kernel versions include symbols such as "+". Yocto L4T kernel is
an example of this behaviour. To fix this error the symbols are replaced
by an underscore.

Signed-off-by: Pablo Rodriguez <paroque28@gmail.com>

This is associated with issue:
https://github.com/kubernetes-sigs/node-feature-discovery/issues/308

### DIFF
--- a/source/kernel/kernel.go
+++ b/source/kernel/kernel.go
@@ -97,6 +97,11 @@ func parseVersion() (map[string]string, error) {
 	}
 
 	full := strings.TrimSpace(string(raw))
+
+	// Replace forbidden symbols
+	fullRegex := regexp.MustCompile("[^-A-Za-z0-9_.]")
+	full = fullRegex.ReplaceAllString(full, "_")
+
 	version["full"] = full
 
 	// Regexp for parsing version components


source/kernel/kernel.go
```diff
@@ -97,6 +101,21 @@ func parseVersion() (map[string]string, error) {
...
+	}
+
+	// Check if string matches standard
+	matched, err := regexp.MatchString(`^[[a-zA-Z0-9]+[a-zA-Z0-9-_.]*[a-zA-Z0-9]+$`, full)
+	if err != nil {
+		log.Printf("ERROR: Failed to read regex for : %s", full)
+	}
+	if !matched {
+		log.Printf("ERROR: Kernel version full should consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character : %s", full)
+	}
```
marquiz, line 117: I think we can make this much simpler. Just replace all forbidden characters without any configurability.

E.g. global
```
re := regexp.MustCompile("[^-A-Za-z0-9_.]")
```

And the for the full version label:
```
fmt.Println(re.ReplaceAllString(value, "_"))
```

paroque28:
Hi @marquiz ,
Thanks for your review, I've taken your suggestions and tested the patch on my cluster:
```
$ kubectl logs
...
2020/05/15 17:54:52 kernel-config.PREEMPT = true
2020/05/15 17:54:52 kernel-version.minor = 9
2020/05/15 17:54:52 kernel-version.revision = 140
2020/05/15 17:54:52 kernel-version.full = 4.9.140
```
```
$ kubectl describe node
...
feature.node.kubernetes.io/kernel-version.full=4.9.140-l4t-r32.3.1_g47e7e1cb0b49
```

PR #310, opened 2020-05-14
Add support for USB device discovery

pmundt:
This builds on the PCI support to enable the discovery of USB devices. While we allow for class-based discovery, many USB devices shift this to the interface level instead and have meaningless class values. The vendor and device IDs, therefore, are the main items of interest. This is primarily useful for the discovery of Edge-based heterogeneous accelerators that are connected via USB, such as the Coral USB Accelerator and the Intel NCS2 - our main motivation for this work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -129,7 +129,7 @@ nfd-worker.
                               in testing
                               [Default: ]
   --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpu,iommu,kernel,local,memory,network,pci,storage,system]
+                              [Default: cpu,iommu,kernel,local,memory,network,pci,storage,system,usb]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
@@ -161,6 +161,7 @@ The current set of feature sources are the following:
 - PCI
 - Storage
 - System
+- USB
 - Local (hooks for user-specific features)
 
 ### Feature labels
@@ -193,6 +194,7 @@ feature logically has sub-hierarchy, e.g. `sriov.capable` and
   "feature.node.kubernetes.io/pci-<device label>.present": "true",
   "feature.node.kubernetes.io/storage-<feature-name>": "true",
   "feature.node.kubernetes.io/system-<feature name>": "<feature value>",
+  "feature.node.kubernetes.io/usb-<device label>.present": "<feature value>",
   "feature.node.kubernetes.io/<file name>-<feature name>": "<feature value>"
 }
 ```
@@ -316,6 +318,28 @@ Matching is done by performing a logical _OR_ between Elements of an Attribute a
 each PCI device in the system.
 At least one Attribute must be specified. Missing attributes will not partake in the matching process.
 
+##### UsbId Rule
+###### Nomenclature
+```
+Attribute   :A USB attribute.
+Element     :An identifier of the USB attribute.
+```
+
+The UsbId Rule allows matching the USB devices in the system on the following Attributes: `class`,`vendor` and
+`device`. A list of Elements is provided for each Attribute.
+
+###### Format
+```yaml
+usbId :
+  class: [<class id>, ...]
+  vendor: [<vendor id>,  ...]
+  device: [<device id>, ...]
+```
+
+Matching is done by performing a logical _OR_ between Elements of an Attribute and logical _AND_ between the specified Attributes for
+each USB device in the system.
+At least one Attribute must be specified. Missing attributes will not partake in the matching process.
+
 ##### LoadedKMod Rule
 ###### Nomenclature
 ```
@@ -342,6 +366,11 @@ custom:
       - pciId:
           vendor: ["15b3"]
           device: ["1014", "1017"]
+  - name: "my.usb.feature"
+    matchOn:
+      - usbId:
+          vendor: ["1d6b"]
+          device: ["0003"]
   - name: "my.combined.feature"
     matchOn:
       - loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
@@ -361,6 +390,8 @@ __In the example above:__
 if the node has `kmod1` _AND_ `kmod2` kernel modules loaded. 
 - A node would contain the label: `feature.node.kubernetes.io/custom-my.pci.feature=true`
 if the node contains a PCI device with a PCI vendor ID of `15b3` _AND_ PCI device ID of `1014` _OR_ `1017`.
+- A node would contain the label: `feature.node.kubernetes.io/custom-my.usb.feature=true`
+if the node contains a USB device with a USB vendor ID of `1d6b` _AND_ USB device ID of `0003`.
 - A node would contain the label: `feature.node.kubernetes.io/custom-my.combined.feature=true`
 if `vendor_kmod1` _AND_ `vendor_kmod2` kernel modules are loaded __AND__ the node contains a PCI device
 with a PCI vendor ID of `15b3` _AND_ PCI device ID of `1014` _or_ `1017`.
@@ -432,6 +463,21 @@ feature.node.kubernetes.io/pci-1200_8086.present=true
 Also  the set of PCI device classes that the feature source detects is
 configurable. By default, device classes (0x)03, (0x)0b40 and (0x)12, i.e.
 GPUs, co-processors and accelerator cards are detected.
+
+### USB Features
+
+| Feature              | Attribute     | Description                               |
+| -------------------- | ------------- | ----------------------------------------- |
+| &lt;device label&gt; | present       | USB device is detected
+
+`<device label>` is composed of raw USB IDs, separated by underscores.
+The set of fields used in `<device label>` is configurable, valid fields being
+`class`, `vendor`, and `device`.
+Defaults are `class`, `vendor` and `device`. An example label using the default
+label fields:
+```
+feature.node.kubernetes.io/usb-fe_1a6e_089a.present=true
+```
 
 See [configuration options](#configuration-options)
 for more information on NFD config.
--- a/source/internal/usb_utils.go
+++ b/source/internal/usb_utils.go
@@ -1,0 +1,138 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package busutils
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"path"
+	"path/filepath"
+	"strings"
+)
+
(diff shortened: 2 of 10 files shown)


pmundt:
I signed it

marquiz:
Thank you for your patch @pmundt !

Some quick comments and questions, without taking a close look at the code, yet:
- The usb matcher for custom source is a good addition 👍 
- I think the new usb source is justifiable. However, I think that needs to have configurable whitelisting (e.g. something similar what that the pci source has). We want to minimize the number of useless labels, e.g. we don't want to see a label from every usb hub attached to the system.
- Are there some other attributes that should be available (like subclass)?
- Is the concept extensible to match/advertise interfaces, too, if that would be needed in the future? I don't necessarily hope for that to happen, but better prepare for the worst 😉

pmundt:
Thanks for the review. In response to your comments:

- I had thought about whitelisting as well, but the class reporting on USB is unfortunately not as straightforward as on PCI - the problem is that the base class can be encoded at the device or per-interface level, while the vendor/product IDs are at the device level. I can do it if necessary, but did not really want to get into doing per-device interface enumeration as this adds more complexity than what we really need - determining whether a vendor/device pair exists. That being said, I would not be opposed to perhaps flipping this on its head and doing a class blacklist instead - things like hubs  tend to reliably report their base classes at the device level often enough that we can use this as a basis for exclusion, for example.
- The other attributes that we could expose would be the subclass and protocol, but I've left this out for the time being as I'm hard pressed to come up with a scenario where anyone would be interested in either of these as a basis for labelling. This is more if you want to single out a keyboard or a mouse attached to a HID device for example. If someone has a real use case for either of these attributes, I'd be happy to add them in.
- As far as match/advertise functionality goes - I think this could be applied to anything that is choosing to walk the sysfs tree as a means of poor man's bus enumeration. There is probably utility in reworking this into a proper interface and just having each bus provide its own implementation, but I don't know that we're really at a point where we can generalize this sufficiently at the moment - the USB case worked out easily because it's similar enough to the PCI case. Others, like i2c, are different enough that the current approach would need to be reworked more substantially.

marquiz:
> * I had thought about whitelisting as well, but the class reporting on USB is unfortunately not as straightforward as on PCI - the problem is that the base class can be encoded at the device or per-interface level, while the vendor/product IDs are at the device level. I can do it if necessary, but did not really want to get into doing per-device interface enumeration as this adds more complexity than what we really need - determining whether a vendor/device pair exists.

Need to think about that. Maybe we should walk thourgh the interfaces, too, and dig out the actual class(es) available 🤔The `vendor-device` based label is very specific and narrow in scope and as such I'm still a bit dubious about that. Can you be more specific about the use case (or how you plan to use the labels), deploying device plugin(s) or something else?

> That being said, I would not be opposed to perhaps flipping this on its head and doing a class blacklist instead - things like hubs  tend to reliably report their base classes at the device level often enough that we can use this as a basis for exclusion, for example.

This might be an option. In any case we need to think carefully what to white/black list. E.g. hubs shouldn't be seen, for sure, and e.g. hid devices neither by default.


> * The other attributes that we could expose would be the subclass and protocol, but I've left this out for the time being as I'm hard pressed to come up with a scenario where anyone would be interested in either of these as a basis for labelling. This is more if you want to single out a keyboard or a mouse attached to a HID device for example. If someone has a real use case for either of these attributes, I'd be happy to add them in.

Sure 👍 Can be added later if a use case arises

> * As far as match/advertise functionality goes - I think this could be applied to anything that is choosing to walk the sysfs tree as a means of poor man's bus enumeration. There is probably utility in reworking this into a proper interface and just having each bus provide its own implementation, but I don't know that we're really at a point where we can generalize this sufficiently at the moment - the USB case worked out easily because it's similar enough to the PCI case. Others, like i2c, are different enough that the current approach would need to be reworked more substantially.

Ach, sorry about my bad working. I was thinkin about the USB interfaces (under a device entry) 😄 


I think I need to do some testing of my own with this and come up with fresh comments next week.

pmundt:
> > * I had thought about whitelisting as well, but the class reporting on USB is unfortunately not as straightforward as on PCI - the problem is that the base class can be encoded at the device or per-interface level, while the vendor/product IDs are at the device level. I can do it if necessary, but did not really want to get into doing per-device interface enumeration as this adds more complexity than what we really need - determining whether a vendor/device pair exists.
> 
> Need to think about that. Maybe we should walk thourgh the interfaces, too, and dig out the actual class(es) available The `vendor-device` based label is very specific and narrow in scope and as such I'm still a bit dubious about that. Can you be more specific about the use case (or how you plan to use the labels), deploying device plugin(s) or something else?

I had a go at this, and it ended up being not too complicated. The additional complexity is really in the fact that devices that pass this down to the interface level may have multiple interfaces with the same or different classes - I've addressed this by exposing each unique class for the device. This seems to work ok now, and makes building a whitelist easier.

Our main use-case is as in the original commit text - the identification of specific accelerators at the node level. Device plugins would be one thing that would use this, but we're also using it as a basis for identifying node capabilities to build optimized containers and then control their targeted deployment onto specific nodes by the orchestrator, as part of our work in the [SODALITE H2020 project](https://sodalite.eu/). In this specific case, we have both NCS2 and Coral AI Accelerators Edge that are USB-attached on the Edge nodes.

> This might be an option. In any case we need to think carefully what to white/black list. E.g. hubs shouldn't be seen, for sure, and e.g. hid devices neither by default.
>
With the interface classes enumerated, doing a whitelist actually turned out easier, there's only a small handful of classes under which accelerators of interest are mapped:

- Video
- Miscellaneous
- Application Specific
- Vendor Specific

This should cut down on the labelling signal-to-noise ratio at least. The full list is here https://www.usb.org/defined-class-codes.

I've now pushed out these changes for consideration.

source/internal/usb_utils.go
```diff
@@ -0,0 +1,134 @@
...
+		// unique interface class.
+		for _, intf := range interfaces {
+			// Determine the interface class
+			attrVal, err := readSingleUsbSysfsAttribute(intf)
+			if err != nil {
+				return classmap, err
+			}
+
+			// Update interface class
+			info["class"] = attrVal
```
marquiz, line 91: Instead of modifying, I think we need to create a new copy of `info` here. Otherwise we end up modifying the same instance of the map.

source/usb/usb.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
marquiz, line 2: nit: copyright year should be 2020(?)

source/internal/usb_utils.go
pmundt: This was intentional, as we copy it on the next line when inserting it into the classmap, which is the thing that is returned from the function. We never use `info` again, so clobbering "class" on each successive iteration is perfectly fine.

source/usb/usb.go
pmundt: Copy and paste error, thanks.

README.md
```diff
@@ -342,6 +366,10 @@ custom:
       - pciId:
           vendor: ["15b3"]
           device: ["1014", "1017"]
+  - name: "my.usb.feature"
+    matchOn:
```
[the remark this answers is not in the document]
pmundt: Looks like it. Fixed now, thanks.

pmundt:
> One thing that I'm still pondering is the default label format. I.e. is the `<class>_<vendor>_<device>` too specific/detailed, or, would a more `<class>_<vendor>` similar to the pci source be sufficient and more generally useful. WDYT? I prolly have to trust in you on this, anyway 😄

We could do `<class>_<vendor>` now that we have the class whitelist, but as far as the usefulness of this approach, I'm not sure. The problem that I see is that the classes in the whitelist are largely arbitrary - e.g. some vendors choose to put their accelerators under vendor-specific, others under application-specific, this placement is entirely at the whim of the vendor, and we can make no real inference from the class type itself. Taking the application-specific base class for example, the subclasses are pretty much all over the place (https://www.usb.org/defined-class-codes#anchor_BaseClassFEh).

For our purpose we only care about the vendor and device, as we need to know the precise device in order to apply optimizations. Even if we were to change the default labelling to only include the vendor, any actual use of this in practice is still going to need the device ID due to the inability to make any reasonable inference from the class/vendor pair. It narrows the scope, certainly, but I don't see any way to avoid this when the class coding tells us absolutely nothing about the underlying devices that may be present.

I could see how the `<class>_<vendor>` case would be useful if e.g. we happened to have a single vendor reporting multiple video cards under the video class, so perhaps this is reasonable as a default, but none of our devices exhibit such consistent or well-defined encoding.

Perhaps a better solution is if we just add labels for both?

source/internal/usb_utils.go
marquiz: The problem with this is that with `classmap[attrVal] = info` you don't create a new copy of the map, merely just a reference is passed. Thus, later changes to info are also reflected in other `classmap` members created in this for loop.

I.e. if you do somehting like:
```
	a := map[string]string{"foo": "1", "bar": "2"}
	b := a
	b["foo"] = "123"
```
then `a["foo"] == "123"`

marquiz:
> We could do `<class>_<vendor>` now that we have the class whitelist, but as far as the usefulness of this approach, I'm not sure. The problem that I see is that the classes in the whitelist are largely arbitrary - e.g. some vendors choose to put their accelerators under vendor-specific, others under application-specific, this placement is entirely at the whim of the vendor, and we can make no real inference from the class type itself. Taking the application-specific base class for example, the subclasses are pretty much all over the place (https://www.usb.org/defined-class-codes#anchor_BaseClassFEh).

Fair enough. Sounds like the fine grained labeling (default) would be justifiable. I think we could go with your preference. What you could do is add the reasoning above^ e.g. to the commit message, to document the rationale with the codebase. WDYT?

> Perhaps a better solution is if we just add labels for both?

I wouldn't probably do that, at least for now.

source/internal/usb_utils.go
pmundt: Wow, I thought this only applied to pointers, but you're right, the same thing applies to maps. Learn something new every day. I've now replaced this with a local copy for each iteration.

pmundt:
> > We could do `<class>_<vendor>` now that we have the class whitelist, but as far as the usefulness of this approach, I'm not sure. The problem that I see is that the classes in the whitelist are largely arbitrary - e.g. some vendors choose to put their accelerators under vendor-specific, others under application-specific, this placement is entirely at the whim of the vendor, and we can make no real inference from the class type itself. Taking the application-specific base class for example, the subclasses are pretty much all over the place (https://www.usb.org/defined-class-codes#anchor_BaseClassFEh).
> 
> Fair enough. Sounds like the fine grained labeling (default) would be justifiable. I think we could go with your preference. What you could do is add the reasoning above^ e.g. to the commit message, to document the rationale with the codebase. WDYT?

That's fine with me, do you want me to squash everything into a single commit with an updated changelog and open a new PR?

source/internal/usb_utils.go
marquiz: 👍 Beware that this applies to slices, too 😉

marquiz:
> That's fine with me, do you want me to squash everything into a single commit with an updated changelog and open a new PR?

Yes, please squash. With a meaningful commit message  I think you have some good explanation/background information in some of the commit message of this patchset so don't lose that information 😄I don't think you need to create a new PR, just squash and force push over your github branch(?)

pmundt:
My apologies for the slow response, I've just had a chance to get back to this now. I've squashed all the commits and force-pushed a single commit with a more comprehensive commit log that should encapsulate what we have discussed here. It should now be ready to merge.

PR #313, opened 2020-05-19
Slight rework of building

marquiz:
Add build and install targets to Makefile and take use of that in the Dockerfile


Ethyling:
Should we add the bin directory to gitignore and dockerignore?

marquiz:
> Should we add the bin directory to gitignore and dockerignore?

Very good observation! Added. And rebased.

Ethyling:
Need to be rebased to pass the CI (to have the fix in the usb custom source) :smile:

marquiz:
> Need to be rebased to pass the CI

Ha, rebased (thanks for the fix!)

marquiz:
Rebased. I needed to add two variables in makefile to preserve separate hostmount prefix for local and container builds. Also added documentation for the new variables (hopefully those make sense)

marquiz:
Rebased

PR #315, opened 2020-05-19
Verify golangci-lint

marquiz:
Integrate golangci-lint with our build. Fix lint errors.

### DIFF
--- a/pkg/cpuid/cpuid_amd64.s
+++ b/pkg/cpuid/cpuid_amd64.s
@@ -16,9 +16,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-TEXT ·cpuidAsm(SB), 4, $0   // 4 = NOSPLIT
-    MOVL    eax_arg+0(FP), AX
-    MOVL    ecx_arg+4(FP), CX
+#include "textflag.h"
+
+TEXT ·cpuidAsm(SB), NOSPLIT, $0
+    MOVL    eaxArg+0(FP), AX
+    MOVL    ecxArg+4(FP), CX
     CPUID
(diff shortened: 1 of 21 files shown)


pkg/cpuid/cpuid_amd64.s
```diff
@@ -16,9 +16,9 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-TEXT ·cpuidAsm(SB), 4, $0   // 4 = NOSPLIT
-    MOVL    eax_arg+0(FP), AX
-    MOVL    ecx_arg+4(FP), CX
+TEXT ·cpuidAsm(SB), 4, $0-24   // 4 = NOSPLIT
```
Ethyling, line 19: Why do we add a `-24` here?

Ethyling:
The CI is not working because `golangci-lint` is misisng. Should we fix this in this PR?

pkg/cpuid/cpuid_amd64.s
marquiz: That was to fix this problem:
```
$ go vet ./pkg/cpuid/
# sigs.k8s.io/node-feature-discovery/pkg/cpuid
pkg/cpuid/cpuid_amd64.s:19:1: [amd64] cpuidAsm: wrong argument size 0; expected $...-24
```

But, good question! There was a more proper way to fix this with `#include "textflag.h"`. Changed accordingly

marquiz:
> The CI is not working because `golangci-lint` is misisng. Should we fix this in this PR?

Yes, stupid mistake. I was too busy filing these to even check that they pass CI 🤦 Should be fixed, now.

PR #317, opened 2020-05-19
source/network: run discovery under /host-sys

marquiz:
The default deployment mounts /sys from the host under /host-sys so
let's use that. No need to use hostNetwork after that so drop it from
the worker deployment templates.

### DIFF
--- a/nfd-daemonset-combined.yaml.template
+++ b/nfd-daemonset-combined.yaml.template
@@ -56,7 +56,6 @@ spec:
         app: nfd
     spec:
       serviceAccount: nfd-master
-      hostNetwork: true
       containers:
         - env:
           - name: NODE_NAME
--- a/nfd-worker-daemonset.yaml.template
+++ b/nfd-worker-daemonset.yaml.template
@@ -14,7 +14,6 @@ spec:
       labels:
         app: nfd-worker
     spec:
-      hostNetwork: true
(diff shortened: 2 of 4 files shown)


Ethyling:
We are looping over the host interfaces using `netInterfaces, err := net.Interfaces()`. Without `hostNetwork` the interfaces are the one from the container. I think we should loop over interfaces using the sysfs instead?

marquiz:
> We are looping over the host interfaces using `netInterfaces, err := net.Interfaces()`. Without `hostNetwork` the interfaces are the one from the container. I think we should loop over interfaces using the sysfs instead?

Yes, I realized this when I was writing the sysfs patch in #318, but didn't have the time to fix this one (should've put the hold label here 😬 ). I now pushed a new version ditching "net" for good and doing all discovery under host sysfs. I haven't tested this properly yet, though.

Ethyling:
LGTM

But golangci-lint is not happy:
```golangci-lint run --timeout 5m0s
source/network/network.go:33:2: `flagBroadcast` is unused (deadcode)
        flagBroadcast
        ^
source/network/network.go:34:2: `flagDebug` is unused (deadcode)
        flagDebug
        ^
```

marquiz:
> But golangci-lint is not happy:

Should've checked that myself as I was the one who mandated that rule. Fixed, now.

PR #318, opened 2020-05-20
source: parametrise host directory paths

marquiz:
Specify and handle system paths we use for discovery in a unified way.

Consistently do all sysfs based feature discovery under the same sysfs directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,9 +12,10 @@ RUN go mod download
 COPY . /go/node-feature-discovery
 
 ARG NFD_VERSION
+ARG HOSTMOUNT_PREFIX
 
 RUN go install \
-  -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$NFD_VERSION -X sigs.k8s.io/node-feature-discovery/source.pathPrefix=/host-" \
+  -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$NFD_VERSION -X sigs.k8s.io/node-feature-discovery/source.pathPrefix=$HOSTMOUNT_PREFIX" \
   ./cmd/*
 RUN install -D -m644 nfd-worker.conf.example /etc/kubernetes/node-feature-discovery/nfd-worker.conf
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ COPY . /go/node-feature-discovery
 ARG NFD_VERSION
 
 RUN go install \
-  -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$NFD_VERSION" \
+  -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$NFD_VERSION -X sigs.k8s.io/node-feature-discovery/source.pathPrefix=/host-" \
   ./cmd/*
 RUN install -D -m644 nfd-worker.conf.example /etc/kubernetes/node-feature-discovery/nfd-worker.conf
 
--- a/source/config.go
+++ b/source/config.go
@@ -1,0 +1,39 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package source
+
+import (
+	"path/filepath"
+)
+
+var (
+	pathPrefix = "/"
+	// BootPath is where the /boot directory of the system to be inspected is located
+	BootDir = HostDir(pathPrefix + "boot")
+	// EtcPath is where the /etc directory of the system to be inspected is located
+	EtcDir = HostDir(pathPrefix + "etc")
+	// SysfsPath is where the /sys directory of the system to be inspected is located
+	SysfsDir = HostDir(pathPrefix + "sys")
+)
+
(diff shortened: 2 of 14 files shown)


source/cpu/cpu.go
```diff
@@ -19,14 +19,11 @@ package cpu
...
 	"log"
-	"path"
 
 	"sigs.k8s.io/node-feature-discovery/source"
 )
 
-const (
-	cpuDevicesBaseDir = "/sys/bus/cpu/devices"
-)
+const ()
```
Ethyling, line 26: Can be removed I think

source/config.go
```diff
@@ -0,0 +1,39 @@
...
+*/
+
+package source
+
+import (
+	"path/filepath"
+)
+
+var (
+	pathPrefix = "/"
```
Ethyling, line 24: Should we set the default value to something that would fail to prevent reading the wrong directories if the variable is not set at build time?

Dockerfile
```diff
@@ -14,7 +14,7 @@ COPY . /go/node-feature-discovery
 ARG NFD_VERSION
 
 RUN go install \
-  -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$NFD_VERSION" \
+  -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$NFD_VERSION -X sigs.k8s.io/node-feature-discovery/source.pathPrefix=/host-" \
```
Ethyling, line 17: I think we could use a variable here and use it in the YAMLs templates too. Maybe use the work done in https://github.com/kubernetes-sigs/node-feature-discovery/pull/313. With this, we could do `make PATH_PREFIX=/my_prefix` and we would have the binary and the templates using the same prefix.

WDYT?

source/cpu/cpu.go
marquiz: Thx, will be dropped! I really thought I removed that 🤔

source/config.go
marquiz: I was thinking about it and ended up with this to enable easier development experience. Building/running nfd-worker locally on a development system will cause it to do discovery on the standard system directories. Prefix is set in the Dockerfile to make container-based NFD behave differently (i.e. look under `/host-*`. Does this make sense?

Dockerfile
marquiz: Mm, agree that this probably a nicer solution. I will change accordingly.

source/cpu/cpu.go
marquiz: Fixed

Dockerfile
marquiz: I added a new commit to address this. I don't know if it looks any good(?)

Ethyling: Looks better to me. And it would be really better with the work done on https://github.com/kubernetes-sigs/node-feature-discovery/pull/313 I think

source/cpu/cpu.go
```diff
@@ -137,14 +132,15 @@ func (s Source) Discover() (source.Features, error) {
 
 // Check if any (online) CPUs have thread siblings
 func haveThreadSiblings() (bool, error) {
-	files, err := ioutil.ReadDir(cpuDevicesBaseDir)
+
+	files, err := ioutil.ReadDir(source.SysfsDir.Path("cpu/devices"))
```
Ethyling, line 136: Previously we were reading `/host-sys/bus/cpu/devices` and now we are reading `/host-sys/cpu/devices` which does not exist AFAIK. I think we are missing the `bus` part here

source/cpu/cpu.go
```diff
@@ -137,14 +132,15 @@ func (s Source) Discover() (source.Features, error) {
...
 	if err != nil {
 		return false, err
 	}
 
 	for _, file := range files {
 		// Try to read siblings from topology
-		siblings, err := ioutil.ReadFile(path.Join(cpuDevicesBaseDir, file.Name(), "topology/thread_siblings_list"))
+		siblings, err := ioutil.ReadFile(source.SysfsDir.Path("cpu/devices", file.Name(), "topology/thread_siblings_list"))
```
Ethyling, line 143: Same as above

source/cpu/power_amd64.go
```diff
@@ -38,13 +38,14 @@ func discoverSSTBF() (bool, error) {
 	nominalBaseFrequency := int(freqInfo.EAX)
 
 	// Loop over all CPUs in the system
-	files, err := ioutil.ReadDir(cpuDevicesBaseDir)
+	files, err := ioutil.ReadDir(source.SysfsDir.Path("cpu/devices"))
```
Ethyling, line 41: Same

source/cpu/power_amd64.go
```diff
@@ -38,13 +38,14 @@ func discoverSSTBF() (bool, error) {
...
+
 	if err != nil {
 		return false, err
 	}
 	for _, file := range files {
 		// Try to read effective base frequency of each cpu in the system
-		filePath := path.Join(cpuDevicesBaseDir, file.Name(), "cpufreq/base_frequency")
+		filePath := source.SysfsDir.Path("cpu/devices", file.Name(), "cpufreq/base_frequency")
```
Ethyling, line 48: Same

Dockerfile
marquiz: Yes, probably so 😄

source/cpu/cpu.go
marquiz: Argh 🤯 This bug obviously creeped in the latest version of my patch where I introduced the `HostDir` type, and, I must've tested the previous oudated version 😭 Thanks for catching this 🎏 Fixed now

source/cpu/cpu.go
marquiz: Fixed

source/cpu/power_amd64.go
marquiz: Fixed

source/cpu/power_amd64.go
marquiz: Fixed

source/config.go
Ethyling: Oh ok, let's keep it like this then

PR #327, opened 2020-06-12
Adds multi-architecture support to the Makefile using docker buildx (amd64, arm64, arm/v7)

MatthewShotton:
This adds an additional `push-multiarch` command to the `Makefile` which builds and pushes multi-architecture images (architectures are listed in the `$MULTIARCH_IMAGE_ARCHITECTURES` variable).

I tried initially to do this as two separate steps, one build step and one push step as for single architecture builds. Alas it seems like for multi-arch builds with buildx you need to push as part of the build if you want to generate a suitable multi-architecture manifest file for a registry.

(There is the "hard way" as outlined here: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/ which might make it possible to run as two steps, but buildx seems to be the future for multi-architecture builds)

**Note**: buildx is still an experimental feature, and needs to be enabled in docker as outlined here: [https://docs.docker.com/buildx/working-with-buildx](https://docs.docker.com/buildx/working-with-buildx/).

**Context** I'm been wanting to experiment with `node-feature-discovery` on a mixed-architecture [k3s](https://k3s.io/) home-lab cluster. It seems like a really cool solution for scheduling nodes based on the devices plugged into the USB ports of the cluster nodes. I noticed you didn't have any arm images built.

Thank you for this cool project!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@ HOSTMOUNT_PREFIX := /host-
 KUBECONFIG :=
 E2E_TEST_CONFIG :=
 
+MULTIARCH_IMAGE_ARCHITECTURES := linux/amd64,linux/arm64,linux/arm/v7
+MULTIARCH_IMAGE_BUILD_CMD := docker buildx build
+
 yaml_templates := $(wildcard *.yaml.template)
 yaml_instances := $(patsubst %.yaml.template,%.yaml,$(yaml_templates))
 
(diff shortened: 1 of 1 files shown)


MatthewShotton:
I've just signed the "Individual Contributor" CLA.

marquiz:
Thank you for the patch @MatthewShotton !

I'm currently on child care leave so I apologies for the unresponsiveness (past and the near future). I'll take a closer look as soon as I can.

Maybe @pmundt could take a look, too, as he has lately been doing multiarch builds

pmundt:
Thanks @MatthewShotton, it's nice to see that others are interested in this direction.
 
The method proposed here is basically the same method I have been using for multiarch builds in my local testing, so I don't have an issue with buildx per se, but it's experimental nature means that there is very poor integration with the rest of the workflow. For example, the image building and pushing is done via Travis, which has its [own method](https://docs.travis-ci.com/user/multi-cpu-architectures/) for doing multi-arch builds, but does not have a method for generating and pushing manifest files to an image registry. This means that if we want to support multiarch in the CI/CD pipeline, which I think we certainly do, we will still need to implement manual manifest wrangling in the `.travis-ci.yml` for the time being.

The addition of a buildx rule in the top-level Makefile, while providing a convenience wrapper for people wanting to do this themselves, unfortunately brings us no closer to multiarch image generation in the existing CI/CD pipeline. It may still be worthwhile having a Makefile target just for testing locally, but I would not expect most users to care about this once real multiarch images are regularly pushed out to the public image registry.

I would suggest that we try to solve this in the CI/CD pipeline first, and then see if there are specific Makefile helpers that could be added to tie things together more coherently. Having two different build/push mechanisms where only one is tied into the CI/CD pipeline is something I think we should try to avoid.

MatthewShotton:
Ok cool! Thanks for the detailed reply. That makes a lot of sense. I unfortunately don't have a whole lot of experience with Travis otherwise I'd try to help getting multi-arch working in there. I'm excited for when it lands though!

I'll close this PR for the time-being.

PR #334, opened 2020-08-11
custom: Add rules for kconfig and cpuid

mythi:


mythi:
@marquiz @Ethyling any comments?

source/internal/kernelutils/kernel_kconfig.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer r.Close()
+
+	return ioutil.ReadAll(r)
+}
+
+// Read kconfig into a map
+func ParseKconfig(searchPaths []string) (map[string]string, error) {
```
marquiz, line 52: nit: it'd be easier to review if moving the code and changing the functionality was done in separate commits

source/custom/rules/kconfig_rule.go
```diff
@@ -0,0 +1,66 @@
...
+	} else {
+		// from k8s.io/systemr-validator used by kubeadm preflight checks
+		searchPaths = []string{
+			"/usr/src/linux-" + kVer + "/.config",
+			"/usr/src/linux/.config",
+			"/usr/lib/modules/" + kVer + "/config",
+			"/usr/lib/ostree-boot/config-" + kVer,
+			"/usr/lib/kernel/config-" + kVer,
+			"/usr/src/linux-headers-" + kVer + "/.config",
+			"/lib/modules/" + kVer + "/build/.config",
```
marquiz, line 53: I think we should use the same logic in kernel feature source and here. Prolly move these to kernelutils(?)

marquiz (commented):
This is pretty straightforward and basically looks good. What I immediately started to think that should we start to use a hierarchy (e.g. something similar that the feature sources and labels have), i.e. pack all CPU rules under a `cpu` matcher rule, kernel rules under `kernel` etc.

More specifically,  instead of  (with some future additions envisioned):
```
- name: "my.label"
  matchOn:
    - loadedKMod: [...]
    - kConfig: [...]
    - kernelVersion: [...]
    - cpuId: [...]
    - cpuModel: [...]
```
we would have 
```
- name: "my.label"
  matchOn:
    - kernel:
      - loadedMod: [...]
      - config: [...]
      - version: [...]
    - cpu
      - cpuId: [...]
      - model: [...]
```
I don't know if this would make it any more understandable/manageable. We already have `loadedKMod`, but, we could duplicate that and deprecate the original and keep it around for few releases.

Another thing is that should we allow matching the value of a kernel config flag(?) E.g. support `<KEY>=<VALUE>` format inside the rule?

source/custom/custom.go
```diff
@@ -76,7 +75,8 @@ func (s Source) Discover() (source.Features, error) {
 	for _, customFeature := range allFeatureConfig {
 		featureExist, err := s.discoverFeature(customFeature)
 		if err != nil {
-			return features, fmt.Errorf("failed to discover feature: %s. %s", customFeature.Name, err.Error())
+			log.Printf("failed to discover feature: %s. %s", customFeature.Name, err.Error())
```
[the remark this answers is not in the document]
mythi: fixed

source/custom/rules/kconfig_rule.go
mythi: moved to kernelutils

source/internal/kernelutils/kernel_kconfig.go
mythi: fixed

mythi:
> Another thing is that should we allow matching the value of a kernel config flag(?)

Good question. However, since `=m` vs `=y` gets lost so in most cases the check would become `KEY=true`.

marquiz:
> However, since `=m` vs `=y` gets lost, in most cases the check would become `KEY=true`

We could support both
1. `[<kconfig_flag>]` would imply `=y` or `=m`
2. `[<kconfig_flag>=<value>]` would match against given value

mythi:
> We could support both

Looks it pretty easy to support:

```
for k, v := range kconfig {
                        if v != "true" {
+                               kConfigs[fmt.Sprintf("%s=%s", k, v)] = struct{}{}
                                continue
                        }
                        kConfigs[k] = struct{}{}
}

**Comment by mythi** [comment]
> We could support both

this is now implemented

**Comment by marquiz** [comment]
Also, what do you think about the hiearchical (per-domain) matcher structure I pondered above?
```
  matchOn:
    - kernel:
      - loadedMod: [...]
      - config: [...]
      - version: [...]
    - pci:
       - ...
    - cpu:
       - ...
```

Would it make the matchers (and the rule configs) more manageable, going in the future, WDYT @mythi @adrianchiris @Ethyling @pmundt ?

File: README.md [hunk 4]
```diff
@@ -378,6 +378,36 @@ loadedKMod : [<kernel module>, ...]
  Matching is done by performing logical _AND_ for each provided Element, i.e the Rule will match if all provided Elements (kernel modules) are loaded
  in the system.
 
+##### CpuId Rule
+###### Nomenclature
+```
+Element     :A CPUID flag
+```
+
+The Rule allows matching the in the system against a provided list of Elements.
```
adrianchiris, line 387: > the in the system against a provided list of Elements.
the CPUID flags ?

README.md
```diff
@@ -378,6 +378,36 @@ loadedKMod : [<kernel module>, ...]
...
+ Matching is done by performing logical _AND_ for each provided Element, i.e the Rule will match if all provided Elements (CPUID flag strings) are available
+ in the system.
+
+##### Kconfig Rule
+###### Nomenclature
+```
+Element     :A Kconfig option
+```
+
+The Rule allows matching the in the system against a provided list of Elements.
```
adrianchiris, line 402: > the in the system against a provided list of Elements.
the running kernel KConfig config options against a provided list of elements ?

README.md
```diff
@@ -421,6 +459,9 @@ with a PCI vendor ID of `15b3` _AND_ PCI device ID of `1014` _or_ `1017`.
 - A node would contain the label: `feature.node.kubernetes.io/custom-my.accumulated.feature=true`
 if `some_kmod1` _AND_ `some_kmod2` kernel modules are loaded __OR__ the node contains a PCI device
 with a PCI vendor ID of `15b3` _AND_ PCI device ID of `1014` _OR_ `1017`.
+- A node would contain the label: `feature.node.kubernetes.io/custom-my.kernel.featureneedscpu=true`
+if `KVM_INTEL` kernel config is enabled __AND__ the node CPU supports `VMX` virtual machine extensions
+- A node would contain the label: `feature.node.kubernetes.io/custom-my.kernel.modulecompiler=true` if `kmod1` kernel module is loaded __AND__ it's built with `GCC_VERSION=100101`.
```
adrianchiris, line 464: > `kmod1` kernel module is loaded __AND__ it's built with `GCC_VERSION=100101`.

This may not always be the case right ? e.g `kmod1` is an out of tree kernel module which may have been compiled with a different compiler than the kernel itself.

adrianchiris:
>Also, what do you think about the hiearchical (per-domain) matcher structure I pondered above?
>
>```
>  matchOn:
>    - kernel:
>      - loadedMod: [...]
>      - config: [...]
>      - version: [...]
>    - pci:
>       - ...
>    - cpu:
>       - ...
>```
>Would it make the matchers (and the rule configs) more manageable, going in the future, WDYT

Grouping the rules by domain sounds like a good idea if we start accumulating too many rules. (are we at this point?)
In general I would not do more than one level of hierarchy unless absolutely needed as it would make configuration harder.

Perhaps before going with API change, we can start with grouping the rules in README.

README.md
mythi: changed to "if the in-tree `kmod1` kernel module is loaded"

README.md
mythi: copy&paste issue. now fixed

source/kernel/kernel.go
```diff
@@ -158,37 +158,45 @@ func readKconfigGzip(filename string) ([]byte, error) {
...
-	// Then, try to read from /proc
-	if raw == nil {
-		raw, err = readKconfigGzip("/proc/config.gz")
-		if err != nil {
-			log.Printf("Failed to read /proc/config.gz: %s", err)
+	for _, path := range append([]string{configPath}, searchPaths...) {
+		if len(path) == 0 {
+			continue
+		}
+		if raw, err = ioutil.ReadFile(path); err == nil {
```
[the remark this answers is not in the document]
mythi: makes sense yes

source/kernel/kernel.go
```diff
@@ -158,37 +158,45 @@ func readKconfigGzip(filename string) ([]byte, error) {
...
-		// Get kernel version
-		unameRaw, err := ioutil.ReadFile("/proc/sys/kernel/osrelease")
-		uname := strings.TrimSpace(string(unameRaw))
-		if err != nil {
-			return nil, err
-		}
-		// Read kconfig
-		raw, err = ioutil.ReadFile(source.BootDir.Path("config-" + uname))
+		log.Printf("ERROR: Failed to read kernel config from %+v:", append([]string{configPath}, searchPaths...))
+		raw, err = readKconfigGzip("/proc/config.gz")
```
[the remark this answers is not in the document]
mythi: right. it gets checked first by `k8s.io/system-validator` too.

source/custom/rules/kconfig_rule.go
```diff
@@ -0,0 +1,51 @@
...
+	return true, nil
+}
+
+func init() {
+	kConfigs = make(map[string]struct{})
+
+	kconfig, err := kernelutils.ParseKconfig("")
+	if err == nil {
+		for k, v := range kconfig {
+			if v != "true" {
```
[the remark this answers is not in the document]
mythi: +1 for future exercise.

source/kernel/kernel.go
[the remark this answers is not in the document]
mythi: It's now part of `seachPaths`

marquiz:
> Grouping the rules by domain sounds like a good idea if we start accumulating too many rules. (are we at this point?)
> In general I would not do more than one level of hierarchy unless absolutely needed as it would make configuration harder.

Agree. Maybe we're not in that point, yet. After thinking about this a bit more I feel that we can go with the flat structure for now. We can introduce the nested structure if/when the number of matchers starts to grow too big. And, that shouldn't break anything as we should easily be able to support both in parallel (the flat structure can be deprecated).

source/custom/rules/kconfig_rule.go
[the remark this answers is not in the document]
marquiz: 👍 Could you open an issue about this?

marquiz (approved):
Thanks @mythi for the patchset, and your patience 😊 
/approve

source/custom/rules/kconfig_rule.go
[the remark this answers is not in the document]
mythi: opened #361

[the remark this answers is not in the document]
marquiz: Thx!

PR #340, opened 2020-08-26
test/e2e: interpret node names in config as regexps

marquiz:
Interpret node names in the e2e test config as regexps instead of plain
strings. This makes it a lot easier to maintain a config file where
multiple nodes share configuration.

### DIFF
--- a/test/e2e/e2e-test-config.exapmle.yaml
+++ b/test/e2e/e2e-test-config.exapmle.yaml
@@ -60,7 +60,8 @@ defaultFeatures:
     - "nfd.node.kubernetes.io/worker.version"
     - "nfd.node.kubernetes.io/feature-labels"
   nodes:
-    my-node-1:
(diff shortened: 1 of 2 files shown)


test/e2e/e2e-test-config.exapmle.yaml
```diff
@@ -60,7 +60,8 @@ defaultFeatures:
     - "nfd.node.kubernetes.io/worker.version"
     - "nfd.node.kubernetes.io/feature-labels"
   nodes:
-    my-node-1:
+    # NOTE: keys here are interpreted as regexps
```
ArangoGutierrez, line 63: not part of the Patch but just noted that this file has a typo in the file name
`test/e2e/e2e-test-config.exapmle.yaml ` 
example...
can you fix this in this patch :smiley:

ArangoGutierrez (approved):
/lgtm

PR #341, opened 2020-08-26
scripts/test-infra: add script for end-to-end testing

marquiz:
Add a new script for running end-to-end tests. Implement a Makefile rule
for polling the container image. We need this for prow as there is no
other way for syncing the jobs (e2e postsubmit job must wait until the
image-pushing job has built and published the image). This is
implemented in the Makefile as all the image-related variables are also
defined/available there.

Also, adapt the TravisCI config to use the new script.


ArangoGutierrez:
I am going to do the same for the operator @marquiz

PR #342, opened 2020-08-28
Update go to 1.14.2

ArangoGutierrez:
Go 1.14.2 is stable enough for us to move from 1.13 at this point
after this PR we could start thinking updating deps for K1.19

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-  - "1.13.12"
+  - "1.14.7"
 
 env:
   global:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build node feature discovery
-FROM golang:1.13.12 as builder
+FROM golang:1.14.7 as builder
 
 # Get (cache) deps in a separate layer
 COPY go.mod go.sum /go/node-feature-discovery/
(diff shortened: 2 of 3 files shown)


marquiz (changes requested):
Thanks for the patch @ArangoGutierrez !

Shouldn't we use 1.14.7 instead, though? Also, could you capitalize the first work of the commit message subject.

ArangoGutierrez:
1.14.2 is the stable version on RPM as of today, that's why I went for it. we can do a newer if you think is stable enough

marquiz:
> 1.14.2 is the stable version on RPM as of today, that's why I went for it. we can do a newer if you think is stable enough

I think it makes sense to take the latest released version when updating. That way we get e.g. all the security fixes that added between the versions. RPMs can cherry-pick/backport important fixes to earlier versions - we can't.

ArangoGutierrez:
> > 1.14.2 is the stable version on RPM as of today, that's why I went for it. we can do a newer if you think is stable enough
> 
> I think it makes sense to take the latest released version when updating. That way we get e.g. all the security fixes that added between the versions. RPMs can cherry-pick/backport important fixes to earlier versions - we can't.

Makes sense, will bump to .7

ArangoGutierrez:
@marquiz -> b172610

marquiz (approved):
Thanks!

/approve
/lgtm

PR #343, opened 2020-09-03
cloudbuild.yaml: build using make

marquiz:
Use make for building our container image (through a new push-image.sh
script) instead of building directly with docker. This way we avoid the
container image tag naming hassle where we build and push the image on
"image-pushing" prow job and need to be able to reliably re-create the
same tag on the "testing" side for end-to-end tests. This also has the
advantage that the same mechanism (make) is used everywhere, i.e. in
image-pushing, testing and locally.


ArangoGutierrez (approved):
/lgtm

PR #345, opened 2020-09-03
scripts: hand-construct a reproducible version

marquiz:
We don't have a full git repo in the gcb builds so our version/tag magic
inside the makefile does not work. Thus, get version information from
the build information - fortunately we can do that by simply stripping
the date prefix from the _GIT_TAG variable.


ArangoGutierrez (approved):
/lgtm

ArangoGutierrez:
jejeje nice bash'ism

PR #347, opened 2020-09-03
scripts: configure docker authentication in push-image.sh

marquiz:
As we're trying to push the images directly from the phase job and not
use 'images:' targets in cloudbuild.yaml.


ArangoGutierrez (approved):
/lgtm

PR #348, opened 2020-09-04
cloudbuild.yaml: set correct container image registry

marquiz:
Use the staging image repository for tagging/pushing images. We aren't
able to (and wouldn't want to) push directly to the production
repository that is the default in the Makefile.


kad (approved):
/lgtm

PR #355, opened 2020-09-08
Fix up Makefile variable initialization and documentation

marquiz:
An attempt to organize the Makefile variables a bit further. The idea is
to only document variables that are really intended for build
customization and initialize all these variables in the Makefile with
'=?' so that they can be specified in the environment (and not just with
variable overrides). Thus, this patch drops some variables from the
README and change the initialization of some variables in the Makefile
so that the two files would be in sync.


ArangoGutierrez (approved):
/lgtm

PR #356, opened 2020-09-09
nfd-worker: add special handling for --sources=all

marquiz:
A new special value 'all' is a shortcut for enabling all feature
sources. It should be the only name specified -- if any other names are
specified 'all' does not take effect, but, we only enable the listed
feature sources. E.g.
  --sources=all enables all sources, but
  --sources=all,cpu only enables the cpu source

### DIFF
--- a/docs/advanced/developer-guide.md
+++ b/docs/advanced/developer-guide.md
@@ -222,8 +222,6 @@ Command line flags of nfd-worker:
 ```bash
 $ docker run --rm ${NFD_CONTAINER_IMAGE} nfd-worker --help
 ...
-nfd-worker.
-
   Usage:
   nfd-worker [--no-publish] [--sources=<sources>] [--label-whitelist=<pattern>]
      [--oneshot | --sleep-interval=<seconds>] [--config=<path>]
@@ -253,8 +251,9 @@ nfd-worker.
   --server-name-override=<name> Name (CN) expect from server certificate, useful
                               in testing
                               [Default: ]
-  --sources=<sources>         Comma separated list of feature sources.
-                              [Default: cpu,custom,iommu,kernel,local,memory,network,pci,storage,system,usb]
+  --sources=<sources>         Comma separated list of feature sources. Special
+                              value 'all' enables all feature sources.
+                              [Default: all]
   --no-publish                Do not publish discovered features to the
                               cluster-local Kubernetes API server.
   --label-whitelist=<pattern> Regular expression to filter label names to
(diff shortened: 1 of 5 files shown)


pkg/nfd-worker/nfd-worker.go
```diff
@@ -140,15 +139,24 @@ func NewNfdWorker(args Args) (NfdWorker, error) {
 		&local.Source{},
 	}
 
-	sourceWhiteList := map[string]struct{}{}
-	for _, s := range args.Sources {
-		sourceWhiteList[strings.TrimSpace(s)] = struct{}{}
-	}
+	// Determine enabled feature
+	if len(args.Sources) == 1 && args.Sources[0] == "all" {
+		nfd.sources = allSources
```
Ethyling, line 144: We should also remove the `panicfake` source too from `allSources`

pkg/nfd-worker/nfd-worker.go
```diff
@@ -140,15 +139,24 @@ func NewNfdWorker(args Args) (NfdWorker, error) {
...
+		sourceWhiteList := map[string]struct{}{}
+		for _, s := range args.Sources {
+			sourceWhiteList[strings.TrimSpace(s)] = struct{}{}
+		}
+
+		nfd.sources = []source.FeatureSource{}
+		for _, s := range allSources {
+			if _, enabled := sourceWhiteList[s.Name()]; enabled {
+				nfd.sources = append(nfd.sources, s)
+			}
```
Ethyling, line 159: Should we print a warning if the user specify a source name that does not match to any source?

pkg/nfd-worker/nfd-worker.go
marquiz: VERY good note, thanks! Fixed

pkg/nfd-worker/nfd-worker.go
marquiz: Mm, reasonable improvement. Fixed

marquiz:
Rebased and addressed comments from @Ethyling

docs/advanced/developer-guide.md
```diff
@@ -166,41 +166,49 @@ Command line flags of nfd-master:
 ```bash
 $ docker run --rm ${NFD_CONTAINER_IMAGE} nfd-master --help
 ...
-Usage:
-  nfd-master [--prune] [--no-publish] [--label-whitelist=<pattern>] [--port=<port>]
+  Usage:
+  nfd-worker [--no-publish] [--sources=<sources>] [--label-whitelist=<pattern>]
```
Ethyling, line 170: I think there is an issue here as it is the nfd-worker help message instead of the nfd-master help message

marquiz: Argh 🤦  Thanks for noticing this! Fixed

PR #360, opened 2020-09-12
Drop gcloud auth for image pushing

ArangoGutierrez:
gcloud auth call in push_image.sh test script is no longer needed


marquiz (approved):
I believe this is not needed 🤞 
/approve
/lgtm

PR #371, opened 2020-10-21
docs: make the list of available versions dynamic

marquiz:
Make the list of available versions dynamic. The items displayed is now
generated with javascript. The parent directory of the site is supposed
to contain versions.js providing getVersionListItems() that returns the
available versions.

The update-gh-pages.sh script is modified to update/create versions.js
on every invocation. It simply lists all directories in the root
directory and adds them to the version list.


ArangoGutierrez (approved):
/lgtm

PR #376, opened 2020-10-30
docs: polish the version menu

marquiz:
Only show the dynamic version menu in production environment where it's
available. Local (development) builds only show a stub version menu with
one item.

Also, fix the theme customization. Append version list items to the
correct element in html so that they are displayed correctly. Drop
unnecessary <div/>.


marquiz:
The end-result is demonstrated in my personal repo:
https://marquiz.github.io/node-feature-discovery/documentation-debug

PR #386, opened 2020-11-18
Add nfd-worker-conf ConfigMap to deployment templates

marquiz:
Add a virtually empty ConfigMap that is mounted inside the workers.
Makes it easier to start customizing the worker deployment e.g. with just:

  `$ kubectl -n ${NFD_NS} edit configmap nfd-worker-conf`

Create a new 'templates' make target for inserting the content of
nfd-worker.conf.example into the configmap spec of the templates. Thus,
'make templates' should be run whenever the example config is update.
Update the verify.sh prow script to check that the templates are up to
date.

This patch also streamlines the documentation about configuration
management, reflecting the changes.

### DIFF
--- a/nfd-daemonset-combined.yaml.template
+++ b/nfd-daemonset-combined.yaml.template
@@ -110,6 +110,9 @@ spec:
             - name: features-d
               mountPath: "/etc/kubernetes/node-feature-discovery/features.d/"
               readOnly: true
+            - name: nfd-worker-conf
+              mountPath: "/etc/kubernetes/node-feature-discovery"
+              readOnly: true
       volumes:
         - name: host-boot
           hostPath:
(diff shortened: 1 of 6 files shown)


nfd-daemonset-combined.yaml.template
```diff
@@ -118,3 +120,14 @@ spec:
...
+          configMap:
+            name: nfd-worker-conf
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nfd-worker-conf
+  namespace: node-feature-discovery
+data:
+  nfd-worker.conf: |
```
ArangoGutierrez, line 133:
```suggestion
  nfd-worker.conf: |
      #uncomment to configure
    #sources:
    #  pci:
    #    deviceClassWhitelist:
    #      - "0200"
    #      - "03"
    #      - "12"
    #    deviceLabelFields:
    #      - "class"
    #      - "vendor"
    #      - "device"
    #      - "subsystem_vendor"
    #      - "subsystem_device"
```

nfd-worker-daemonset.yaml.template
```diff
@@ -77,10 +79,21 @@ spec:
...
 #          secret:
 #            secretName: nfd-worker-cert
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nfd-worker-conf
+  namespace: node-feature-discovery
+data:
+  nfd-worker.conf: |
```
ArangoGutierrez, line 99:
```suggestion
  nfd-worker.conf: |
      #uncomment to configure
    #sources:
    #  pci:
    #    deviceClassWhitelist:
    #      - "0200"
    #      - "03"
    #      - "12"
    #    deviceLabelFields:
    #      - "class"
    #      - "vendor"
    #      - "device"
    #      - "subsystem_vendor"
    #      - "subsystem_device"
```

nfd-worker-job.yaml.template
```diff
@@ -73,3 +89,21 @@ spec:
...
+#          secret:
+#            secretName: nfd-worker-cert
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nfd-worker-conf
+  namespace: node-feature-discovery
+data:
+  nfd-worker.conf: |
```
ArangoGutierrez, line 109:
```suggestion
  nfd-worker.conf: |
      #uncomment to configure
    #sources:
    #  pci:
    #    deviceClassWhitelist:
    #      - "0200"
    #      - "03"
    #      - "12"
    #    deviceLabelFields:
    #      - "class"
    #      - "vendor"
    #      - "device"
    #      - "subsystem_vendor"
    #      - "subsystem_device"
```

ArangoGutierrez (approved):
The operator does the same, an empty file, maybe we could, in the name of documentation, not do it via a virtual empty but a commented file (wish has the same effect) , and will help people

marquiz:
I was thinking about something like the changes you suggested, but, decided otherwise. I wanted to have all the available options visible in the configmap, this now happens when `make yamls`. And, I wanted to maintain the example configmap only in one place i.e. `nfd-worker.example.conf`. Lastly, with your additions the result of `make yamls` becomes quite ugly as both that example and the content of `nfd-worker.conf.example` gets into the configmap.

But, on a second thought it would be really nice to have the example also in the templates. I'll sketch up something. BBL
/hold

marquiz:
Changed the patch so that the configmaps in templates have same data than `nfd-worker.conf.example`. There's a new `templates` target for updating them whenever the example config is updated

ArangoGutierrez (approved):
Your example is waaaay better than mine
/lgtm

PR #389, opened 2020-11-18
Go 1.15

ArangoGutierrez:
Now that we are using deps from K8s 1.19 and the cert issue has been solved we are good to bump to go 1.15

this patch bumps both the Dockerfile and go.mod go version to 1.15


marquiz:
Thanks @ArangoGutierrez ! You were faster than me 😸 
/lgtm
/approve

PR #393, opened 2020-11-19
Fix a thing or two under test

ArangoGutierrez:
This patch fix typo in a file name
Updates default image on e2e tests
Adds an extra file that should be ignored to gitignore


test/e2e/node_feature_discovery.go
```diff
@@ -44,8 +44,8 @@ import (
 )
 
 var (
-	dockerRepo    = flag.String("nfd.repo", "quay.io/kubernetes_incubator/node-feature-discovery", "Docker repository to fetch image from")
-	dockerTag     = flag.String("nfd.tag", "e2e-test", "Docker tag to use")
+	dockerRepo    = flag.String("nfd.repo", "k8s.gcr.io/nfd/node-feature-discovery", "Docker repository to fetch image from")
+	dockerTag     = flag.String("nfd.tag", "v0.6.0", "Docker tag to use")
```
marquiz, line 48: Hmm, we should probably use `gcr.io/k8s-staging-nfd/node-feature-discovery:master` as the default here as a past release is not guaranteed to be compatible with tests from `master`. Another option would be to leave the both arg defaults empty

ArangoGutierrez: I like master as default, the `Makefile` uses a proper flag to point it to the dirty git stage (when developing something) 
I was here really thinking in the casual users trying to understand the project.
will set it to `master`

test/e2e/node_feature_discovery.go
```diff
@@ -45,7 +45,7 @@ import (
 
 var (
 	dockerRepo    = flag.String("nfd.repo", "k8s.gcr.io/nfd/node-feature-discovery", "Docker repository to fetch image from")
-	dockerTag     = flag.String("nfd.tag", "v0.6.0", "Docker tag to use")
+	dockerTag     = flag.String("nfd.tag", "master", "Docker tag to use")
```
marquiz, line 48: You need to change `nfd.repo`, too as the production repo `k8s.gcr.io/nfd/node-feature-discovery`  does not have `master` tag

test/e2e/node_feature_discovery.go
marquiz: Mm, that's a fair point. This is just yet another place which needs to be modified when creating a release branch. I prolly need to create a helper script for this 🤔

test/e2e/node_feature_discovery.go
ArangoGutierrez: oh staging .... sorry missed that part

marquiz:
BTW, created script for helping with the release preparations: #397

marquiz:
Looks good, could you squash these into one? I'd like to preserve the commit msg of the first commit 😄

ArangoGutierrez:
> Looks good, could you squash these into one? I'd like to preserve the commit msg of the first commit smile

Done!

PR #395, opened 2020-11-20
Make all mounts in deployment templates read-only

marquiz:
NFD-Worker does not need write access.


Ethyling:
Seems to be working fine
/lgtm

PR #396, opened 2020-11-20
docs: improve documentation of the custom source

marquiz:
Better explain where the rules configuration is supposed to be
specified.


eero-t:
I hadn't done anything with NFD before, but needed custom rules for a new label.  Current documentation gave no hints where those rules should be put & with what spec path => I think this improvement is very much needed.

PR #403, opened 2020-11-24
test/e2e: fix e2e-config node matching

marquiz:
Pick the correct rule when multiple node rules are present.


ArangoGutierrez (approved):
/lgtm

ArangoGutierrez:
now `make e2e-test` works on multi node clusters, thanks

PR #407, opened 2020-12-01
not need close

goushicui:


ArangoGutierrez:
Hey @goushicui thanks for the patch, could you check the CLA info first, and second, please add more info to both the PR header and the commit. a 2 or 3 liner explanation on why this is needed, thanks again!

ArangoGutierrez (changes requested):
Please amend the commit with some verbose explanation on why this is needed

/hold

ArangoGutierrez:
the assignee is meant to carry the patch process and reviews, is not the same PR Author
/unassign goushicui

goushicui:
close or not close the chan may not affect WaitForReady func, i think it is not necessary

marquiz:
> close or not close the chan may not affect WaitForReady func, i think it is not necessary

It is necessary and deliberate. The ready signal is emitted only once. Without closing the channel further calls `WaitForReady()` would fail with a timeout (as it would wait for the signal that won't be emitted). Suggesting to close this PR

PR #408, opened 2020-12-03
scripts/update-gh-pages: only update docs if version has been released

marquiz:
Only update documentation if the version has been released (i.e. a git
tag corresponding the version we're updating exists) or it is for the
master branch. Prevents us from pointing users to documentation of an
unreleased version.

Also updates github workflow to fetch full commit history (including
tags) so that git describe works correctly.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -14,6 +14,8 @@ jobs:
 
     - name: Check out repo
       uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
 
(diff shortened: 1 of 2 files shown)


ArangoGutierrez (commented):
Looks sane, can we squash first?

ArangoGutierrez (approved):
Good point, we don't want brand new, non released features to be in the documentation, that can be confusing.
/lgtm

marquiz:
> Looks sane, can we squash first?

Squashed. It was two unrelated changes so I opened #409

PR #410, opened 2020-12-03
Fix auto-generation of nfd-prune.yaml

marquiz:
Change the template so that our sed magic updates the container image
name correctly.


ArangoGutierrez (approved):
works as expected now
/lgtm

PR #412, opened 2020-12-03
Add namespace and RBAC rules to nfd-prune.yaml.template

marquiz:
Makes it possible to run prune on its own. Reflect this change in the
documentation.


ArangoGutierrez (approved):
I can smell that fresh new release been cook in the oven 
/lgtm

PR #417, opened 2020-12-04
scripts/update-gh-pages: add more details when stopping

marquiz:


ArangoGutierrez (approved):
verbosity never hurt anyone 
/lgtm

Issue #413, opened 2020-12-04
Dependency: update klauspost/cpuid to v2


ArangoGutierrez:
**What would you like to be added**:
- https://github.com/klauspost/cpuid/issues/28
We need to watch for `klauspost/cpuid` dependency, once it cut a next release we should update go.mod 

**Why is this needed**:
Is always a good practice to keep deps up to date and pointing to a release not a commit hash

marquiz:
[v2.0.1](https://github.com/klauspost/cpuid/releases/tag/v2.0.1) is out

PR #421, opened 2020-12-08
go.mod: update to klauspost/cpuid/v2@v2.0.2

mythi:
Closes: #413


marquiz:
I think you need to run `go mod tidy`

mythi:
> I think you need to run `go mod tidy`

Hmm...I thought I did that. Let me check

Issue #422, opened 2020-12-13
Add Helm Chart to NFD


adrianchiris:
**What would you like to be added**:
A helm chart to be used by Helm for deploying NFD in a K8s cluster

**Why is this needed**:
To ease NFD deployment and upgrade.

Currently there are several self brewed charts to deploy NFD. we should have one source.

https://github.com/Mellanox/network-operator/tree/master/deployment/network-operator/charts/node-feature-discovery
https://github.com/k8s-at-home/charts/tree/master/charts/node-feature-discovery
https://github.com/NVIDIA/gpu-operator/tree/master/deployments/gpu-operator/charts/node-feature-discovery

cpanato:
@adrianchiris what is the idea for this, we create the chart in this repository?

adrianchiris:
yes, maintain a chart a part of this repository, github can also be used as a helm repo to serve the chart.

cpanato:
yes, we can release the chart to another branch.

would you mind if I work on this issue?

adrianchiris:
was planning to take it from [network-operator](https://github.com/Mellanox/network-operator/tree/master/deployment/network-operator/charts/node-feature-discovery) and propose a PR, if you want to propose a PR for this im fine with it :)

cpanato:
@adrianchiris if you already have things you can move forward, i will help to review! thanks!

e0ne:
Since #423 is merged, let's close this issue

PR #423, opened 2020-12-16
Add helm chart for NFD

adrianchiris:
This commit adds Helm chart for node-feature-discovery

Signed-off-by: Adrian Chiris <adrianc@nvidia.com>

### DIFF
--- a/deployment/node-feature-discovery/Chart.yaml
+++ b/deployment/node-feature-discovery/Chart.yaml
@@ -1,0 +1,15 @@
+apiVersion: v2
+appVersion: master
+description: |
+  Detects hardware features available on each node in a Kubernetes cluster, and advertises
+  those features using node labels.
+name: node-feature-discovery
+sources:
+ - https://github.com/kubernetes-sigs/node-feature-discovery
+home: https://github.com/kubernetes-sigs/node-feature-discovery
+keywords:
+  - feature-discovery
+  - feature-detection
+  - node-labels
+type: application
+version: 0.1.0
--- a/deployment/node-feature-discovery/templates/_helpers.tpl
+++ b/deployment/node-feature-discovery/templates/_helpers.tpl
@@ -1,0 +1,63 @@
+{{/* vim: set filetype=mustache: */}}
+{{/*
+Expand the name of the chart.
+*/}}
+{{- define "node-feature-discovery.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "node-feature-discovery.fullname" -}}
+{{- if .Values.fullnameOverride -}}
+{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- $name := default .Chart.Name .Values.nameOverride -}}
+{{- if contains $name .Release.Name -}}
+{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
+{{- else -}}
+{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create chart name and version as used by the chart label.
+*/}}
+{{- define "node-feature-discovery.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Common labels
+*/}}
+{{- define "node-feature-discovery.labels" -}}
+helm.sh/chart: {{ include "node-feature-discovery.chart" . }}
+{{ include "node-feature-discovery.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end -}}
+
+{{/*
+Selector labels
+*/}}
+{{- define "node-feature-discovery.selectorLabels" -}}
+app.kubernetes.io/name: {{ include "node-feature-discovery.name" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- end -}}
+
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "node-feature-discovery.serviceAccountName" -}}
+{{- if .Values.serviceAccount.create -}}
+    {{ default (include "node-feature-discovery.fullname" .) .Values.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.serviceAccount.name }}
+{{- end -}}
+{{- end -}}
--- a/deployment/node-feature-discovery/templates/master.yaml
+++ b/deployment/node-feature-discovery/templates/master.yaml
@@ -1,0 +1,83 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name:  {{ include "node-feature-discovery.fullname" . }}-master
+  labels:
+    {{- include "node-feature-discovery.labels" . | nindent 4 }}
+    role: master
+spec:
+  replicas: {{ .Values.master.replicaCount }}
+  selector:
+    matchLabels:
+      {{- include "node-feature-discovery.selectorLabels" . | nindent 6 }}
+      role: master
+  template:
+    metadata:
+      labels:
+        {{- include "node-feature-discovery.selectorLabels" . | nindent 8 }}
+        role: master
+      annotations:
+        {{- toYaml .Values.master.annotations | nindent 8 }}
+    spec:
+    {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+      serviceAccountName: {{ include "node-feature-discovery.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.master.podSecurityContext | nindent 8 }}
+      containers:
+        - name: master
+          securityContext:
+            {{- toYaml .Values.master.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          ports:
+          - containerPort: 8080
+            name: grpc
+          env:
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+          command:
+            - "nfd-master"
+          resources:
+            {{- toYaml .Values.master.resources | nindent 12 }}
+## Enable TLS authentication
+## The example below assumes having the root certificate named ca.crt stored in
+## a ConfigMap named nfd-ca-cert, and, the TLS authentication credentials stored
+## in a TLS Secret named nfd-master-cert.
+## Additional hardening can be enabled by specifying --verify-node-name in
+## args, in which case every nfd-worker requires a individual node-specific
+## TLS certificate.
+#          args:
+#            - "--ca-file=/etc/kubernetes/node-feature-discovery/trust/ca.crt"
+#            - "--key-file=/etc/kubernetes/node-feature-discovery/certs/tls.key"
+#            - "--cert-file=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
+#          volumeMounts:
+#            - name: nfd-ca-cert
+#              mountPath: "/etc/kubernetes/node-feature-discovery/trust"
+#              readOnly: true
+#            - name: nfd-master-cert
+#              mountPath: "/etc/kubernetes/node-feature-discovery/certs"
+#              readOnly: true
+#      volumes:
+#        - name: nfd-ca-cert
+#          configMap:
+#            name: nfd-ca-cert
+#        - name: nfd-master-cert
+#          secret:
+#            secretName: nfd-master-cert
+    {{- with .Values.master.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    {{- with .Values.master.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.master.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
--- a/deployment/node-feature-discovery/templates/nfd-worker-conf.yaml
+++ b/deployment/node-feature-discovery/templates/nfd-worker-conf.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.worker.configmapName }}
+  labels:
+  {{- include "node-feature-discovery.labels" . | nindent 4 }}
+data:
+  nfd-worker.conf: |
+  {{ .Values.worker.config | indent 4 }}
--- a/deployment/node-feature-discovery/templates/worker.yaml
+++ b/deployment/node-feature-discovery/templates/worker.yaml
@@ -1,0 +1,119 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name:  {{ include "node-feature-discovery.fullname" . }}-worker
+  labels:
+    {{- include "node-feature-discovery.labels" . | nindent 4 }}
+    role: worker
+spec:
+  selector:
+    matchLabels:
+      {{- include "node-feature-discovery.selectorLabels" . | nindent 6 }}
+      role: worker
+  template:
+    metadata:
+      labels:
+        {{- include "node-feature-discovery.selectorLabels" . | nindent 8 }}
+        role: worker
+      annotations:
+        {{- toYaml .Values.worker.annotations | nindent 8 }}
+    spec:
+      dnsPolicy: ClusterFirstWithHostNet
+    {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+      securityContext:
+        {{- toYaml .Values.worker.podSecurityContext | nindent 8 }}
+      containers:
+      - name: worker
+        securityContext:
+          {{- toYaml .Values.worker.securityContext | nindent 12 }}
+        image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
+        imagePullPolicy: {{ .Values.image.pullPolicy }}
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        resources:
+        {{- toYaml .Values.worker.resources | nindent 12 }}
+        command:
+        - "nfd-worker"
+        args:
+        - "--sleep-interval=60s"
+        - "--server={{ include "node-feature-discovery.fullname" . }}-master:{{ .Values.master.service.port }}"
+## Enable TLS authentication (1/3)
+## The example below assumes having the root certificate named ca.crt stored in
+## a ConfigMap named nfd-ca-cert, and, the TLS authentication credentials stored
+## in a TLS Secret named nfd-worker-cert
+#          - "--ca-file=/etc/kubernetes/node-feature-discovery/trust/ca.crt"
+#          - "--key-file=/etc/kubernetes/node-feature-discovery/certs/tls.key"
+#          - "--cert-file=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
+        volumeMounts:
+        - name: host-boot
+          mountPath: "/host-boot"
+          readOnly: true
+        - name: host-os-release
+          mountPath: "/host-etc/os-release"
+          readOnly: true
+        - name: host-sys
+          mountPath: "/host-sys"
+          readOnly: true
+        - name: source-d
+          mountPath: "/etc/kubernetes/node-feature-discovery/source.d/"
+          readOnly: true
+        - name: features-d
+          mountPath: "/etc/kubernetes/node-feature-discovery/features.d/"
+          readOnly: true
+        - name: nfd-worker-conf
+          mountPath: "/etc/kubernetes/node-feature-discovery"
+          readOnly: true
+## Enable TLS authentication (2/3)
+#        - name: nfd-ca-cert
+#          mountPath: "/etc/kubernetes/node-feature-discovery/trust"
+#          readOnly: true
+#        - name: nfd-worker-cert
+#          mountPath: "/etc/kubernetes/node-feature-discovery/certs"
+#          readOnly: true
+      volumes:
+        - name: host-boot
+          hostPath:
+            path: "/boot"
+        - name: host-os-release
+          hostPath:
+            path: "/etc/os-release"
+        - name: host-sys
+          hostPath:
+            path: "/sys"
+        - name: source-d
+          hostPath:
+            path: "/etc/kubernetes/node-feature-discovery/source.d/"
+        - name: features-d
+          hostPath:
+            path: "/etc/kubernetes/node-feature-discovery/features.d/"
+        - name: nfd-worker-conf
+          configMap:
+            name: {{ .Values.worker.configmapName }}
+            items:
+              - key: nfd-worker.conf
+                path: nfd-worker.conf
+## Enable TLS authentication (3/3)
+#        - name: nfd-ca-cert
+#          configMap:
+#            name: nfd-ca-cert
+#        - name: nfd-worker-cert
+#          secret:
+#            secretName: nfd-worker-cert
+    {{- with .Values.worker.nodeSelector }}
+      nodeSelector:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+    {{- with .Values.worker.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+    {{- with .Values.worker.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
--- a/deployment/node-feature-discovery/values.yaml
+++ b/deployment/node-feature-discovery/values.yaml
@@ -1,0 +1,203 @@
+image:
+  repository: gcr.io/k8s-staging-nfd/node-feature-discovery
+  # This should be set to 'IfNotPresent' for released version
+  pullPolicy: Always
+  # tag, if defined will use the given image tag, else Chart.AppVersion will be used
+  # tag
+imagePullSecrets: []
+  
+serviceAccount:
+  # Specifies whether a service account should be created
+  create: true
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name:
+
+nameOverride: ""
+fullnameOverride: ""
+
+master:
+  replicaCount: 1
+
+  podSecurityContext: {}
+    # fsGroup: 2000
+
+  securityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop: [ "ALL" ]
+    readOnlyRootFilesystem: true
+    runAsNonRoot: true
+    # runAsUser: 1000
+
+  service:
+    type: ClusterIP
+    port: 8080
+
+  resources: {}
+    # We usually recommend not to specify default resources and to leave this as a conscious
+    # choice for the user. This also increases chances charts run on environments with little
+    # resources, such as Minikube. If you do want to specify resources, uncomment the following
+    # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+    # limits:
+    #   cpu: 100m
+    #   memory: 128Mi
+    # requests:
+    #   cpu: 100m
+    #   memory: 128Mi
+
+  nodeSelector: {}
+
+  tolerations:
+  - key: "node-role.kubernetes.io/master"
+    operator: "Equal"
+    value: ""
+    effect: "NoSchedule"
+
+  annotations: {}
+
+  affinity:
+    nodeAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 1
+          preference:
+            matchExpressions:
+              - key: "node-role.kubernetes.io/master"
+                operator: In
+                values: [""]
+
+worker:
+  configmapName: nfd-worker-conf
+  config: |### <NFD-WORKER-CONF-START-DO-NOT-REMOVE>
+    #core:
+    #  labelWhiteList:
+    #  noPublish: false
+    #  sleepInterval: 60s
+    #  sources: [all]
+    #sources:
+    #  cpu:
+    #    cpuid:
+    ##     NOTE: whitelist has priority over blacklist
+    #      attributeBlacklist:
+    #        - "BMI1"
+    #        - "BMI2"
+    #        - "CLMUL"
+    #        - "CMOV"
+    #        - "CX16"
+    #        - "ERMS"
+    #        - "F16C"
+    #        - "HTT"
+    #        - "LZCNT"
+    #        - "MMX"
+    #        - "MMXEXT"
+    #        - "NX"
+    #        - "POPCNT"
+    #        - "RDRAND"
+    #        - "RDSEED"
+    #        - "RDTSCP"
+    #        - "SGX"
+    #        - "SSE"
+    #        - "SSE2"
+    #        - "SSE3"
+    #        - "SSE4.1"
+    #        - "SSE4.2"
+    #        - "SSSE3"
+    #      attributeWhitelist:
+    #  kernel:
+    #    kconfigFile: "/path/to/kconfig"
+    #    configOpts:
+    #      - "NO_HZ"
+    #      - "X86"
+    #      - "DMI"
+    #  pci:
+    #    deviceClassWhitelist:
+    #      - "0200"
+    #      - "03"
+    #      - "12"
+    #    deviceLabelFields:
+    #      - "class"
+    #      - "vendor"
+    #      - "device"
+    #      - "subsystem_vendor"
+    #      - "subsystem_device"
+    #  usb:
+    #    deviceClassWhitelist:
+    #      - "0e"
+    #      - "ef"
+    #      - "fe"
+    #      - "ff"
+    #    deviceLabelFields:
+    #      - "class"
+    #      - "vendor"
+    #      - "device"
+    #  custom:
+    #    - name: "my.kernel.feature"
+    #      matchOn:
+    #        - loadedKMod: ["example_kmod1", "example_kmod2"]
+    #    - name: "my.pci.feature"
+    #      matchOn:
+    #        - pciId:
+    #            class: ["0200"]
+    #            vendor: ["15b3"]
+    #            device: ["1014", "1017"]
+    #        - pciId :
+    #            vendor: ["8086"]
+    #            device: ["1000", "1100"]
+    #    - name: "my.usb.feature"
+    #      matchOn:
+    #        - usbId:
+    #          class: ["ff"]
+    #          vendor: ["03e7"]
+    #          device: ["2485"]
+    #        - usbId:
+    #          class: ["fe"]
+    #          vendor: ["1a6e"]
+    #          device: ["089a"]
+    #    - name: "my.combined.feature"
+    #      matchOn:
+    #        - pciId:
+    #            vendor: ["15b3"]
+    #            device: ["1014", "1017"]
+    #          loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
+### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+
+  podSecurityContext: {}
+    # fsGroup: 2000
+
+  securityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop: [ "ALL" ]
+    readOnlyRootFilesystem: true
+    runAsNonRoot: true
+    # runAsUser: 1000
+
+  resources: {}
+    # We usually recommend not to specify default resources and to leave this as a conscious
+    # choice for the user. This also increases chances charts run on environments with little
+    # resources, such as Minikube. If you do want to specify resources, uncomment the following
+    # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+    # limits:
+    #   cpu: 100m
+    #   memory: 128Mi
+    # requests:
+    #   cpu: 100m
+    #   memory: 128Mi
+
+  nodeSelector: {}
+
+  tolerations: {}
+
+  annotations: {}
+
+## RBAC parameteres
+## https://kubernetes.io/docs/reference/access-authn-authz/rbac/
+##
+rbac:
+  create: true
+  serviceAccountName:
+  ## Annotations for the Service Account
+  ##
+  serviceAccountAnnotations: {}
--- a/docs/get-started/deployment-and-usage.md
+++ b/docs/get-started/deployment-and-usage.md
@@ -99,6 +99,101 @@ this approach does not guarantee running once on every node. For example,
 tainted, non-ready nodes or some other reasons in Job scheduling may cause some
 node(s) will run extra job instance(s) to satisfy the request.
 
+### Deployment with Helm
+
+Node Feature Discovery Helm chart allow to easily deploy and manage NFD.
+
+#### Prerequisites
+
+[Helm package manager](https://helm.sh/) should be installed.
+
+#### Deployment with Helm
+
+To install the chart with the release name node-feature-discovery:
+
+```bash
+git clone https://github.com/kubernetes-sigs/node-feature-discovery/
+cd node-feature-discovery/deployment
+export NFD_NS=node-feature-discovery
+helm install node-feature-discovery ./node-feature-discovery/ --namespace $NFD_NS --create-namespace
+```
+
+The command deploys Node Feature Discovery on the Kubernetes cluster in the default configuration.
+The Configuration section describes how it can be configured during installation.
+
+#### Configuration
+
+You can override values from `values.yaml` and provide a file with custom values:
+
+```bash
+export NFD_NS=node-feature-discovery
+helm install node-feature-discovery ./node-feature-discovery/ -f <path/to/custom/values.yaml> --namespace $NFD_NS --create-namespace
+```
+
+To specify each parameter separately you can provide them to helm install command:
+
+```bash
+export NFD_NS=node-feature-discovery
+helm install node-feature-discovery ./node-feature-discovery/ --set nameOverride=NFDinstance --set master.replicaCount=2 --namespace $NFD_NS --create-namespace
+```
+
+#### Uninstalling the Chart
+
+To uninstall the `node-feature-discovery` deployment:
+
+```bash
+export NFD_NS=node-feature-discovery
+helm uninstall node-feature-discovery --namespace $NFD_NS
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+#### Chart Parameters
+
+In order to tailor the deployment of the Node Feature Discovery to your cluster needs
+We have introduced the following Chart parameters.
+
+##### General parameters
+
+| Name | Type | Default | description |
+| ---- | ---- | ------- | ----------- |
+| `image.repository` | string | `gcr.io/k8s-staging-nfd/node-feature-discovery` | NFD image repository |
+| `image.pullPolicy` | string | `Always` | Image pull policy |
+| `imagePullSecrets` | list | [] | ImagePullSecrets is an optional list of references to secrets in the same namespace to use for pulling any of the images used by this PodSpec. If specified, these secrets will be passed to individual puller implementations for them to use. For example, in the case of docker, only DockerConfig type secrets are honored. [https://kubernetes.io/docs/concepts/containers/images#specifying-imagepullsecrets-on-a-pod](More info) |
+| `serviceAccount.create` | bool | true | Specifies whether a service account should be created |
+| `serviceAccount.annotations` | dict | {} | Annotations to add to the service account |
+| `serviceAccount.name` | string |  | The name of the service account to use. If not set and create is true, a name is generated using the fullname template |
+| `rbac` | dict |  | RBAC [parameteres](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) |
+| `nameOverride` | string |  | Override the name of the chart |
+| `fullnameOverride` | string |  | Override a default fully qualified app name |
+
+
+##### Master pod parameters
+
+| `master.*` | dict |  | NFD master deployment configuration |
+| `master.replicaCount` | integer | 1 | Number of desired pods. This is a pointer to distinguish between explicit zero and not specified |
+| `master.podSecurityContext` | dict | {} | SecurityContext holds pod-level security attributes and common container settings |
+| `master.service.type` | string | ClusterIP | NFD master service type |
+| `master.service.port` | integer | port | NFD master service port |
+| `master.resources` | dict | {} | NFD master pod [resources management](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) |
+| `master.nodeSelector` | dict | {} | NFD master pod [node selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) |
+| `master.tolerations` | dict | _Scheduling to master node is disabled_ | NFD master pod [tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) |
+| `master.annotations` | dict | {} | NFD master pod [metadata](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) |
+| `master.affinity` | dict |  | NFD master pod required [node affinity](https://kubernetes.io/docs/tasks/configure-pod-container/assign-pods-nodes-using-node-affinity/) |
+
+##### Worker pod parameters
+
+| `worker.*` | dict |  | NFD master daemonset configuration |
+| `worker.configmapName` | string | `nfd-worker-conf` | NFD worker pod ConfigMap name |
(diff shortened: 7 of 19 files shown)


adrianchiris:
Addresses issue #422

marquiz:
Thanks @adrianchiris for the PR! I'll take a closer look soon

/ok-to-test

deployment/node-feature-discovery/Chart.yaml
```diff
@@ -0,0 +1,13 @@
...
+appVersion: v0.7.0
+description: |
+  Detects hardware features available on each node in a Kubernetes cluster, and advertises
+  those features using node labels.
+name: node-feature-discovery
+sources:
+ - https://github.com/kubernetes-sigs/node-feature-discovery
+home: https://github.com/kubernetes-sigs/node-feature-discovery
+keywords:
+  - feature-discovery
```
ArangoGutierrez, line 11: ``
```suggestion
  - feature-discovery
  - feature-detection
  - node-labels
```

deployment/node-feature-discovery/templates/nfd-worker-conf.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nfd-worker-conf
+  namespace: {{ .Release.Namespace }}
+  labels:
+  {{- include "node-feature-discovery.labels" . | nindent 4 }}
+data:
+  nfd-worker.conf: |
+  {{ .Values.worker.config | indent 4 }}
```
ArangoGutierrez, line 10: Insert new line to make the linter happy

ArangoGutierrez (changes requested):
Looking good just a couple of suggestions

deployment/node-feature-discovery/templates/nfd-worker-conf.yaml
adrianchiris: Done

deployment/node-feature-discovery/Chart.yaml
adrianchiris: Sure, will add these labels as well

adrianchiris:
Hi,
A pair of eyes on this PR is appreciated  :)

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,207 @@
+image:
+  repository: gcr.io/k8s-staging-nfd/node-feature-discovery
+  pullPolicy: IfNotPresent
```
marquiz, line 3: If using the staging repo the pull policy should prolly be `Always`

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,207 @@
...
+  podSecurityContext: {}
+    # fsGroup: 2000
+
+  securityContext: {}
+    # capabilities:
+    #   drop:
+    #   - ALL
+    # readOnlyRootFilesystem: true
+    # runAsNonRoot: true
+    # runAsUser: 1000
```
marquiz, line 30: Is there any reason we shouldn't enable these, by default, except for the `runAsUser` setting?

Also, `allowPrivilegeEscalation: false`

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,207 @@
...
```
marquiz, line 165: Same thing as for nfd-master, shouldn't we enable these (sans `runAsUser`) spiced up with `allowPrivilegeEscalation: false`

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,207 @@
...
+    #   cpu: 100m
+    #   memory: 128Mi
+
+  nodeSelector: {}
+
+  tolerations:
+  - key: "node-role.kubernetes.io/master"
+    operator: "Equal"
+    value: ""
+    effect: "NoSchedule"
```
marquiz, line 185: I'm not sure if we should run nfd-worker on dedicated master nodes, by default. WDYT?

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+  affinity:
+    nodeAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 1
+          preference:
+            matchExpressions:
+              - key: "node-role.kubernetes.io/master"
+                operator: In
+                values: [""]
```
marquiz, line 197: We shouldn't have this for nfd-worker

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,207 @@
...
+
+## RBAC parameteres
+## https://kubernetes.io/docs/reference/access-authn-authz/rbac/
+##
+rbac:
+  create: true
+  serviceAccountName: default
```
marquiz, line 204: Shouldn't this be in sync with the serviceAccount settings above?

marquiz (commented):
Apologies for the delayed review. I had long holidays...

I'm not an expert on helm charts but basically looks good. Just had a few small comments and questions. One thing that I'd like to ask is to update documentation. I.e. a new section in `docs/get-started/deployment-and-usage.md` describing helm deployment.

In general I think it would be good to keep the raw deployment yaml templates and helm charts as much in sync as possible. I don't know if there's anything we can really do to make to help with this but to just state it somewhere (e.g. comments in all yaml files or smth).

deployment/node-feature-discovery/values.yaml
ArangoGutierrez: ++

deployment/node-feature-discovery/values.yaml
ArangoGutierrez: maybe you want to look at https://github.com/kubernetes-sigs/node-feature-discovery-operator/blob/master/assets/worker/0700_worker_daemonset.yaml#L16 to set this rules on how nfd-worker will be placed

deployment/node-feature-discovery/templates/nfd-worker-conf.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nfd-worker-conf
+  namespace: {{ .Release.Namespace }}
```
marquiz, line 5: Could you educate me on helm: is the namespace needed to be defined? If so, why?

deployment/node-feature-discovery/templates/worker.yaml
```diff
@@ -0,0 +1,113 @@
...
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        resources:
+        {{- toYaml .Values.worker.resources | nindent 12 }}
+        command:
+        - "nfd-worker"
+        args:
+        - "--sleep-interval=60s"
```
e0ne, line 44: Should we add other startup options mentioned in https://kubernetes-sigs.github.io/node-feature-discovery/v0.6/advanced/#nfd-worker?

deployment/node-feature-discovery/templates/nfd-worker-conf.yaml
adrianchiris: it does not i will remove, helm will set the namespace during installation

deployment/node-feature-discovery/templates/worker.yaml
adrianchiris: IMO we should start with the defaults and add configuration per request if needed, keeping things simple as much as possible.

deployment/node-feature-discovery/values.yaml
adrianchiris: i see, nfd-worker is not expected to run on master by default. i will fix

deployment/node-feature-discovery/templates/nfd-worker-conf.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: nfd-worker-conf
```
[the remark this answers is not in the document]
adrianchiris: yes i will align the name.

deployment/node-feature-discovery/templates/master.yaml
```diff
@@ -0,0 +1,84 @@
...
+          - containerPort: 8080
+            name: grpc
+          env:
+          - name: NODE_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: spec.nodeName
+          command:
+            - "nfd-master"
+
```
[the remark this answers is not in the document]
adrianchiris: will remove

deployment/node-feature-discovery/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v2
+appVersion: v0.7.0
```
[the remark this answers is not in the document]
adrianchiris: looking at helm docs i see:

`appVersion: The version of the app that this contains (optional). This needn't be SemVer.`
so as this PR is intended to work with master branch i will set this to `master` and align the images in the chart.

as a second step when we setup a helm repository for this, the release chart would need to have the actual release verison.

deployment/node-feature-discovery/templates/_helpers.tpl
```diff
@@ -0,0 +1,63 @@
...
+{{- end -}}
+
+{{/*
+Common labels
+*/}}
+{{- define "node-feature-discovery.labels" -}}
+helm.sh/chart: {{ include "node-feature-discovery.chart" . }}
+{{ include "node-feature-discovery.selectorLabels" . }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
```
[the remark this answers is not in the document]
adrianchiris: we can, does `node-feature-discovery` as app name works ?

deployment/node-feature-discovery/templates/master.yaml
```diff
@@ -0,0 +1,84 @@
...
+          resources:
+            {{- toYaml .Values.master.resources | nindent 12 }}
+## Enable TLS authentication
+## The example below assumes having the root certificate named ca.crt stored in
+## a ConfigMap named nfd-ca-cert, and, the TLS authentication credentials stored
+## in a TLS Secret named nfd-master-cert.
+## Additional hardening can be enabled by specifying --verify-node-name in
+## args, in which case every nfd-worker requires a individual node-specific
+## TLS certificate.
+#          args:
```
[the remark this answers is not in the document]
adrianchiris: This can be done so deployers can provide additional args.

however as i see the original intention of this section is to support nfd worker auth, this would not help in enabling it as certificates would still need to be mounted to the master.

deploying the needed constructs for worker authentication can be simplified with helm (add a section for this in values.yaml with all needed configuration, create config maps and add mounts to master etc..)
This specific feature can be addressed in a separate PR IMO.

deployment/node-feature-discovery/templates/worker.yaml
adrianchiris: Thinking on this more, i can expose `args` to be overwritten by deployment WDYT ?

deployment/node-feature-discovery/values.yaml
adrianchiris: Hmm, i missed that, yes it should !

deployment/node-feature-discovery/values.yaml
adrianchiris: I can find this beneficial for deploying in a development cluster. real deployment wouldn't need nfd-worker running on the master IMO, so thinking to set the defaults as to not run on master and give the user an option to override WDYT?

@ArangoGutierrez  is the node-feature-discovery operator intended to deploy NFD via OLM ?

ArangoGutierrez: NFD can be deployed on OpenShift via OperatorHub or at any Kubernetes cluster via the Operator-framework community operators.
Notice that recently we added support for nfd-workers to run on master nodes, by request of the developers working on kubernetes edge support, this will be single node clusters, so some times, NFD master and worker would end up running on the same node. 
Upstream operator -> https://github.com/operator-framework/community-operators/tree/master/upstream-community-operators/nfd-operator
OpenShift-community operator -> https://github.com/operator-framework/community-operators/tree/master/community-operators/openshift-nfd-operator

deployment/node-feature-discovery/values.yaml
ArangoGutierrez: well, can this be set via args on HELM? edge users are running single node clusters and they require NFD to run master and worker on the same node bc of that

deployment/node-feature-discovery/values.yaml
adrianchiris: So i would need some education here on which image repos NFD has

in general i think we want this (as part of master) to deploy the latest master image.
what repo should we use ? (staging ok i guess for master builds)

and also when a new NFD version is released which repo should be used ?  (will add as comment)

currently the image tag is set to the chart `AppVersion`.  per other comment this will no longer be the case.
I will add to this section a `tag` attribute so the container image can be properly identified.

Will change the pull policy to Always and add a comment that for released versions this should be set to IfNotPresent.
(as a tagged version is going to be used)

deployment/node-feature-discovery/templates/worker.yaml
```diff
@@ -0,0 +1,113 @@
...
+          readOnly: true
+        - name: host-os-release
+          mountPath: "/host-etc/os-release"
+          readOnly: true
+        - name: host-sys
+          mountPath: "/host-sys"
+        - name: source-d
+          mountPath: "/etc/kubernetes/node-feature-discovery/source.d/"
+        - name: features-d
+          mountPath: "/etc/kubernetes/node-feature-discovery/features.d/"
```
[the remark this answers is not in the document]
adrianchiris: i will align this

adrianchiris (commented):
Will update the PR during next week. 

>Apologies for the delayed review. I had long holidays...

No worries, hope you enjoyed the holidays ! :)


>I'm not an expert on helm charts but basically looks good. Just had a few small comments >and questions. One thing that I'd like to ask is to update documentation. I.e. a new section in >docs/get-started/deployment-and-usage.md describing helm deployment.
>

Will add doc as well. this was actually in my TODO list for this PR. for now ill cover how to deploy with helm from master. as a second step we should look into creating a helm package for each release and setup a helm repository.

>In general I think it would be good to keep the raw deployment yaml templates and helm >charts as much in sync as possible. I don't know if there's anything we can really do to make >to help with this but to just state it somewhere (e.g. comments in all yaml files or smth).

This would need to be looked after in PR reviews unfortunately, but ill add a comment in the yaml templates saying that if you change it, helm deployment templates should be changed as well.

deployment/node-feature-discovery/values.yaml
adrianchiris: yes, my intention is to allow the deployer to customize it to allow this use-case (also for myself personally when i work in an AIO setup :))

deployment/node-feature-discovery/values.yaml
adrianchiris: i will try to align the defaults with the templates

deployment/node-feature-discovery/values.yaml
adrianchiris: i will try to align the defaults with the templates

ArangoGutierrez:
> Will update the PR during next week.
> 
> > Apologies for the delayed review. I had long holidays...
> 
> No worries, hope you enjoyed the holidays ! :)
> 
> > I'm not an expert on helm charts but basically looks good. Just had a few small comments >and questions. One thing that I'd like to ask is to update documentation. I.e. a new section in >docs/get-started/deployment-and-usage.md describing helm deployment.
> 
> Will add doc as well. this was actually in my TODO list for this PR. for now ill cover how to deploy with helm from master. as a second step we should look into creating a helm package for each release and setup a helm repository.
> 
> > In general I think it would be good to keep the raw deployment yaml templates and helm >charts as much in sync as possible. I don't know if there's anything we can really do to make >to help with this but to just state it somewhere (e.g. comments in all yaml files or smth).
> 
> This would need to be looked after in PR reviews unfortunately, but ill add a comment in the yaml templates saying that if you change it, helm deployment templates should be changed as well.

Awesome, thanks for your contribution, is really appreciated!

deployment/node-feature-discovery/templates/_helpers.tpl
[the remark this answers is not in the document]
marquiz: Good, And yes, I think it `node-feature-discovery` would be in line with the rest of this PR

deployment/node-feature-discovery/Chart.yaml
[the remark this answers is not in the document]
marquiz: > as a second step when we create a release package (and possibly make it available in a helm repository), the release chart would need to have the actual release verison.

Exactly. There is `scripts/prepare-release.sh` that is supposed to do most of that magic. Could you try to update that, too? It should basically be adding yet another sed rule.

deployment/node-feature-discovery/templates/master.yaml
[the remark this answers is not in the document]
marquiz: Fair enough

deployment/node-feature-discovery/values.yaml
marquiz: 👍

deployment/node-feature-discovery/values.yaml
marquiz: 👍

deployment/node-feature-discovery/values.yaml
marquiz: > IMO, so thinking to set the defaults as to not run on master and give the user an option to override WDYT?

That would sound good to me. To keep the two deployment methods (raw templates and helm) similar in this aspect. If we want to change the default behavior that should be done in both, then, I think.

deployment/node-feature-discovery/values.yaml
marquiz: 👍

deployment/node-feature-discovery/templates/worker.yaml
marquiz: Sounds good to me 👍  But, I think this can be left as a future enhancement, too, if you like it better that way

marquiz:
> Will add doc as well. this was actually in my TODO list for this PR. for now ill cover how to deploy with helm from master. as a second step we should look into creating a helm package for each release and setup a helm repository.

OK, good



> This would need to be looked after in PR reviews unfortunately, but ill add a comment in the yaml templates saying that if you change it, helm deployment templates should be changed as well.

I think this would be enough, thanks

deployment/node-feature-discovery/templates/worker.yaml
```diff
@@ -59,10 +59,12 @@ spec:
           readOnly: true
         - name: host-sys
           mountPath: "/host-sys"
+          readOnly: true
         - name: source-d
           mountPath: "/etc/kubernetes/node-feature-discovery/source.d/"
```
marquiz, line 64: `source-d` should be read-only like the other mounts

deployment/node-feature-discovery/values.yaml
marquiz: What about this? I still fail to understand why nfd-worker would need to have affinity towards master nodes

nfd-worker-daemonset.yaml.template
```diff
@@ -29,6 +31,11 @@ spec:
               drop: ["ALL"]
             readOnlyRootFilesystem: true
             runAsNonRoot: true
+          tolerations:
```
adrianchiris, line 34: I think we said that by default it should not run on master, am i wrong ?

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,208 @@
...
+    # limits:
+    #   cpu: 100m
+    #   memory: 128Mi
+    # requests:
+    #   cpu: 100m
+    #   memory: 128Mi
+
+  nodeSelector: {}
+
+  tolerations:
```
adrianchiris, line 182: I think we want to keep this empty for the worker.

users can then override if they have all-in-one node or want to run on master

deployment/node-feature-discovery/templates/_helpers.tpl
```diff
@@ -0,0 +1,66 @@
...
+{{- end }}
+{{- if .Chart.Name }}
+app.kubernetes.io/name: {{ .Chart.Name | quote }}
+{{- end }}
```
marquiz, line 45: Argh, my previous comment on this looks like amiss 😬  Now that I look at this the `app.kubernetes.io/name` label seems to always be present via `include "node-feature-discovery.selectorLabels"` above, and, this could be removed. Am I right?

deployment/node-feature-discovery/templates/worker.yaml
marquiz: > OK, let's do it in a separate PR

That's fine. nfd-master args could be handled in the same (future) pr

deployment/node-feature-discovery/values.yaml
marquiz: > in general i think we want this (as part of master) to deploy the latest master image.
> what repo should we use ? (staging ok i guess for master builds)

Yes, staging is the correct one, as is the case in this PR
 
> and also when a new NFD version is released which repo should be used ? (will add as comment)
 
Release images are served under `k8s.gcr.io/nfd/node-feature-discovery`. Actually it would be good to add this to the `prepare-release.sh` script, too.
 
> currently the image tag is set to the chart `AppVersion`. per other comment this will no longer be the case.
> I will add to this section a `tag` attribute so the container image can be properly identified.
> 
> Will change the pull policy to Always and add a comment that for released versions this should be set to IfNotPresent.
> (as a tagged version is going to be used)

👍  We should add this (changing pull policy) to `prepare-release.sh` as well.

scripts/prepare-release.sh
```diff
@@ -68,6 +68,9 @@ sed s"!node-feature-discovery/v.*/!node-feature-discovery/$release/!" -i README.
 echo Patching '*.yaml.template' to use $container_image
 sed -E s",^([[:space:]]+)image:.+$,\1image: $container_image," -i *yaml.template
 
+# Patch Helm chart
+sed -e s"/appVersion:.*/appVersion: $release/" -i deployment/node-feature-discovery/Chart.yaml
```
marquiz, line 72: This is good. From the other comments I realized that we need to patch the `repository` and `pullPolicy` fields, too

scripts/prepare-release.sh
```diff
@@ -70,6 +70,8 @@ sed -E s",^([[:space:]]+)image:.+$,\1image: $container_image," -i *yaml.template
 
 # Patch Helm chart
 sed -e s"/appVersion:.*/appVersion: $release/" -i deployment/node-feature-discovery/Chart.yaml
+sed -e s"/pullPolicy: Always$/pullPolicy: IfNotPresent/" -i deployment/node-feature-discovery/values.yaml
```
marquiz, line 73: nit: you could combine all three rules into a single sed call, especially as you already use `-e`. Like this:
```
sed -e <rule1> \
    -e <rule-2> \
    -i <file>
```

nit: it would be more fool-proof to use wildcards similar to appVersion, i.e. `s"/pullPolicy:.*$/pullPolicy: IfNotPresent/"`

scripts/prepare-release.sh
```diff
@@ -70,6 +70,8 @@ sed -E s",^([[:space:]]+)image:.+$,\1image: $container_image," -i *yaml.template
...
+sed -e s"/k8s-staging-nfd\/node-feature-discovery$/k8s-staging-nfd\/node-feature-discovery:$release/" -i deployment/node-feature-discovery/values.yaml
```
marquiz, line 74: The production repo is `k8s.gcr.io/nfd/node-feature-discovery`. nit: you could use some other separator than slash in order to improve readability
```suggest
    -e s"!gcr.io/k8s-staging-nfd/node-feature-discovery!k8s.gcr.io/nfd/node-feature-discovery!" \
    -i deployment/node-feature-discovery/values.yaml
```

scripts/prepare-release.sh
marquiz:
```suggest
sed -e s"/pullPolicy:.*/pullPolicy: IfNotPresent/" \
```

scripts/prepare-release.sh
e0ne: Thanks for the tip

marquiz (changes requested):
Argh, there is still one thing that I forgot, and, which I hit myself in #433 (and you in this PR, too, I think): There is the `templates` makefile target which updates the ConfigMap spec in the yaml templates. That is, it copies the contents of `nfd-worker.conf.example` to the tempates. We also have a CI rule in place that checks that the templates are up-to-date.

I'd like to have this same mechanism to update the Helm chart (`values.yaml` to be precise, I think), too. That would make maintenance a tad easier, removing one point of possible human error. Currently (for the yaml templates) the auto-update is done with a rather simple-and-stupid one-liner sed script. I think this does not work as is for the helm yaml so something else needs to be devised. One possible easy-ish solution would be to add start and end markers (something like `### NFD-WORKER-CONF-START (DO NOT REMOVE)`) and replace the content between with sed or awk or smth. I'd add the same markers for yaml templates and helm file(s) so that the same script could be used to update all of them.

The other thing still missing is the documentation.

But other than these two things I think this would be ready 👍

marquiz:
> I'd like to have this same mechanism to update the Helm chart (`values.yaml` to be precise, I think), too. That would make maintenance a tad easier, removing one point of possible human error. Currently (for the yaml templates) the auto-update is done with a rather simple-and-stupid one-liner sed script. I think this does not work as is for the helm yaml so something else needs to be devised. One possible easy-ish solution would be to add start and end markers (something like `### NFD-WORKER-CONF-START (DO NOT REMOVE)`) and replace the content between with sed or awk or smth. I'd add the same markers for yaml templates and helm file(s) so that the same script could be used to update all of them.

Related to this, I started to really dislike how the update works at the moment. Submitted #434 which uses start and end markers. I don't know if this is much nicer but at least it should be directly usable for updating the helm chart, too

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,194 @@
...
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 1
+          preference:
+            matchExpressions:
+              - key: "node-role.kubernetes.io/master"
+                operator: In
+                values: [""]
+
+worker:
+  config: |
```
marquiz, line 70: After the latest change in master (#434) you should now update the `make templates` target in Makefile to update the configmapcontent 

```suggest
  config: | ### <NFD-WORKER-CONF-START-DO-NOT-REMOVE>
```

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,194 @@
...
+    #          class: ["fe"]
+    #          vendor: ["1a6e"]
+    #          device: ["089a"]
+    #    - name: "my.combined.feature"
+    #      matchOn:
+    #        - pciId:
+    #            vendor: ["15b3"]
+    #            device: ["1014", "1017"]
+    #          loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
+
```
marquiz, line 156: Related to the auto-update with `make templates`:
```suggest
### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
```

marquiz (changes requested):

nfd-daemonset-combined.yaml.template
```diff
@@ -141,7 +141,8 @@ metadata:
   name: nfd-worker-conf
   namespace: node-feature-discovery
 data:
-  nfd-worker.conf: | ### <NFD-WORKER-CONF-START-DO-NOT-REMOVE>
+  nfd-worker.conf: |
+    ### <NFD-WORKER-CONF-START-DO-NOT-REMOVE>
```
marquiz, line 145: Please don't edit the templates. I think it's nicer that these comments do not end up in the actual config map that gets created

nfd-daemonset-combined.yaml.template
```diff
@@ -228,4 +229,4 @@ data:
     #            vendor: ["15b3"]
     #            device: ["1014", "1017"]
     #          loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
-### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+    ### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
```
marquiz, line 232: Same here. The (un-)indentation was intentional

nfd-daemonset-combined.yaml.template
e0ne: Such formatting was generated by 'make templates' execution. should I fix it?

marquiz: OK, sorry 😬  I'll check that, don't change it

e0ne: Please, ignore my comment above. It was an issue on my environment

marquiz: Hmm, I can't possibly see how this could happen by the sed script 🤔  What OS and sed version are you using? Are you using some kind of automatic YAML formatter tool?

`values.yaml` did not get automatically mangled like this. Please change these back as they were.

nfd-worker.conf.example
```diff
@@ -1,3 +1,4 @@
+# All changes in this template should be applied to Helm chart too.
```
marquiz, line 1: Please remove this line. It's not needed as `values.yaml` is update by `make templates`

e0ne:
@marquiz I'm confused now. After reading https://github.com/kubernetes-sigs/node-feature-discovery/pull/423#discussion_r571973093 I was thinking that values.yaml should be updated too. Should I rollback changes that generates config in values.yaml? In a such case, we need to sync config to chart manually

marquiz:
> @marquiz I'm confused now. After reading [#423 (comment)](https://github.com/kubernetes-sigs/node-feature-discovery/pull/423#discussion_r571973093) I was thinking that values.yaml should be updated too. Should I rollback changes that generates config in values.yaml? In a such case, we need to sync config to chart manually

No no no, I just missed your previous comment. You were too quick to respond 😄 It's good now 👍

marquiz:
There's only one nitty-gritty open comment (remove the first line in nfd-worker.conf.example). In addition to that, could you please squash the PR into one commit. I think this looks really good now (👍 ) and I don't think I'll have any further comments to the "code". 

We're still missing the documentation part, though

e0ne:
> We're still missing the documentation part, though

Should I add it in the current pull request or create a new one?

marquiz:
> Should I add it in the current pull request or create a new one?

Please add it to this one. It can be a separate commit, though, if you wish

marquiz:
And please remove the first line from nfd-worker.conf.example

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,51 @@ this approach does not guarantee running once on every node. For example,
 tainted, non-ready nodes or some other reasons in Job scheduling may cause some
 node(s) will run extra job instance(s) to satisfy the request.
 
+### Helm Chart
+
+This chart bootstraps a Node Feature deployment on a Kubernetes cluster using the Helm package manager.
```
adrianchiris, line 104: suggested reword:

>This chart bootstraps a Node Feature deployment

Node Feature Discovery Helm chart allow to easily deploy and manage NFD ....

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,51 @@ this approach does not guarantee running once on every node. For example,
...
+
+#### Prerequisites
+
+[Helm package manager](https://helm.sh/) should be installed.
+
+#### Installing the Chart
+
+To install the chart with the release name network-operator:
```
adrianchiris, line 112: > network-operator

nfd ? or node-feature-discovery ?

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,51 @@ this approach does not guarantee running once on every node. For example,
 tainted, non-ready nodes or some other reasons in Job scheduling may cause some
 node(s) will run extra job instance(s) to satisfy the request.
 
+### Helm Chart
```
adrianchiris, line 102: Similarly to Deployment Templates heading above , perhaps : Deployment with Helm

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,51 @@ this approach does not guarantee running once on every node. For example,
...
+[Helm package manager](https://helm.sh/) should be installed.
+
+#### Installing the Chart
+
+To install the chart with the release name network-operator:
+
+```bash
+git clone https://github.com/kubernetes-sigs/node-feature-discovery/
+cd node-feature-discovery/deployment
+helm install network-operator ./network-operator/
```
adrianchiris, line 117: > helm install network-operator ./network-operator/

helm install node-feature-discovery ./node-feature-discovery/

IMO you should replace below as well  (network-operator -> node-feature-discovery)

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,51 @@ this approach does not guarantee running once on every node. For example,
...
+#### Uninstalling the Chart
+
+To uninstall the `network-operator` deployment:
+
+```bash
+helm uninstall network-operator 
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
```
adrianchiris, line 146: Should we have a section which describes the chart parameters similarly to what we have in ?

https://github.com/Mellanox/network-operator/tree/master/deployment/network-operator#chart-parameters

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,195 @@
+image:
+  repository: gcr.io/k8s-staging-nfd/node-feature-discovery
```
ArangoGutierrez, line 2: I think we should add image tag as a value, with `master` as default

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,195 @@
...
+  # Annotations to add to the service account
+  annotations: {}
+  # The name of the service account to use.
+  # If not set and create is true, a name is generated using the fullname template
+  name:
+
+nameOverride: ""
+fullnameOverride: ""
+
+master:
```
ArangoGutierrez, line 19: this is defaulting `nfd-master` as a replicaset and not daemonset, why?

deployment/node-feature-discovery/values.yaml
```diff
@@ -0,0 +1,195 @@
...
+
+  tolerations:
+  - key: "node-role.kubernetes.io/master"
+    operator: "Equal"
+    value: ""
+    effect: "NoSchedule"
+
+  annotations: {}
+
+  affinity:
```
ArangoGutierrez, line 59: this is ok for master
we should have also a way to set values for allocation values for `nfd-worker` 
example
```yaml
      tolerations:
      - operator: "Exists" 
        effect: "NoSchedule"
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
              - matchExpressions:
                  - key: node-role.kubernetes.io/master
                    operator: DoesNotExist
              - matchExpressions:
                  - key: node-role.kubernetes.io/worker
                    operator: Exists
```

deployment/node-feature-discovery/templates/worker.yaml
```diff
@@ -0,0 +1,116 @@
...
+            path: "/sys"
+        - name: source-d
+          hostPath:
+            path: "/etc/kubernetes/node-feature-discovery/source.d/"
+        - name: features-d
+          hostPath:
+            path: "/etc/kubernetes/node-feature-discovery/features.d/"
+        - name: nfd-worker-conf
+          configMap:
+            name: nfd-worker-conf
```
ArangoGutierrez, line 97:
```suggestion
            name: node-feature-discovery
            items:
              - key: nfd-worker.conf
                path: nfd-worker.conf
```

ArangoGutierrez (changes requested):
Some comments after testing this my self on a 1.20+ cluster

deployment/node-feature-discovery/templates/master.yaml
```diff
@@ -0,0 +1,83 @@
...
+        {{- toYaml . | nindent 8 }}
+    {{- end }}
+      serviceAccountName: {{ include "node-feature-discovery.serviceAccountName" . }}
+      securityContext:
+        {{- toYaml .Values.master.podSecurityContext | nindent 8 }}
+      containers:
+        - name: master
+          securityContext:
+            {{- toYaml .Values.master.securityContext | nindent 12 }}
+          image: "{{ .Values.image.repository }}:{{ .Chart.AppVersion }}"
```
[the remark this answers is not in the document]
e0ne: OK, let's add 'version' entry

deployment/node-feature-discovery/values.yaml
e0ne: I'm not sure that I understood you right. nfd-master is a Deployment: https://github.com/kubernetes-sigs/node-feature-discovery/pull/423/files#diff-a4f5c7e13b9887fcc6c41a624fa5b34e46042d4b1dcaf2401c78e98c7451ccedR2

deployment/node-feature-discovery/templates/master.yaml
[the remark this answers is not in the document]
adrianchiris: > OK, let's add 'version' entry

this can be done similarly as we did in network-operator 

https://github.com/Mellanox/network-operator/blob/1b733d443861a84531116bc8fb2f1c09873e1284/deployment/network-operator/templates/operator.yaml#L49

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,65 @@ this approach does not guarantee running once on every node. For example,
...
+The command deploys Node Feature Discovery on the Kubernetes cluster in the default configuration.
+The Configuration section describes how it can be configured during installation.
+
+#### Configuration
+
+You can override values from `values.yaml` and provide a file with custom values:
+
+```bash
+helm install node-feature-discovery ./node-feature-discovery/ -f values.yaml
```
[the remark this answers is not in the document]
adrianchiris: @ArangoGutierrez  you should create a `.yaml` file with overrides to fields for helm deployment then provide it with -f

is the comment here that the path is wrong and it should be `./values.yaml` ?

also the WARNING bit is related to your k8s deployment IMO (Openshift?)

deployment/node-feature-discovery/values.yaml
adrianchiris: So should tolerations be empty by default ? currently this will add a toleration to any taint with an effect of "NoSchedule"

deployment/node-feature-discovery/values.yaml
```diff
@@ -1,5 +1,6 @@
 image:
   repository: gcr.io/k8s-staging-nfd/node-feature-discovery
+  version: master
```
adrianchiris, line 3: Currently `appVersion` is set to `master` in `Chart.yaml`
When a chart is released `appVersion` will be updated with nfd release tag so the correct image is pulled.
with this change, this field would need to be updated as well.

IMO this should be empty by default and used as image tag only if specified by the user. (i.e image tag override.)

deployment/node-feature-discovery/templates/worker.yaml
```diff
@@ -94,7 +94,10 @@ spec:
             path: "/etc/kubernetes/node-feature-discovery/features.d/"
         - name: nfd-worker-conf
           configMap:
-            name: nfd-worker-conf
+            name: node-feature-discovery
+              items:
+                - key: nfd-worker.conf
```
adrianchiris, line 99: this is no longer aligned with the templates : https://github.com/kubernetes-sigs/node-feature-discovery/blob/2960f63ae5d51bd77015aa4adfe96a90bf753faa/nfd-worker-daemonset.yaml.template#L87

I thought we want to be aligned with the templates as much as possible in the chart.

docs/get-started/deployment-and-usage.md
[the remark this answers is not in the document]
ArangoGutierrez: yeah don't mind the WARNING, I was pointing out that the path is wrong, and since this is user documentation, we should be more explicit so the user doesn't run into a `file not found` error log

deployment/node-feature-discovery/values.yaml
ArangoGutierrez: Is ok, I am super used to run NFD as a daemon set, but this is ok and works just fine

deployment/node-feature-discovery/values.yaml
ArangoGutierrez: But what if an user/developer wants to use HELM to deploy a devel version of NFD, how can I override the TAG value? is there a way we can set this from the command line? `helm install .... --set image.tag=devel` ?

deployment/node-feature-discovery/templates/worker.yaml
ArangoGutierrez: In deed, but I noticed this line https://github.com/adrianchiris/node-feature-discovery/blob/add-helm-chart/deployment/node-feature-discovery/templates/worker.yaml#L97 so it never mounts the config map, bc is a diff name

deployment/node-feature-discovery/templates/master.yaml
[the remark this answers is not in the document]
ArangoGutierrez: Is a good approach, anything that allows users/developers to set custom TAG values, is well received

docs/get-started/deployment-and-usage.md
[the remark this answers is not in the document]
adrianchiris: I see , thanks for clarifying !

deployment/node-feature-discovery/values.yaml
```diff
@@ -180,7 +181,19 @@ worker:
...
+    affinity:
+      nodeAffinity:
+        requiredDuringSchedulingIgnoredDuringExecution:
+          nodeSelectorTerms:
+            - matchExpressions:
+                - key: node-role.kubernetes.io/master
+                  operator: DoesNotExist
+            - matchExpressions:
+                - key: node-role.kubernetes.io/worker
+                  operator: Exists
```
marquiz, line 196: See above comment about tolerations. I suggest to make this change in a separate PR, changing the templates at the same time.

deployment/node-feature-discovery/values.yaml
marquiz: I think this has been addressed. Resolving the conversation

deployment/node-feature-discovery/values.yaml
adrianchiris: yes exactly. 

a user may override. but if not specified, use `.Chart.AppVersion`

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,65 @@ this approach does not guarantee running once on every node. For example,
...
+In order to tailor the deployment of the Node Feature Discovery to your cluster needs
+We have introduced the following Chart parameters.
+
+| Name | Type | Default | description |
+| ---- | ---- | ------- | ----------- |
+| `image.repository` | string | `gcr.io/k8s-staging-nfd/node-feature-discovery` | NFD image repository |
+| `image.pullPolicy` | string | `Always` | Image pull policy |
+| `master.*` | dict |  | NFD master deployment configuration |
+| `worker.*` | dict |  | NFD master daemonset configuration |
+| `worker.config` | string | `` | NFD worker service configuration |
```
marquiz, line 158: This is really good 👍  I think we should add `version` here as well.

In addition, could we open up the `master.*` and `worker.*` parameters, too(?)

marquiz (changes requested):
Thanks for the documentation! Still some minor comments that need to be addressed

docs/get-started/deployment-and-usage.md
e0ne: @marquiz do we need to add all master.* and worker.* params?

e0ne: master.* and worker.*  contain general kubernetes  settings and they are not NFD-specific. That's why I didn't add details to them

deployment/node-feature-discovery/templates/worker.yaml
ArangoGutierrez: yeah maybe my comment should have been against line https://github.com/adrianchiris/node-feature-discovery/blob/add-helm-chart/deployment/node-feature-discovery/templates/worker.yaml#L97  and not here to make things similar to the templated examples

deployment/node-feature-discovery/values.yaml
ArangoGutierrez: I can take that as a TODO for sure, this add NFD support for people using it on single node deployments and edge cases, so is on my priority list

marquiz: K, thanks

docs/get-started/deployment-and-usage.md
marquiz: I think it would be very useful for the reader (like me) to read here what aspects of the deployment are configurable via Helm parameters. Or is it common Helm practice/knowledge that these (and only these) general deployment settings are configurable via parameters of the same name? If so, a link to this information would be nice.

deployment/node-feature-discovery/values.yaml
marquiz: Are you happy with this? Can we resolve the conversation?

ArangoGutierrez: yes

marquiz:
Thanks @e0ne! Looks like the only open comment is changing (back) the name of the (worker config) configmap to `nfd-worker-conf`

deployment/node-feature-discovery/templates/nfd-worker-conf.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "node-feature-discovery.fullname" . }}
```
marquiz, line 4: Hmm, this must be aligned with the mount in nfd-worker. A separate configuration parameter in values.yaml e.g. `worker.configmapName`(?)

marquiz (changes requested):
Thanks for the updates @e0ne. Still found few minor things to fix. After these are done you could probably squash the PR again and we should be good to go 🤞 😄

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,96 @@ this approach does not guarantee running once on every node. For example,
...
+[Helm package manager](https://helm.sh/) should be installed.
+
+#### Deployment with Helm
+
+To install the chart with the release name node-feature-discovery:
+
+```bash
+git clone https://github.com/kubernetes-sigs/node-feature-discovery/
+cd node-feature-discovery/deployment
+helm install node-feature-discovery ./node-feature-discovery/
```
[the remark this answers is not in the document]
ArangoGutierrez: I agree, I am not knowledgeable in Helm, and the first time testing this PR NFD got installed in another Project nameSpace I was working in. (I thought Helm would auto-generate a new NS) so for people like me that are new to Helm is good to be explicit

deployment/node-feature-discovery/templates/nfd-worker-conf.yaml
e0ne: Good catch, will fix it asap

e0ne:
@marquiz could you please help me with CI failure troubleshooting? 'make templates' on my local env doesn't generate any new changes

ArangoGutierrez:
```bash
[eduardo@fedora-ws node-feature-discovery]$ ./scripts/test-infra/verify.sh 
golangci/golangci-lint info checking GitHub for tag 'v1.30.0'
golangci/golangci-lint info found version: 1.30.0 for v1.30.0/linux/amd64
golangci/golangci-lint info installed /home/eduardo/go/bin/golangci-lint
golangci-lint run --timeout 5m0s
[eduardo@fedora-ws node-feature-discovery]$ git branch 
* (HEAD detached at adrianchiris/add-helm-chart)*
```
/test pull-node-feature-discovery-verify

marquiz:
> @marquiz could you please help me with CI failure troubleshooting? 'make templates' on my local env doesn't generate any new changes

Ach, this must be caused by the latest changes in master. Prow doesn't actually test the revision here, but, this revision merged to master. Rebase on top of master and try `make templates` again.

e0ne:
@ArangoGutierrez  thank you. I executed it locally and it didn't find any issue. Let's see if retriggered job passes

ArangoGutierrez:
Yeah @marquiz is right, maybe you need a soft rebase, that should fix it

e0ne:
@marquiz  thanks!

ArangoGutierrez:
Makes me wonder, why Prow is not setting the `needs-rebase` label? 🤔

ArangoGutierrez:
Rebase did the trick

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,98 @@ this approach does not guarantee running once on every node. For example,
...
+You can override values from `values.yaml` and provide a file with custom values:
+
+```bash
+helm install node-feature-discovery ./node-feature-discovery/ -f <path/to/custom/values.yaml>
+```
+
+To specify each parameter separately you can provide them to helm install command:
+
+```bash
+helm install node-feature-discovery ./node-feature-discovery/ --set nameOverride=NFDinstance --set master.replicaCount=2
```
marquiz, line 135: ...and here
```suggestion
helm install node-feature-discovery ./node-feature-discovery/ --namespace $NFD_NS --create-namespace --set nameOverride=NFDinstance --set master.replicaCount=2
```

docs/get-started/deployment-and-usage.md
```diff
@@ -99,6 +99,98 @@ this approach does not guarantee running once on every node. For example,
...
+```
+
+#### Uninstalling the Chart
+
+To uninstall the `node-feature-discovery` deployment:
+
+```bash
+helm uninstall node-feature-discovery
```
marquiz, line 143: ...and here
```suggestion
helm uninstall node-feature-discovery --namespace $NFD_NS
```

marquiz (changes requested):

marquiz:
> Makes me wonder, why Prow is not setting the `needs-rebase` label? 🤔

Because from the git pow, no rebase is needed, merge succeeds just fine. It's just our testing (or verify step) wasn't happy with the merged result. Prow doesn't know the cause behind this failure

marquiz:
I don't have anything else to add 😊  Great work @e0ne and @adrianchiris! 🎉   Very good to have helm support for NFD, finally...
/approve

marquiz (approved):

ArangoGutierrez (approved):
tested on k 1.20+ 
/lgtm

e0ne:
@marquiz @ArangoGutierrez thanks for the guiding me to make this merge happen.

adrianchiris:
Closes: #422

Issue #427, opened 2021-02-01
Two different instances NFD should not override labels


e0ne:
**What would you like to be added**: NDF should be able to work in a concurrency with several NFD instances and don't override labels

**Why is this needed**: In a complex deployments we can have several instances of NFD deployed and configured by different operators or k8s cluster is used for multiple purposes

e0ne:
Possible solutions how to implement this feature:

1. Add `--label-ignorelist` which will work in pretty same way as `--label-whitelist` but just will read labels from node and keep them in the list.
2. Add new `labels` source where we can specify which labels we should leave 'as is'
3. nfd-worker should store labels list somewhere in k8s(e.g. shared ConfigMap between several instances) and nfd-master will read them before applying to the node.

marquiz:
This might be a bit more involved than it seems at first. I think we'd want to maintain the relatively simple stateless operation of NFD. NFD-Worker is a fully stateless daemon with no connection to the K8s control plane which seems to rule out # 3 of the possible solutions above. # 2 I don't fully understand but looks like similarly confrontational. The problem with # 1 is that we store the some metadata in node annotations. E.g. the names of nfd-managed labels in `nfd.node.kubernetes.io/feature-labels`. One way to extend this could be e.g. something like `--instance` flag in nfd-master which would change the annotation namespace/prefix to `<instance>.nfd.node.kubernetes.io`. Do you really need to filter out the labels on the nfd-master side? I think it would be more versatile and maintainable to configure it on the worker side of each parallel deployment. I think the command line flags are awkward to use and e.g. `--label-whitelist` is a legacy flag that I'm thinking we might want to deprecate.

What is the main restriction why you couldn't use just one deployment of NFD? Related to configuration handling or something different? I'm just thinking that what would be the other, possibly simpler, ways to implement your use case.

e0ne:
@marquizWe can't filter labels on the worker side because each instance has only a list of own labels.

I agree it's better do not use CLI args in k8s but it's out of the scope of this patch.

We've got several NFD instances controlled by different operators. We don't want to mix them together because they're responsible for different areas

moshe010:
In our case with have Network Operator which provide nfd to label network stuff (nfd Network config)  and GPU operator which provide nfd lablel for GPU stuff (nfd GPU config). both can work together or standalone. We don't wont to force dependency. Some provide their own nfd hook as well so the other is not aware of it.

marquiz:
[...]

PR #428, opened 2021-02-01
[WIP] Add '--label-ignorelist' param

e0ne:
'--label-ignorelist' param is used when you need to run several
NFD in the one Kubernetes cluster, so NFD will not override labels
created by another instance.

Close: issue #427


e0ne:
Pull request #431 fixes the issue, so we don't need this anymore

PR #429, opened 2021-02-03
Added nodename rule to custom source

slintes:
# What
With this PR we add a new `nodename` rule for the `custom` source for the NFD worker, which will apply features based on  nodes' hostnames.

For making more dynamic configuration by potentially multiple parties easier, `custom` source rules can also be read from the `/etc/kubernetes/node-feature-discovery/custom.d` directory now (probabaly mostly populated by mounted ConfigMaps).

Also an optional `value` field was added, which allows overruling the default `true` value for matching features if set.

# Details

## Example
Consider this config:
```
- name: "feature.by.nodename"
  # default value is true
  matchOn:
    - nodename:
      - worker-0
      - worker-1
- name: "another.features.by.nodename.regex"
  value: customValue
  matchOn:
    - nodename:
      - wor.*-0
```

The config will apply the `feature.by.nodename=true` feature to the `worker-0` and `worker-1` nodes, and the `another.features.by.nodename.regex=customValue` feature to all nodes matching the `wor.*-0` regex.

## Implementation
- the mounting of additional ConfigMaps into the NFD worker pod is a bit cumbersome, but can be automated with the NFD operator in a follow up (e.g. mount all ConfigMaps in the NFD namespace which match a certain name pattern).

# Usecase
A customer with many nodes is able to set the nodename according the node's purpose during node deployment, but not to add the needed labels for workload scheduling. The nodename is the only metadata available for determining the node's purpose.
With this new source the customer can put the desired nodename - label mapping(s) into a ConfigMap, and let NFD do the work of putting the labels on the nodes.

/cc @zvonkok @ArangoGutierrez @marquiz

### DIFF
--- a/docs/get-started/features.md
+++ b/docs/get-started/features.md
@@ -134,6 +134,17 @@ examples how to set-up and manage the worker configuration.
 To aid in making Custom Features clearer, we define a general and a per rule
 nomenclature, keeping things as consistent as possible.
 
+#### Additional configuration directory
+
+Additionally to the rules defined in the nfd-worker configuration file, the
+Custom feature can read more configuration files located in the
+`/etc/kubernetes/node-feature-discovery/custom.d/` directory. This makes more
+dynamic and flexible configuration easier. This directory must be available
+inside the NFD worker container, so Volumes and VolumeMounts must be used for
+mounting e.g. ConfigMap(s). The example deployment manifests provide an example
+(commented out) for providing Custom configuration with an additional
+ConfigMap, mounted into the `custom.d` directory.
+
 #### General Nomenclature & Definitions
 
 ```
@@ -151,6 +162,7 @@ file.
 sources:
   custom:
   - name: <feature name>
+    value: <optional feature value, defaults to "true">
     matchOn:
     - <Rule-1>: <Rule-1 Input>
       [<Rule-2>: <Rule-2 Input>]
@@ -291,6 +303,26 @@ Matching is done by performing logical _AND_ for each provided Element, i.e the
 Rule will match if all provided Elements (kernel config options) are enabled
 (`y` or `m`) or matching `=<value>` in the kernel.
 
+##### Nodename Rule
+
+###### Nomenclature
+
+```
+Element     :A nodename regexp pattern
+```
+
+The Rule allows matching the node's name against a provided list of Elements.
+
+###### Format
+
+```yaml
+nodename: [ <nodename regexp pattern>, ... ]
+```
+
+Matching is done by performing logical _OR_ for each provided Element, i.e the
+Rule will match if one of the provided Elements (nodename regexp pattern)
+matches the node's name.
+
 #### Example
 
 ```yaml
@@ -328,6 +360,10 @@ custom:
     matchOn:
       - kConfig: ["GCC_VERSION=100101"]
         loadedKMod: ["kmod1"]
+  - name: "my.datacenter"
+    value: "datacenter-1"
+    matchOn:
+      - nodename: [ "node-datacenter1-rack.*-server.*" ]
 ```
 
 __In the example above:__
@@ -360,6 +396,10 @@ __In the example above:__
   `feature.node.kubernetes.io/custom-my.kernel.modulecompiler=true` if the
   in-tree `kmod1` kernel module is loaded __AND__ it's built with
   `GCC_VERSION=100101`.
+- A node would contain the label:
+  `feature.node.kubernetes.io/my.datacenter=datacenter-1` if the node's name
+  matches the `node-datacenter1-rack.*-server.*` pattern, e.g.
+  `node-datacenter1-rack2-server42`
 
 #### Statically defined features
 
--- a/nfd-daemonset-combined.yaml.template
+++ b/nfd-daemonset-combined.yaml.template
@@ -115,6 +115,11 @@ spec:
             - name: nfd-worker-conf
               mountPath: "/etc/kubernetes/node-feature-discovery"
               readOnly: true
+## Example for more custom configs in an additional configmap (1/3)
+## Mounting into subdirectories of custom.d makes it easy to use multiple configmaps
+#            - name: custom-source-extra-rules
+#              mountPath: "/etc/kubernetes/node-feature-discovery/custom.d/extra-rules-1"
+#              readOnly: true
       volumes:
         - name: host-boot
           hostPath:
@@ -134,6 +139,10 @@ spec:
         - name: nfd-worker-conf
           configMap:
             name: nfd-worker-conf
+## Example for more custom configs in an additional configmap (2/3)
+#        - name: custom-source-extra-rules
+#          configMap:
+#            name: custom-source-extra-rules
 ---
 apiVersion: v1
 kind: ConfigMap
@@ -232,4 +241,25 @@ data:
     #            vendor: ["15b3"]
     #            device: ["1014", "1017"]
     #          loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
+    #    - name: "feature.by.nodename"
+    #      value: customValue
+    #      matchOn:
+    #        - nodename: ["worker-0", "my-.*-node"]
 ### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+---
+## Example for more custom configs in an additional configmap (3/3)
+#apiVersion: v1
+#kind: ConfigMap
+#metadata:
+#  name: custom-source-extra-rules
+#  namespace: node-feature-discovery
+#data:
+## Filename doesn't matter, and there can be multiple. They just need to be unique.
+#  custom.conf: |
+#    - name: "more.kernel.features"
+#      matchOn:
+#      - loadedKMod: ["example_kmod3"]
+#    - name: "more.features.by.nodename"
+#      value: customValue
+#      matchOn:
+#      - nodename: ["special-.*-node-.*"]
--- a/nfd-worker-daemonset.yaml.template
+++ b/nfd-worker-daemonset.yaml.template
@@ -68,6 +68,11 @@ spec:
             - name: nfd-worker-conf
               mountPath: "/etc/kubernetes/node-feature-discovery"
               readOnly: true
+## Example for more custom configs in an additional configmap (1/3)
+## Mounting into subdirectories of custom.d makes it easy to use multiple configmaps
+#            - name: custom-source-extra-rules
+#              mountPath: "/etc/kubernetes/node-feature-discovery/custom.d/extra-rules-1"
+#              readOnly: true
 ## Enable TLS authentication (2/3)
 #            - name: nfd-ca-cert
 #              mountPath: "/etc/kubernetes/node-feature-discovery/trust"
@@ -94,6 +99,10 @@ spec:
         - name: nfd-worker-conf
           configMap:
             name: nfd-worker-conf
+## Example for more custom configs in an additional configmap (2/3)
+#        - name: custom-source-extra-rules
+#          configMap:
+#            name: custom-source-extra-rules
 ## Enable TLS authentication (3/3)
 #        - name: nfd-ca-cert
 #          configMap:
@@ -199,4 +208,25 @@ data:
     #            vendor: ["15b3"]
     #            device: ["1014", "1017"]
     #          loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
+    #    - name: "feature.by.nodename"
+    #      value: customValue
+    #      matchOn:
+    #        - nodename: ["worker-0", "my-.*-node"]
 ### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+---
+## Example for more custom configs in an additional configmap (3/3)
+#apiVersion: v1
+#kind: ConfigMap
+#metadata:
+#  name: custom-source-extra-rules
+#  namespace: node-feature-discovery
+#data:
+## Filename doesn't matter, and there can be multiple. They just need to be unique.
+#  custom.conf: |
+#    - name: "more.kernel.features"
+#      matchOn:
+#      - loadedKMod: ["example_kmod3"]
+#    - name: "more.features.by.nodename"
+#      value: customValue
+#      matchOn:
+#      - nodename: ["special-.*-node-.*"]
--- a/nfd-worker-job.yaml.template
+++ b/nfd-worker-job.yaml.template
@@ -72,6 +72,11 @@ spec:
             - name: nfd-worker-conf
               mountPath: "/etc/kubernetes/node-feature-discovery"
               readOnly: true
+## Example for more custom configs in an additional configmap (1/3)
+## Mounting into subdirectories of custom.d makes it easy to use multiple configmaps
+#            - name: custom-source-extra-rules
+#              mountPath: "/etc/kubernetes/node-feature-discovery/custom.d/extra-rules-1"
+#              readOnly: true
 ## Enable TLS authentication (2/3)
 #            - name: nfd-ca-cert
 #              mountPath: "/etc/kubernetes/node-feature-discovery/trust"
@@ -99,6 +104,10 @@ spec:
         - name: nfd-worker-conf
           configMap:
             name: nfd-worker-conf
+## Example for more custom configs in an additional configmap (2/3)
+#        - name: custom-source-extra-rules
+#          configMap:
+#            name: custom-source-extra-rules
 ## Enable TLS authentication (3/3)
 #        - name: nfd-ca-cert
 #          configMap:
@@ -204,4 +213,25 @@ data:
     #            vendor: ["15b3"]
     #            device: ["1014", "1017"]
     #          loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
+    #    - name: "feature.by.nodename"
+    #      value: customValue
+    #      matchOn:
+    #        - nodename: ["worker-0", "my-.*-node"]
 ### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+---
+## Example for more custom configs in an additional configmap (3/3)
+#apiVersion: v1
+#kind: ConfigMap
+#metadata:
+#  name: custom-source-extra-rules
+#  namespace: node-feature-discovery
+#data:
+## Filename doesn't matter, and there can be multiple. They just need to be unique.
+#  custom.conf: |
+#    - name: "more.kernel.features"
+#      matchOn:
+#      - loadedKMod: ["example_kmod3"]
+#    - name: "more.features.by.nodename"
+#      value: customValue
+#      matchOn:
+#      - nodename: ["special-.*-node-.*"]
--- a/source/custom/custom.go
+++ b/source/custom/custom.go
@@ -18,6 +18,7 @@ package custom
 
 import (
 	"log"
+	"reflect"
 
 	"sigs.k8s.io/node-feature-discovery/source"
 	"sigs.k8s.io/node-feature-discovery/source/custom/rules"
@@ -30,10 +31,12 @@ type MatchRule struct {
 	LoadedKMod *rules.LoadedKModRule `json:"loadedKMod,omitempty"`
 	CpuID      *rules.CpuIDRule      `json:"cpuId,omitempty"`
 	Kconfig    *rules.KconfigRule    `json:"kConfig,omitempty"`
+	Nodename   *rules.NodenameRule   `json:"nodename,omitempty"`
 }
 
 type FeatureSpec struct {
 	Name    string      `json:"name"`
+	Value   *string     `json:"value"`
 	MatchOn []MatchRule `json:"matchOn"`
 }
 
@@ -72,6 +75,7 @@ func (s *Source) SetConfig(conf source.Config) {
 func (s Source) Discover() (source.Features, error) {
 	features := source.Features{}
 	allFeatureConfig := append(getStaticFeatureConfig(), *s.config...)
+	allFeatureConfig = append(allFeatureConfig, getDirectoryFeatureConfig()...)
 	log.Printf("INFO: Custom features: %+v", allFeatureConfig)
 	// Iterate over features
 	for _, customFeature := range allFeatureConfig {
@@ -81,7 +85,11 @@ func (s Source) Discover() (source.Features, error) {
 			continue
 		}
 		if featureExist {
-			features[customFeature.Name] = true
+			var value interface{} = true
+			if customFeature.Value != nil {
+				value = *customFeature.Value
+			}
+			features[customFeature.Name] = value
 		}
 	}
 	return features, nil
@@ -90,58 +98,37 @@ func (s Source) Discover() (source.Features, error) {
 // Process a single feature by Matching on the defined rules.
 // A feature is present if all defined Rules in a MatchRule return a match.
 func (s Source) discoverFeature(feature FeatureSpec) (bool, error) {
-	for _, rule := range feature.MatchOn {
-		// PCI ID rule
-		if rule.PciID != nil {
-			match, err := rule.PciID.Match()
-			if err != nil {
-				return false, err
-			}
-			if !match {
-				continue
-			}
+	for _, matchRules := range feature.MatchOn {
+
+		allRules := []rules.Rule{
+			matchRules.PciID,
+			matchRules.UsbID,
+			matchRules.LoadedKMod,
+			matchRules.CpuID,
+			matchRules.Kconfig,
+			matchRules.Nodename,
 		}
-		// USB ID rule
-		if rule.UsbID != nil {
-			match, err := rule.UsbID.Match()
-			if err != nil {
-				return false, err
+
+		// return true, nil if all rules match
+		matchRules := func(rules []rules.Rule) (bool, error) {
+			for _, rule := range rules {
+				if reflect.ValueOf(rule).IsNil() {
+					continue
+				}
+				if match, err := rule.Match(); err != nil {
+					return false, err
+				} else if !match {
+					return false, nil
+				}
 			}
-			if !match {
-				continue
-			}
+			return true, nil
 		}
-		// Loaded kernel module rule
-		if rule.LoadedKMod != nil {
-			match, err := rule.LoadedKMod.Match()
-			if err != nil {
-				return false, err
-			}
-			if !match {
-				continue
-			}
+
+		if match, err := matchRules(allRules); err != nil {
+			return false, err
+		} else if match {
+			return true, nil
 		}
-		// cpuid rule
-		if rule.CpuID != nil {
-			match, err := rule.CpuID.Match()
-			if err != nil {
-				return false, err
-			}
-			if !match {
-				continue
-			}
-		}
-		// kconfig rule
-		if rule.Kconfig != nil {
-			match, err := rule.Kconfig.Match()
-			if err != nil {
-				return false, err
-			}
-			if !match {
-				continue
-			}
-		}
-		return true, nil
 	}
 	return false, nil
 }
--- a/source/custom/directory_features.go
+++ b/source/custom/directory_features.go
@@ -1,0 +1,88 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package custom
+
+import (
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"sigs.k8s.io/yaml"
+)
+
+const Directory = "/etc/kubernetes/node-feature-discovery/custom.d"
+
+// getDirectoryFeatureConfig returns features configured in the "/etc/kubernetes/node-feature-discovery/custom.d"
+// host directory and its 1st level subdirectories, which can be populated e.g. by ConfigMaps
+func getDirectoryFeatureConfig() []FeatureSpec {
+	features := readDir(Directory, true)
+	//log.Printf("DEBUG: all configmap based custom feature specs: %+v", features)
+	return features
+}
+
+func readDir(dirName string, recursive bool) []FeatureSpec {
+	features := make([]FeatureSpec, 0)
+
+	log.Printf("DEBUG: getting files in %s", dirName)
+	files, err := ioutil.ReadDir(dirName)
+	if err != nil {
+		if os.IsNotExist(err) {
+			log.Printf("DEBUG: custom config directory %q does not exist", dirName)
+		} else {
+			log.Printf("ERROR: unable to access custom config directory %q, %v", dirName, err)
+		}
+		return features
+	}
+
+	for _, file := range files {
+		fileName := filepath.Join(dirName, file.Name())
+
+		if file.IsDir() {
+			if recursive {
+				//log.Printf("DEBUG: going into dir %q", fileName)
+				features = append(features, readDir(fileName, false)...)
+				//} else {
+				//	log.Printf("DEBUG: skipping dir %q", fileName)
+			}
+			continue
+		}
+		if strings.HasPrefix(file.Name(), ".") {
+			//log.Printf("DEBUG: skipping hidden file %q", fileName)
+			continue
+		}
+		//log.Printf("DEBUG: processing file %q", fileName)
+
+		bytes, err := ioutil.ReadFile(fileName)
+		if err != nil {
+			log.Printf("ERROR: could not read custom config file %q, %v", fileName, err)
+			continue
+		}
+		//log.Printf("DEBUG: custom config rules raw: %s", string(bytes))
+
+		config := &[]FeatureSpec{}
+		err = yaml.UnmarshalStrict(bytes, config)
+		if err != nil {
+			log.Printf("ERROR: could not parse custom config file %q, %v", fileName, err)
+			continue
+		}
+
+		features = append(features, *config...)
+	}
+	return features
+}
--- a/source/custom/rules/nodename_rule.go
+++ b/source/custom/rules/nodename_rule.go
@@ -1,0 +1,50 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package rules
+
+import (
+	"log"
+	"os"
+	"regexp"
+)
+
+var (
+	nodeName = os.Getenv("NODE_NAME")
+)
+
+// Rule that matches on nodenames configured in a ConfigMap
+type NodenameRule []string
+
+// Force implementation of Rule
+var _ Rule = NodenameRule{}
+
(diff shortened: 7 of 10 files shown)


marquiz:
Thanks @slintes for the patchset!

I see the use case. What I immediately started to think is that would it be possible to extend the existing `custom` source, instead of adding a new one(?) Is there need for multiple configmaps or could the single nfd-worker configmap instance be used instead? If multiple configs would be required we could enable a config dir something like `/etc/kubernetes/node-feature-discovery/custom.d` (similar to `source.d` and `features.d`, currently). Currently, the limitation with the custom feature rules is that there can only be one label per rule but I think that could be extended with a `labels` field (similar to what you have in this PR). WDYT?

ArangoGutierrez (changes requested):
Patch is looking good, I need to run a couple of tests before digging deeper.
At a first glance, this patch requires unit testing for the big chunks of GO code being added, could you please include at least some minor unit tests for this new features, thanks

marquiz:
> Patch is looking good, I need to run a couple of tests before digging deeper.
> At a first glance, this patch requires unit testing for the big chunks of GO code being added, could you please include at least some minor unit tests for this new features, thanks

I really want to explore extending the existing source(s) (`custom`) before going forward with this as is

slintes:
@marquiz Thanks for the feedback, and I understand it is preferable to extend an existing source.

The hostname - label matching rules are potentially very dynamic, and there might even be multiple parties which want to add or modify the rules (the exact requirements are vague unfortunately, that's why I went with a very flexible implementation).
Having to modify the main worker config for this would need some careful coordination. Having dedicated ConfigMaps for each party simplifies configuration from the user's POV a lot in this scenario IMHO.

For having multiple ConfigMaps you proposed a `/etc/kubernetes/node-feature-discovery/custom.d` dir... do I understand correctly that
- those additional ConfigMaps should be mounted into that directory
- the custom source will use them additionally to the existing config (similar to how it's implemented in the my current approach)
- but in contrast to the current impl those ConfigMaps will contain the rules in the same format like in the existing worker config?

I think multiple labels per rule are not really needed... it was just easy to implement. But I don't see an actual usecase. If you actually need multiple labels, it will be very few only, which can still be configured with multiple rules. Not worth a modification of the current config structure IMHO.

@ArangoGutierrez yes, I will add unit tests as soon as we have an agreement to the way forward. I expected the e2e test to run on the PR as well, that's why I focused on that one in the 1st iteration.

marquiz:
> The hostname - label matching rules are potentially very dynamic, and there might even be multiple parties which want to add or modify the rules (the exact requirements are vague unfortunately, that's why I went with a very flexible implementation).

Yes, I understand this, and makes sense. That's why I proposed to extend the custom source to support multiple config files.

> For having multiple ConfigMaps you proposed a `/etc/kubernetes/node-feature-discovery/custom.d` dir... do I understand correctly that
> 
> * those additional ConfigMaps should be mounted into that directory

Yes, that's right.

> * the custom source will use them additionally to the existing config (similar to how it's implemented in the my current approach)

Yes, rules could be defined in the nfd-worker's main config file like we already have. In addition to that, additional rules could be specified by adding rule files under the `custom.d`.

> * but in contrast to the current impl those ConfigMaps will contain the rules in the same format like in the existing worker config?

Yes, the format would be similar. That would make it possible to define any rules there and for example combine hostname/nodename matching with other rules. E.g.
```yaml
- name: "my.host.feature"
  matchOn:
    - hostName: ["host-1", "host-2"]
    - nodeName: ["host-1", "host-2"]
- name: "my.kernel.feature"
  matchOn:
    - loadedKMod: ["kmod-1"]
- name: "my.combo.feature"
  matchOn:
    - hostName: ["host-1", "host-2"]
    - pciId:
        vendor: ["8086"]
        device: ["1234"]
```
I think the `hostName`/`nodeName` could be a treated as a regexp to make matching super flexible.

> I think multiple labels per rule are not really needed... it was just easy to implement.

OK, even better. And, I think that could be added later, then, if there's a need for that. We might want to add a `value` field for the custom config, though (IIRC, we don't we have that at the moment). All labels just get value `true`.

slintes:
> I think the hostName/nodeName could be a treated as a regexp to make matching super flexible.

Yes, that's a requirement actually :)

> We might want to add a value field ... All labels just get value true

oh, good point! Didn't notice that yet.

slintes:
- refactored to custom source
- added possibility to set an optional label value, defaults to the old `true` value
- rebased
- unit test and doc update pending
- e2e test succeeds in a local test

slintes:
forgot this:

- added possibility to read more custom source configs from additional ConfigMaps
- updated PR description to reflect the changes

:)

marquiz:
Really cool, thanks @slintes ! I'll take a detailed look as soon as I have time. Trying to do that on Friday but I'm afraid tomorrow is lost to other things

source/custom/custom.go
```diff
@@ -141,6 +148,16 @@ func (s Source) discoverFeature(feature FeatureSpec) (bool, error) {
 				continue
 			}
 		}
+		// hostname rule
```
adrianchiris, line 151: not related to this commit, but we could possibly do something nifty with reflect package to avoid this repetitive code.

source/custom/rules/hostname.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	"log"
+	"os"
+	"regexp"
+)
+
+var (
+	// TODO is it safe to assume nodeName == hostname??
+	nodeName = os.Getenv("NODE_NAME")
```
adrianchiris, line 26: any reason not to use `/etc/hostname` ?

`NODE_NAME` is set as the k8s node object name, im sure there is some kubelet flag that allows to advertise the node under a different name somewhere.

from kubelet --help:
`--hostname-override string                                 If non-empty, will use this string as identification instead of the actual hostname.`

If these are indeed non synonymous maybe the Rule should be NodeNameRule ?

nfd-daemonset-combined.yaml.template
```diff
@@ -227,4 +227,10 @@ data:
     #            vendor: ["15b3"]
     #            device: ["1014", "1017"]
     #          loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
-### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+    #    - name: "feature-by-hostname"
```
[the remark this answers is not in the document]
slintes: done at all places, thanks for the heads up

nfd-worker-daemonset.yaml.template
```diff
@@ -187,4 +195,31 @@ data:
...
+#  - name: "more-features-by-hostname"
+#    # default value is true
+#    matchOn:
+#    - hostname:
+#      - worker-0
+#      - worker-1
+#  - name: "more-features-by-hostname-regex"
+#    value: customValue
+#    matchOn:
+#    - hostname:
```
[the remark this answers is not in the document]
slintes: done

nfd-worker-job.yaml.template
```diff
@@ -197,4 +197,10 @@ data:
...
     #            device: ["1014", "1017"]
     #          loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
-### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+    #    - name: "feature-by-hostname"
+    #      value: customValue
+    #      matchOn:
+    #        - hostname:
+    #          - worker-0
+    #          - wo.*r-1
+### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
```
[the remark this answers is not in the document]
slintes: fixed

source/custom/configmap_features.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+
+		bytes, err := ioutil.ReadFile(configFile)
+		if err != nil {
+			log.Printf("ERROR: could not read custom config file %s,\n%v\n", configFile, err)
+			continue
+		}
+		log.Printf("DEBUG: custom config rules raw: %s\n", string(bytes))
+
+		config := &config{}
```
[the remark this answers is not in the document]
slintes: fair point, done

source/custom/custom.go
slintes: I gave it a try. It works, but the code is a bit ugly IMHO. But that might be personal preference. You can see it at [1].
In this PR I went with a different approach for code de-deduplication, see last commit. It still needs one line with reflection, I didn't find another to filter out the nil values, but overall its better readable. With the downside of still needing to touch it for new rules though.

WDYT?

[1] https://github.com/slintes/node-feature-discovery/commit/ad705f9e14d71508d367ba7cc83a13e0e198f0dd#diff-1b283f9d332ba27ed8b09eb6642c7bf9c9d552d45f89d485c590c7c1619f6b7dR101

source/custom/rules/hostname.go
slintes: Very good point, thanks! I need to clarify this.

nfd-worker-daemonset.yaml.template
```diff
@@ -187,4 +195,25 @@ data:
     #            vendor: ["15b3"]
     #            device: ["1014", "1017"]
     #          loadedKMod : ["vendor_kmod1", "vendor_kmod2"]
+    #    - name: "feature.by.hostname"
+    #      value: customValue
+    #      matchOn:
+    #        - hostname: ["worker-0", "my-.*-node"]
 ### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+---
+## Example for more custom configs in an additional configmap (3/3)
```
ArangoGutierrez, line 204: Could this be a stand alone file, sameas the `nfd-worker-conf.example` file?

ArangoGutierrez (changes requested):
When generating the yamls `make yamls` only the example `nfd-worker-daemonset.yaml.template ` get's populated with the correct mounts and extra ConfigMap , do we want this, or we want all generated examples to carry around empty custom-conf files plus commented lines on how to mount this?

nfd-worker-daemonset.yaml.template
```diff
@@ -61,6 +61,10 @@ spec:
             - name: nfd-worker-conf
               mountPath: "/etc/kubernetes/node-feature-discovery"
               readOnly: true
+## Example for more custom configs in an additional configmap (1/3)
+#            - name: more-custom-configs
+#              mountPath: "/etc/kubernetes/node-feature-discovery/custom.d/more-custom-configs"
```
ArangoGutierrez, line 66:
```suggestion
#              mountPath: "/etc/kubernetes/node-feature-discovery/custom.d"
```
just like this, kubernetes will mount the configMap data at `/etc/kubernetes/node-feature-discovery/custom.d/custom.conf`

nfd-worker-daemonset.yaml.template
```diff
@@ -187,4 +195,25 @@ data:
...
+## Example for more custom configs in an additional configmap (3/3)
+#apiVersion: v1
+#kind: ConfigMap
+#metadata:
+#  name: more-custom-configs
+#  namespace: node-feature-discovery
+#data:
+## Filename doesn't matter, and there can be multiple. They just need to be unique across all additional ConfigMaps
+#  custom.conf: |
+#  - name: "more.kernel.features"
```
ArangoGutierrez, line 213: just a yaml lint here, line 213 and below need to be moved to the right 2 spaces.
when moved to a single file, you get `error: error parsing custom-configs.yaml: error converting YAML to JSON: yaml: line 9: did not find expected key` meaning everything from line 9 is miss indented

ArangoGutierrez (changes requested):
Looking good,
```bash
2021/02/16 15:37:18 DEBUG: processing file /etc/kubernetes/node-feature-discovery/custom.d/..data
2021/02/16 15:37:18 DEBUG: skipping hidden file /etc/kubernetes/node-feature-discovery/custom.d/..data
2021/02/16 15:37:18 DEBUG: processing file /etc/kubernetes/node-feature-discovery/custom.d/custom.conf
2021/02/16 15:37:18 DEBUG: custom config rules raw: - name: "more.kernel.features"
  matchOn:
  - loadedKMod: ["example_kmod3"]
- name: "more.features.by.hostname"
  matchOn:
  - hostname: ["ip*"]
 ```
then
```bash
[eduardo@fedora-ws node-feature-discovery-operator]$ oc get no -o json | jq .items[].metadata.labels |grep  custom-more
  "feature.node.kubernetes.io/custom-more.features.by.hostname": "true",
  "feature.node.kubernetes.io/custom-more.features.by.hostname": "true",
```
A small request, could you add some documentation on how to get here, I had to manually create a new configMap, deploy it into the nameSpace I had NFD
```yaml
---
apiVersion: v1
kind: ConfigMap
metadata:
  name: more-custom-configs
  namespace: openshift-nfd
data:
# Filename doesn't matter, and there can be multiple. They just need to be unique across all additional ConfigMaps
  custom.conf: |
    - name: "more.kernel.features"
      matchOn:
      - loadedKMod: ["example_kmod3"]
    - name: "more.features.by.hostname"
      matchOn:
      - hostname: ["ip*"]
```
then un-comment lines x,y,z from the examples. That should be enough for users trying to leverage this new feature

source/custom/custom.go
adrianchiris: > I gave it a try. It works, but the code is a bit ugly IMHO

Doesnt seem too bad to me, but your current change is OK if you feel it is better.

marquiz:
Thanks for the reviews @adrianchiris and @ArangoGutierrez ! I need to call it a day, but, I'll take a look tomorrow

source/custom/custom.go
marquiz: I think the refactor in the last patch of this PR is good. I'm not a big fan of reflect but I guess there's no nice and easy way around it here.

marquiz (changes requested):
Thanks @slintes for the re-implementation!

Looks really good, overall. Adding the e2e test is really great 👍 

Just a few comments and nits. I also mentioned this in the code comments but I think it'd be better to rename the configMap* terminology to configDir* or smth (you may come up with a better name). This would also mean, renaming `configmap_features.go`

As a side note, there is a lot of debug logging which is basically good, I think. However, our logging is very rudimentary and probably need to address that in a separate PR, adding some control of logger verbosity/filtering.

nfd-worker-daemonset.yaml.template
slintes: oh yes, thanks, in the e2e test I do it correct already :grin:

source/custom/configmap_features.go
```diff
@@ -0,0 +1,81 @@
...
+	"strings"
+
+	"sigs.k8s.io/yaml"
+)
+
+const ConfigMapMountDir = "/etc/kubernetes/node-feature-discovery/custom.d"
+
+// getConfigMapFeatureConfig returns features configured in additional ConfigMaps, which are mounted to
+// /etc/kubernetes/node-feature-discovery/custom.d
+func getConfigMapFeatureConfig() []FeatureSpec {
```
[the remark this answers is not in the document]
slintes: I will go with `getDirectoryFeatureConfig` to be consistent with `getStaticFeatureConfig`

source/custom/configmap_features.go
```diff
@@ -0,0 +1,81 @@
...
+
+	features := make([]FeatureSpec, 0)
+
+	log.Printf("DEBUG: getting files in %s\n", ConfigMapMountDir)
+	files, err := ioutil.ReadDir(ConfigMapMountDir)
+	if err != nil {
+		if os.IsNotExist(err) {
+			log.Printf("DEBUG: hostname config directory %v does not exist\n", ConfigMapMountDir)
+		} else {
+			log.Printf("ERROR: unable to access hostname config directory %v\n", ConfigMapMountDir)
```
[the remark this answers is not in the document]
slintes: :+1:  (and I need to get used to use %q for strings, lazy me almost always uses %v for everything :grin: )

source/custom/rules/hostname.go
slintes: after clarifying I switch to nodename everywhere :)

source/custom/custom.go
slintes: I'll keep the current version version then :slightly_smiling_face:

source/custom/rules/hostname.go
```diff
@@ -0,0 +1,54 @@
...
+var _ Rule = HostnameRule{}
+
+func (h HostnameRule) Match() (bool, error) {
+
+	for _, hostnamePattern := range h {
+		log.Printf("DEBUG: matchHostname %s\n", hostnamePattern)
+
+		match, err := regexp.MatchString(hostnamePattern, nodeName)
+		if err != nil {
+			log.Printf("ERROR: error testing regex: pattern %s, nodename %s, %v", hostnamePattern, nodeName, err)
```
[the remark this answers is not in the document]
slintes: makes sense :+1:

slintes:
Thanks for the reviews and feedback all :+1: 
Most things are addressed:
- Renamed hostname -> nodename at all places
- Use more generic name for the additional configs (ConfigMap -> Directory)
- Improved and less logging (I commented some debug logs for reducing amount of logs, can be re-enabled when switching to a logging framework with configurable verbosity... but that's out of scope of this PR :slightly_smiling_face: )
- and some smaller things

About the manifests: should I add the ConfigMap and its mounts to all manifests? I'd prefer to mention the daemonset in documentation as example instead of copying more yaml snippets around :thinking:  

Talking about doc: that's the next item on my list...

slintes:
Updated PR title and description to reflect recent changes.
And updated doc :slightly_smiling_face:

ArangoGutierrez:
meh, ok @slintes a simple rebase will fix the CI after your PR update

slintes:
rebased, and squashed commits

ArangoGutierrez:
I think during the squash event you removed the tests

docs/get-started/features.md
```diff
@@ -124,15 +124,25 @@ See the full list in [github.com/klauspost/cpuid][klauspost-cpuid].
...
+#### Additional configuration directory
+
+Additionally to the rules defined in the nfd-worker configuration file, the Custom feature can read more configuration
+files located in the `/etc/kubernetes/node-feature-discovery/custom.d/` directory. This makes more dynamic and flexible
+configuration easier. This directory must be available inside the NFD worker container, so Volumes and VolumeMounts must
+be used for mounting e.g. ConfigMap(s). The
+[nfd-worker deployment template](https://github.com/kubernetes-sigs/node-feature-discovery/blob/{{ site.release
+}}/nfd-worker-daemonset.yaml.template#L204)
+provides an example for providing Custom configuration with an additional ConfigMap, mounted into the `custom.d`
+directory.
```
marquiz, line 145: nit: could you wrap the lines to 80 chars (also elsewhere in this file)? Markdown linter (`mdl`) bugs about these. I haven't integrated that to CI because we don't fully pass muster

nfd-worker-daemonset.yaml.template
```diff
@@ -68,6 +68,10 @@ spec:
             - name: nfd-worker-conf
               mountPath: "/etc/kubernetes/node-feature-discovery"
               readOnly: true
+## Example for more custom configs in an additional configmap (1/3)
+#            - name: more-custom-configs
+#              mountPath: "/etc/kubernetes/node-feature-discovery/custom.d/more-custom-configs"
+#              readOnly: true
```
marquiz, line 74: For consistency, add this to the "combined" and "job" templates, too

nfd-worker-daemonset.yaml.template
```diff
@@ -94,6 +98,10 @@ spec:
         - name: nfd-worker-conf
           configMap:
             name: nfd-worker-conf
+## Example for more custom configs in an additional configmap (2/3)
+#        - name: more-custom-configs
+#          configMap:
+#            name: more-custom-configs
```
marquiz, line 104: For consistency, add this to the "combined" and "job" templates, too

nfd-worker-daemonset.yaml.template
marquiz: Mm, why not. At least, for consistency, this example configmap should be added to the "combined" and "job" templates, too

nfd-worker-daemonset.yaml.template
```diff
@@ -199,4 +207,25 @@ data:
...
+    #      value: customValue
+    #      matchOn:
+    #        - nodename: ["worker-0", "my-.*-node"]
 ### <NFD-WORKER-CONF-END-DO-NOT-REMOVE>
+---
+## Example for more custom configs in an additional configmap (3/3)
+#apiVersion: v1
+#kind: ConfigMap
+#metadata:
+#  name: more-custom-configs
```
marquiz, line 220: nit: maybe it's my problem, but this is just somehow funny name 😄 . How about e.g. `custom-source-extra-config`, `custom-source-extra-rules` , `custom-extra-rule` or smth (you can probably come up with better name)?

marquiz (changes requested):
@slintes thank you for the refresh. Few small comments, nothing big.

One thing I'm a bit dubious are the now-commented-out log messages. But, I think that in the light of the much-needed (and upcoming) logging rework we can leave them there (and uncomment with the logger patch)

docs/get-started/features.md
slintes: sure, thx, done

nfd-worker-daemonset.yaml.template
slintes: done

slintes:
Thanks for review, all remarks are addressed.

> One thing I'm a bit dubious are the now-commented-out log messages. But, I think that in the light of the much-needed (and upcoming) logging rework we can leave them there (and uncomment with the logger patch)

yes, that's the idea

docs/get-started/features.md
```diff
@@ -328,6 +361,10 @@ custom:
     matchOn:
       - kConfig: ["GCC_VERSION=100101"]
         loadedKMod: ["kmod1"]
+  - name: "my.datacenter"
+    value: "datacenter-1"
+    matchOn:
+      - hostname: [ "node-datacenter1-rack.*-server.*" ]
```
adrianchiris, line 367: nodename ?

docs/get-started/features.md
```diff
@@ -291,6 +304,26 @@ Matching is done by performing logical _AND_ for each provided Element, i.e the
...
 
+##### Nodename Rule
+
+###### Nomenclature
+
+```
+Element     :A nodename regexp pattern
+```
+
+The Rule allows matching the node's name against a provided list of Elements.
```
adrianchiris, line 315: Should a note be added here clarifying that this matches on the node's name as registered in the kubernetes cluster ?
or do you think its clear they way it is now ?

nfd-daemonset-combined.yaml.template
```diff
@@ -232,4 +240,25 @@ data:
...
+#data:
+## Filename doesn't matter, and there can be multiple. They just need to be unique across all additional ConfigMaps
+#  custom.conf: |
+#    - name: "more.kernel.features"
+#      matchOn:
+#      - loadedKMod: ["example_kmod3"]
+#    - name: "more.features.by.nodename"
+#      value: customValue
+#      matchOn:
+#      - nodename: ["special-.*-node-.*"]
```
adrianchiris, line 264: nit: missing newline

nfd-worker-job.yaml.template
```diff
@@ -204,4 +212,25 @@ data:
...
+#data:
+## Filename doesn't matter, and there can be multiple. They just need to be unique across all additional ConfigMaps
+#  custom.conf: |
+#    - name: "more.kernel.features"
+#      matchOn:
+#      - loadedKMod: ["example_kmod3"]
+#    - name: "more.features.by.nodename"
+#      value: customValue
+#      matchOn:
+#      - nodename: ["special-.*-node-.*"]
```
adrianchiris, line 236: nit: missing newline

source/custom/custom.go
```diff
@@ -90,58 +98,35 @@ func (s Source) Discover() (source.Features, error) {
...
-		if rule.CpuID != nil {
-			match, err := rule.CpuID.Match()
-			if err != nil {
+			if match, err := rule.Match(); err != nil {
 				return false, err
-			}
-			if !match {
-				continue
+			} else if !match {
+				ruleFailed = true
```
adrianchiris, line 120: not sure if its favored but you could use `continue <label>` statement to continue to the next `matchRule` on the outer loop.

You would not need the `ruleFailed` bits in this case.
(just a suggestion here)

adrianchiris:
commit message contains some typos, can be fixed when merged or if an update is pushed to this PR

docs/get-started/features.md
```diff
@@ -134,6 +134,18 @@ examples how to set-up and manage the worker configuration.
...
 
+#### Additional configuration directory
+
+Additionally to the rules defined in the nfd-worker configuration file, the
+Custom feature can read morddde configuration files located in the
+`/etc/kubernetes/node-feature-discovery/custom.d/` directory. This makes more
+dynamic and flexible configuration easier. This directory must be available
+inside the NFD worker container, so Volumes and VolumeMounts must be used for
+mounting e.g. ConfigMap(s). The
+[nfd-worker deployment template](https://github.com/kubernetes-sigs/node-feature-discovery/blob/{{ site.release }}/nfd-worker-daemonset.yaml.template#L204)
```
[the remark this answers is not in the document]
slintes: Fair point. But I removed the link completely, because the example config is in all manifests now.

docs/get-started/features.md
slintes: good catch, thanks!

docs/get-started/features.md
slintes: IMHO node name is clear...?

nfd-daemonset-combined.yaml.template
slintes: fixed

slintes:
> Overall looks OK, some minor nits.

thanks! Pushed another new version :slightly_smiling_face:

docs/get-started/features.md
adrianchiris: ack

docs/get-started/features.md
```diff
@@ -134,6 +134,17 @@ examples how to set-up and manage the worker configuration.
 To aid in making Custom Features clearer, we define a general and a per rule
 nomenclature, keeping things as consistent as possible.
 
+#### Additional configuration directory
+
+Additionally to the rules defined in the nfd-worker configuration file, the
+Custom feature can read morddde configuration files located in the
```
adrianchiris, line 140: >  morddde
typo? must have missed it last round of review.

source/custom/custom.go
adrianchiris: yea no worries, im fine either way.

marquiz: > Personally I dislike those labels...

I'm not a big fan either, even if I've used them one or two times. You could achieve the above without labels e.g. by moving the inner loop into a separate function. But I think it's fine the way it is now and leave this to you to decide

docs/get-started/features.md
slintes: You didn't miss it, I messed it up somehow during reformat in the latest push. Thx and Fixed.

source/custom/directory_features.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		return features
+	}
+
+	for _, file := range files {
+		configFile := filepath.Join(Directory, file.Name())
+
+		if file.IsDir() {
+			//log.Printf("DEBUG: skipping dir %q", configFile)
+			continue
```
marquiz, line 53: Hmm, I started to think that maybe we should recurse into subdirectories as well. At least one level deep. This would make it much more user-friendly to drop in (mount) multiple config files under `custom.d/`. Each configmap would just be mounted in a separate subdir, that is.

I think with the current scheme it is only possible to use multiple configmaps with `subPath` mount which a bit cumbersome (IMO), and more importantly, prevents nfd-worker from getting confimap updates (i.e. the config inside the container is static even if the configmap(s) changed).

source/custom/custom.go
slintes: I like the idea of moving the inner loop into a func, will do that, thanks

source/custom/directory_features.go
slintes: Oh, good catch! I thought I can mount multiple CMs into the same path as long as the filenames in their data is unique, but that doesn't work. Will fix this and extend the e2e test to use 2 CMs.

slintes:
- refactored the nested loops in custom.go
- read config files from 1st level subdirs of custom.d and test multiple CMs in e2e test

For easier review commits are not squashed yet, happy to do so after review

source/custom/directory_features.go
marquiz: > I thought I can mount multiple CMs into the same path as long as the filenames in their data is unique, but that doesn't work.

Unfortunately that's not the case 😞  It's possible with the `subPath` directive, but, it has it's drawbacks as explained above.

> Will fix this and extend the e2e test to use 2 CMs.

That would be great 👍

marquiz (changes requested):
Nice work @slintes, and really fast response times(!) 😄  I came up with two more comments. We're really close, I think...

nfd-daemonset-combined.yaml.template
```diff
@@ -115,6 +115,11 @@ spec:
             - name: nfd-worker-conf
               mountPath: "/etc/kubernetes/node-feature-discovery"
               readOnly: true
+## Example for more custom configs in an additional configmap (1/3)
+## When using multiple configmaps, mount them into subdirectories of custom.d
+#            - name: custom-source-extra-rules
+#              mountPath: "/etc/kubernetes/node-feature-discovery/custom.d/"
+#              readOnly: true
```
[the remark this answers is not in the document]
slintes: works for me, done

test/e2e/node_feature_discovery.go
```diff
@@ -581,23 +590,183 @@ var _ = framework.KubeDescribe("[NFD] Node Feature Discovery", func() {
...
+
+				// We need a valid nodename for the configmap
+				nodeList, err := f.ClientSet.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(len(nodeList.Items)).ToNot(BeZero())
+
+				targetNodeName := ""
+				for _, node := range nodeList.Items {
+					for k := range node.Labels {
+						if k == "node-role.kubernetes.io/worker" {
```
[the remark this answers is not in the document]
slintes: good point, thanks and done

test/e2e/node_feature_discovery.go
```diff
@@ -626,7 +626,7 @@ var _ = framework.KubeDescribe("[NFD] Node Feature Discovery", func() {
 				targetNodeName := ""
 				for _, node := range nodeList.Items {
 					for k := range node.Labels {
-						if k == "node-role.kubernetes.io/worker" {
+						if k != "node-role.kubernetes.io/master" {
```
marquiz, line 629: There are many labels which are not `node-role.kubernetes.io/master` 😉 

Suggest to do smth like this
```
for _, node := range nodeList.Items {
	if _, ok := node.Labels["node-role.kubernetes.io/master"]; !ok {
		targetNodeName = node.Name
		break
	}
}
```

slintes: LOL, sorry... I should wait for finishing the 1st coffee in the morning before doing such "trivial changes" :D

slintes: fixed, thx

marquiz:
Looks good and I don't think I've anything to add. I think you could squash the commits, now

slintes:
> Looks good and I don't think I've anything to add. I think you could squash the commits, now

thanks & done :)

marquiz (changes requested):
Ach, one more nit, about the commit message.

I've been quite lazy in reviewing the commit messages, probably because that's somewhat inconvenient in Github. I think it would be good practice for us to try to follow the k8s [commit message guidelines](https://www.kubernetes.dev/docs/guide/pull-requests/#commit-message-guidelines).

I'd like to address at least *Use imperative mood in your commit message subject* and *Providing additional context* by changing the subject to something like:
`source/custom: add nodename rule`

Also the commit message body could be a bit more better read but I think I can live with that. Even if it's quite draft-like and I'm not a big fan of a bullet list only.

Another note: there are some trailing whitespace in `test/e2e/node_feature_discovery.go`

marquiz:
> Also the commit message body could be a bit more better read but I think I can live with that. Even if it's quite draft-like and I'm not a big fan of a bullet list only.

Talking about the commit message body, you already addressed *[Use the commit message body to explain the what and why of the commit](https://www.kubernetes.dev/docs/guide/pull-requests/#use-the-commit-message-body-to-explain-the-_what_-and-_why_-of-the-commit)* in the [description of this PR](https://github.com/kubernetes-sigs/node-feature-discovery/pull/429#issue-566745454). Some of that could be re-used in the commit message, I think.

test/e2e/node_feature_discovery.go
```diff
@@ -581,23 +590,181 @@ var _ = framework.KubeDescribe("[NFD] Node Feature Discovery", func() {
...
+						Name: "custom-config-extra-" + string(uuid.NewUUID()),
+					},
+					Data: data1,
+				}
+				cm1, err = f.ClientSet.CoreV1().ConfigMaps(f.Namespace.Name).Create(context.TODO(), cm1, metav1.CreateOptions{})
+				Expect(err).NotTo(HaveOccurred())
+
+				data2 := make(map[string]string)
+				data2["custom1.conf"] = `
+- name: ` + targetLabelNameWildcard + `
```
marquiz, line 667: trailing whitespace

test/e2e/node_feature_discovery.go
```diff
@@ -581,23 +590,181 @@ var _ = framework.KubeDescribe("[NFD] Node Feature Discovery", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+
+				data2 := make(map[string]string)
+				data2["custom1.conf"] = `
+- name: ` + targetLabelNameWildcard + `  
+  value: ` + targetLabelValueWildcard + `
+  matchOn:
+  - nodename:
+    - ` + targetNodeNameWildcard + `
+- name: ` + targetLabelNameNegative + `
```
marquiz, line 672: trailing whitespace

slintes:
fixed whitespaces and improved commit message

ArangoGutierrez:
```bash
Ran 3 of 220 Specs in 62.471 seconds
SUCCESS! -- 3 Passed | 0 Failed | 0 Pending | 217 Skipped
--- PASS: TestE2E (62.47s)
PASS
ok      sigs.k8s.io/node-feature-discovery/test/e2e     62.575s
```

source/custom/rules/nodename_rule.go
```diff
@@ -0,0 +1,51 @@
...
+	for _, nodenamePattern := range n {
+		log.Printf("DEBUG: matchNodename %s", nodenamePattern)
+
+		match, err := regexp.MatchString(nodenamePattern, nodeName)
+		if err != nil {
+			log.Printf("ERROR: nodename rule: invalid nodename regexp %q: %v", nodenamePattern, err)
+			continue
+		}
+		if !match {
+			log.Printf("DEBUG: nodename rule: No match for for pattern %q with node %q", nodenamePattern, nodeName)
```
ArangoGutierrez, line 44: Double `for for` in log string

ArangoGutierrez (changes requested):
One last catch

source/custom/rules/nodename_rule.go
slintes: fixed... and in the light of "too many log messages" I commented it and the other one out

source/custom/rules/nodename_rule.go
```diff
@@ -0,0 +1,50 @@
...
+		match, err := regexp.MatchString(nodenamePattern, nodeName)
+		if err != nil {
+			log.Printf("ERROR: nodename rule: invalid nodename regexp %q: %v", nodenamePattern, err)
+			continue
+		}
+		if !match {
+			//log.Printf("DEBUG: nodename rule: No match for pattern %q with node %q", nodenamePattern, nodeName)
+			continue
+		}
+		//log.Printf("DEBUG: nodename rule: Match for pattern %q with node %q", nodenamePattern, nodeName)
```
ArangoGutierrez, line 46: @marquiz you think that after the #445 this 2 log's could be DEBUG lvl, only seen when requested?

marquiz: Yeah, I think so

marquiz (approved):
Thank you @slintes for your efforts and patience! 😄  This is a great enhancement to NFD
/approve

ArangoGutierrez (approved):
Awesome, can't wait to support this via the Operator as well 
/lgtm

slintes:
Thanks all for the very helpful and constructive reviews :+1:

PR #430, opened 2021-02-03
Set imagePullPolicy in the deployment templates to Always

marquiz:
Update prepare-release.sh so that the policy will be set to IfNotPresent
for releases.

Inspired by the discussion in #423


ArangoGutierrez (approved):
Can't like and approve this more!!!!
/lgtm

PR #431, opened 2021-02-03
nfd-master: implement --instance flag

marquiz:
This can be used to help running multiple parallel NFD deployments in
the same cluster. The flag changes the node annotation namespace to
<instance>.nfd.node.kubernetes.io allowing different nfd-master intances
to store metadata in separate annotations.

Related to #427

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -66,7 +66,7 @@ func argsParse(argv []string) (master.Args, error) {
   %s [--prune] [--no-publish] [--label-whitelist=<pattern>] [--port=<port>]
      [--ca-file=<path>] [--cert-file=<path>] [--key-file=<path>]
      [--verify-node-name] [--extra-label-ns=<list>] [--resource-labels=<list>]
-     [--kubeconfig=<path>]
+     [--kubeconfig=<path>] [--instance=<name>]
   %s -h | --help
   %s --version
 
@@ -75,6 +75,9 @@ func argsParse(argv []string) (master.Args, error) {
   --version                       Output version and exit.
   --prune                         Prune all NFD related attributes from all nodes
                                   of the cluster and exit.
+  --instance=<name>               Instance name. Used to separate annotation
+                                  namespaces for multiple parallel deployments.
+                                  [Default: ]
   --kubeconfig=<path>             Kubeconfig to use [Default: ]
                                   of the cluster and exit.
   --port=<port>                   Port on which to listen for connections.
@@ -109,6 +112,12 @@ func argsParse(argv []string) (master.Args, error) {
 
 	// Parse argument values as usable types.
 	var err error
+	args.Instance = arguments["--instance"].(string)
+	if ok, _ := regexp.MatchString(`^([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$`, args.Instance); args.Instance != "" && !ok {
+		return args, fmt.Errorf("invalid --instance %q: instance name "+
+			"must start and end with an alphanumeric character and may only contain "+
+			"alphanumerics, `-`, `_` or `.`", args.Instance)
+	}
 	args.CaFile = arguments["--ca-file"].(string)
 	args.CertFile = arguments["--cert-file"].(string)
 	args.KeyFile = arguments["--key-file"].(string)
--- a/docs/advanced/master-commandline-reference.md
+++ b/docs/advanced/master-commandline-reference.md
@@ -48,6 +48,22 @@ Example:
 nfd-master --port=443
 ```
 
+### --instance
+
+The `--instance` flag makes it possible to run multiple NFD deployments in
+parallel. In practice, it separates the node annotations between deployments so
+that each of them can store metadata independently. The instance name must
+start and end with an alphanumeric character and may only contain alphanumeric
+characters, `-`, `_` or `.`.
+
+Default: *empty*
+
+Example:
+
+```bash
+nfd-master --instance=network
+```
+
 ### --ca-file
 
 The `--ca-file` is one of the three flags (together with `--cert-file` and
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -175,7 +175,7 @@ func (m *nfdMaster) Run() error {
 		serverOpts = append(serverOpts, grpc.Creds(credentials.NewTLS(tlsConfig)))
 	}
 	m.server = grpc.NewServer(serverOpts...)
-	pb.RegisterLabelerServer(m.server, &labelerServer{args: m.args, apiHelper: m.apihelper})
+	pb.RegisterLabelerServer(m.server, m)
 	stdoutLogger.Printf("gRPC server serving on port: %d", m.args.Port)
 	return m.server.Serve(lis)
 }
@@ -310,15 +310,9 @@ func filterFeatureLabels(labels Labels, extraLabelNs map[string]struct{}, labelW
 	return outLabels, extendedResources
 }
 
-// Implement LabelerServer
-type labelerServer struct {
-	args      Args
-	apiHelper apihelper.APIHelpers
-}
-
-// Service SetLabels
-func (s *labelerServer) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.SetLabelsReply, error) {
-	if s.args.VerifyNodeName {
+// SetLabels implements LabelerServer
+func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.SetLabelsReply, error) {
+	if m.args.VerifyNodeName {
 		// Client authorization.
 		// Check that the node name matches the CN from the TLS cert
 		client, ok := peer.FromContext(c)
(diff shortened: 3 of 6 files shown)


e0ne:
It doesn't fixes #427. Adding new annotations works well, but different NDF instances will override each other labels. It means any software which relates on standard NFD behaviour and labels won't work good in case when two or more NFD are deployed.

marquiz:
> It doesn't fixes #427. Adding new annotations works well, but different NDF instances will override each other labels. It means any software which relates on standard NFD behaviour and labels won't work good in case when two or more NFD are deployed.

My point is that you need to configure the "label separation" somewhere, either master or worker side. Does it really matter if it would be on worker side? Worker side is much more flexible, e.g. you can directly disable all unneeded feature sources.

e0ne:
@marquiz is it a supported configuration to have 2+ workers on the same node with different configuration?

marquiz:
> override each other labels

What kind of label overwrite are you concerned about? If two instances write cpuid.XYZ or kernel version it shouldn't even matter as the value from both sources should be the same. When running multiple instances of NFD in parallel there is really no reliable way prevent overlap, i.e. prevent different instances modifying the same labels. An incorrect deployment/configuration of one NFD instance will ruin everything.

marquiz:
> @marquiz is it a supported configuration to have 2+ workers on the same node with different configuration?

Yes, sure. They are separate instances (pods) so they may have different configuration. The "configuration" is basically the command line flags plus the configuration file. Configuration file is typically taken from a config map. Just use different configmap for different instances (e.g. run them in separate k8s namespaces)

e0ne:
E..g.:
instance1: --label-whitelist=pci-0300_102b.present"
instance2: --label-whitelist=pci-8086.sriov.capable"

As a result, I would like to have a list of labels which contains both pci-0300_102b and pci-8086.sriov features discovered

e0ne:
In other words, master should merge labels provided by different workers

zvonkok:
@e0ne Why would different workers have different labels? Two instances of NFD would be running in e.g. different namespaces, not on different nodes.

zvonkok:
Each worker would discover the very same features. @e0ne Are you planning to run the second instance of NFD on a subset of available nodes? Usually NFD is deployed to all workers.

e0ne:
@zvonkok here is an example how we deploy NFD: https://github.com/kubernetes-sigs/node-feature-discovery/issues/427#issuecomment-772554744

zvonkok:
Ah got it. You have different configurations for each instance of NFD.

e0ne:
> Ah got it. You have different configurations for each instance of NFD.

Exactly, that's why we need several instances of NDF

marquiz:
> In other words, master should merge labels provided by different workers

As I commented in #427, this totally different and orthogonal AFAICS. Could you open a separate issue about this where we could discuss it?

e0ne:
we can add 'instance' prefix to labels, not for annotations and it should work in a such way: each NFD instance will override labels with 'instance' prefix and can deploy a worker without prefix as a default

marquiz:
> we can add 'instance' prefix to labels, not for annotations and it should work in a such way: each NFD instance will override labels with 'instance' prefix and can deploy a worker without prefix as a default

This sounds like not a very good idea and would just add confusion, I think.

I still fail to see what you'd gain with this. If two deployments would create the same label (name) with a different value this is clearly a configuration error. They should use a different label name, e.g. by using a dedicated label ns like `vendor-1.com/feature-x` and `vendor-2.com/feature-x`. If you're talking about the "built-in" feature labels like `cpu-cpuid.XYZ` or `pci-<device>.present` etc the value really should not be different between deployments

moshe010 (approved):
Thanks for working on this. We tested it in it solves our problem :)

marquiz:
> Thanks for working on this. We tested it in it solves our problem :)

Thanks for testing. I haven't found any problems in my own test setup. Just waiting for a review....

What I really want to have is an e2e-test for this but I think that could be addressed in a separate PR

moshe010:
@zvonkok, Can you please review this PR?

marquiz:
@moshe010 you can also take a look at the changes and provide comments, that would be much appreciated 😉

pkg/nfd-master/nfd-master.go
```diff
@@ -46,20 +46,19 @@ const (
 	LabelNs = "feature.node.kubernetes.io"
 
 	// Namespace for all NFD-related annotations
```
adrianchiris, line 48: nit: base Namespace ?

cmd/nfd-master/main.go
```diff
@@ -109,6 +112,10 @@ func argsParse(argv []string) (master.Args, error) {
 
 	// Parse argument values as usable types.
 	var err error
+	args.Instance = arguments["--instance"].(string)
+	if ok, _ := regexp.MatchString(`^([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9]$`, args.Instance); args.Instance != "" && !ok {
+		return args, fmt.Errorf("invalid --instance %q", args.Instance)
```
e0ne, line 117: It would be good to add some help message here or in line 80 and provide details what instance name format is expected

pkg/nfd-master/nfd-master.go
```diff
@@ -73,6 +72,7 @@ type Annotations map[string]string
 
 // Command line arguments
 type Args struct {
+	Instance       string
```
e0ne, line 75: nit: do we want to keep all items in an alphabetical order?

pkg/nfd-master/nfd-master.go
```diff
@@ -92,18 +93,26 @@ type NfdMaster interface {
...
 	nfd := &nfdMaster{args: args,
 		nodeName: os.Getenv("NODE_NAME"),
-		ready:    make(chan bool, 1)}
+		ready:    make(chan bool, 1),
+	}
+
+	if args.Instance == "" {
+		nfd.annotationNs = AnnotationNsBase
+	} else {
+		nfd.annotationNs = args.Instance + "." + AnnotationNsBase
```
adrianchiris, line 114: nit:
should you use strings.join() like in other places in this file ?

pkg/nfd-master/nfd-master.go
```diff
@@ -496,7 +515,7 @@ func addNs(src string, nsToAdd string) string {
 	if strings.Contains(src, "/") {
 		return src
 	}
-	return filepath.Join(nsToAdd, src)
+	return path.Join(nsToAdd, src)
```
adrianchiris, line 518: ran this on windows ?

nit: not directly related to this commit

test/e2e/node_feature_discovery.go
```diff
@@ -393,7 +393,7 @@ func cleanupNode(cs clientset.Interface) {
 
 			// Remove annotations
 			for key := range node.Annotations {
-				if strings.HasPrefix(key, master.AnnotationNs) {
+				if strings.HasPrefix(key, master.AnnotationNsBase) {
```
adrianchiris, line 396: in case instance is used it will break cleanup right ?

if/when support is added to instance flag in e2e test this would need to be modified.

adrianchiris (commented):
Changes look good IMO,

first commits are refactors which are not necessarily needed for support for instance flag, but are sensible and simplifies code.

just a few very minor nits from me plus a typo in cmdline reference.

marquiz:
Thanks for the review @e0ne and @adrianchiris ! Addressed the comments in the topmost patch

I want to squash the fixes in the other patches before merging
/hold

test/e2e/node_feature_discovery.go
marquiz: Yes, correct. My plan was to update/extend the e2e tests in a separate PR

pkg/nfd-master/nfd-master.go
marquiz: > nit: not directly related to this commit

True. Reverted this change here and addressed in a separate PR (#435)

pkg/nfd-master/nfd-master.go
marquiz: The other places have a slice to join. I originally used strings.Join but decided chaged it. Here we only have two values so IMHO this is more readable and practical than doing
```
strings.Join([args.Instance, AnnotationNsBase], ".")
```

WDYT?

pkg/nfd-master/nfd-master.go
marquiz: Yes, fixed

docs/advanced/master-commandline-reference.md
```diff
@@ -48,6 +48,22 @@ Example:
 nfd-master --port=443
 ```
 
+### --instance
+
+The `--instance` flag makes it possible to run multiple NFD deployments in
+parallel. In practice, it separates the the node annotations between deployments
```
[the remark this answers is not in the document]
marquiz: Thanks, fixed

pkg/nfd-master/nfd-master.go
marquiz: Yes, certainly, thanks for noting this. Fixed

cmd/nfd-master/main.go
marquiz: Indeed 😄  Added an explanation to the error message

marquiz:
> first commits are refactors which are not necessarily needed for support for instance flag, but are sensible and simplifies code.

True, the refactors are not strictly necessary, but without them the code would far uglier as the annotation ns mangling messes things up. I tried and didn't like it 😄  I didn't see much point in putting the refactors in a separate PR, but, could be easily done if people like it better that way

adrianchiris:
> True, the refactors are not strictly necessary, but without them the code would far uglier as the annotation ns mangling messes things up. I tried and didn't like it 😄 I didn't see much point in putting the refactors in a separate PR, but, could be easily done if people like it better that way

Im totally fine with having in the same PR.

pkg/nfd-master/nfd-master.go
adrianchiris: Make sense, just wanted to point it out and i see you already considered it when coding.

adrianchiris (approved):
Thanks for addressing my comments @marquiz  !

marquiz:
Needed to rebase. Squashed the review fixes in the patchset.
/unhold

e0ne:
@marquiz thanks for addressing my comments!

PR #432, opened 2021-02-04
nfd-worker: use fsnotify for watching for config file changes

marquiz:
Add support for detecting configuration file changes via file system
notifications (fsnotify). Watches are added for the whole directory
chain (up to root directory) so that all changes (even directory
renames) affecting the given configuration file path are captured.

Previously dynamic (re-)configuration of nfd-worker was implemented by
(re-)reading the configuration file on every labeling pass. This was
simple and effective, even if a bit wasteful. However, it didn't provide
asynchronous configuration updates that will be required for e.g.
controlling the "sleep-interval" parameter dynamically which will be
implemented by later patches.

### DIFF
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -286,6 +286,10 @@ func (w *nfdWorker) Run() error {
 
 		case <-configTrigger:
 			w.configure(configFilePath, w.args.Options)
+
+			// Always re-label after a re-config event. This way the new config
+			// comes into effect even if the sleep interval is long (or infinite)
+			labelTrigger = time.After(0)
 		}
 	}
 }
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -191,31 +191,31 @@ func (w *nfdWorker) Run() error {
 	}
 	defer w.disconnect()
 
+	trigger := time.After(0)
 	for {
-		// Parse and apply configuration
-		w.configure(w.args.ConfigFile, w.args.Options)
+		select {
+		case <-trigger:
+			// Parse and apply configuration
+			w.configure(w.args.ConfigFile, w.args.Options)
 
-		// Get the set of feature labels.
-		labels := createFeatureLabels(w.sources, w.labelWhiteList)
+			// Get the set of feature labels.
+			labels := createFeatureLabels(w.sources, w.labelWhiteList)
 
-		// Update the node with the feature labels.
-		if w.client != nil {
-			err := advertiseFeatureLabels(w.client, labels)
-			if err != nil {
-				return fmt.Errorf("failed to advertise labels: %s", err.Error())
+			// Update the node with the feature labels.
+			if w.client != nil {
+				err := advertiseFeatureLabels(w.client, labels)
+				if err != nil {
+					return fmt.Errorf("failed to advertise labels: %s", err.Error())
+				}
 			}
-		}
 
-		if w.args.Oneshot {
-			break
-		}
+			if w.args.Oneshot {
+				return nil
+			}
 
-		if w.args.SleepInterval > 0 {
-			time.Sleep(w.args.SleepInterval)
-		} else {
-			w.disconnect()
-			// Sleep forever
-			select {}
+			if w.args.SleepInterval > 0 {
+				trigger = time.After(w.args.SleepInterval)
+			}
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


pkg/nfd-worker/nfd-worker.go
```diff
@@ -214,11 +256,33 @@ func (w *nfdWorker) Run() error {
...
+				// Blindly remove existing watch and add a new one
+				if err := configWatch.Close(); err != nil {
+					stderrLogger.Printf("WARNING: failed to close fsnotify watcher: %v", err)
+				}
+				configWatch, paths, err = addConfigWatch(configFilePath)
+				if err != nil {
+					return err
+				}
+
+				configTrigger = time.After(time.Second)
```
Ethyling, line 276: I'm not sure to understand why we are not directly reloading the configuration here. Do we need another channel for this?

pkg/nfd-worker/nfd-worker.go
```diff
@@ -281,6 +281,10 @@ func (w *nfdWorker) Run() error {
 
 		case <-configTrigger:
 			w.configure(configFilePath, w.args.Options)
+
+			// Always re-label after a re-config event. This way the new config
+			// comes into effect even if the sleep interval is long (or infinite)
+			labelTrigger = time.After(1 * time.Second)
```
Ethyling, line 287: Should we use `time.After(0)` here?

Ethyling (commented):
I did a first review, I still need to test it now. Also there is an issue on your last commit message :)

pkg/nfd-worker/nfd-worker.go
```diff
@@ -191,34 +227,66 @@ func (w *nfdWorker) Run() error {
...
-				return fmt.Errorf("failed to advertise labels: %s", err.Error())
+			if w.args.SleepInterval > 0 {
+				labelTrigger = time.After(w.args.SleepInterval)
 			}
-		}
 
-		if w.args.Oneshot {
-			break
-		}
+		case e := <-configWatch.Events:
```
Ethyling, line 262: I think we should add a log here to specify that we have detected a change in the configuration file

Ethyling:
Some small comments, but it seems to be working as expected :smiley:

pkg/nfd-worker/nfd-worker.go
marquiz: I'm not sure. The idea behind this was a 1 second grace period/rate limiter, e.g. if there were multiple filesystem operations in a quick succession, each of which would cause an (unnecessary) config reload. So I'd probably keep it this way. But, a comment describing this would be in order. WDYT?

pkg/nfd-worker/nfd-worker.go
marquiz: At this point we only have detected a change somewhere in the directory tree. E.g. if the config would be `/foo/bar/baz.conf`, any change in any file under `/foo/` or `/foo/bar/` will create an event. We don't want to log all of them, just those concerning `baz.conf` or the directories directly.

But yes, a log message makes sense. I'll add a log a bit later after the `if`

pkg/nfd-worker/nfd-worker.go
marquiz: Woops, me neither, anymore 🙄 Very good comment, thanks for pointing this out. Doesn't seem to make any sense, I'll change this

I really don't remember anymore as the first versions date back to last November or so. But, I think this is just some stale leftover ripples from some early prototype that I became blind with.

marquiz: OK, now that I took a more complete look at the code I start to remember 😄  This is actually the grace period / rate limiter of numerous events in quick succession. We actually want this. I'll add a comment

pkg/nfd-worker/nfd-worker.go
marquiz: With a fresh look, you're correctl. `0` is better here. I'll change it. We do the rate limiting elsewhere.

pkg/nfd-worker/nfd-worker.go
marquiz: Changed to `0`

pkg/nfd-worker/nfd-worker.go
marquiz: Changed to `0`

pkg/nfd-worker/nfd-worker.go
marquiz: Added a log message below

pkg/nfd-worker/nfd-worker.go
marquiz: Added a comment explaining this use of double event channels

marquiz:
Rebased and addressed review comments

PR #433, opened 2021-02-04
nfd-worker: add core config

marquiz:
Allows dynamic (re-)configuration of most nfd-worker options. The goal
is to have most configuration parameters specified in the configuration
file and deprecate most of the command line flags. The priority is
intended to be such that command line flags override whatever is
specified in the configuration file. Thus, specifying something on the
command line effectively disables dynamic configurability of that
parameter.

Add some "core" config options:
- `core.noPublish`
  A new config file option for (dynamically) controlling the "dry-run" mode of
nfd-worker, i.e. whether to send labeling requests to nfd-master.
The `--no-publish` command line flag takes precedence over this config file option.
- `core.sleepInterval`
  A new config file option for (dynamically) controlling the sleep
interval. At the same time, deprecate the `--sleep-interval` command line
flag. The command line flag takes precedence over the config file option.
- `core.labelWhiteList`
  A new config file option for label whitelisting. Deprecate the
`--label-whitelist` command line flag. Note that the command line flag has
higher priority than the config file option.
- `core.sources`
  A new config file option for controlling the enabled feature sources,
aimed at replacing the `--sources` command line flag which is now marked
as deprecated. The command line flag takes precedence over the config
file option.

Also, adds a patch to bail out on invalid config file:
- Changes the behaviour so that if the specified configuration file exists
it must be valid. Error out at startup if the config is invalid.
Similarly, exit with an error at runtime if the config file becomes
invalid. Bailing out, instead of just printing an error, was a
deliberate choice in order to make configuration mistakes evident.
- Having no configuration file is tolerated, however. If the specified
configuration file does not exists nfd-worker resorts to default
settings.


marquiz:
Updated documentation. Still missing unit tests

ArangoGutierrez (commented):
Looking good, tested both manually and with the operator, I think we are good once unit tests are in place.

Question: should we handle the case when the user might update the ConifgMap with a new Config file, loading the new config, or rebooting the pod? Or should this be handled in the operator instead? 

`(not reviwing with approve or lgtm since the PR is on hold)`

image:
- quay.io/eduardoarango/node-feature-discovery:433

marquiz:
Thanks for the review @ArangoGutierrez ! I've been working on the tests but there's still a little bit of work to do there.

> Question: should we handle the case when the user might update the ConifgMap with a new Config file, loading the new config, or rebooting the pod? Or should this be handled in the operator instead?

I don't know if I fully understand the question. But if I do, I think this should be handled by the fsnotify watch (I don't think we should restart the pod): configmap changes -> eventually the file inside the container changes (via `mounts`) -> fsnotify detects this change -> we re-read the config

ArangoGutierrez:
> fsnotify

fsnotify is the answer I was looking for, I totally overlooked that. 👍🏽

marquiz:
Documentation and unit tests updated
/unhold

ArangoGutierrez:
Go linter is not happy :(

ArangoGutierrez:
```bash
 golangci-lint run --timeout 5m0s
pkg/nfd-worker/nfd-worker-internal_test.go:241:12: Error return value of `w.Run` is not checked (errcheck)
			go w.Run()
			        ^
pkg/nfd-worker/nfd-worker-internal_test.go:220:7: ineffectual assignment to `err` (ineffassign)
			_, err = f.WriteString(data)
			   ^
pkg/nfd-worker/nfd-worker.go:300:8: S1005: unnecessary assignment to the blank identifier (gosimple)
		case _ = <-w.stop:
		     ^
source/fake/fake.go:27:2: structtag: struct field labels has json tag but is not exported (govet)
	labels map[string]string `json:"labels"`
	^
make: *** [Makefile:115: ci-lint] Error 1 
```
Still not happy

marquiz:
> Still not happy

looks better, now 🤞

ArangoGutierrez:
`make test e2e-test > test.txt` 

report:
```bash
Ran 2 of 219 Specs in 22.944 seconds
SUCCESS! -- 2 Passed | 0 Failed | 0 Pending | 217 Skipped
--- PASS: TestE2E (22.94s)
PASS
ok      sigs.k8s.io/node-feature-discovery/test/e2e     23.016s
```

Output logs : [test.txt](https://github.com/kubernetes-sigs/node-feature-discovery/files/5998622/test.txt)

After testing on kubernetes 1.20+, and reading the patch I can be sure to 
/lgtm

PR #434, opened 2021-02-04
Change the worker conf update mechanism of templates

marquiz:
Rely on markers instead which makes the update work also in the middle
of a template file. Previously it relied on the configuration data to be
at the very end of the file to be updated.


ArangoGutierrez (approved):
good old `sed` magic
/lgtm

PR #438, opened 2021-02-10
Update gogo/protobuf and golang.org/x/text

ArangoGutierrez:
Why:

- CVE-2020-28851
- CVE-2021-3121

This CVE's do not impact NFD directly, but we are importing those deps, so is better to keep them up to date


marquiz:
Thanks @ArangoGutierrez 
/lgtm

PR #443, opened 2021-02-19
Add NFD master '--instance' flag support to Helm chart

e0ne:
--instance flag was introduce in the e52ec3480fc586211bbdea3c99caa47bf810c72d commit.

This patch adds ability to configure it in the Helm chard during deployment.

### DIFF
--- a/deployment/node-feature-discovery/templates/master.yaml
+++ b/deployment/node-feature-discovery/templates/master.yaml
@@ -44,6 +44,10 @@ spec:
             - "nfd-master"
           resources:
             {{- toYaml .Values.master.resources | nindent 12 }}
+          args:
+            { { - if .Values.master.instance | empty | not } }
+            - "--instance={{ .Values.master.instance }}"
+            { { - end } }
 ## Enable TLS authentication
 ## The example below assumes having the root certificate named ca.crt stored in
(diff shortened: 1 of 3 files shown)


ArangoGutierrez:
Thanks for this patch @e0ne!
could you add a proper PR description, so it creates a nice git history when merged
Cheers

e0ne:
@ArangoGutierrez thanks for the feedback. I updated PR description.

ArangoGutierrez (commented):
Patch looks good to me
/lgtm

deployment/node-feature-discovery/templates/master.yaml
```diff
@@ -44,14 +44,17 @@ spec:
             - "nfd-master"
           resources:
             {{- toYaml .Values.master.resources | nindent 12 }}
+          args:
+            { { - if .Values.master.instance | empty | not } }
```
moshe010, line 48: will it works when .Values.master.instance is empty? won't you have args: with empty flags?

e0ne: It will pass empty flag, in this case no args will be rendered. Let's add empty value for consistency with orhter

marquiz:
Thanks @e0ne ! I haven't tested/verified this myself, but looks good
/approve

adrianchiris (approved):

PR #446, opened 2021-02-23
Fix daemonset deployment template

marquiz:
Remove incorrect tolerations that accidentally crept in with the helm
charts and break the deployment.


ArangoGutierrez (approved):
/lgtm

PR #449, opened 2021-02-24
docs: align docs with the single-dash command line flags

marquiz:
After moving to the flags package for command line argument parsing NFD
accepts command line arguments (flags) starting with a single dash (e.g.
-no-publish in addition to --no-publish). Even if double-dash can be
used the single-dash version is printed e.g. in the usage string (from
-h, -help) so align documentation with that.


ArangoGutierrez (approved):
No brainier PR
/lgtm

PR #452, opened 2021-02-25
test/e2e: make openshift bits optional

marquiz:
Only enabled if OPENSHIFT is specified in the environment when running
make e2e-test, for example:

```
$ OPENSHIFT=y make e2e-test
```


ArangoGutierrez (approved):
openshift specific bit of test was introduced on #429 , good catch 
/lgtm

PR #458, opened 2021-03-01
nfd-worker: stop masking crashes in feature discovery

marquiz:
The code should be stable enough. If there are fatal bugs causing the
discovery to panic/segfault that should be made visible instead of
semi-siently hiding it. Also, this caused one (negative) test case to
fail undetected which is now fixed.


pkg/nfd-worker/nfd-worker-internal_test.go
```diff
@@ -374,9 +372,6 @@ func TestCreateFeatureLabels(t *testing.T) {
 			})
 		})
 		Convey("When fake feature source is configured with a whitelist that doesn't match", func() {
-			fakeFeatureSource := source.FeatureSource(new(fake.Source))
```
adrianchiris, line 377: so when you removed the `recover()` from `getFeatureLabels()` you hit a panic as the fake Source did not have its config initialized right ?

marquiz: Yup, exactly

Issue #439, opened 2021-02-10
Add additional cpu features


bfournie:
In order to provide for more granularity in scheduling decisions, we would like to add the following cpu features:

- feature.node.kubernetes.io/cpu-performance_policy (_performance_ or _powersave_) based on /sys/devices/system/cpu/cpufreq/policy
- feature.node.kubernetes.io/cpu-max_cstate (integer) based on /sys/module/intel_idle/parameters/max_cstate
- feature.node.kubernetes.io/cpu-virtualization (_VT-x_ or _VT-d_) based on output of lscpu
- feature.node.kubernetes.io/cpu-pstate.disable (true or false) based on /sys/module/acpi_cpufreq/parameters/acpi_pstate_strict

ArangoGutierrez:
Awesome @bfournie !!

marquiz:
> * feature.node.kubernetes.io/cpu-performance_policy (_performance_ or _powersave_) based on /sys/devices/system/cpu/cpufreq/policy

You mean the scaling governor? We prolly need to check that all cpus (policies) have the same governor activated.



> * feature.node.kubernetes.io/cpu-virtualization (_VT-x_ or _VT-d_) based on output of lscpu

Hmm, need to check if there's an easy way to discover these without resorting to `lscpu` binary

bfournie:
> > * feature.node.kubernetes.io/cpu-performance_policy (_performance_ or _powersave_) based on /sys/devices/system/cpu/cpufreq/policy
> 
> You mean the scaling governor? We prolly need to check that all cpus (policies) have the same governor activated.
> 
Yes the intent was to use the scaling governor, but need to figure out what to report if all cpus don't have the same policy.

> > * feature.node.kubernetes.io/cpu-virtualization (_VT-x_ or _VT-d_) based on output of lscpu
> 
> Hmm, need to check if there's an easy way to discover these without resorting to `lscpu` binary

Right, if don't want use lscpu, could check /proc/cpuinfo for the 'vmx' flag which would show if VT-x is enabled.  Again not sure how to report this if cpus have different settings though.

MadhavJivrajani:
Hi! Could I work on this if it's available?

mythi:
> feature.node.kubernetes.io/cpu-virtualization (VT-x or VT-d) 

`VMX` and `HYPERVISOR` bits are supported by cpu source.

bfournie:
> Hi! Could I work on this if it's available?

[...]

PR #463, opened 2021-03-02
Support for additional cpu features

bfournie:
This adds additional cpu features:
- pstate enable from status of intel_pstate driver
- pstate performance and powersave settings from scaling_governor
- cstate enable from max_cstates in intel_idle driver

Closes: #439

### DIFF
--- a/docs/get-started/features.md
+++ b/docs/get-started/features.md
@@ -61,7 +61,10 @@ The **cpu** feature source supports the following labels:
 | cpuid                   | &lt;cpuid flag&gt; | CPU capability is supported
 | hardware_multithreading |                    | Hardware multithreading, such as Intel HTT, enabled (number of logical CPUs is greater than physical CPUs)
 | power                   | sst_bf.enabled     | Intel SST-BF ([Intel Speed Select Technology][intel-sst] - Base frequency) enabled
-| [pstate][intel-pstate]  | turbo              | Set to 'true' if turbo frequencies are enabled in Intel pstate driver, set to 'false' if they have been disabled.
+| [pstate][intel-pstate]  | status             | The status of the Intel pstate driver when in use and enabled, either 'active' or 'passive'.
+|                         | turbo              | Set to 'true' if turbo frequencies are enabled in Intel pstate driver, set to 'false' if they have been disabled.
+|                         | scaling_governor   | The value of the Intel pstate scaling_governor when in use, either 'powersave' or 'performance'.
+| cstate                  | enabled            | Set to 'true' if cstates are set in the intel_idle driver, otherwise set to 'false'.
 | [rdt][intel-rdt]        | RDTMON             | Intel RDT Monitoring Technology
 |                         | RDTCMT             | Intel Cache Monitoring (CMT)
 |                         | RDTMBM             | Intel Memory Bandwidth Monitoring (MBM)
--- a/source/cpu/cstate.go
+++ b/source/cpu/cstate.go
@@ -1,0 +1,51 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cpu
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strconv"
+	"strings"
+
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+// Discover if c-states are enabled
+func detectCstate() (bool, error) {
+	// When the intel_idle driver is in use (default), check setting of max_cstates
+	driver, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpuidle/current_driver"))
+	if err != nil {
+		return false, fmt.Errorf("cannot get driver for cpuidle: %s", err.Error())
+	}
+
+	if strings.TrimSpace(string(driver)) != "intel_idle" {
+		// Currently only checking intel_idle driver for cstates
+		return false, fmt.Errorf("intel_idle driver is not in use: %s", string(driver))
+	}
+
+	data, err := ioutil.ReadFile(source.SysfsDir.Path("module/intel_idle/parameters/max_cstate"))
+	if err != nil {
+		return false, fmt.Errorf("cannot determine cstate from max_cstates: %s", err.Error())
+	}
+	cstates, err := strconv.Atoi(strings.TrimSpace(string(data)))
+	if err != nil {
+		return false, fmt.Errorf("non-integer value of cstates: %s", err.Error())
+	}
+
+	return cstates > 0, nil
+}
--- a/source/cpu/pstate.go
+++ b/source/cpu/pstate.go
@@ -20,6 +20,9 @@ import (
 	"fmt"
 	"io/ioutil"
 	"runtime"
+	"strings"
+
+	"k8s.io/klog/v2"
 
 	"sigs.k8s.io/node-feature-discovery/source"
 )
@@ -32,14 +35,72 @@ func detectPstate() (map[string]string, error) {
 		return nil, nil
 	}
 
-	// Only looking for turbo boost for now...
+	// Get global pstate status
+	data, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/intel_pstate/status"))
+	if err != nil {
+		return nil, fmt.Errorf("could not read pstate status: %s", err.Error())
+	}
+	status := strings.TrimSpace(string(data))
+	if status == "off" {
+		// No need to check other pstate features
+		klog.Infof("intel_pstate driver is not in use")
+		return nil, nil
+	}
+	features := map[string]string{"status": status}
+
+	// Check turbo boost
 	bytes, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/intel_pstate/no_turbo"))
 	if err != nil {
-		return nil, fmt.Errorf("can't detect whether turbo boost is enabled: %s", err.Error())
+		klog.Errorf("can't detect whether turbo boost is enabled: %s", err.Error())
+	} else {
+		features["turbo"] = "false"
+		if bytes[0] == byte('0') {
+			features["turbo"] = "true"
+		}
 	}
-	features := map[string]string{"turbo": "false"}
-	if bytes[0] == byte('0') {
-		features["turbo"] = "true"
+
+	if status != "active" {
+		// Don't check other features which depend on active state
+		return features, nil
+	}
+
+	// Determine scaling governor that is being used
+	policies, err := ioutil.ReadDir(source.SysfsDir.Path("devices/system/cpu/cpufreq"))
+	if err != nil {
+		klog.Errorf("failed to read cpufreq directory: %s", err.Error())
+		return features, nil
+	}
+
+	scaling := ""
+	for _, policy := range policies {
+		// Ensure at least one cpu is using this policy
+		cpus, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpufreq", policy.Name(), "affected_cpus"))
+		if err != nil {
+			klog.Errorf("could not read cpufreq policy %s affected_cpus", policy.Name())
+			continue
+		}
+		if strings.TrimSpace(string(cpus)) == "" {
+			klog.Infof("policy %s has no associated cpus", policy.Name())
+			continue
+		}
+
+		data, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpufreq", policy.Name(), "scaling_governor"))
+		if err != nil {
+			klog.Errorf("could not read cpufreq policy %s scaling_governor", policy.Name())
+			continue
+		}
+		policy_scaling := strings.TrimSpace(string(data))
+		// Check that all of the policies have the same scaling governor, if not don't set feature
+		if scaling != "" && scaling != policy_scaling {
+			klog.Infof("scaling_governor for policy %s doesn't match prior policy", policy.Name())
+			scaling = ""
+			break
+		}
+		scaling = policy_scaling
+	}
+
+	if scaling != "" {
+		features["scaling_governor"] = scaling
 	}
 
 	return features, nil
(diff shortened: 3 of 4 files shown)


docs/get-started/features.md
```diff
@@ -61,7 +61,11 @@ The **cpu** feature source supports the following labels:
 | cpuid                   | &lt;cpuid flag&gt; | CPU capability is supported
 | hardware_multithreading |                    | Hardware multithreading, such as Intel HTT, enabled (number of logical CPUs is greater than physical CPUs)
 | power                   | sst_bf.enabled     | Intel SST-BF ([Intel Speed Select Technology][intel-sst] - Base frequency) enabled
-| [pstate][intel-pstate]  | turbo              | Set to 'true' if turbo frequencies are enabled in Intel pstate driver, set to 'false' if they have been disabled.
+| [pstate][intel-pstate]  | enabled            | Set to 'true' if the Intel pstate driver is in use, otherwise set to 'false'.
```
marquiz, line 64: Just thinking, would it be useful to advertise the status 🤔  I.e. pstate driver has been enabled if the label exists. The value of the label would indicate the status i.e. `active` or `passive`. The existence of the label could be used to check if enabled and the value for more detailed affinity rules. WDYT?

docs/get-started/features.md
```diff
@@ -61,7 +61,11 @@ The **cpu** feature source supports the following labels:
...
+|                         | turbo              | Set to 'true' if turbo frequencies are enabled in Intel pstate driver, set to 'false' if they have been disabled.
+|                         | powersave          | Set to 'true' if the scaling policy is set to 'powersave'.
+|                         | performance        | Set to 'true' if the scaling policy is set to 'performance'.
```
marquiz, line 67: Again, just thinking should we only have one `governor` label that would be set to the active scaling_governor, assuming all cpus (policies) use the same governor. This label would not exist if different policies used different governors. I doubt that there is any practical use case in K8s scope to have the system partitioned by the scaling governor (at least scheduling on the desired cpus would be very hard)?

Another consideration: the practical operation of the scaling governor depends quite heavily on the mode (active/passive). Should these labels only exist if in the "active" mode?

WDYT?

source/cpu/cstate.go
```diff
@@ -0,0 +1,58 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
marquiz, line 2: 2021

marquiz (commented):
Thanks @bfournie for the PR! Some quick comments, need to take another look a bit later.

/ok-to-test

docs/get-started/features.md
bfournie: Good idea. I will change the label to "status" with values active/passive based on the intel_pstate status.

docs/get-started/features.md
bfournie: Sure, will change to have one label instead of two.  Maybe call it "policy" instead of "governor"?  It will be set to the value of scaling_governor. And yes also it should only be set if the status of intel_pstate is active according to https://www.kernel.org/doc/html/v4.12/admin-guide/pm/intel_pstate.html.

Thank you for the review.

source/cpu/cstate.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Discover if c-states are enabled
+func detectCstate() (bool, error) {
+	// When the intel_idle driver is in use (default), check setting of max_cstates
+	driver, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpuidle/current_driver"))
+	if err != nil {
+		return false, fmt.Errorf("cannot get driver for cpuidle: %s", err.Error())
+	}
+
+	if strings.TrimSpace(string(driver)) == "intel_idle" {
```
dhellmann, line 36: It's usual in go code to check for reasons to return early and keep the "happy path" through a function indented as far to the left as possible. So, if you reverse this condition and move the else block up, the bulk of this function doesn't need to be inside an if/else statement.

source/cpu/cstate.go
```diff
@@ -0,0 +1,57 @@
...
+		data, err := ioutil.ReadFile(source.SysfsDir.Path("module/intel_idle/parameters/max_cstate"))
+		if err != nil {
+			return false, fmt.Errorf("cannot determine cstate from max_cstates: %s", err.Error())
+		}
+		cstates, err := strconv.Atoi(strings.TrimSpace(string(data)))
+		if err != nil {
+			return false, fmt.Errorf("non-integer value of cstates: %s", err.Error())
+		}
+
+		if cstates > 0 {
```
dhellmann, line 46: I think this if/else block can be condensed to

```
return cstates > 0, nil
```

source/cpu/pstate.go
```diff
@@ -32,12 +33,64 @@ func detectPstate() (map[string]string, error) {
...
+	policies, err := ioutil.ReadDir(source.SysfsDir.Path("devices/system/cpu/cpufreq"))
+	if err != nil {
+		return nil, fmt.Errorf("cpufreq directory not available to check pstate: %s", err.Error())
+	}
+
+	features := map[string]string{}
+	for _, policy := range policies {
+		// Ensure at least one cpu is using this policy
+		cpus, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpufreq", policy.Name(), "affected_cpus"))
+		if err != nil || strings.TrimSpace(string(cpus)) == "" {
```
dhellmann, line 53: Do we need to log the error for debugging?

source/cpu/pstate.go
```diff
@@ -32,12 +33,64 @@ func detectPstate() (map[string]string, error) {
...
+			continue
+		}
+
+		// Check if policy is using intel_pstate scaling driver
+		driver, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpufreq", policy.Name(), "scaling_driver"))
+		if err != nil {
+			continue
+		}
+
+		if strings.TrimSpace(string(driver)) == "intel_pstate" {
```
dhellmann, line 63: You can apply similar changes "happy path" changes to the logic here.

source/cpu/cstate.go
```diff
@@ -0,0 +1,57 @@
...
+	// When the intel_idle driver is in use (default), check setting of max_cstates
+	driver, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpuidle/current_driver"))
+	if err != nil {
+		return false, fmt.Errorf("cannot get driver for cpuidle: %s", err.Error())
+	}
+
+	if strings.TrimSpace(string(driver)) == "intel_idle" {
+		data, err := ioutil.ReadFile(source.SysfsDir.Path("module/intel_idle/parameters/max_cstate"))
+		if err != nil {
+			return false, fmt.Errorf("cannot determine cstate from max_cstates: %s", err.Error())
```
dhellmann, line 39: Other modules in this project use `errors.Wrap()` to extend error messages in this way. Using the same approach here would ensure that errors are formatted consistently.

source/cpu/cstate.go
```diff
@@ -0,0 +1,57 @@
...
+		if cstates > 0 {
+			return true, nil
+
+		} else {
+			return false, nil
+
+		}
+	} else {
+		// Currently only checking intel_idle driver for cstates
+		return false, fmt.Errorf("intel_idle driver is not in use: %s", string(driver))
```
dhellmann, line 55: Do we need to return an error in this case? Or is it enough to return false?

source/cpu/pstate.go
```diff
@@ -32,12 +33,64 @@ func detectPstate() (map[string]string, error) {
...
+	data, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/intel_pstate/status"))
+	if err != nil {
+		return nil, fmt.Errorf("could not read pstate status: %s", err.Error())
+	}
+	status := strings.TrimSpace(string(data))
+
+	// Check that intel_pstate driver is being used by looping through policies
+	policies, err := ioutil.ReadDir(source.SysfsDir.Path("devices/system/cpu/cpufreq"))
+	if err != nil {
+		return nil, fmt.Errorf("cpufreq directory not available to check pstate: %s", err.Error())
```
dhellmann, line 46: The phrasing of this error message assumes the reason for the error returned from ReadDir(). Could the problem be related to permissions, or something else other than the directory not being there? Maybe here wrap the existing error with a "failed to read cpufreq..." general message and let the error from ReadDir report the reason.

source/cpu/pstate.go
```diff
@@ -32,12 +33,64 @@ func detectPstate() (map[string]string, error) {
...
+		}
+
+		if strings.TrimSpace(string(driver)) == "intel_pstate" {
+			data, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/intel_pstate/status"))
+			if err != nil {
+				continue
+			}
+			status := strings.TrimSpace(string(data))
+			if status == "active" || status == "passive" {
+				features["enabled"] = "true"
```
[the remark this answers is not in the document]
bfournie: Fixed

source/cpu/cstate.go
bfournie: Fixed

source/cpu/cstate.go
bfournie: Done

source/cpu/cstate.go
bfournie: Good point, done

source/cpu/cstate.go
bfournie: I think its useful to return the driver string stored in current_driver, by default it should be intel_idle.

source/cpu/pstate.go
```diff
@@ -32,12 +35,68 @@ func detectPstate() (map[string]string, error) {
...
+
+		if status == "active" {
+			// Check powersave/performance setting
+			data, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpufreq", policy.Name(), "scaling_governor"))
+			scaling := strings.TrimSpace(string(data))
+			if err == nil {
+				if scaling == "powersave" || scaling == "performance" {
+					features["policy"] = scaling
+				}
+			}
```
marquiz, line 83: I think we need to check that all policies share the same governor and not just take what happens to be the governor of the last policy

source/cpu/pstate.go
```diff
@@ -32,12 +35,68 @@ func detectPstate() (map[string]string, error) {
...
+		// Check if policy is using intel_pstate scaling driver
+		driver, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpufreq", policy.Name(), "scaling_driver"))
+		if err != nil {
+			klog.Errorf("cpufreq scaling driver cannot be read: %s", err.Error())
+			continue
+		}
+
+		if strings.TrimSpace(string(driver)) != "intel_pstate" {
+			continue
+		}
```
marquiz, line 69: I'm not sure that we need/want to check this. Isn't the driver a global setting(?) That is, if `intel_pstate/status != off` then that implies that it is the acting scaling driver for all cpus. WDYT(?)

docs/get-started/features.md
```diff
@@ -61,7 +61,10 @@ The **cpu** feature source supports the following labels:
 | cpuid                   | &lt;cpuid flag&gt; | CPU capability is supported
 | hardware_multithreading |                    | Hardware multithreading, such as Intel HTT, enabled (number of logical CPUs is greater than physical CPUs)
 | power                   | sst_bf.enabled     | Intel SST-BF ([Intel Speed Select Technology][intel-sst] - Base frequency) enabled
-| [pstate][intel-pstate]  | turbo              | Set to 'true' if turbo frequencies are enabled in Intel pstate driver, set to 'false' if they have been disabled.
+| [pstate][intel-pstate]  | status             | The status of the Intel pstate driver when in use and enabled, either 'active' or 'passive'.
+|                         | turbo              | Set to 'true' if turbo frequencies are enabled in Intel pstate driver, set to 'false' if they have been disabled.
+|                         | scaling            | The value of the Intel pstate scaling_governor when in use, either 'powersave' or 'performance'.
```
marquiz, line 66: I'd just use the full original name `scaling_governor`. WDYT?

source/cpu/pstate.go
bfournie: Done

source/cpu/pstate.go
```diff
@@ -32,12 +35,68 @@ func detectPstate() (map[string]string, error) {
...
+				}
+			}
+		}
+
+		// No need to check other policies
+		break
+	}
+
+	if !(features["status"] == "active" || features["status"] == "passive") {
+		return nil, fmt.Errorf("intel_pstate driver not in use or disabled: %s", err.Error())
```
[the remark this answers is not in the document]
bfournie: Done

source/cpu/pstate.go
```diff
@@ -32,12 +35,68 @@ func detectPstate() (map[string]string, error) {
...
+			continue
+		}
+
+		if strings.TrimSpace(string(driver)) != "intel_pstate" {
+			continue
+		}
+
+		if status == "active" || status == "passive" {
+			features["status"] = status
+		}
```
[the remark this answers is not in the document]
bfournie: Done

source/cpu/pstate.go
```diff
@@ -32,12 +35,68 @@ func detectPstate() (map[string]string, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("cpufreq directory not available to check pstate: %s", err.Error())
+	}
+
+	features := map[string]string{}
+	for _, policy := range policies {
+		// Ensure at least one cpu is using this policy
+		cpus, err := ioutil.ReadFile(source.SysfsDir.Path("devices/system/cpu/cpufreq", policy.Name(), "affected_cpus"))
+		if err != nil || strings.TrimSpace(string(cpus)) == "" {
+			klog.Errorf("cpufreq policy %s has no affected cpus", policy.Name())
```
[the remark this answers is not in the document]
bfournie: Done

source/cpu/pstate.go
bfournie: Sounds good. Removed this check.

docs/get-started/features.md
bfournie: Sounds good. Changed to use scaling_governor.

Thanks for the review.

source/cpu/pstate.go
bfournie: OK, changed to check all policies and only set feature if all polices have the same scaling_governor.

ArangoGutierrez:
> If you ask me, this looks ready for merging. Thank you @bfournie for the hard work and you persistence smile
> /approve

Awesome! 
Any comments @kad @dhellmann ?

bfournie:
> If you ask me, this looks ready for merging. Thank you @bfournie for the hard work and you persistence 
> /approve

Thank you!

marquiz:
@dhellmann @zvonkok any comments? Feel free to lgtm

dhellmann:
I'm happy with this version. Thanks, @bfournie !

/lgtm

Issue #247, opened 2019-05-14
Better document the usage of hooks and feature files


marquiz:
Inspired by https://github.com/kubernetes-sigs/node-feature-discovery/pull/221#issuecomment-492280978

1. We should better document the runtime environment(s) supported for hooks (python, bash, c++, go), i.e. limitations of the NFD (default) image
2. Document examples for deploying hooks (`source.d`) and features files (`features.d`). E.g. (from @zvonkok's comment) "*to use a side-car container to inject feature labels into NFD*"

marquiz:
Not going to abandon this

/remove-lifecycle stale

marquiz:
Made an attempt at fixing this in #285

marquiz:
#285 addresses the two concerns mentioned in the issue description. Closing this issue as #285 was merged. Please re-open if you're not happy

PR #469, opened 2021-03-09
Dockerfile: add minimal image

marquiz:
Build a "minimal" variant of the nfd image based on
gcr.io/distroless/base. The motivations behind the minimal image are
image hardening (security) and reducing the image footprint (from ca.
108MB down to about 40MB).

The practical effect of deploying the minimal image is that no runtimes
for running worker hooks are present, not even a shell. This means that
only statically linked linked hook binaries are supported. Also, because
of the image hardening live debugging of the minimal image by attaching
to the container is not possible, and, the "full" image needs to be used
for that purpose.

This doesn't change the default usage/deployment of NFD in any way. Just adds a new image variant that can be alternatively deployed. The minimal variant might become the default in the future, though

### DIFF
--- a/docs/get-started/deployment-and-usage.md
+++ b/docs/get-started/deployment-and-usage.md
@@ -23,6 +23,27 @@ sort: 3
 1. [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl)
    (properly set up and configured to work with your Kubernetes cluster)
 
+## Image variants
+
+NFD currently offers two variants of the container image. The "full" variant is
+currently deployed by default.
+
+### Full
+
+This image is based on
+[debian:buster-slim](https://hub.docker.com/_/debian) and contains a full Linux
+system for running shell-based nfd-worker hooks and doing live debugging and
+diagnosis of the NFD images.
+
+### Minimal
+
+This is a minimal image based on
+[gcr.io/distroless/base](https://github.com/GoogleContainerTools/distroless/blob/master/base/README.md)
+and only supports running statically linked binaries.
+
+The container image tag has suffix `-minimal`
+(e.g. `{{ site.container_image }}-minimal`)
+
 ## Deployment options
 
 ### Operator
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,8 +19,20 @@ RUN make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
 RUN make test
 
 
-# Create production image for running node feature discovery
-FROM debian:buster-slim
+# Create full variant of the production image
+FROM debian:buster-slim as full
+
+# Run as unprivileged user
+USER 65534:65534
+
+# Use more verbose logging of gRPC
+ENV GRPC_GO_LOG_SEVERITY_LEVEL="INFO"
+
+COPY --from=builder /go/node-feature-discovery/nfd-worker.conf.example /etc/kubernetes/node-feature-discovery/nfd-worker.conf
+COPY --from=builder /go/bin/* /usr/bin/
+
+# Create minimal variant of the production image
+FROM gcr.io/distroless/base as minimal
 
 # Run as unprivileged user
 USER 65534:65534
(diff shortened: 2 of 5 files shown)


marquiz:
Pondering that should the minimal image be based on `vgcr.io/distroless/base`, instead. That would add a bit more flexibility to the hooks, supporting statically linked C binaries, for example.
/hold

zvonkok:
Makes sense. Moving forward if it becomes "the" default we should explicitly mention it in the documentation. We had https://github.com/kubernetes-sigs/node-feature-discovery/issues/247 for explaining the runtime options. When already minimal we could also add `readOnlyFilesystem: true` to the security context.

marquiz:
> Moving forward if it becomes "the" default we should explicitly mention it in the documentation.

Yes, sure. This PR adds a note about lack of runtimes in the minimal image .

ArangoGutierrez:
Debugging minimal images is...fun, but this PR looks good
/lgtm

marquiz:
@zvonkok @mythi @kad @adrianchiris any thoughts on `distroless/static` vs. `distroless/base`?

mythi:
> @zvonkok @mythi @kad @adrianchiris any thoughts on distroless/static vs. distroless/base

Would `distroless/static` set too tight restrictions to existing non-static source hooks, i.e., users cannot move to `minimal` without also updating their own hooks?

adrianchiris:
> @zvonkok @mythi @kad @adrianchiris any thoughts on distroless/static vs. distroless/base

So, going through this PR, i did not find the motivation to provide such an image (also no related issue on it).
e.g smaller footprint, security (is that obvious and should not be mentioned ?)

My main concern with this (and in the future moving to this image by default) are:

1. Possibly breaking existing hooks which may not run or may run and fail due to the change to minimal image. for sure these would need to be recompiled statically.
2. Feature sources cannot rely on OS tooling (i know they dont today, but this kinda shuts the door down the road)
3.  Debugging

In regards to static vs base : i would go with base as to give hooks that are not written in go a chance.

marquiz:
> So, going through this PR, i did not find the motivation to provide such an image (also no related issue on it).

Yeah, this was kind of quick spin-off from #221.

> e.g smaller footprint, security (is that obvious and should not be mentioned ?)

Yes, these are the motivations behind this. I need to explain that in the commit message.
 
> My main concern with this (and in the future moving to this image by default) are:
> 
> 1. Possibly breaking existing hooks which may not run or may run and fail due to the change to minimal image. These would need to be recompiled statically.

Yes. This might be too strict and `base` would be a better choice

> 2. Feature sources cannot rely on OS tooling (i know they dont today, but this kinda shuts the door down the road)

True. We have been deliberately avoiding this dependency. I think it's good as we (at least so far) have detected low-level features that are mainly available from kernel interfaces.

This doesn't shut the door for using such tools in the future, though (if we choose the `base` image). We just need to install the tool and the libraries it depends on. I actually think that this is also good as it really makes you consider what and how to contain in the image.

> 3. Debugging

The full image can be used for debugging. I think it's actually desirable from the security point of view to not be able to attach to running production containers. It vastly reduces the attack surface.

> In regards to static vs base : i would go with base as to give hooks that are not written in go a chance.

Yes, after sleeping over this PR I'm inclined to agree with this.

marquiz:
> Would `distroless/static` set too tight restrictions to existing non-static source hooks, i.e., users cannot move to `minimal` without also updating their own hooks?

Yes, I think so. Only supporting "fully static" binaries is probably unnecessarily strict, limiting hooks practically only to statically linked Go binaries. With `base` we will have gllibc which improves the support considerably

marquiz:
> My main concern with this (and in the future moving to this image by default)

Moving to minimal image by default must be carefully considered and I don't see it happening anytime soon. We're kind of held captive by the hooks, which I don't have any idea how widespread their usage is 🤔 Might be that nobody uses them and would even notice the change

marquiz:
Changed the base image to be base on distroless/base. Improved commit message of the first patch.

adrianchiris:
Thanks for addressing my comments @marquiz. 

/lgtm

marquiz:
Looks like there's a sort of consensus on this. And, the minimal image is optional, anyway.
/unhold

docs/get-started/deployment-and-usage.md
```diff
@@ -23,6 +23,27 @@ sort: 3
...
+system for running shell-based nfd-worker hooks and doing live debugging and
+diagnosis of the NFD images.
+
+### Minimal
+
+This is a minimal image based on
+[gcr.io/distroless/base](https://github.com/GoogleContainerTools/distroless/blob/master/base/README.md)
+and only supports running statically linked binaries.
+
+The container image tag has suffix `-minimal`
```
mythi, line 44: Forgot to add that should this section also explain how to generate the .yamls with this image: `make yamls IMAGE_NAME=node-feature-discovery-minimal`?

marquiz: Addressed that in a way in #471

PR #475, opened 2021-03-11
pkg/utils: show correct source file in gRPC logs

marquiz:
Unwind two call frames so that the source (file:line) of the log message
is correctly displayed.

### DIFF
--- a/pkg/utils/grpc_log.go
+++ b/pkg/utils/grpc_log.go
@@ -17,6 +17,8 @@ limitations under the License.
 package utils
 
 import (
+	"fmt"
+
 	"google.golang.org/grpc/grpclog"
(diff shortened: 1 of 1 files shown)


zvonkok:
Hmm, this is really pretty nice, I think I need to have a deep look into klog, I like the fact that you can set the frame depth dynamically for unwinding.  What is the default? 1 or complete dump of the stack frames?

marquiz:
0 is the default, i.e. you'll get the location of the actual klog.Func() call

ArangoGutierrez (commented):
After dealing with a "too many logs" issue on the operator, this type of PR's bring me joy
/lgtm

PR #483, opened 2021-03-17
Remove --sleep-interval flag from worker deployment templates

marquiz:
We don't want that as it makes dynamic configuration of the
sleep-interval via the config file impossible. It was set to the
internal default value, anyway.


ArangoGutierrez (approved):
wow we are all in core config way 
/lgtm

PR #486, opened 2021-03-18
scripts/update-gh-pages: fix helm repo update

marquiz:
Whitespace in release metadata json was not correctly handled.

Not surprisingly, we failed magnificently in production 🤦 
https://github.com/kubernetes-sigs/node-feature-discovery/runs/2137744191?check_suite_focus=true


adrianchiris:
Hmm, why didnt it fail when testing this on a fork ?

adrianchiris (commented):
reproduced locally by having the `GITHUB_*` env variables pointing to the upstream repo
it works as expected.

/lgtm

PR #488, opened 2021-03-18
github: two more steps to release process

marquiz:


.github/ISSUE_TEMPLATE/new-release.md
```diff
@@ -38,6 +38,10 @@ Please do not remove items from the checklist
 - [ ] Add a link to the tagged release in this issue.
 - [ ] Send an announcement email to `kubernetes-dev@googlegroups.com` with the subject `[ANNOUNCE] node-feature-discovery $VERSION is released`
 - [ ] Add a link to the release announcement in this issue
+- [ ] For a major release, update README in master branch
+  - [ ] Update references to point to `$VERSION` and submit a PR
```
adrianchiris, line 42: what are these references ?

should this be : run `prepare_release.sh $VERSION <KEY ID>` and only commit the README ?

.github/ISSUE_TEMPLATE/new-release.md
```diff
@@ -38,6 +38,10 @@ Please do not remove items from the checklist
 - [ ] Add a link to the tagged release in this issue.
 - [ ] Send an announcement email to `kubernetes-dev@googlegroups.com` with the subject `[ANNOUNCE] node-feature-discovery $VERSION is released`
 - [ ] Add a link to the release announcement in this issue
+- [ ] For a major release, update README in master branch
```
adrianchiris, line 41: what about a minor release ? e.g once v0.8.0 is released, say we need to release `v0.8.1` what will update helm chart in this case ?

.github/ISSUE_TEMPLATE/new-release.md
marquiz: > should this be : run `prepare_release.sh $VERSION <KEY ID>` and only commit the README ?
> 
> i assume it to replace any instance of current release with the new release e.g 0.7.0 -> 0.8.0

Basically yes, can be done with `prepare_release.sh`, too

.github/ISSUE_TEMPLATE/new-release.md
marquiz: Haven't had any, yet 🤞  But yes, a better solution would be nice

marquiz: Found out that we can actually trigger on publish release event 🤦 
https://github.com/kubernetes-sigs/node-feature-discovery/pull/489

PR #491, opened 2021-03-18
docs: describe Helm repo as the primary option for Helm

marquiz:


adrianchiris:
Now that we have the first release with Helm support, this doc change is certainly welcome.

/lgtm

PR #495, opened 2021-03-18
[helm] fix nfd worker tolerations value

moshe010:
the tolerations value is array therefore we need to pass
[] and not {}

Signed-off-by: Moshe Levi <moshele@nvidia.com>


marquiz:
Thanks @moshe010 
/ok-to-test
/approve

PR #496, opened 2021-03-23
Helm chart: Fix configMap indenting

bjw-s:
Currently when the `worker.config` key is being filled with a YAML string literal, it does not render properly and causes Helm errors. This is due to the way the indenting and whitespace trimming is handled.

This PR aims to fix that, by stripping the leading whitespace from the `worker.config` value, and forcing a newline.

Expected to work (snippet):
```yaml
worker:
  config: |-
    core:
      sources:
      - custom
      - pci
      - usb
    sources:
      usb:
```

Currently required to work (snippet), notice the indenting of `core:` vs the rest of the config file:
```yaml
worker:
  config: |-
    core:
        sources:
        - custom
        - pci
        - usb
      sources:
        usb:
```

### DIFF
--- a/deployment/node-feature-discovery/Chart.yaml
+++ b/deployment/node-feature-discovery/Chart.yaml
@@ -12,4 +12,4 @@ keywords:
   - feature-detection
   - node-labels
 type: application
-version: 0.1.0
+version: 0.1.1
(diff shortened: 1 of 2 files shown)


marquiz:
@bjw-s thanks for your patch!

/ok-to-test

marquiz (approved):
/lgtm

PR #502, opened 2021-03-31
scripts/prepare-release: option to only create assets

marquiz:
Add '-a' command line option to only create release assets and not patch
any files in the repo. E.g.

  $ scripts/prepare-release.sh -a v0.8.0+1 user@example.com

### DIFF
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -9,6 +9,7 @@ Usage: $this [-h] RELEASE_VERSION GPG_KEY
 
 Options:
   -h         show this help and exit
+  -a         only generate release assets, do not patch files in the repo
 
 Example:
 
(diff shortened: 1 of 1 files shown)


scripts/prepare-release.sh
```diff
@@ -79,34 +83,36 @@ else
...
+    echo Patching README.md to refer to $release
+    sed s"!node-feature-discovery/v.*/!node-feature-discovery/$release/!" -i README.md
+
+    # Patch deployment templates
+    echo Patching '*.yaml.template' to use $container_image
+    sed -E -e s",^([[:space:]]+)image:.+$,\1image: $container_image," \
+           -e s",^([[:space:]]+)imagePullPolicy:.+$,\1imagePullPolicy: IfNotPresent," \
+           -i *yaml.template
+
+    # Patch Helm chart
```
adrianchiris, line 104: No need to patch helm chart before creating the assets ?

marquiz: > No need to patch helm chart before creating the assets ?

Nope because we don't want to change the `appVersion`. Image pull policy and image registry have already been patched in the release branch

adrianchiris (commented):
/lgtm

Issue #511, opened 2021-04-14
When --verify-node-name is on, respect either SubjectAlternativeName and/or wildcards


robertdavidsmith:
Hi,

Using --verify-node-name is currently pretty painful, while it's easy enough to produce a certificate per node with Helm/CertManager, there is no clean way to get a different certificate into every pod with a DaemonSet.

It would be better if one certificate could cover all nodes.

Can we have an alternative way of securing this?
- CommonName wildcard support
- SubjectAlternativeName support
- An alternative auth mechanism, maybe based on a shared secret

I can put in a PR to change https://github.com/kubernetes-sigs/node-feature-discovery/blob/ab78df346679a6df0e1f5286874fb191022404ba/pkg/nfd-master/nfd-master.go#L369 to support wildcards or SubjectAlternativeName. Would you accept such as PR?

Thanks,

Rob

marquiz:
@robertdavidsmith thanks for bringing this up

> Using --verify-node-name is currently pretty painful, while it's easy enough to produce a certificate per node with Helm/CertManager, there is no clean way to get a different certificate into every pod with a DaemonSet.

I'm painfully aware of this 😓 

I think `--verify-node-name` is actually very badly out-of-date as SSL common name is kind of deprecated 😬 

> It would be better if one certificate could cover all nodes.
> 
> Can we have an alternative way of securing this?
> 
> * CommonName wildcard support
> * SubjectAlternativeName support
> * An alternative auth mechanism, maybe based on a shared secret
> 
> I can put in a PR to change
> 
> https://github.com/kubernetes-sigs/node-feature-discovery/blob/ab78df346679a6df0e1f5286874fb191022404ba/pkg/nfd-master/nfd-master.go#L369
> 
> to support wildcards or SubjectAlternativeName.

I think we should definitely add support for SubjectAlternativeName. How about wildcards in SubjectAlternativeName? We could keep the CommonName for some time for backwards compatibility but I wouldn't add wildcard support there. Regarding the  additional auth mechanisms, I think those could be left as a future exercise. Add the later if needed. WDYT?


> Would you accept such as PR?

Definitely! That would be much appreciated 😊 👍

robertdavidsmith:
Ok, I'll do this, might be a few days though due to other commitments

Suggest just using https://golang.org/pkg/crypto/x509/#Certificate.VerifyHostname - this uses SubjectAlternativeName, and it looks like it supports limited wildcards (* on left hand side only).

[...]

PR #514, opened 2021-04-16
Accept client certs based on SAN, not just CN

robertdavidsmith:
Fix for https://github.com/kubernetes-sigs/node-feature-discovery/issues/511

When VerifyNodeName is on, this change respects SubjectAltName (rather than just CN) in the client cert 

The VerifyHostname method used respects wildcards on the left hand side, so for example a cert with a SAN of *.mycluster.org would match node1.mycluster.org

I've left in the existing common name check to avoid breaking existing clients (Certificate.VerifyHostname does not respect CN).


marquiz:
Cool @robertdavidsmith 😎 Will take a closer look next week
/ok-to-test

You need to sign the cla, too

marquiz (approved):
Nice work @robertdavidsmith! Oh man this was simple and clean 😊 
/lgtm

PR #519, opened 2021-04-21
Mount /usr inside the Pod

jschintag:
Mount /usr as /host-usr inside the pod to allow NFD to find the kernel
configuration file inside /usr.
This solves the problem of the kernel config file not being present in /boot
on s390x RHCOS.

### DIFF
--- a/nfd-daemonset-combined.yaml.template
+++ b/nfd-daemonset-combined.yaml.template
@@ -104,6 +104,12 @@ spec:
             - name: host-sys
               mountPath: "/host-sys"
               readOnly: true
+            - name: host-usr-lib
+              mountPath: "/host-usr/lib"
+              readOnly: true
+            - name: host-usr-src
+              mountPath: "/host-usr/src"
+              readOnly: true
             - name: source-d
               mountPath: "/etc/kubernetes/node-feature-discovery/source.d/"
               readOnly: true
@@ -128,6 +134,12 @@ spec:
         - name: host-sys
           hostPath:
             path: "/sys"
+        - name: host-usr-lib
+          hostPath:
+            path: "/usr/lib"
+        - name: host-usr-src
+          hostPath:
+            path: "/usr/src"
         - name: source-d
           hostPath:
             path: "/etc/kubernetes/node-feature-discovery/source.d/"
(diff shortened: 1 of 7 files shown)


nfd-daemonset-combined.yaml.template
```diff
@@ -104,6 +104,9 @@ spec:
             - name: host-sys
               mountPath: "/host-sys"
               readOnly: true
+            - name: host-usr
```
adrianchiris, line 107: Greetings !

can you update : https://github.com/kubernetes-sigs/node-feature-discovery/blob/09cb512f22bae327ac38699f63312728c39ceb38/deployment/node-feature-discovery/templates/worker.yaml#L60  as well ?

jschintag: Done. Thanks for the catch

marquiz:
Thanks for the patch @jschintag!

I just ponder if we should follow the principle of "least privileges" and only expose `/usr/lib` and `/usr/src`. Opinions?

jschintag:
@marquiz I would say that this can be decided by the user, as it only affects the deployment configuration, not the code.
What about `/usr/lib64`?

marquiz:
> @marquiz I would say that this can be decided by the user, as it only affects the deployment configuration, not the code.
> What about `/usr/lib64`?

Yeah, sure. But the default deployment that we offer should only mount/expose what is really necessary. I don't this we want `lib64`, I don't see any reason to to access actual libraries from the host

jschintag:
> > @marquiz I would say that this can be decided by the user, as it only affects the deployment configuration, not the code.
> > What about `/usr/lib64`?
> 
> Yeah, sure. But the default deployment that we offer should only mount/expose what is really necessary. I don't this we want `lib64`, I don't see any reason to to access actual libraries from the host

I changed the default to only mount `/usr/lib` and `/usr/src`

adrianchiris:
nit: appreciate if you update the commit message to reflect the new mounts.

otherwise no other comments :)

jschintag:
I changed the commit message to reflect the change.

adrianchiris (approved):
/lgtm

marquiz:
@jschintag thanks for the patch! This was a fix definitely needed.
/approve

PR #523, opened 2021-05-12
WIP: Add detection of IBM Z Crypto Express Cards

jschintag:
Add feature detection for IBM Crypto Express Cards.
Detect the available crypto cards, their type and the assigned domains.

### DIFF
--- a/source/crypto/cex_s390x.go
+++ b/source/crypto/cex_s390x.go
@@ -1,0 +1,77 @@
+// +build s390x
+
+/*
+Copyright 2021 The Kubernetes Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package crypto
+
+import (
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+
+	"k8s.io/klog/v2"
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+type CEX struct {
+	Name string
+	Type string
+}
+
+func discoverCEX() ([]*CEX, error) {
+	cards := []*CEX{}
+	sysfsBasePath := source.SysfsDir.Path("bus/ap/devices")
+
+	devices, err := ioutil.ReadDir(sysfsBasePath)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return []*CEX{}, nil
+		} else {
+			return nil, err
+		}
+	}
+
+	// Iterate over devices
+	for _, device := range devices {
+		if !strings.HasPrefix(device.Name(), "card") {
+			continue
+		}
+		card, err := readCex(sysfsBasePath, device.Name())
+		if err != nil {
+			klog.Errorf("Failed to read Crypto Express %s: %v", device.Name(), err)
+			continue
+		}
+		cards = append(cards, card)
+	}
+
+	return cards, nil
+}
+
+func readCex(basePath string, name string) (*CEX, error) {
+	card := new(CEX)
+	cardPath := path.Join(basePath, name)
+
+	card.Name = name
+	data, err := ioutil.ReadFile(cardPath + "/type")
+	if err != nil {
+		return nil, err
+	}
+	cardType := string(data)
+	card.Type = strings.Trim(cardType, "\n")
+	return card, nil
+}
--- a/docs/get-started/features.md
+++ b/docs/get-started/features.md
@@ -525,6 +525,14 @@ The **system** feature source supports the following labels:
 |             | VERSION_ID.major | First component of the OS version id (e.g. '6')
 |             | VERSION_ID.minor | Second component of the OS version id (e.g. '7')
 
+### Crypto
+
+The **crypto** feature source supports the following labels:
+
+| Feature     | Attribute        | Description                                 |
+| ----------- | ---------------- | --------------------------------------------|
+| cex         | present          | Node has at least one Crypto Express Adapter
+
 ### Local -- user-specific features
 
 NFD has a special feature source named *local* which is designed for getting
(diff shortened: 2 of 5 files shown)


jschintag:
TODO:

- [ ] Add documentation

source/crypto/cex_s390x.go
```diff
@@ -0,0 +1,91 @@
+// +build s390x
+
+/*
+Copyright 2021 IBM
```
marquiz, line 4: In order to merge this, the copyright should be changed, s/IBM/The Kubernetes Authors/

jschintag: Thank you, i changed it.

jschintag:
I've added the documentation.

One question i have is if it should be it's own source as it now is, or wether it might be added to system? Though i don't know if system should be a sort of catch-all for features that don't fit into other sources.

marquiz:
> I've added the documentation.

Thanks!
 
> One question i have is if it should be it's own source as it now is, or wether it might be added to system? Though i don't know if system should be a sort of catch-all for features that don't fit into other sources.

Not under `system`. However, I was thinking about the name of the source myself. I'm not at all familiar with the `/sys/bus/ap` devices. Is it a separate physical bus so that the devices are not visible under pci bus for example? Is it something more widely used or specific to IBM Z and the crypto applications? Also, I was pondering the name `cex`, is it something "commonly known" or very specific to one driver/device vendor?

What is the designed use case/usage scenario for this? For example, are these labels supposed to be used by end-user workloads directly or by a device plugin or something else?

It also looks like this patch has a potential to spit out a long list of labels. Just wondering are all the details necessary?

zvonkok:
@marquiz Looking at this PR I wondered do we want to have better management of arch specific features. Going forward we will have more architectures and obviously more arch specifics to expose. 

In some cases, we just print out an error on the console if a feature is not available. Should we think of disabling code-path on some archs where they do not make any sense? 

In good old C/C++ days we just used `#ifdef`s on the arch .. :)

zvonkok:
Several times when I looked at the output of NFD I thought I was doing something wrong having so many messages. Adding more features that will just error our on most of the cases could overload the logging .

jschintag:
> Not under `system`. However, I was thinking about the name of the source myself. I'm not at all familiar with the `/sys/bus/ap` devices. Is it a separate physical bus so that the devices are not visible under pci bus for example? Is it something more widely used or specific to IBM Z and the crypto applications? Also, I was pondering the name `cex`, is it something "commonly known" or very specific to one driver/device vendor?
The Crypto Cards under the `/sys/bus/ap` are not visible under pci bus. I think this https://www.kernel.org/doc/html/latest/s390/vfio-ap.html# might explain it in more detail.
The Name CEX refers to the IBM Crypto Express Adapters.
https://www.ibm.com/security/cryptocards

> What is the designed use case/usage scenario for this? For example, are these labels supposed to be used by end-user workloads directly or by a device plugin or something else?
The labels are supposed to allow workloads to be scheduled on nodes with crypto capabilities. This can either happen through the user directly or be automated by custom schedulers.

> It also looks like this patch has a potential to spit out a long list of labels. Just wondering are all the details necessary?
This could indeed split out a long list of labels for a node. However for cryptographic operations it would be necessary to know what mode the card is currently in (Coprocessor, Accelerator or EP11), as well at what domains are available, as they have individual master keys and unless configured with the same master keys, stored keys would not be portable between domains.

jschintag:
> Several times when I looked at the output of NFD I thought I was doing something wrong having so many messages. Adding more features that will just error our on most of the cases could overload the logging .

@zvonkok Do you see any errors/new messages with this PR? I haven't tested it on archs other than s390x, but it should not do anything on them.

zvonkok:
@jschintag Can you provide here an output on how this would look like? I am following also another use-case where such things like the CEX labels could help. 

Will check it on my systems, too. I was relating to other features that log an Error if some things are missing or not configured.

jschintag:
@zvonkok Here. Though i doubt anyone would have that many cards. At most 3 cards, one for each mode. I have here multiple generations for testing purposes. As you can see they all have only one domain attached to the host.
```
                    feature.node.kubernetes.io/crypto-cex.card00=CEX6A
                    feature.node.kubernetes.io/crypto-cex.card01=CEX5A
                    feature.node.kubernetes.io/crypto-cex.card02=CEX5C
                    feature.node.kubernetes.io/crypto-cex.card03=CEX5P
                    feature.node.kubernetes.io/crypto-cex.card04=CEX7A
                    feature.node.kubernetes.io/crypto-cex.card06=CEX7C
                    feature.node.kubernetes.io/crypto-cex.card08=CEX7P
                    feature.node.kubernetes.io/crypto-cex.card0a=CEX7A
                    feature.node.kubernetes.io/crypto-cex.domain_00.003d=true
                    feature.node.kubernetes.io/crypto-cex.domain_01.003d=true
                    feature.node.kubernetes.io/crypto-cex.domain_02.003d=true
                    feature.node.kubernetes.io/crypto-cex.domain_03.003d=true
                    feature.node.kubernetes.io/crypto-cex.domain_04.003d=true
                    feature.node.kubernetes.io/crypto-cex.domain_06.003d=true
                    feature.node.kubernetes.io/crypto-cex.domain_08.003d=true
                    feature.node.kubernetes.io/crypto-cex.domain_0a.003d=true
                    feature.node.kubernetes.io/crypto-cex.present=true
```

marquiz:
> @marquiz Looking at this PR I wondered do we want to have better management of arch specific features. Going forward we will have more architectures and obviously more arch specifics to expose.
> 
> In some cases, we just print out an error on the console if a feature is not available. Should we think of disabling code-path on some archs where they do not make any sense?

Ha, how come I was thinking about exactly the same annoyance when looking at this patch 😅  I think it's not that much about arch-specific code but just system configurations and hw-differences which we do not handle in a precise manner.

Anyway, I think we need to do something about that. Those errors are very annoying and misleading (e.g. many systems don't have selinux and those always get an selinux related error 😬 ). Most of those are caused by old-ish code, sort of legacy.

jschintag:
Regarding the labels, i had a discussion with a few of my colleagues, we will take a look at the labels and see if we need to change them for the CEX enablement on s390x.

I'll keep the PR as WIP for now.

marquiz:
@jschintag still working on this?

jschintag:
At the moment i'm not working on this. It might be picked up again in the future, but right now CEX and Kubernetes are more at Home here https://github.com/ibm-s390-cloud/k8s-cex-dev-plugin

marquiz:
> At the moment i'm not working on this. It might be picked up again in the future, but right now CEX and Kubernetes are more at Home here https://github.com/ibm-s390-cloud/k8s-cex-dev-plugin

👍 Thanks for the update.

You might want some detection of the hardware in order to "optimize" the deployment, only deploying the device plugin on the nodes that actually have the HW it is managing(?)

PR #524, opened 2021-05-13
Fix nfd worker tolerations value

e0ne:
Tolerations value is array therefore we need to pass
[] and not {} as a default value.


marquiz (changes requested):
Hmm, this patch is doing exactly the opposite?? 🤔

marquiz:
This should've been fixed by https://github.com/kubernetes-sigs/node-feature-discovery/commit/57575231ab9fd7dba205e048f0b1a86d22444c30

PR #536, opened 2021-05-26
nfd-master: allow sub-namespaces of the default label ns

### LINKED ISSUE #477: Allow sub-namespaces of default label namespace

We could allow `*.feature.node.kubernetes.io` sub-namespaces for feature labels, in addition to the default `feature.node.kubernetes.io`. That is, these namespaces would be automatically allowed, without any extra configuration/options. Arbitrary namespaces (like `vendor.com/`) could be enabled with `-extra-label-ns` just like before.

**Why is this needed**:

Make it easier to do vendor/user specific differentiation/partitioning.

marquiz:
Allow <sub-ns>.feature.node.kubernetes.io label namespaces. Makes it
possible to have e.g. vendor specific label ns without the need to user
-extra-label-ns.

Fixes #477

### DIFF
--- a/docs/advanced/master-commandline-reference.md
+++ b/docs/advanced/master-commandline-reference.md
@@ -175,9 +175,10 @@ nfd-master -label-whitelist='.*cpuid\.'
 
 The `-extra-label-ns` flag specifies a comma-separated list of allowed feature
 label namespaces. By default, nfd-master only allows creating labels in the
-default `feature.node.kubernetes.io` label namespace. This option can be used
-to allow vendor-specific namespaces for custom labels from the local and custom
-feature sources.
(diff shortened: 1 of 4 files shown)


ArangoGutierrez (changes requested):
PR looks good, maybe the only nit would be to add a comment on documentation about this? (let's keep documentation up to date )

marquiz:
> PR looks good, maybe the only nit would be to add a comment on documentation about this? (let's keep documentation up to date )

Hmm, very good point! Need to fix that. I'm pretty sure I've written the documentation once but clearly not committed it 🤔  😬

marquiz:
Documentation now added

ArangoGutierrez (approved):
/lgtm

PR #541, opened 2021-06-10
Makefile: add apigen target

marquiz:
For auto-generating api(s).

Also, re-generate/refresh the gRPC with `make apigen` (with protoc
v3.17.3 and protoc-gen-go from github.com/golang/protobuf v1.5.2) to
sync up things.


pkg/labeler/labeler.pb.go
```diff
@@ -1,155 +1,300 @@
+//
+//Copyright 2019 The Kubernetes Authors.
```
ArangoGutierrez, line 2: 2021?

marquiz: This is autogenerated file. The year is copied from `labeler.proto`. I don't think it makes sense to modify the proto file itself just to update the copyright year

ArangoGutierrez (approved):
/lgtm

PR #549, opened 2021-07-07
Makefile: add lint target

marquiz:
Shortcut for running golint.


ArangoGutierrez:
Maybe we should address some of them.... 😅 
```bash
make lint
[---]
Found 121 lint suggestions; failing.
make: *** [Makefile:139: lint] Error 1
```

ArangoGutierrez (approved):
/lgtm

Issue #468, opened 2021-03-09
Support CRDs for custom label rules


marquiz:
**What would you like to be added**:

An idea I've been pondering to make adding new rules for the `custom` source easier, without the need to add any new mounts or anything. NFD would specify a new CRD and watch for objects. Not sure what would be the best way to implement this:

1. NFD worker instances would watch the CRs directly themselves
1. NFD master would watch the CRs and communicate rules via gRPC
1. NFD master would watch the CRs and update a predefined ConfigMap that is mounted on the worker containers
1. Do rule processing on master side: NFD worker sends "raw" features to master. Master watches CRs, performs rule processing/matching against the "raw" features and creates labels.

Option 1 would be the obvious simplest choice even if it opens up direct communication of the worker to the control plane. 

**Why is this needed**:

Make dropping in vendor/user specific custom label rules easier, as simple as `kubectl apply -f <rules.yaml>`

**List of individual PRs**:

- ~#543~
- ~#544~
- ~#596~
- ~#597~
- ~#601~
- ~#596~
- ~#642~
- ~#604~
- ~#639~
- ~#464~
- ~#646~
- ~#653~
- ~#654~
- ~#655~
- ~#656~
- ~#553~

Adjacent PRs, building on top and extending the functionality:

- ~#649~
- ~#660~
- ~#661~
- ~#550~
- ~#663~
- #605
- #662

zvonkok:
etcd has the 1.5 MiB object limit size which corresponds to roughly 750 pages of text. I think this should be enough to hold enough rules.

zvonkok:
Regarding 1. does it mean we would have 1 CR that all nfd-workers would read and report the labels ... If we have several hundreds of nodes, how will the API behave is several hundred nfd-workers are reading from the very same CR? 

Should we add a random startup time?

marquiz:
> Regarding 1. does it mean we would have 1 CR that all nfd-workers would read and report the labels ... 

Yes, all workers will need read that config (CR)

> If we have several hundreds of nodes, how will the API behave is several hundred nfd-workers are reading from the very same CR at the same time? Should we add a random startup time?

This is a fair concern. Unfortunately I don't have access to such a cluster. It's not just startup time but any time a new CR is added (or old one updated)

marquiz:
> Regarding 3. only the master would read the CR and the load would be distributed to each node, update of the CM...

Yes. Point of stress would be moved to the configmap

[...]

PR #553, opened 2021-07-08
CRD-based custom node labeling

marquiz:
Enable Custom Resource based label creation in nfd-master. This extends
the previously implemented controller stub for watching NodeFeatureRule
objects. NFD-master watches NodeFeatureRule objects in the cluster and
processes the rules on every incoming labeling request from workers.
The functionality relies on the "raw features" (identical to how
nfd-worker handles custom rules) submitted by nfd-worker, making it
independent of the label source configuration of the worker. This means
that the labeling functions as expected even if all sources in the
worker are disabled.

NOTE: nfd-master is stateless and re-labeling only happens on the
reception of SetLabelsRequest from the worker – i.e. on intervals
specified by the core.sleepInterval configuration option (or
-sleep-interval cmdline flag) of each nfd-worker instance. This means
that modification/creation of NodeFeatureRule objects does not
automatically update the node labels. Instead, the changes only come
visible when workers send their labeling requests.

This PR implements #468, i.e. CRD-based custom node labeling, and builds on top of previous work:
- #464
- #646
- #653
- #654
- #655
- #656

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -39,6 +39,7 @@ import (
 	api "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
 
@@ -179,6 +180,7 @@ func (m *nfdMaster) Run() error {
 		if err != nil {
 			return err
 		}
+		klog.Info("starting nfd LabelRule controller")
 		m.nfdController = newNfdController(kubeconfig)
 	}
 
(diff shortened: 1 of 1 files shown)


marquiz:
All feedback is very much appreciated. Especially on the soundness of the whole concept.

Some concerns/questions from the top of my head below, just talking out loud...

### Data format

I've been pondering the data representation of "raw" features, i.e. is it good (enough). In this patchset NFD now under the hood uses a unified data model of all the "raw features" (not talking about labels here) it is detecting.

Basically we divide feature types into three categories:

1. "key-only features": features that don't have any value other than "present" or "not present". Examples are CPUID flags or loaded kernel nodules
2. "value features": features that have a value associated with it. Examples are kernel config flags or os-release information.
3. "instance features": features that cannot be represented by a single key-value pair (or key-only) but consist of mulitple attributes (key-value pairs). Examples are pci and usb devices

On the higher level, data representation of all detected features (from all features sources) is fixed into a three level hierarchy:
`<source-name>.<feature-name>.<single-feature-instance>`.

In go code this is a bit more complicated looks something like this (in a pseudo-ish golang):

```
// Name of the feature source e.g. "cpu" or "pci"
type FeatureSourceName string

// Name of the feature under the feature source e.g. "cpuid" (cpu) or "device" (pci)
type FeatureName string

// All features from all the feature sources
type AllFeatures map[FeatureSourceName]struct {
	// "key-only" features like cpuid or loaded kernel modules
	Keys map[FeatureName]struct {
		Features map[string]struct{}
	}

	// "key-value" features e.g. kernel version or kernel config flags
	Values map[FeatureName]struct {
		Features map[string]string
	}

	// "instance features" like pci or usb devices
	Instances map[FeatureName]struct {
		Features []struct {
			Attributes map[string]string
		}
	}
}
```

This data representation trickles down everywhere and to the rule format and CRD, too. The unified representation is nice in the sense that it reduces the lines of code dramatically and also simplifies the CRD schema plus makes it stable. Basically any new features added to NFD is automatically supported (because of the unified data representation and unified rule schema) with essentially no changes required in the CRD spec or rules.

My only concern is that is this flexible enough as we're fixed with the three-level hierarchy and cannot specify arbitrary data structures for the features. For example, Intel Speed Select Technology we now have `cpu["sst"].Features["bf.enabled"]` even though `cpu.sst.bf.enabled` (just nested structs and a bool in the end) might be more logical representation. However, the downside is that every feature needs a rule type of its own, similar to what we currently have in master (the custom rules). This also means that any added feature potentially changes the CRD API making is potentially way messier.

### MatchExpressions

The `key` attribute of matchexpressions is implicit os the set of expressions are are presented as a map - the map key is used as the `key` attribute of the matchexpression. This has the limitation that you cannot specify more than one expression per key. I originally thought that this does not matter but now realized that there is actually one case where you would need this property: if you want to check that a value is between two values - in this case you'd need both `Lt` and `Gt` expressions. Talking out loud: maybe this should be changed...

### Naming

I'm not happy with all the terms/naming that I've come up with. Need to think through it.

ArangoGutierrez:
Not yet ready for review? 
I want to jump into this PR as soon as it moves out of draft !

marquiz:
> Not yet ready for review?
> I want to jump into this PR as soon as it moves out of draft !

#464 would be the first part. But even that is not ready yet, missing unit tests and documentation. I will work on these in the coming weeks, for sure, as I want to get these finally in shape, too.

marquiz:
Removing draft status. However, I'll split multiple smaller PRs out of this
/hold

marquiz:
Still based on #464 but I dropped all the non-essential/auxiliary/adjacent stuff (like templating, backreferences etc) to make this PR more focused. I will still split out smaller pieces of this. The "adjacencies" are not lost either - I will submit those as separate PRs, too.

ArangoGutierrez:
@marquiz rebase needed

marquiz:
#656 was merged -> rebased and PR description updated
/unhold

PR #555, opened 2021-08-02
go.mod: update to klauspost/cpuid/v2@v2.0.9

mythi:
The new version adds AVX512-FP16 (half precision floating-point)
support available on Sapphire Rapids.

Signed-off-by: Mikko Ylinen <mikko.ylinen@intel.com>


ArangoGutierrez (approved):
/lgtm

PR #563, opened 2021-08-10
Add Readiness and liveliness probes to nfd-master

### LINKED ISSUE #522: Readiness and liveliness probes
Readiness and Liveliness probes for the daemonset and the manager
**Why is this needed**:
Good practice

ArangoGutierrez:
Fixes #522 

This patch seeks to:

- https://github.com/kubernetes-sigs/node-feature-discovery/commit/273cfffe62936300f575bca72205bff0cc8bae3d enable readinessprobe for nfd-master
- https://github.com/kubernetes-sigs/node-feature-discovery/pull/563/commits/f4e8e20ea5c963c0cb4610f702abf0ed889b90a9 add livenessprobe for nfd-master

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,11 @@ ARG BASE_IMAGE_MINIMAL
 # Build node feature discovery
 FROM golang:1.16.7-buster as builder
 
+# Download the grpc_health_probe bin
+RUN GRPC_HEALTH_PROBE_VERSION=v0.3.1 && \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
+    chmod +x /bin/grpc_health_probe
+
 # Get (cache) deps in a separate layer
 COPY go.mod go.sum /go/node-feature-discovery/
 
@@ -21,7 +26,6 @@ RUN make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
 
 RUN make test
 
-
 # Create full variant of the production image
 FROM ${BASE_IMAGE_FULL} as full
 
@@ -33,6 +37,7 @@ ENV GRPC_GO_LOG_SEVERITY_LEVEL="INFO"
 
 COPY --from=builder /go/node-feature-discovery/deployment/components/worker-config/nfd-worker.conf.example /etc/kubernetes/node-feature-discovery/nfd-worker.conf
 COPY --from=builder /go/bin/* /usr/bin/
+COPY --from=builder /bin/grpc_health_probe /usr/bin/grpc_health_probe
 
 # Create minimal variant of the production image
 FROM ${BASE_IMAGE_MINIMAL} as minimal
@@ -45,3 +50,4 @@ ENV GRPC_GO_LOG_SEVERITY_LEVEL="INFO"
 
 COPY --from=builder /go/node-feature-discovery/deployment/components/worker-config/nfd-worker.conf.example /etc/kubernetes/node-feature-discovery/nfd-worker.conf
 COPY --from=builder /go/bin/* /usr/bin/
+COPY --from=builder /bin/grpc_health_probe /usr/bin/grpc_health_probe
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -36,6 +36,8 @@ import (
 	api "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 
+	"google.golang.org/grpc/health"
+	"google.golang.org/grpc/health/grpc_health_v1"
 	"sigs.k8s.io/node-feature-discovery/pkg/apihelper"
 	pb "sigs.k8s.io/node-feature-discovery/pkg/labeler"
 	"sigs.k8s.io/node-feature-discovery/pkg/utils"
@@ -186,6 +188,7 @@ func (m *nfdMaster) Run() error {
 	}
 	m.server = grpc.NewServer(serverOpts...)
 	pb.RegisterLabelerServer(m.server, m)
+	grpc_health_v1.RegisterHealthServer(m.server, health.NewServer())
 	klog.Infof("gRPC server serving on port: %d", m.args.Port)
 
 	// Run gRPC server
--- a/deployment/base/master-worker-combined/master-worker-daemonset.yaml
+++ b/deployment/base/master-worker-combined/master-worker-daemonset.yaml
@@ -19,6 +19,17 @@ spec:
       - name: nfd-master
         image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
         imagePullPolicy: Always
+          livenessProbe:
+            exec:
+              command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
+            initialDelaySeconds: 10
+            periodSeconds: 10
+          readinessProbe:
+            exec:
+              command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            failureThreshold: 10
         command:
         - "nfd-master"
       - name: nfd-worker
--- a/deployment/base/master/master-deployment.yaml
+++ b/deployment/base/master/master-deployment.yaml
@@ -20,6 +20,17 @@ spec:
         - name: nfd-master
           image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
           imagePullPolicy: Always
+          livenessProbe:
+            exec:
+              command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
+            initialDelaySeconds: 10
+            periodSeconds: 10
+          readinessProbe:
+            exec:
+              command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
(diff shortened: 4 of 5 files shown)


zvonkok:
Hmm, why are we introducing a new HTTP server when we can do grpc liveness probes? https://kubernetes.io/blog/2018/10/01/health-checking-grpc-servers-on-kubernetes/

pkg/nfd-master/nfd-master.go
```diff
@@ -195,6 +197,21 @@ func (m *nfdMaster) Run() error {
 		grpcErr <- m.server.Serve(lis)
 	}()
 
+	// We are ready to start listening
+	// Readiness probe
+	r := ready.NewFileReady("nfd-master")
+	err = r.Set()
+	if err != nil {
+		klog.Error(err, "")
+		os.Exit(1)
```
Ethyling, line 206: Do we really want to exit instead of returning the error?

nfd-daemonset-combined.yaml.template
```diff
@@ -69,6 +69,22 @@ spec:
                 fieldPath: spec.nodeName
           image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
           imagePullPolicy: Always
+<<<<<<< HEAD
```
zvonkok, line 72: Merge not done?

ArangoGutierrez: Rebase not completed :(

ArangoGutierrez: Done now thanks for the catch

ArangoGutierrez:
@zvonkok PTAL , thanks for the previous reviews

pkg/ready/ready.go
```diff
@@ -0,0 +1,70 @@
...
+// to determine that the operator is ready.
+func (r fileReady) Set() error {
+	f, err := os.Create(r.FileName)
+	if err != nil {
+		if os.IsExist(err) {
+			return nil
+		}
+		return err
+	}
+	klog.Infof("Readiness Probe set to %s", r.FileName)
```
marquiz, line 60: Should this be debug (i.e. `V(1)`), wdyt?

pkg/ready/ready.go
```diff
@@ -0,0 +1,70 @@
...
+	// FileName represent the full path to the file for determining ready status
+	FileName string
+}
+
+// Set creates a file on disk whose presence can be used by a readiness probe
+// to determine that the operator is ready.
+func (r fileReady) Set() error {
+	f, err := os.Create(r.FileName)
+	if err != nil {
+		if os.IsExist(err) {
```
marquiz, line 55: Should we log this, too? WDYT?

marquiz (commented):
Thanks @ArangoGutierrez for this, looks good! Just few comments/questions

nfd-master.yaml.template
```diff
@@ -102,6 +115,13 @@ spec:
             runAsNonRoot: true
           command:
             - "nfd-master"
+## Enable Readiness probe
+          volumeMounts:
+          - mountPath: /tmp
+            name: tmp
+      volumes:
+      - name: tmp
+        emptyDir: {}
```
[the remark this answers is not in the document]
ArangoGutierrez: is a READ only container, only volumes are writable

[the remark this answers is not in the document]
marquiz: Ach, yes of course 🤦

marquiz (commented):
Does the readiness probe actually bring any value to us, or just complication and more LOC? We set the "flag" at startup and unset on os.Exit but no toggling during runtime. Thus, I think liveness probe covers everything and the readiness probe could/should be dropped. WDYT?

pkg/ready/ready.go
ArangoGutierrez: done

pkg/ready/ready.go
ArangoGutierrez: done

pkg/nfd-master/nfd-master.go
```diff
@@ -205,10 +225,18 @@ func (m *nfdMaster) Run() error {
...
+				klog.Error(err, "")
+				os.Exit(1)
+			}
 			return fmt.Errorf("gRPC server exited with an error: %v", err)
 
 		case <-m.stop:
 			klog.Infof("shutting down nfd-master")
+			if err = r.Unset(); err != nil {
+				klog.Error(err, "")
+				os.Exit(1)
```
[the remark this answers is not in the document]
ArangoGutierrez: done

pkg/nfd-master/nfd-master.go
```diff
@@ -205,10 +225,18 @@ func (m *nfdMaster) Run() error {
 			}
 
 		case <-grpcErr:
+			if err = r.Unset(); err != nil {
+				klog.Error(err, "")
+				os.Exit(1)
```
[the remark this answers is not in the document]
ArangoGutierrez: done

nfd-daemonset-combined.yaml.template
```diff
@@ -78,13 +91,22 @@ spec:
...
             valueFrom:
               fieldRef:
                 fieldPath: spec.nodeName
           image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
           imagePullPolicy: Always
+          livenessProbe:
+            exec:
+              command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
+            initialDelaySeconds: 10
+            periodSeconds: 10
```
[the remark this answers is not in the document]
ArangoGutierrez: good catch

Dockerfile
```diff
@@ -21,6 +21,10 @@ RUN make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
 
 RUN make test
 
+# Download the grpc_health_probe bin
+RUN GRPC_HEALTH_PROBE_VERSION=v0.3.1 && \
+    wget -qO/bin/grpc_health_probe https://github.com/grpc-ecosystem/grpc-health-probe/releases/download/${GRPC_HEALTH_PROBE_VERSION}/grpc_health_probe-linux-amd64 && \
+    chmod +x /bin/grpc_health_probe
```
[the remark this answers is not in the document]
ArangoGutierrez: done

ArangoGutierrez:
> Does the readiness probe actually bring any value to us, or just complication and more LOC? We set the "flag" at startup and unset on os.Exit but no toggling during runtime. Thus, I think liveness probe covers everything and the readiness probe could/should be dropped. WDYT?

Well, by definition the `ReadinessProbe` just tells Kubernetes that the service is ready to start getting traffic, that is why the ready func is after a successful `Run gRPC server` run, and we remove the file from the volume when the service is being shut down, in case the volume will be re-used for example in a rolling-update event. I think is serves it's purpose as it is. 

Worth mentioning this is just for the Master (that is the service that gets traffic) I didn't added this checks to the worker bc it can be used in a one-shot run, so is no worth adding long run checks to it.

(Addressed all your comments 😃 )

marquiz:
> Well, by definition the `ReadinessProbe` just tells Kubernetes that the service is ready to start getting traffic, that is why the ready func is after a successful `Run gRPC server` run, and we remove the file from the volume when the service is being shut down, in case the volume will be re-used for example in a rolling-update event. I think is serves it's purpose as it is.

My thinking was that we virtually have no time when the pod is running but it's not "Ready". The startup time of nfd-master is very fast and we never toggle the readiness on/off at runtime, just turn if of at pod termination time. I'm not sure if the added LOC (with maintenance burden( brings any added value. If we want to have a readiness probe, maybe se should just re-use the grpc probe because that **really** tells if we're ready to accept traffic. WDYT?

> Worth mentioning this is just for the Master (that is the service that gets traffic) I didn't added this checks to the worker bc it can be used in a one-shot run, so is no worth adding long run checks to it.

Yeah, I don't think we need to try to add any checks there for now.

ArangoGutierrez:
> > Well, by definition the `ReadinessProbe` just tells Kubernetes that the service is ready to start getting traffic, that is why the ready func is after a successful `Run gRPC server` run, and we remove the file from the volume when the service is being shut down, in case the volume will be re-used for example in a rolling-update event. I think is serves it's purpose as it is.
> 
> My thinking was that we virtually have no time when the pod is running but it's not "Ready". The startup time of nfd-master is very fast and we never toggle the readiness on/off at runtime, just turn if of at pod termination time. I'm not sure if the added LOC (with maintenance burden( brings any added value. If we want to have a readiness probe, maybe se should just re-use the grpc probe because that **really** tells if we're ready to accept traffic. WDYT?
> 
> > Worth mentioning this is just for the Master (that is the service that gets traffic) I didn't added this checks to the worker bc it can be used in a one-shot run, so is no worth adding long run checks to it.
> 
> Yeah, I don't think we need to try to add any checks there for now.

Readiness also via Grpc, I agree, working on the fix

marquiz (changes requested):
Looks good and simple now 👍 

Just one thing: I think we could drop ` pkg/ready/ready.go` now.

ArangoGutierrez:
> Looks good and simple now +1
> 
> Just one thing: I think we could drop ` pkg/ready/ready.go` now.

Done

marquiz (approved):
This is a nice improvement @ArangoGutierrez!
/lgtm

dbaker-rh:
FYI - this probe addition has broken installs that make use of TLS (grpc_health_check also needs to use tls if the listener is).  I offer #710 to address this in the kustomize framework, and will probably bash together a patch for the helm chart shortly to do the same.

PR #570, opened 2021-08-12
Better error reporting of kernel and cpu feature sources

### LINKED ISSUE #527: Get rid of spurious feature detection errors

NFD should not print errors when detection of some feature fails because it's non-existent. For example, currently we get selinux related errors if selinux hasn't been enabled in the kernel. Another example: when detecting `foo` we possibly want to print an error if `sysfs` is not available (mount is missing) but not in the case that  `/sys/bus/foo` is missing.

Inspired by discussion in #523, specifically https://github.com/kubernetes-sigs/node-feature-discovery/pull/523#issuecomment-842296406

**Why is this needed**:

Users don't want to see misleading/fallacious errors

marquiz:
- source/cpu: better error reporting
  Drop confusing errors in the log when intel pstate or cstate driver is
  not enabled in the system. However, we still log an error if sysfs is
  not available at all, in which case we're not able to detect these
  correctly.
- source/kernel: better error reporting
  Get rid of distracting error in the log in case selinux is not enabled
  in the kernel. Still print an error only if sysfs/fs directory is not
  available, though, which indicates that we're not able to correctly
  detect the presence of selinux.

Fixes #527

### DIFF
--- a/source/cpu/cpu.go
+++ b/source/cpu/cpu.go
@@ -147,10 +147,10 @@ func (s *Source) Discover() (source.Features, error) {
 	}
 
 	// Detect cstate configuration
-	cstate, err := detectCstate()
+	cstate, ok, err := detectCstate()
 	if err != nil {
-		klog.Errorf("failed to detect cstate: %v", err)
-	} else {
+		klog.Error(err)
+	} else if ok {
 		features["cstate.enabled"] = cstate
 	}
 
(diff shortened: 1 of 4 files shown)


marquiz:
I want to do some verification, still
/hold

marquiz:
Seems to work as expected
/unhold

ArangoGutierrez (approved):
/lgtm

PR #571, opened 2021-08-13
cstate/pstate: Skip check on non intel arches

jschintag:
Intel driver is not available on other arches, skip check.


marquiz:
#570 addresses the logging issues.

We can disable it altogether, but I think it would be bette to use conditional build with filename suffixes, like we do for the other features (like cpuid). And, do the same for pstate as well. I.e. have files `cstate_amd64.go` and `cstate_stub.go` (and pstate similarly).

jschintag:
@marquiz I'll rebase and make the changes after #570 is merged.
/hold

marquiz:
> @marquiz I'll rebase and make the changes after #570 is merged.

Perfect, thanks!

marquiz (approved):
Fast response, @jschintag 👍  At a quick glance looks good to me
/ok-to-test

source/cpu/cstate_amd64.go
```diff
@@ -1,3 +1,5 @@
+// +build amd64
+
```
marquiz, line 2: nit: redundant in the file suffixed `_amd64`, could be dropped

source/cpu/pstate_amd64.go
```diff
@@ -1,3 +1,5 @@
+// +build amd64
+
```
marquiz, line 2: nit: redundant in the file suffixed `_amd64`, could be dropped

marquiz (changes requested):
Actually, found one small nit

jschintag:
Yeah, i forgott to remove `runtime` from pstate_amd64.go
Regarding the build flags, i copied them from rdt. I can leave them out as well.

marquiz:
> Yeah, i forgott to remove `runtime` from pstate_amd64.go

OK

> Regarding the build flags, i copied them from rdt. I can leave them out as well.

Yeah, please remove (as e.g. `source/cpu/power_amd64.go` does not have that). We should prolly remove those from rdt, too (in a separate PR)

marquiz (approved):
Thanks @jschintag for this optimization 👍 

/lgtm

PR #558, opened 2021-08-05
Move to kustomize

ArangoGutierrez:
Fixes #398 

This patch moves the templating format currently used to kustomize

PR #573, opened 2021-08-13
Move to kustomize

marquiz:
Another attempt (in addition to #558) at converging to kustomize. This approach tries to leverage re-usable bases and components in order to avoid duplication which should make the maintenance of the deployment templates easier far less error-prone. What complicates the setup is that we offer various alternative deployment scenarios (daemonset, job, combined daemonset) - I wanted to preserve this, at least for now.

Convert to kustomize-based deployment in multiple steps (separate patches):
- helm: move files under deployment/helm
- deployment: add kustomize base
  Implement functionality virtually replicating deployment templates for
  nfd-master and nfd-worker daemonset (nfd-master.yaml.template and 
  nfd-worker-daemonset.yaml.template)

  We split the resources into multiple bases (rbac, master and
  worker-daemonset) so that relevant parts are re-usable in
  other deployment scenarios added later (e.g. "one-shot job", and "combined daemonset").
- kustomize: add support for cert-manager
  Add a kustomize component and an example overlay for enabling
  cert-manager for NFD.
- kustomize: add an example custom rules configmap
  Add example kustomize component and overlay for deploying a configmap
  specifying extra rules for the custom feature source.
- kustomize: add Job example deployment
  Add a new base kustomization for worker Job and an overlay stitching up
  the complete deployment.
- kustomize: add prune overlay
  Add an overlay for deploying "nfd-master --prune".
- kustomize: add standard-combined overlay
  Replicates nfd-daemonset-combined.yaml.template.
- Drop deployment templates and resort to kustomize, instead.


marquiz:
Documentation is not yet finished.

/hold

marquiz:
Documentation now fully updated
/unhold

ArangoGutierrez (approved):
I approve this PR, I will let @mythi to lgtm

marquiz:
Any thoughts on the directory structure or naming? 🤔  Don't really know if `standard` is a good descriptive name - would e.g default be better...

deployment/base/rbac/master-clusterrole.yaml
```diff
@@ -0,0 +1,15 @@
...
+- apiGroups:
+  - ""
+  resources:
+  - nodes
+  verbs:
+  - get
+  - patch
+  - update
+  # List only needed for --prune
+  - list
```
mythi, line 15: patch it via `deployment/overlays/prune`?

mythi (commented):
> Any thoughts on the directory structure or naming? 

`base/rbac` seems to be used by all the overlays. I was thinking should it be part of `base/master` but OTOH keeping it separate seems to be common. 

One other thing: what would you think of providing cert-manager and custom-rules as `overlay/samples/` (or examples). For custom-rules, I have a use case that I wrote in https://github.com/kubernetes-sigs/node-feature-discovery/pull/558#discussion_r688326594. How about add docs how to do that?

> Don't really know if standard is a good descriptive name - would e.g default be better...

My preference would be default and it seems to be commonly used elsewhere too (e.g., kubebuilder gives you default).

mythi:
> Hmm, I have overlays for both cert-manager and custom-rules. Do you mean just moving them into a separate subdirectory or smth else?

I was thinking `overlays/samples/cert-manager` and `overlays/samples/custom-rules` but I'm also OK as it is now

marquiz:
> I was thinking `overlays/samples/cert-manager` and `overlays/samples/custom-rules` but I'm also OK as it is now

Mm, I hear you. I think that makes sense, and hopefully reduces confusion, as these are really samples. I'll move them.

The corresponding patches add new components (in addition to the overlays) that the overlays then use – was thinking about re-usability here, too. Would it be better just to scrap the components and put the stuff under the overlay, WDYT?

docs/get-started/deployment-and-usage.md
```diff
@@ -317,24 +319,24 @@ of its certificate.
...
-$ openssl genrsa -out ca.key 2048
-$ openssl req -x509 -new -nodes -key ca.key -subj "/CN=nfd-ca" -days 10000 -out ca.crt
-$ sed s"/tls.key:.*/tls.key: $(cat ca.key|base64 -w 0)/" -i nfd-cert-manager.yaml
-$ sed s"/tls.crt:.*/tls.crt: $(cat ca.crt|base64 -w 0)/" -i nfd-cert-manager.yaml
-$ kubectl apply -f nfd-cert-manager.yaml
+kubectl apply -f https://github.com/jetstack/cert-manager/releases/download/v1.2.0/cert-manager.yaml
+openssl genrsa -out deployment/components/cert-manager/tls.key 2048
+openssl req -x509 -new -nodes -key deployment/components/cert-manager/tls.key -subj "/CN=nfd-ca" \
+        -days 10000 -out deployment/components/cert-manager/tls.crt
+kubectl apply -k deployment/overlays/cert-manage
```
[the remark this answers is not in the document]
marquiz: Fixed

deployment/base/rbac/master-clusterrole.yaml
marquiz: Changed accordingly: prune overlay now patches the clusterrole

mythi:
> Would it be better just to scrap the components and put the stuff under the overlay, WDYT?

no strong opinion but if we consider cert-manager and custom-config as samples, perhaps their components are not justified either and it'd be better to have everything under their respective overlays.

marquiz:
> no strong opinion but if we consider cert-manager and custom-config as samples, perhaps their components are not justified either and it'd be better to have everything under their respective overlays.

I'm inclined to agree with that – I was probably just awaiting your confirmation on this 😇  I think custom-rules is a clear case, cert-manager not that much but it's better to move it out of components as it's not readily usable as it (you need to manually add the CA files in order to use it). Anyways, I'll remove both of those components

marquiz:
OK, I now removed the cert-manager and custom-rule components and moved the stuff over to the respective overlays. I think this starts to be ready for merging, unless further review comments arise, of course

docs/advanced/developer-guide.md
```diff
@@ -54,32 +54,19 @@ attribute in the spec template(s) to the new location
...
+The `yamls` makefile generates a `kustomization.yaml` matching your locally
+built image and using the `deploy/overlays/default` deployment. See
+[build customization](#customizing-the-build) below for configurability, e.g.
+changing the deployment namespace.
 
 ```bash
 K8S_NAMESPACE=my-ns make yamls
-kubectl apply -f nfd-master.yaml
-kubectl apply -f nfd-worker-daemonset.yaml
+kubectl apply -k kustomize.yaml
```
mythi, line 64: this seems to have issues.

after `kubectl apply -k .` (should be . instead of kustomize.yaml) I can see the rbac rules are still created in `node-feature-discovery` ns.

mythi:
> I think this starts to be ready for merging, unless further review comments arise, of course

Looks good to me now. One small nit above.

docs/advanced/developer-guide.md
marquiz: Really good catch! This is now fixed in a separate patch on top of the old ones.

marquiz:
> Looks good to me now. One small nit above.

Thanks for the help @mythi. That was a really relevant and important "nit" 😸  It lead to a clear improvement in our rbac rules, I think. Changing the namespace should now work as expected.

marquiz:
Hmm, I'm still getting some errors
/hold

marquiz:
My first attempt at fixing the rbac issue failed. But, now it really should work as expected

/unhold

mythi (approved):
LGTM!

ArangoGutierrez (approved):
/lgtm

PR #601, opened 2021-09-20
source: introduce FeatureSource interface

marquiz:
Specify a new interface for managing "raw" feature data. This is the
first step to separate raw feature data from node labels. None of the
feature sources implement this interface, yet.

This patch unifies the data format of "raw" features by dividing them
into three different basic types.
- `keys`, a set of names without any associated values, e.g. CPUID flags
  or loaded kernel modules
- `values`, a map of key-value pairs, for features with a single value,
  e.g. kernel config flags or os version
- `instances`, a list of instances each of which has multiple attributes
  (key-value pairs of their own), e.g. PCI or USB devices

The new feature data types are defined in a new "pkg/api/feature"
package, catering decoupling and re-usability of code e.g. within future
extentions of the NFD gRPC API.

Rename the `Discover()` method of LabelSource interface to `GetLabels()`.

This is part of #464 - a separate smaller PR hopefully making review easier.

### DIFF
--- a/pkg/api/feature/feature.go
+++ b/pkg/api/feature/feature.go
@@ -1,0 +1,36 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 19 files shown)


ArangoGutierrez (approved):
/lgtm

PR #602, opened 2021-09-21
Utilize go generate

marquiz:
Use 'go generate' for auto-generating code. Drop the old 'mock' and
'apigen' makefile targets. Those are replaced with a single
  make generate

which (re-)generates everything.


pkg/apihelper/apihelpers.go
```diff
@@ -16,6 +16,8 @@ limitations under the License.
 
 package apihelper
 
+//go:generate mockery --name=APIHelpers --inpkg
```
zvonkok, line 19: Leftover?

ArangoGutierrez (changes requested):
Just  a nit/question, should we add this new make target to `https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/scripts/prepare-release.sh` ? so all is freshly generated when cutting a release (if not, this lgtm)

marquiz:
> Just a nit/question, should we add this new make target to `https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/scripts/prepare-release.sh` ? so all is freshly generated when cutting a release (if not, this lgtm)

I wouldn't add it there – the auto-generated parts should be aligned before cutting the release. I've been thinking about adding a verify step, though. To check that everything's in sync. However, we're not quite ready for that yet. We have no nice way of ensuring that we use the correct version of the auto-generation tools. Thus, all of this will be left as a future enhancement.

ArangoGutierrez (approved):
My concern has been addressed 
/lgtm

PR #618, opened 2021-10-04
nfd-master: drop the -nrt-namespace cmdline flag

marquiz:
Always create the CR in the namespace we're running in. NFD owns the CR
so it should create it in its own namespace rather than messing around
in other namespaces. Also makes cleanup less error-prone. Deleting the
nfd namespace removes the CRs too, not leaving them abandoned in some
random corner of the cluster.

Update RBAC rules accordingly as nfd-master doesn't require cluster-wide
privileges to NodeResourceTopology objects, anymore.


deployment/base/rbac/master-clusterrole.yaml
```diff
@@ -12,11 +12,3 @@ rules:
   - patch
   - update
   - list
-- apiGroups:
```
ArangoGutierrez, line 15: This `topology.node.k8s.io` changes don't align with the PR description

ArangoGutierrez (changes requested):
Looks good, just a minor question

deployment/base/rbac/master-clusterrole.yaml
marquiz: Updated commit message

marquiz:
Superseded/fixed by #652 
/close

PR #621, opened 2021-10-05
Detect and label CPU Model Name#270

AshwiniGhanwat:
**Changes made in the following files:**
- deployment/components/common/worker-mounts-combined.yaml
- deployment/components/common/worker-mounts.yaml
- deployment/helm/node-feature-discovery/templates/worker.yaml
- pkg/nfd-client/worker/nfd-worker.go
- source/config.go
- source/source_test.go
- test/e2e/e2e-test-config.example.yaml
- test/e2e/node_feature_discovery.go

**New file to generate the detect CPU Info:**
- source/cpuinfo/cpuinfo.go

**Please find below tested output:**
image


digambar15:
Request Maintainer to add /ok-to-test tag to initiate CI tests

zvonkok:
@digambar15 Can you run `make gofmt-verify` before checking in?

test/e2e/e2e-test-config.example.yaml
```diff
@@ -77,12 +82,22 @@ defaultFeatures:
         "feature.node.kubernetes.io/cpu-cpuid.MPX": "true"
         "feature.node.kubernetes.io/cpu-cpuid.RTM": "true"
         "feature.node.kubernetes.io/cpu-hardware_multithreading": "true"
+        "feature.node.kubernetes.io/cpuinfo-cpuinfo.model": "142"
+        "feature.node.kubernetes.io/cpuinfo-cpuinfo.cpu_MHz": "1896.002"
+        "feature.node.kubernetes.io/cpuinfo-cpuinfo.cpu_family": "6"
+        "feature.node.kubernetes.io/cpuinfo-cpuinfo.vendor_id": "GenuineIntel"
+        "feature.node.kubernetes.io/cpuinfo-cpuinfo.stepping": "10"
```
ArangoGutierrez, line 89: This test could fail if not run on the same cpu-arch

ArangoGutierrez (changes requested):
badly indented
Also a comment about the test file

source/cpuinfo/cpuinfo.go
```diff
@@ -0,0 +1,100 @@
+/*
+Copyright 2018-2021 The Kubernetes Authors.
```
ArangoGutierrez, line 2: If this is a new file, only 2021

ArangoGutierrez (changes requested):

source/cpuinfo/cpuinfo.go
ArangoGutierrez: PING*

ArangoGutierrez (changes requested):

marquiz (changes requested):
Thanks for the PR @AshwiniGhanwat 

I think this should be implemented as part of the `source/cpu` instead of a completely new source. Even the label name suggests that: `cpu-cpuinfo.model` makes more sense (IMO) than `cpuinfo-cpuinfo.model`.

Also, I'm not entirely sure how useful `cpu_MHz` label really is. Any thoughts?

ArangoGutierrez:
> Thanks for the PR @AshwiniGhanwat
> 
> I think this should be implemented as part of the `source/cpu` instead of a completely new source. Even the label name suggests that: `cpu-cpuinfo.model` makes more sense (IMO) than `cpuinfo-cpuinfo.model`.
> 
> Also, I'm not entirely sure how useful `cpu_MHz` label really is. Any thoughts?

*PING* @AshwiniGhanwat

AshwiniGhanwat:
Closing this PR. Included all comments in https://github.com/kubernetes-sigs/node-feature-discovery/pull/700

Issue #626, opened 2021-10-20
deployment: Allow dictionary-like configuration of nfd-workers in Helm


eliaskoromilas:
**What would you like to be added**:

The `worker.config` key in Helm values should accept an object representing the `nfd-worker.conf`, instead of a multi-line YAML string.

**Why is this needed**:

Would make configuration of nfd-workers simpler in Helm-based deployments, enabling advanced installations/revisions using just the CLI.

e.g. To enable only `pci`-related labels:

`helm install --generate-name nfd/node-feature-discovery --set 'worker.config.core.sources={pci}'`

Issue #485, opened 2021-03-18
Config file for nfd-master


marquiz:
**What would you like to be added**:

Now that nfd-worker has quite extensive support for dynamic run-time configurability through its config file, I started to think would it be good to have something similar for nfd-master as well 🤔  A similar json/yaml based configuration file with fsnotify watching for changes. Could allow dynamic control of e.g. logging parameters, allowed namespaces, extended resources and label whitelisting.

**Why is this needed**:

Would make configuration of nfd-master easier, enabling dynamic (run-time) configurability.

ArangoGutierrez:
++ I like this idea

marquiz:
Moving milestone to v0.12.0
/remove-lifecycle stale

marquiz:
We still want this (I think) but might need to adjust the milestone...

/remove-lifecycle rotten

PR #627, opened 2021-10-20
deployment: Simplify NFD worker configuration in Helm

eliaskoromilas:
fixes https://github.com/kubernetes-sigs/node-feature-discovery/issues/626

1. `toYaml` function is used to transform the map into a YAML string.
2. `<NFD-WORKER-CONF-*-DO-NOT-REMOVE>` tags are preserved for templating the sample config in `values.yaml`.
3. The default `nfd-worker.conf` file now contains `null` (and not the sample #commented config), but users should prefer Helm revisions instead of manually editing the `ConfigMap`.

### DIFF
--- a/docs/get-started/deployment-and-usage.md
+++ b/docs/get-started/deployment-and-usage.md
@@ -269,7 +269,7 @@ We have introduced the following Chart parameters.
 | ---- | ---- | ------- | ----------- |
 | `worker.*` | dict |  | NFD worker daemonset configuration |
 | `worker.configmapName` | string | `nfd-worker-conf` | NFD worker pod ConfigMap name |
-| `worker.config` | dict |  | NFD worker [service configuration](../advanced/worker-configuration-reference.md) |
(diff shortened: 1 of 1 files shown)


eliaskoromilas:
> I'd suggest to use smth like `{{ include "node-feature-discovery.fullname" . }}-worker-conf`, though. Could you submit a separate PR about this?

The `-worker-conf` suffix makes sense, but I noticed that you also plan to introduce a master config file (https://github.com/kubernetes-sigs/node-feature-discovery/issues/485), so I guess the question is whether you prefer to store them in separate `ConfigMap`s or under the same resource:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ include "node-feature-discovery.fullname" . }}
  labels:
  {{- include "node-feature-discovery.labels" . | nindent 4 }}
data:
  nfd-master.conf: |-
    {{- .Values.master.config | toYaml | nindent 4 }}
  nfd-worker.conf: |-
    {{- .Values.worker.config | toYaml | nindent 4 }}
```

But let's discuss it further in a new PR, as you suggested.

PR #628, opened 2021-10-21
deployment/helm: don't force sleep-interval in  worker cmdline flags

marquiz:
Drop --sleep-interval from the template. We really don't want to do that
as. First, it's the default value so no use repeating that in the
template. And more importantly, the commandline flag will override
anything that will be provided in the worker config file, making it
impossible for users to specify the sleep interval (other than by
editing the template directly).


marquiz:
ping @eliaskoromilas

PR #640, opened 2021-11-03
deployment: Implicitly generate the worker ConfigMap name

eliaskoromilas:
fixes https://github.com/kubernetes-sigs/node-feature-discovery/pull/627#issuecomment-948316754

1. `"node-feature-discovery.fullname"` is used to prefix the worker ConfigMap name
2. `worker.configmapName` Helm parameter is dropped
3. This change is documented.

**Comments**:

Since the template filename is still `nfd-worker-conf.yaml`, it makes sense to have the name suffix `-worker-conf`.

Generally, we should consider renaming the template file, to e.g. `configmap.yaml`, and drop that resource name suffix. This would make that ConfigMap resource the single source of information for any (future) nfd pod configuration (worker, master, etc.). Should I proceed with this here?

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/nfd-worker-conf.yaml
+++ b/deployment/helm/node-feature-discovery/templates/nfd-worker-conf.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 kind: ConfigMap
 metadata:
-  name: {{ .Values.worker.configmapName }}
+  name: {{ include "node-feature-discovery.fullname" . }}-worker-conf
   labels:
   {{- include "node-feature-discovery.labels" . | nindent 4 }}
 data:
--- a/deployment/helm/node-feature-discovery/templates/worker.yaml
+++ b/deployment/helm/node-feature-discovery/templates/worker.yaml
@@ -109,7 +109,7 @@ spec:
             path: "/etc/kubernetes/node-feature-discovery/features.d/"
         - name: nfd-worker-conf
           configMap:
-            name: {{ .Values.worker.configmapName }}
+            name: {{ include "node-feature-discovery.fullname" . }}-worker-conf
             items:
               - key: nfd-worker.conf
                 path: nfd-worker.conf
--- a/deployment/helm/node-feature-discovery/values.yaml
+++ b/deployment/helm/node-feature-discovery/values.yaml
@@ -83,7 +83,6 @@ master:
                 values: [""]
 
 worker:
-  configmapName: nfd-worker-conf
   config: ### <NFD-WORKER-CONF-START-DO-NOT-REMOVE>
     #core:
     #  labelWhiteList:
--- a/docs/get-started/deployment-and-usage.md
+++ b/docs/get-started/deployment-and-usage.md
@@ -314,7 +314,6 @@ We have introduced the following Chart parameters.
 | Name | Type | Default | description |
 | ---- | ---- | ------- | ----------- |
 | `worker.*` | dict |  | NFD worker daemonset configuration |
-| `worker.configmapName` | string | `nfd-worker-conf` | NFD worker pod ConfigMap name |
 | `worker.config` | dict |  | NFD worker [configuration](../advanced/worker-configuration-reference.md) |
 | `worker.podSecurityContext` | dict | {} | SecurityContext holds pod-level security attributes and common container settings |
 | `worker.securityContext` | dict | {} | Container [security settings](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod) |
@@ -464,11 +463,15 @@ preferred method is to use a ConfigMap which provides easy deployment and
 re-configurability.
 
 The provided nfd-worker deployment templates create an empty configmap and
-mount it inside the nfd-worker containers. Configuration can be edited with:
+mount it inside the nfd-worker containers. In kustomize deployments,
+configuration can be edited with:
 
 ```bash
 kubectl -n ${NFD_NS} edit configmap nfd-worker-conf
 ```
+
+In Helm deployments, [Worker pod parameter](#worker-pod-parameters)
+`worker.config` can be used to edit the respective configuration.
 
 See
 [nfd-worker configuration file reference](../advanced/worker-configuration-reference.md)


marquiz:
Thanks @eliaskoromilas for this

> Generally, we should consider renaming the template file, to e.g. `configmap.yaml`, and drop that resource name suffix. This would make that ConfigMap resource the single source of information for any (future) nfd pod configuration (worker, master, etc.). Should I proceed with this here?

I think this sounds reasonable, I don't oppose 😄  Maybe in a separate PR, though. What I'm just thinking that we should change the kustomize-based deployment at the same time (to keep these in sync)

/ok-to-test

eliaskoromilas:
> What I'm just thinking that we should change the kustomize-based deployment at the same time (to keep these in sync)

The problem is that in kustomize deployments everything is almost statically defined... and then patched. So everything is pretty match allowed.

The [default](https://github.com/kubernetes-sigs/node-feature-discovery/tree/v0.9.0/deployment/overlays/default) overlay uses the [worker-config](https://github.com/kubernetes-sigs/node-feature-discovery/tree/v0.9.0/deployment/components/worker-config) component and that is why the documentation states that users may edit `configmap/nfd-worker-conf`. Of course, in a **kustom** (not the default) deployment this may be incorrect.

I think that the kustomize deployment reflects the project's initial intention to serve as a K8s extension that needs to be deployed just once in a cluster. Helm changes that and transforms NFD to a K8s application that can be deployed as many times (even in the same namespace), as different NFD instances, each (probably) serving different purposes.

Having both kustomize and Helm as deployment options, makes the projects intentions unclear. For example, `/etc/kubernetes/node-feature-discovery/*` volume mounts are not much aligned with the multi-instance deployments that Helm offers. My opinion is that dropping  kustomize support would allow the project to evolve and better design decisions to be made.

Let me know if you want me to elaborate more with examples.

marquiz:
> I think that the kustomize deployment reflects the project's initial intention to serve as a K8s extension that needs to be deployed just once in a cluster.

IMO this is still the intention and objective. Nodes are "cluster scoped", so are their features/configuration and so should NFD be. Multiple parallel NFD deployments should be a rare exception. There are use cases that currently cannot be nicely covered with one NFD deployment.

> Helm changes that and transforms NFD to a K8s application that can be deployed as many times (even in the same namespace), as different NFD instances, each (probably) serving different purposes.

Trying to move away from this e.g. with #468 


> For example, `/etc/kubernetes/node-feature-discovery/*` volume mounts are not much aligned with the multi-instance deployments that Helm offers.

Hmm, that is true 🧐  Maybe there should be an option to disable these mounts or use `/etc/kubernetes/node-feature-discovery-{{ .Values.instance }}/` if instance has been specified?

> My opinion is that dropping kustomize support would allow the project to evolve and better design decisions to be made.

I wouldn't drop it. Makes it easy to deploy without any other tools than kubectl. The selection of deployment option(s)/method(s) really should not restrict the evolution of the project, dictate technical decisions or hold us back. We can loosen the coupling between kustomize and helm deployments if it causes pain.

marquiz (approved):

eliaskoromilas:
> IMO this is still the intention and objective. Nodes are "cluster scoped", so are their features/configuration and so should NFD be. 

I realize now that my problem is not the **kustomize** deployment but the **Helm** one. Helm doesn't "support" cluster-scoped charts, neither cluster-scoped chart dependencies.

This means that Helm apps, like the NVIDIA [GPU Operator](https://github.com/NVIDIA/gpu-operator/blob/v1.8.2/deployments/gpu-operator/values.yaml#L234-L250) or the InAccel [FPGA Operator](https://github.com/inaccel/helm/blob/fpga-operator-2.6.0/charts/fpga-operator/values.yaml#L23-L49), need to deploy their own NFD instances (with custom configuration) to make sure they will get the nodes labelled.

I didn't mean to question the project's intentions, I got confused by the available Helm options that do not imply cluster-scoped usage.

> Trying to move away from this e.g. with #468

But still such chart packages, that want to expose custom labels, can't depend on something they can't add as a dependency.

> I wouldn't drop it. Makes it easy to deploy without any other tools than kubectl. The selection of deployment option(s)/method(s) really should not restrict the evolution of the project, dictate technical decisions or hold us back.

I agree. I just wasn't sure about project's objective. I believe though that offering strict deployment options (usually) helps users and not the opposite.

> We can loosen the coupling between kustomize and helm deployments if it causes pain.

I believe this is inevitable. For example, many of the different overlays (`default`, `default-combined`, etc) that the kustomize deployment supports are not (yet ?) modeled in the Helm chart.

marquiz:
> I realize now that my problem is not the **kustomize** deployment but the **Helm** one. Helm doesn't "support" cluster-scoped charts, neither cluster-scoped chart dependencies.
> 
> This means that Helm apps, like the NVIDIA [GPU Operator](https://github.com/NVIDIA/gpu-operator/blob/v1.8.2/deployments/gpu-operator/values.yaml#L234-L250) or the InAccel [FPGA Operator](https://github.com/inaccel/helm/blob/fpga-operator-2.6.0/charts/fpga-operator/values.yaml#L23-L49), need to deploy their own NFD instances (with custom configuration) to make sure they will get the nodes labelled.

Currently, this is a problem


> I didn't mean to question the project's intentions, I got confused by the available Helm ~options~ parameters that do not imply cluster-scoped usage.

No worries, I appreciate that somebody asks this questions ☺️  As it makes me really think about the scope/direction, too

> > Trying to move away from this e.g. with #468
> 
> But still such chart packages, that want to expose custom labels, can't depend on something they can't add as a dependency.

Could they depend on [NFD operator](https://github.com/kubernetes-sigs/node-feature-discovery-operator)? Could the operators deploy cluster-scoped CRs?

#468 is probably not a full solution, yet. But hopefully a big piece of it.

> > I wouldn't drop it. Makes it easy to deploy without any other tools than kubectl. The selection of deployment option(s)/method(s) really should not restrict the evolution of the project, dictate technical decisions or hold us back.
> 
> I agree. I just wasn't sure about project's objective. I believe though that offering strict deployment options (usually) helps users and not the opposite.

By "strict", do you mean little configurability or just one support deployment method (e.g. kustomize vs. helm vs. operator)?

> > We can loosen the coupling between kustomize and helm deployments if it causes pain.
> 
> I believe this is inevitable. For example, many of the different overlays (`default`, `default-combined`, etc) that the kustomize deployment supports are not (yet ?) modeled in the Helm chart.

That's true. And I don't think we should model them there. I kind of dislike the multitude of overlays we currently have, but decided to keep them for backwards compatibility when switching over to kustomize. Maybe they should be moved to `samples/` 🤔

eliaskoromilas:
> Could they depend on NFD operator? Could the operators deploy cluster-scoped CRs?
> 
> #468 is probably not a full solution, yet. But hopefully a big piece of it.

That's true, OLM operators are meant for that. OLM supports dependency resolution, which means that other OLM operators may add specific versions of the CRDs, that NFD owns, (or even the whole NFD operator package) as a (cluster-scoped) dependency.

Helm, on the other hand, doesn't offer great support for CRDs and lifecycle management becomes too complicated.

> By "strict", do you mean little configurability or just one support deployment method (e.g. kustomize vs. helm vs. operator)?

I mean that we should probably accept the different deployment options for what they are. For example:

* *kustomize* as the simplest "non-packaged" deployment method, aligned with the "deploy once" principle of the project, with sane defaults, but manual configuration and upgrades.
* *OLM* operator (maybe) as the preferred way to deploy NFD cluster-scoped resources. Configuration is done using NFD CRDs, that other operators may add as a dependency (which is a very important feature).
* *Helm* chart as a nice declarative way to deploy/revise NFD on any cluster, with easy configuration management and upgrades, but (currently) not ideal for other Helm apps to add as a dependency (since we may end up with the same NFD functionality deployed more than once in the same cluster).

Having said that, it seems that Helm is the NFD deployment option with the less clear purpose.

* Helm can be used to deploy cluster-scoped resources (like NFD does), there isn't a real restriction and many charts do it. The problems usually appear when somebody deploys such charts twice.
* Other Helm apps may also want to add NFD as a dependency. Actually, `custom`/`local` NFD sources look like they were designed for that reason. What needs to be done here is to make sure that there is a configuration for adding NFD as a Helm chart dependency in a way that it won't overlap with other NFD Helm chart dependencies and/or other NFD instances that may be present in the cluster (e.g. the volume mounts that configure the `local` source is an example of such an overlap, and this PR is another).

What I propose is not to drop one method in favor of another, but to offer clear a strategy for all the available deployment methods.

marquiz:
Really good discussion. And also insightful for me 🙄  For example, there are aspects that I haven't even thought about much earlier, like dependencies and cluster-scoped resources with Helm.

> mean that we should probably accept the different deployment options for what they are.

I agree.

> * _kustomize_ as the simplest "non-packaged" deployment method, aligned with the "deploy once" principle of the project, with sane defaults, but manual configuration and upgrades.
> * _OLM_ operator (maybe) as the preferred way to deploy NFD cluster-scoped resources. Configuration is done using NFD CRDs, that other operators may add as a dependency (which is a very important feature).
> * _Helm_ chart as a nice declarative way to deploy/revise NFD on any cluster, with easy configuration management and upgrades, but (currently) not ideal for other Helm apps to add as a dependency (since we may end up with the same NFD functionality deployed more than once in the same cluster).

Well put. We should prolly try to incorporate that^ somehow even somewhere in the documentation 😊 

 
> Having said that, it seems that Helm is the NFD deployment option with the less clear purpose.

I think the original motivator for adding a Helm chart was to offer an "official" alternative for the multitude of downstream NFD Helm charts (and try to convince downstream to use ours).

> * Helm can be used to deploy cluster-scoped resources (like NFD does), there isn't a real restriction and many charts do it. The problems usually appear when somebody deploys such charts twice.

Regarding #468, we should be a bit safer as we would be only deploying the CRD (definintion), not any CRs (objects). But the nfd-topology-updater can and will cause problems if deployed multiple times in parallel. However, it's disabled by default.

> * Other Helm apps may also want to add NFD as a dependency. Actually, `custom`/`local` NFD sources look like they were designed for that reason. What needs to be done here is to make sure that there is a configuration for adding NFD as a Helm chart dependency in a way that it won't overlap with other NFD Helm chart dependencies and/or other NFD instances that may be present in the cluster (e.g. the volume mounts that configure the `local` source is an example of such an overlap, and this PR is another).

Yes, this needs to be considered. With parallel deployments one the `-instance` flag should be used. Maybe the mounts could be prefixed too, i.e. `/etc/kubernetes/node-feature-discovery-$INSTANCE/`? Created #650 to track this.

> What I propose is not to drop one method in favor of another, but to offer a clear strategy for all the available deployment methods.

👍  Created #651 to keep this in mind

PR #642, opened 2021-11-09
pkg/api/feature: small improvements

marquiz:
Slight improvements on the internal pkg/api/feature api.
- use pointers of structs:
  Make it easier to mutate the feature sets.
- generator functions with initial values:
  Flavor the generator helper functions with arguments for specifying the
set of features to put into the generated objects.


marquiz:
Missed one func
/hold

ArangoGutierrez (approved):
/lgtm

PR #643, opened 2021-11-10
test/e2e: make e2e tests run on single-node cluster

marquiz:
Lift the restriction to run custom rule tests on non-master node. Try to
find one but do not fail if that fails. Makes the end-to-end tests
runnable on single-node clusters such a simple minikube deployments.


ArangoGutierrez (approved):
/lgtm

PR #644, opened 2021-11-10
test/e2e: drop /boot mount

marquiz:
This is not currently needed by end-to-end tests. Dropping it enables
testing in restricted environments that don't have /boot directory.


ArangoGutierrez (approved):
/lgtm

Issue #638, opened 2021-11-02
Unexpected Kubernetes node feature-labeling as sgx-capable when SGX feature is turned off in BIOS and host has no /dev/sgx_* devices


MustDie95:
**What happened**:

Kubernetes node has  **feature.node.kubernetes.io/custom-intel.sgx: 'true'**  even if SGX support was forcibly turned off in BIOS

**What you expected to happen**:

Kubernetes node is not feature-labeled with SGX or this label was removed when SGX was off in BIOS and there is no /dev/sgx_* devices on host (even if this feature supported by CPU and OS).

**How to reproduce it (as minimally and precisely as possible)**:

Install server, setup Kubernetes and deploy Node Feature Discovery with 'kubectl apply -k https://github.com/kubernetes-sigs/node-feature-discovery/deployment/overlays/default?ref=v0.22.0' 

All nodes are marked with feature.node.kubernetes.io/custom-intel.sgx: true' despite BIOS settings.  

And  Intel **sgxdeviceplugin-sample** constantly trying to start (because has nodeSelector based on this feature) on this node but failing.

**Anything else we need to know?**:

nfd-worker.conf :

sources:
  custom:
  - name: "intel.sgx"
    matchOn:
      - kConfig: ["X86_SGX"]
        cpuId: ["SGX", "SGXLC"]

I think we should add more rules to make sure Intel SGX is really functional on the host. For example, examine the presence of /dev/sgx_* devices

**Environment**:
- Kubernetes version: 1.22.2
- Cloud provider or hardware configuration: Intel M50CYP1UR212 server (2 x Intel Xeon Platinum 8352S @ 2.2GHz, 512GB RAM)
- OS: Ubuntu Server 20.04.3 LTS
- Kernel: 5.11.0-38-generic

marquiz:
@mythi any thoughts?

In principle, this is not a bug of NFD itself, but merely in the rule configuration. Where is that config maintained? However, there might be something that would be good to add to NFD to make proper detection of SGX easier (possible without custom hooks or sidecar containers).

MustDie95:
@marquiz, I think the rule metioned above in nfd-worker.conf was added by "Intel Software Guard Extensions (SGX) device plugin for Kubernetes" (https://github.com/intel/intel-device-plugins-for-kubernetes/blob/main/deployments/sgx_nfd/nfd-worker.conf) so you are right, NFD itself is not responsible for this behavior.

[...]

PR #647, opened 2021-11-12
source/cpu: detect Intel SGX

mythi:
Fixes: #130, #638

### DIFF
--- a/source/cpu/sgx_amd64.go
+++ b/source/cpu/sgx_amd64.go
@@ -1,0 +1,46 @@
+//go:build amd64
+// +build amd64
+
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cpu
+
+import (
+	"github.com/klauspost/cpuid/v2"
+)
+
+func discoverSGX() map[string]string {
+	var epcSize uint64
+	sgx := make(map[string]string)
+
+	if cpuid.CPU.SGX.Available {
+		for _, s := range cpuid.CPU.SGX.EPCSections {
+			epcSize += s.EPCSize
+		}
+	}
+
+	// Set to 'true' based a non-zero sum value of SGX EPC section sizes. The
+	// kernel checks for IA32_FEATURE_CONTROL.SGX_ENABLE MSR bit but we can't
+	// do that as a normal user. Typically the BIOS, when enabling SGX,
+	// allocates "Processor Reserved Memory" for SGX EPC so we rely on > 0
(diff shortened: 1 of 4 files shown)


source/cpu/sgx_amd64.go
```diff
@@ -0,0 +1,46 @@
...
+	}
+
+	// Set to 'true' based a non-zero sum value of SGX EPC section sizes. The
+	// kernel checks for IA32_FEATURE_CONTROL.SGX_ENABLE MSR bit but we can't
+	// don't that as a normal user. Typically the BIOS, when enabling SGX,
+	// allocates "Processor Reserved Memory" for SGX EPC so we rely on > 0
+	// size here to set "SGX = enabled".
+	if epcSize > 0 {
+		sgx["enabled"] = "true"
+	}
```
marquiz, line 43: Should we advertise `epcSize` as the raw feature, directly?

marquiz (commented):
Basically looks good and simple to me.

Would it be useful to add other SGX attributes from klauspost/cpuid as raw features, now that we add this, WDYT? I'm thinking about these max enclave size etc.

source/cpu/sgx_amd64.go
mythi: I wanted to start with something minimal/basic and `enabled` is consistent with the other `cpu`source features. The boolean value can be taken into `nodeSelector` directly which is my use case at least until #464 lands. Would there be any use for `epcSize` other than using it in, e.g., `{op: Gt, value: ["0"]}`?

> I'm thinking about these max enclave size etc.

I'm not sure how useful they are. We could label, e.g., `sgx2=true` but to actually use the feature the kernel support needs to be there and we cannot detect that. max enclave size value is a bit value for `2^(<value>)` computation.

marquiz: > I wanted to start with something minimal/basic and `enabled` is consistent with the other `cpu`source features. The boolean value can be taken into `nodeSelector` directly which is my use case at least until #464 lands. Would there be any use for `epcSize` other than using it in, e.g., `{op: Gt, value: ["0"]}`?

Yeah, for the label `enabled` makes sense. I might be wrong but my intuition is that the better direction would be to expose "raw features" as raw as possible, avoiding any translations. More likely to avoid problems like the kernel module values (currently `=y` and `=m` now translate to `true` and we have no means of telling between y or m). As you said, in this case `enabled` would translate to `{op: Gt, value: ["0"]}` in the custom rules. That said, I don't know what use people could find for the numerical value (expose it as an extended resource?). Maybe we should expose both, `enabled` and `epcSize`?

> > I'm thinking about these max enclave size etc. 
> 
> I'm not sure how useful they are. We could label, e.g., `sgx2=true` but to actually use the feature the kernel support needs to be there and we cannot detect that. max enclave size value is a bit value for `2^(<value>)` computation.

K, can be added later, if needed

marquiz (commented):
/ok-to-test

source/cpu/sgx_amd64.go
mythi: > expose it as an extended resource?). Maybe we should expose both, `enabled` and `epcSize`?

We have an NFD source hook to read `epcSize` and that value is registered as `sgx.intel.com/epc`. I guess NFD could register the value too but why not then use the NFD source hook because it comes with the device plugin that is also needed.

mythi:
rebased to latest master and force pushed

marquiz (approved):
Thanks @mythi for this enhancement 👍 Let's merge this and expand functionality in future, if needed

/lgtm

PR #649, opened 2021-11-15
source/storage: implement FeatureSource

marquiz:
Separate feature discovery and creation of feature labels. Generalize
the feature discovery so that block devices can be matched in custom
label rules in a similar fashion as pci and usb devices. This extends
the discovery to other block queue attributes than 'rotational': now we
also detect 'dax', 'nr_zones' and 'zoned'.

Labels created by the storage feature source are unchanged. The new

Example custom rules:

```
  - name: "my block rule 1"
    labels:
      my-block-feature-1: "true"
    matchFeatures:
      - feature: storage.block
          "rotational": {op: In, value: ["0"]}

  - name: "my block rule 2"
    labels:
      my-block-feature-2: "true"
    matchFeatures:
      - feature: storage.block
          "zoned": {op: In, value: [“host-aware”, “host-managed”]}
```

Depends/builds on top of #464

### DIFF
--- a/deployment/base/nfd-crds/cr-sample.yaml
+++ b/deployment/base/nfd-crds/cr-sample.yaml
@@ -46,6 +46,11 @@ spec:
             major: {op: In, value: ["5"]}
             minor: {op: Gt, value: ["10"]}
 
+        - feature: storage.block
+          matchExpressions:
+            rotational: {op: In, value: ["0"]}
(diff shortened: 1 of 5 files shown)


marquiz:
Waiting for #464 
/hold

ArangoGutierrez (approved):
/lgtm

PR #653, opened 2021-11-16
 specify CRD for custom labeling rules

marquiz:
Add a cluster-scoped Custom Resource Definition for specifying labeling
rules. Nodes (node features, node objects) are cluster-level objects and
thus the natural and encouraged setup is to only have one NFD deployment
per cluster - the set of underlying features of the node stays the same
independent of how many parallel NFD deployments you have. Our extension
points (hooks, feature files and now CRs) can be be used by multiple
actors (depending on us) simultaneously. Having the CRD cluster-scoped
hopefully drives deployments in this direction. It also should make
deployment of vendor-specific labeling rules easy as there is no need to
worry about the namespace.

This patch virtually replicates the source.custom.FeatureSpec in a CRD
API (located in the pkg/apis/nfd/v1alpha1 package) with the notable
exception that "MatchOn" legacy rules are not supported. Legacy rules
are left out in order to keep the CRD simple and clean.

The duplicate functionality in source/custom will be dropped by upcoming
patches.

This patch utilizes controller-gen (from sigs.k8s.io/controller-tools)
for generating the CRD and deepcopy methods. Code can be (re-)generated
with "make apigen". Install controller-gen with:

```
  go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0
```

Update kustomize and helm deployments to deploy the CRD.


marquiz:
Still want to check some details
/hold

marquiz:
I changed to CRD to cluster-scoped

marquiz:
Added a Helm parameter to make CRD generation configurable (true by default).
/unhold

PR #654, opened 2021-11-16
source/custom: move rule matching to pkg/apis/nfd

marquiz:
Move the rule processing of matchFeatures and matchAny from
source/custom package over to pkg/apis/nfd, aiming to better integrity
and re-usability of the code. Does not change the API as such, just adds
more supportive functions.

Depends/builds on #653


zvonkok:
#653 merged 
/unhold

marquiz:
Rebased, fixed wording in the commit msg

PR #660, opened 2021-11-18
source/network: implement FeatureSource

marquiz:
Separate feature discovery and creation of feature labels. Generalize
the feature discovery so that network devices can be matched in custom
label rules in a similar fashion as pci and usb devices. Available
attributes for matching are:

- operstate
- speed
- sriov_numvfs
- sriov_totalvfs

Labels created by the network feature source are unchanged. The new

Example custom rule:

```
  - name: "my network rule"
    labels:
      my-network-feature: "true"
    matchFeatures:
      - feature: network.device
        matchExpressions:
          "operstate": { op: In, value: ["up"] }
          "sriov_numvfs": { op: Gt, value: ["9"] }
```

### DIFF
--- a/deployment/base/nfd-crds/cr-sample.yaml
+++ b/deployment/base/nfd-crds/cr-sample.yaml
@@ -51,6 +51,11 @@ spec:
             rotational: {op: In, value: ["0"]}
             dax: {op: In, value: ["0"]}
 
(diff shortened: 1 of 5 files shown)


ArangoGutierrez (approved):
/lgtm

PR #662, opened 2021-11-19
Wildcard for matching against all keys in label rules

marquiz:
Extend the feature rule format with a special wildcard that can be used
in matchExpressions. By using '*' as the field name in an expressoin
makes is possible to run the expression agains all feature names,
instead of the value of one feature.

This can be useful in template rules for creating per-feature labels,
based on feature names (instead of values) and in "normal"
(non-template) rules for checking if (at least) one of certain feature
names are present.

An example of creating an "avx512" label if any AVX512* CPUID feature is
present:

```
  - name: "avx512-any-feature"
    labels: {avx512: "true"}
    matchFeatures:
      - feature: cpu.cpuid:
        matchExpressions:
          "*": {op: InRegexp, value: ["^AVX512"]}
```

An example of a template rule creating a dynamic set of labels  based on
the existence of kconfig options.

```
  - name: "kconfig-template-features"
    labelsTemplate: |
      {{range .kernel.config}}kconfig-{{.Name}}={{.Value}}
      {{end}}
    matchFeatures:
      - feature: kernel.config
        matchExpressions:
          "*": {op: In, value: [SWAP, X86, ARM]}
```

Depends on #550

### DIFF
--- a/deployment/base/nfd-crds/cr-sample.yaml
+++ b/deployment/base/nfd-crds/cr-sample.yaml
@@ -88,6 +88,14 @@ spec:
               vendor: {op: In, value: ["8086"]}
               class: {op: In, value: ["02"]}
 
+    - name: "my AVX feature"
+      labels:
+        "my-avx-feature": "true"
+      matchFeatures:
+        - feature: cpu.cpuid
+          matchExpressions:
+            '*': {op: InRegexp, value: ["^AVX512"]}
+
     # The following features demonstreate label templating capabilities
     - name: "my system template feature"
       labelsTemplate: |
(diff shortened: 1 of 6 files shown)


marquiz:
I don't like this approach myself. Possibly something like this would be more comprehensive:

```
  - name: "avx512-any-feature"
    labels: {avx512: "true"}
    matchFeatures:
      - feature: cpu.cpuid:
        matchNames: {op: InRegexp, value: ["^AVX512"]}
```

marquiz:
Moving milestone to v0.12.0

marquiz:
Supserseded by #788
/close

PR #665, opened 2021-11-23
source/kernel: don't advertise selinux.enabled=false

marquiz:
Commit 094501916155c183775e4e941d55a4d9ca5f38a0 changed the behavior so
that NFD started to advertise also "false" status of selinux.enabled
label. This patch reverts this behavior (i.e. we only have
selinux.enabled=true). The rationale behind is avoiding any excess
labels - selinux.enabled=false label would be pointless noise in most
deployments.


zvonkok:
SELinux not advertised 
/lgtm

PR #667, opened 2021-11-23
Add deploy-prune makefile rule to ease devel processes

ArangoGutierrez:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,14 @@ image: yamls
 	    $(foreach tag,$(IMAGE_EXTRA_TAGS),-t $(tag)-minimal) \
 	    $(IMAGE_BUILD_EXTRA_OPTS) ./
 
+# clean NFD labels on all nodes
+# devel only
+deploy-prune:
+	kubectl apply -k deployment/overlays/prune/
+	kubectl wait --for=condition=complete job -l app=nfd -n node-feature-discovery
+	kubectl delete -k deployment/overlays/prune/
+
+
 yamls:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -86,6 +86,12 @@ image: yamls
 	    $(foreach tag,$(IMAGE_EXTRA_TAGS),-t $(tag)-minimal) \
 	    $(IMAGE_BUILD_EXTRA_OPTS) ./
 
+# clean NFD labels on all nodes
+# devel only
+clean-labels:
+	kubectl get no -o yaml | sed -e '/^\s*nfd.node.kubernetes.io/d' -e '/^\s*feature.node.kubernetes.io/d' | kubectl replace -f -
```
marquiz, line 92: This is not needed if we do prune

Makefile
```diff
@@ -86,6 +86,12 @@ image: yamls
...
+	kubectl apply -k deployment/overlays/prune/
```
marquiz, line 93: Hmm, need to think about this. We should probably do a `kubectl wait` on the completion of the job, or smth, and then do `k delete -k deployment/overlays/prune`. Otherwise subsequent `make clean-labels` don't have an effect (no new prune jobs will be created)

ArangoGutierrez: how about now?

ArangoGutierrez:
Changed to `prune-deployment` as discussed

marquiz (approved):

ArangoGutierrez:
@marquiz @zvonkok the bot didn't got the approved label... sorry for the inconveniences, someone has to re-approve

PR #668, opened 2021-11-24
pkg/apis/nfd: stricter format checking for template labels

marquiz:
Require that the expanded LabelsTemplate has values. That is, the
(expanded) template must consist of key=value pairs separated by
newlines. No default value will be assigned and we now return an error
if a (non-empty) line not conforming with the key=value format is
encountered.

Commit c8d73666d described that the value defaults to "true" if not
specified. That was not the case and we defaulted to an empty string,
instead.

An example:

```
  - name: "my rule"
    labelsTemplate: |
      my.label.1=foo
      my.label.2=
```

Would create these labels:

```
  "my.label.1": "foo"
  "my.label.2": ""
```

Further, the following:

```
  - name: "my failing rule"
    labelsTemplate: |
      my.label.3
```

will cause an error in the rule processing.

### DIFF
--- a/pkg/apis/nfd/v1alpha1/rule.go
+++ b/pkg/apis/nfd/v1alpha1/rule.go
@@ -85,14 +85,14 @@ func (r *Rule) executeLabelsTemplate(in matchedFeatures, out map[string]string) 
 	if r.labelsTemplate == nil {
 		t, err := newTemplateHelper(r.LabelsTemplate)
 		if err != nil {
-			return err
+			return fmt.Errorf("failed to parse LabelsTemplate: %w", err)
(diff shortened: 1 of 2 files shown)


marquiz:
Will change this 
/hold

marquiz:
> An alternative to solution would be to ditch the default value ("true") and e.g. just error out if the value part is missing (i.e. the expanded template value does not have an equals sign).

Changed the PR to behave like this.

/unhold

zvonkok:
We can make the API more restrictive and loosen up when needed, the other way around could be problematic.
Like this way much better.

/lgtm

PR #671, opened 2021-11-25
Use single-dash format of cmdline flags

marquiz:
Use the single-dash (i.e. '-option' instead of '--option') format
consistently accross log messages and documentation. This is the format
that was mostly used, already, and shown by command line help of the
binaries, for example.

PR #673, opened 2021-11-25
nfd-worker: rename 'sources' config option

marquiz:
Rename `core.sources` config file option to `core.labelSources` and `-sources` command line flag to `-label-sources`. Also provide backwards compatibility via deprecated options. Multiple smaller pieces as separate commits:

- rename config option `sources` to `labelSources`
  The goal is to make the name more descriptive. Also keeping in mind a
possible future addition a 'featureSources' option (or similar) for
controlling the feature discovery.
- provide deprecated core.sources config option 
  Provide backwards compatibility via a deprecated 'core.sources' config
file option. This will override 'core.labelSources'. A warning is
printed in the log if this option is detected.
- introduce -label-sources cmdline flag 
  Useful for development, testing and debugging.
- update docs on label-sources option 
  Update documentation on core.labelSources/core.sources config file
options and -label-sources/-sources command line flags.

### DIFF
--- a/cmd/nfd-worker/main.go
+++ b/cmd/nfd-worker/main.go
@@ -80,6 +80,8 @@ func parseArgs(flags *flag.FlagSet, osArgs ...string) *worker.Args {
 		switch f.Name {
 		case "no-publish":
 			args.Overrides.NoPublish = overrides.NoPublish
+		case "label-sources":
+			args.Overrides.LabelSources = overrides.LabelSources
 		case "label-whitelist":
 			klog.Warningf("-label-whitelist is deprecated, use 'core.labelWhiteList' option in the config file, instead")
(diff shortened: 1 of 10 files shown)


marquiz:
Let's see what I think about this after sleeping over it 🧐
/hold

marquiz:
Depends on #671

marquiz:
Rebased.

Any thoughts @ArangoGutierrez @zvonkok, does this make sense?

ArangoGutierrez:
> Rebased.
> 
> Any thoughts @ArangoGutierrez @zvonkok, does this make sense?

being more explicit never harm anyone in the past, it is indeed a label source, we have just been calling it source for, I guess, short naming conventions. I am ok with this
@zvonkok ? 
also @marquiz unhold

marquiz:
> also @marquiz unhold

/unhold

PR #680, opened 2021-11-30
Make URL correct

ancientstraits:
The URL in the tutorial isn't correct. This makes it correct by adding `tree/master` to the URL.


ancientstraits:
I have the CLA now

ancientstraits:
apparently the url was right all along. Please reply if this PR was actually correct.

marquiz:
> apparently the url was right all along. Please reply if this PR was actually correct.

Yeah, kubectl/kustomize understands it correctly

PR #682, opened 2021-11-30
source/kernel: drop length check of kconfig values

marquiz:
Do not do length checking here. We do not need/want to limit the values
here because they could still be used in custom rules. Moreover, we do
more proper validation of label all label values in nfd-worker, anyway.


ArangoGutierrez (approved):
/lgtm

PR #683, opened 2021-11-30
source/kernel: ditch regexp in kconfig parsing

marquiz:
Regexp is just a resource hog here and does not simplify the code.


marquiz:
ping @mythi 

/assign zvonkok

PR #684, opened 2021-11-30
source/kernel: unmangled kconfig values for custom rules 

marquiz:
Stop converting `=y` and `=m` to `"true"` for the raw feature values used
in "kernel.config" custom rule processing.

In practice, this means that to check if a kernel config flag has been
set to "y" or "m", one needs to explicitly check for both of the values:

```
  matchFeatures:
    - feature: kernel.config
      matchExpressions:
        FOO: {op: In, value: ["y", "m"]}
```

instead of (how it used to be):

```
  matchFeatures:
    - feature: kernel.config
      matchExpressions:
        FOO: {op: IsTrue}
```

The legacy kconfig custom rule is unchanged as are the
kernel-config.<flag> feature labels.

Builds on top of #683

### DIFF
--- a/source/custom/rules/kconfig_rule.go
+++ b/source/custom/rules/kconfig_rule.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 
 	nfdv1alpha1 "sigs.k8s.io/node-feature-discovery/pkg/apis/nfd/v1alpha1"
-	"sigs.k8s.io/node-feature-discovery/source"
 	"sigs.k8s.io/node-feature-discovery/source/kernel"
 )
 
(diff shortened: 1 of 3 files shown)


marquiz:
Depends on #683
/hold

mythi (commented):
LGTM

would it make sense to add a comment somewhere to explain what is `legacyKconfig`?

marquiz:
Thanks for the review @mythi 

> would it make sense to add a comment somewhere to explain what is `legacyKconfig`?

Yeah, makes sense. Also, the name `legacyKconfig` is not necessarily the most understandable but the best I could come up with ;) I added some comments (also to `parseKconfig()`)

marquiz:
Rebased.

#683 merged ->
/unhold

mythi:
LGTM

zvonkok:
@marquiz please rebase :)

marquiz:
> @marquiz please rebase :)

Done!

PR #688, opened 2021-12-02
scripts: increase e2e-test image poll timeout to 12mins

marquiz:
Rare timeouts (narrowly missing) with 10 minutes observed, try
increasing by 2 minutes.

Also fix printing of error message in case of a timeout.


ArangoGutierrez (approved):
/lgtm

PR #695, opened 2021-12-09
source/local: don't prefix label names with the filename

marquiz:
Implicitly injecting the filename of the hook/featurefile into the name
of the label is usually confusing, counter-intuitive and unnecessarily
complex to understand. It's much clearer to advertise features and
labels as presented in the feature file / output of the hook.

NOTE: this breaks backwards compatibility with usage scenarios that rely
on prefixing the label with the filename.

### DIFF
--- a/docs/get-started/features.md
+++ b/docs/get-started/features.md
@@ -569,20 +569,15 @@ update them to match your needs.
 In both cases, the labels can be binary or non binary, using either `<name>` or
 `<name>=<value>` format.
 
-Unlike the other feature sources, the name of the file, instead of the name of
-the feature source (that would be `local` in this case), is used as a prefix in
-the label name, normally. However, if the `<name>` of the label starts with a
-slash (`/`) it is used as the label name as is, without any additional prefix.
-This makes it possible for the user to fully control the feature label names,
-e.g. for overriding labels created by other feature sources.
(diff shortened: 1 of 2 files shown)


marquiz:
/assign @zvonkok @ArangoGutierrez 

This is a backwards-incompatible change which I believe makes for better future, however. WDYT?

shivamerla:
> @shivamerla @cdesiniotis PTAL this should not affect GFD since we're using the fully qualified names for the features.

ACK. This should not affect labels added by GFD as we don't prefix with GFD feature file name.

```
nvidia.com/gfd.timestamp=1639085271
nvidia.com/cuda.driver.major=470
nvidia.com/gpu.family=turing
nvidia.com/cuda.runtime.major=11
nvidia.com/gpu.compute.major=7
nvidia.com/gpu.product=Tesla-T4
nvidia.com/gpu.memory=15109
nvidia.com/gpu.machine=g4dn.xlarge
nvidia.com/cuda.driver.minor=82
nvidia.com/cuda.driver.rev=01
nvidia.com/cuda.runtime.minor=4
nvidia.com/gpu.compute.minor=5
nvidia.com/mig.strategy=single
nvidia.com/gpu.count=1
```

ArangoGutierrez (approved):
It works for our most critical path
/lgtm

PR #697, opened 2021-12-09
nfd-worker: drop 'custom-' prefix from matchFeatures custom rules

marquiz:
Do not prefix label names from the new matchFeatures/matchAny custom
rules with "custom-". We want to have the same result (set of labels)
from a rule independent of whether it has been specified in worker
config or in a NodeFeatureRule CRs. Legacy matchOn rules (not available
in NodeFeatureRule CRs) are intact, i.e. still prefixed, in order to
retain backwards compatibility.


marquiz:
Fixed
/unhold

ArangoGutierrez:
retesting after #695 
/test all

ArangoGutierrez (approved):
/lgtm

PR #582, opened 2021-08-23
WIP: Test if buildx is enabled at Prow

ArangoGutierrez:
Fixes #426 

THIS IS A WIP PR, TO TEST PROW FUNCTIONALITY TOWARDS DOCKER BUILDX

> DO NOT LGTM OR MERGE

Signed-off-by: Carlos Eduardo Arango Gutierrez <carangog@redhat.com>

PR #698, opened 2021-12-10
Multi ARCH build amd64 for now

zvonkok:


ArangoGutierrez:
Thanks for this PR @zvonkok , I started to play around `buildx` a time ago see here :  https://github.com/kubernetes-sigs/node-feature-discovery/pull/582

happy to close that PR and merge efforts here

zvonkok:
@marquiz PTAL, should we try to keep the old structure with build and push command? Should we consider how podman fits into this picture (@ArangoGutierrez)? I think as a start we can keep amd64 and arm64 and then extend if needed.

zvonkok:
Just checked the building would work with podman as eays as: 
```
IMAGE_BUILD_CMD ?= podman build --platform=${PLATFORMS} $(OUTPUT) --progress=$(PROGRESS) --pull
```

marquiz:
> @marquiz PTAL, should we try to keep the old structure with build and push command? Should we consider how podman fits into this picture (@ArangoGutierrez)? I think as a start we can keep amd64 and arm64 and then extend if needed.

Thanks for working on this @zvonkok! Good questions above. I think it would probably be handy to have separate build targets for building only one image (defaulting to the current os/arch) and all archs, i.e. `make image`and `make image-all` or smth (with possible pattern rule for `image-$(os)/$(arch):` targets). This would be handy for development. Also, would there be a nice way to avoid buildx for native builds?

My test build fails, probably because of some http(s) proxy issue. Need to investigate/fix that:

```
 => ERROR [linux/arm64 internal] load metadata for docker.io/library/debian:buster-slim                                                                       30.0s
 => CANCELED [linux/arm64 internal] load metadata for docker.io/library/golang:1.17.2-buster                                                                  30.0s
 => CANCELED [linux/amd64 internal] load metadata for docker.io/library/golang:1.17.2-buster                                                                  29.9s
 => CANCELED [linux/amd64 internal] load metadata for docker.io/library/debian:buster-slim                                                                    29.9s
------
 > [linux/arm64 internal] load metadata for docker.io/library/debian:buster-slim:
------
Dockerfile:30
--------------------
  28 |     
  29 |     # Create full variant of the production image
  30 | >>> FROM ${BASE_IMAGE_FULL} as full
  31 |     
  32 |     # Run as unprivileged user
--------------------
error: failed to solve: debian:buster-slim: failed to do request: Head "https://registry-1.docker.io/v2/library/debian/manifests/buster-slim": dial tcp 34.238.187.50:443: i/o timeout
make: *** [Makefile:91: image] Error 1
```

zvonkok:
@marquiz What about keeping the old Makefile intact. This enables native builds if you're e.g running on arm64 you do just a make image etc. and those commands would stay the same. 

Additionally to that we can create as you suggested some new targets for cross builds with buildx. I will push another version soon.

zvonkok:
This keeps the original Makefile almost completely intact, we're adding an include for the multi arch Makefile. 
```
# Build all preconfigure OS/ARCH combinations (linux + amd64, arm64) 
make image-all
# Push all
make push-all


# Build a specific OS/ARCH combination 
OS=linux ARCH=arm64 make image-${OS}/${ARCH}
# Push linux/arm64
OS=linux ARCH=arm64 make push-${OS}/${ARCH}
```

zvonkok:
Actually we do not need the extra targets. We could simply do: 
```
PLATFORMS=linux/amd64 make image-all
# or only arm64
PLATFORMS=linux/arm64 make image-all
# or arm64,amd64,s390x
PLATFORMS=linux/arm64,linux/amd64,linux/s390x make image-all
```

zvonkok:
@marquiz Which one do you prefer?

zvonkok:
@ArangoGutierrez Any comments from your side?

ArangoGutierrez:
> Actually we do not need the extra targets. We could simply do:
> 
> ```
> PLATFORMS=linux/amd64 make image-all
> # or only arm64
> PLATFORMS=linux/arm64 make image-all
> # or arm64,amd64,s390x
> PLATFORMS=linux/arm64,linux/amd64,linux/s390x make image-all
> ```

I would only set the Arch, unless we are planning on adding support for non-linux NFD deployments (we don't wanna do there right?) 

For the rest I like the idea of the ENV VAR

`ARCH=arm64 make image-${ARCH}`

zvonkok:
@ArangoGutierrez I would leave it as 
```
PLATFORMS=linux/arm64,linux/amd64,linux/s390x make image-all
```
1. The `PLATFORM` is the cmdline for `docker` and `podman` no changes needed for `--platform=$(PLATFORM)`
2. If we only go for arch e.g. `ARCH=amd64,arm64 make image-${ARCH}` we would need any combination of archs as targets. 
3. I do not want to implement some parsing list handling in the Makefile. 

@marquiz WDYT?

marquiz:
Hmm, I now took a closer look at this. I don't mind refactoring the existing Makefile. It's not perfect by any means 😇

Looking at this I think that it might be better to have separate `make image-cross` (or `make image-multiarch` or smth.) target instead of exploiting the existing `make image` target. Just to keep it simple and easier to understand (and maintain). We could have a separate `hack/build.sh` script or capture the common builder args into `IMAGE_BUILD_ARGS` variable or similar.

Also, I'm not sure how/if the image promotion in https://github.com/kubernetes/k8s.io will work 🧐 Maybe we should build explicit `node-feature-discovery-$(ARCH):$(tag)` (and `node-feature-discovery-$(ARCH):$(tag)-minimal`) images instead? In practice I think, this would mean having pattern rules like `image-cross-%`, `push-cross-%`. Then we'd have something like

```
image-cross-all: $(foreach arch,$(ARCHS),image-cross-$(arch))

image-cross-%: yamls
        $(IMAGE_BUILD_CMD_CROSS) 
        ...
```

BTW: shouldn't the GRPC probes be updated? 🤔 

A bit unrelated note: if multiarch `push` also builds the image, I'd be inclined to change the normal `make push` to behave similarly. This can addressed in a separate PR, though.

Any thoughts?

zvonkok:
For mulitarch the image registries support manifest lists so essentially you have one tag that represents several architectures. Depending on the arch you're running on a 
```
<WHATEVER> pull node-feature-discovery:vX.Y
```
will just work.

marquiz:
> For mulitarch the image registries support manifest lists so essentially you have one tag that represents several architectures. Depending on the arch you're running on a

Yeah, I know that. I'm just not sure how https://github.com/kubernetes/k8s.io/blob/main/k8s.gcr.io/images/k8s-staging-nfd/images.yaml fits into that picture.

ArangoGutierrez:
> 

https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

marquiz:
> https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

Yup 👍  We figured this out with @zvonkok that buildx does create the manifest list, AND, k8s test-infra image promoter should indeed handle those without issues.

zvonkok:
Mildly interesting: Set buildx as the default builder: https://github.com/docker/cli/pull/3314

ArangoGutierrez:
We also need a solution to address https://github.com/kubernetes-sigs/node-feature-discovery/pull/563 when doing Multi-Arch builds, I propose not pulling a binary but pulling the source, and compiling during build.

fcami:
The quick solution to always get the tip of the master branch like this:
- Make sure you have git and go installed.
- Run: go get github.com/grpc-ecosystem/grpc-health-probe
- This will compile the binary into your $GOPATH/bin (or $HOME/go/bin).
As listed in the grpc-health-probe documentation.

fcami:
The longer solution is to have a variable with the version we're interested in and pull from that git tag.

marquiz:
We definitely want a specific version, for reproducibility, like we currently do. I guess the simplest thing to do is just do:

```
RUN go install github.com/grpc-ecosystem/grpc-health-probe@${GRPC_HEALTH_PROBE_VERSION}
```

in the dockerfile

marquiz:
Issue with gprc probes has been addressed. Also, the image promotion of multiarch images (which I was worried about) is not problem.

hack/init-buildx.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
+# Copyright 2020 The Kubernetes Authors.
```
ArangoGutierrez, line 2: Update year

ArangoGutierrez (changes requested):

ArangoGutierrez:
After cutting v0.10 let's re test
/test all

Makefile
```diff
@@ -32,7 +32,7 @@ JEKYLL_OPTS := -d '$(SITE_DESTDIR)' $(if $(SITE_BASEURL),-b '$(SITE_BASEURL)',)
 
 VERSION := $(shell git describe --tags --dirty --always)
 
-IMAGE_REGISTRY ?= k8s.gcr.io/nfd
+IMAGE_REGISTRY ?= quay.io/zvonkok
```
marquiz, line 35: ??

Makefile
```diff
@@ -69,22 +101,15 @@ install:
...
-	    -t $(IMAGE_TAG)-minimal \
-	    $(foreach tag,$(IMAGE_EXTRA_TAGS),-t $(tag)-minimal) \
-	    $(IMAGE_BUILD_EXTRA_OPTS) ./
+	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_TAG_FULL)
+	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_TAG_MINIMAL)
+
+image-all: OUTPUT=--pull
+image-all: IMAGE_BUILD_CMD=$(IMAGE_BUILDX_CMD)
+image-all: ensure-buildx yamls
+	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_TAG_FULL)
```
marquiz, line 110: We could possibly get rid of the `OUTPUT` variable. WDYT?

```suggestion
image-all: IMAGE_BUILD_CMD=$(IMAGE_BUILDX_CMD)
image-all: ensure-buildx yamls
	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_TAG_FULL) --pull
```

Makefile
```diff
@@ -59,6 +59,38 @@ E2E_TEST_CONFIG ?=
...
+ensure-buildx:
+	./hack/init-buildx.sh
+
+IMAGE_BUILDX_CMD ?= docker buildx build --platform=${PLATFORMS} $(OUTPUT) --progress=$(PROGRESS)
+
+IMAGE_BUILD_ARGS = --build-arg VERSION=$(VERSION) \
+	    	   --build-arg HOSTMOUNT_PREFIX=$(CONTAINER_HOSTMOUNT_PREFIX) \
+	    	   --build-arg BASE_IMAGE_FULL=$(BASE_IMAGE_FULL) \
+	    	   --build-arg BASE_IMAGE_MINIMAL=$(BASE_IMAGE_MINIMAL)
+IMAGE_TAG_FULL = --target full \
```
marquiz, line 84: nit: Pondering the name of the var. `IMAGE_BUILD_ARGS_FULL`, WDYT?

Makefile
ArangoGutierrez: Zvonko secretly moving us to a new world

Makefile
zvonkok: Yeah, i realized that I needed to move the `--target` thing also into the tag definition so technically it is not a `TAG` only vars.
`IMAGE_BUILD_ARGS_FULL` + `IMAGE_BUILD_ARGS_MINIMAL`?

zvonkok: Update @marquiz PTAL

Makefile
marquiz: I'd suggest to change this back to `k8s.gcr.io/nfd` 😉

scripts/test-infra/push-image.sh
```diff
@@ -5,5 +5,8 @@
 # container image tag
 VERSION_OVERRIDE=${_GIT_TAG+VERSION=${_GIT_TAG:10}}
 
-make image $VERSION_OVERRIDE
-make push $VERSION_OVERRIDE
+make image-all $VERSION_OVERRIDE
+make push-all $VERSION_OVERRIDE
```
marquiz, line 9: `push-all` always also builds images, right(?) So `make image-all` is kinda redundant here?

```suggestion
make push-all $VERSION_OVERRIDE
```

Makefile
```diff
@@ -59,6 +59,39 @@ E2E_TEST_CONFIG ?=
...
+# required to enable buildx
+export DOCKER_CLI_EXPERIMENTAL=enabled
+
+# build with buildx
+PLATFORMS ?= linux/amd64,linux/arm64
+PROGRESS   = auto
+# buildx keeps the build in the cache no default output
+# use --push as OUTPUT to push to registry
+# or --load to have it available as an image
+OUTPUT =
```
marquiz, line 72: Should/could we get rid of this `OUTPUT` variable, looks kinda odd and stray. What about just appending `--pull`/`--push` in the corresponding builder commands, e.g.

```
$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL) --pull
```

zvonkok:
Just confused two parameters, the `--pull` just says to pull the latest images that are used during the build. The `--load` parameter makes the images available to e.g. `docker images`, manifest-lists are not yet implemented for `docker` see https://github.com/docker/buildx/issues/59 so we're omitting it during the build. The registry supports manifest-lists so we can `push`.

Makefile
```diff
@@ -59,6 +59,35 @@ E2E_TEST_CONFIG ?=
 
 LDFLAGS = -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$(VERSION) -X sigs.k8s.io/node-feature-discovery/source.pathPrefix=$(HOSTMOUNT_PREFIX)"
 
+## multi-arch build
+# required to enable buildx
+export DOCKER_CLI_EXPERIMENTAL=enabled
+
+# build with buildx
+PLATFORMS ?= linux/amd64,linux/arm64
+PROGRESS   = auto
```
marquiz, line 68: Is the `PROGRESS` variable necessary? I'd drop it. Or at least rename it to something like `BUILDX_PROGRESS`

Makefile
```diff
@@ -59,6 +59,35 @@ E2E_TEST_CONFIG ?=
 
 LDFLAGS = -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$(VERSION) -X sigs.k8s.io/node-feature-discovery/source.pathPrefix=$(HOSTMOUNT_PREFIX)"
 
+## multi-arch build
+# required to enable buildx
+export DOCKER_CLI_EXPERIMENTAL=enabled
+
+# build with buildx
+PLATFORMS ?= linux/amd64,linux/arm64
```
marquiz, line 67: Hmm, bike-shedding/nitting, but what do you think about renaming this to e.g. `BUILDX_PLATFORMS`, `IMAGE_ALL_PLATFORMS` or smth(?) Because if you do e.g. `make build` your build isn't affected.

marquiz (commented):
Nice @zvonkok, I think we're almost there 😸 

One thing is that it would be nice to get some mention about the multiarch builds in `docs/advanced/developer-guide.md`. And document the new variables like `IMAGE_BUILDX_CMD` there, too.

Makefile
zvonkok: Yeah was thinking of this as well :)

Makefile
zvonkok: Right, will update asap.

Makefile
```diff
@@ -4,7 +4,7 @@
 GO_CMD ?= go
 GO_FMT ?= gofmt
 
-IMAGE_BUILD_CMD ?= docker build
+IMAGE_BUILD_CMD ?= DOCKER_BUILDKIT=1 docker build
```
marquiz, line 7: Hmm, I think this is unnecessary(?)

Makefile
```diff
@@ -69,22 +97,14 @@ install:
...
-	    -t $(IMAGE_TAG)-minimal \
-	    $(foreach tag,$(IMAGE_EXTRA_TAGS),-t $(tag)-minimal) \
-	    $(IMAGE_BUILD_EXTRA_OPTS) ./
+	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
+	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
+
+image-all: IMAGE_BUILD_CMD=$(IMAGE_BUILDX_CMD) # --load : not implemented yet see: https://github.com/docker/buildx/issues/59
+image-all: ensure-buildx yamls
+	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
+	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
```
marquiz, line 106: Wouldn't this simplify a bit, still(?)

```suggestion
image-all: ensure-buildx yamls
	# --load : not implemented yet see: https://github.com/docker/buildx/issues/59
	$(IMAGE_BUILDX_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
	$(IMAGE_BUILDX_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
```

Makefile
```diff
@@ -166,6 +186,13 @@ push:
 	$(IMAGE_PUSH_CMD) $(IMAGE_TAG)-minimal
 	for tag in $(IMAGE_EXTRA_TAGS); do $(IMAGE_PUSH_CMD) $$tag; $(IMAGE_PUSH_CMD) $$tag-minimal; done
 
+push-all: IMAGE_PUSH_CMD=$(IMAGE_BUILDX_CMD) --push
+push-all:
+	$(IMAGE_PUSH_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
+	$(IMAGE_PUSH_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
```
marquiz, line 192: Clean-up, simplify(?)

```suggestion
push-all:
	$(IMAGE_BUILDX_CMD) --push $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
	$(IMAGE_BUILDX_CMD) --push $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
```

Makefile
```diff
@@ -166,6 +186,13 @@ push:
...
+
+
+
```
marquiz, line 195: nit: stale empty lines

```suggestion
```

Makefile
```diff
@@ -59,6 +59,34 @@ E2E_TEST_CONFIG ?=
 
 LDFLAGS = -ldflags "-s -w -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$(VERSION) -X sigs.k8s.io/node-feature-discovery/source.pathPrefix=$(HOSTMOUNT_PREFIX)"
 
+## multi-arch build
+# required to enable buildx
+export DOCKER_CLI_EXPERIMENTAL=enabled
```
marquiz, line 64: Hmm, do we want to export this globally or just set it when needed? Smth like
```
IMAGE_BUILDX_CMD ?= DOCKER_CLI_EXPERIMENTAL=enabled docker buildx build...
```

scripts/test-infra/push-image.sh
```diff
@@ -5,5 +5,7 @@
 # container image tag
 VERSION_OVERRIDE=${_GIT_TAG+VERSION=${_GIT_TAG:10}}
 
-make image $VERSION_OVERRIDE
-make push $VERSION_OVERRIDE
+make push-all $VERSION_OVERRIDE
+
+
+
```
marquiz, line 11: nit: stale linefeeds
```suggestion
```

marquiz (commented):
Thanks for the update! Few more small comments

And the docs...

docs/advanced/developer-guide.md
```diff
@@ -125,6 +127,8 @@ makefile overrides.
 | BASE_IMAGE_FULL            | Container base image for target image full (--target full)        | debian:buster-slim
 | BASE_IMAGE_MINIMAL         | Container base image for target image minimal (--target minimal)  | gcr.io/distroless/base
 
+
+
```
marquiz, line 131: mdlint hates extra empty lines 😄 

```suggestion
```

marquiz (commented):
One nit to make mdlint happy. Other than that, I think you could squash these into one commit and we could get it merged 👍

docs/advanced/developer-guide.md
zvonkok: Yeah I should take a look at my warnings .... Sorry about that!

marquiz (commented):
Could you squash a bit more, one commit should be enough (the second commit *Added additional Makefile for mulit arch builds* doesn't make much sense) 😉

zvonkok:
@marquiz should be good now? :)

marquiz (approved):
Thanks @zvonkok for the effort! This has been on the wishlist of people for quite some time
/lgtm

PR #699, opened 2021-12-13
deployment/helm: disable nfr controller for parallel instances

marquiz:
Change the helm chart so that the NodeFeatureRule controller will be
disabled for other than the default deployment (i.e. all deployments
where master.instance is non-empty), unless explicitly set to true. With
this we try to ensure that there is only on controller instance for the
CR, avoiding contention and conflicts.


marquiz:
/assign @ArangoGutierrez 

ping @eliaskoromilas, any feedback

marquiz:
Rebased.

ping @eliaskoromilas @dbaker-rh any feedback?

deployment/helm/node-feature-discovery/templates/master.yaml
```diff
@@ -78,7 +78,11 @@ spec:
             {{- if .Values.master.extraLabelNs | empty | not }}
             - "--extra-label-ns={{- join "," .Values.master.extraLabelNs }}"
             {{- end }}
+            {{- if .Values.master.featureRulesController | kindIs "invalid" | not }}
             - "-featurerules-controller={{ .Values.master.featureRulesController }}"
+            {{- else if .Values.master.instance | empty | not }}
+            - "-featurerules-controller=false"
```
eliaskoromilas, line 84: I was confused when I first saw this, but then I checked the code and the default flag value is actually `true` (which is not really common for a boolean flag).

```suggestion
            {{- else }}
            - "-featurerules-controller={{ .Values.master.instance | empty }}"
```

For some reason, I don't feel confident about this, however, it would have made perfect sense if the case was the opposite (the flag were `false` by default and we skipped setting it :). What do you think?

dbaker-rh:
> 

Sorry - this isn't part of nfd that I've had any need to look at yet, so I don't really have an opinion yet.

deployment/helm/node-feature-discovery/templates/master.yaml
marquiz: Thanks for looking into this @eliaskoromilas! I amended your suggestion.

Having bool flags that default `true` is not that uncommon in K8s at least (kubelet, kube-apiserver etc).

You propose changing the cmdline flag to something like `--disable-featurerules-controller` which would default to false? Need to think about that but I feel a bit reluctant to do that just for this edge case. Another solution might be to rename the Helm field to something like `master.forceFeatureRulesController` but dunno if that's any better.

eliaskoromilas: > You propose changing the cmdline flag to something like `--disable-featurerules-controller` which would default to false? Need to think about that but I feel a bit reluctant to do that just for this edge case. Another solution might be to rename the Helm field to something like `master.forceFeatureRulesController` but dunno if that's any better.

No, I think it's just fine as is. As you noted, agents tend to use that kind of explicit language in their configuration.

I suggested that change for clarity.

marquiz: > No, I think it's just fine as is. As you noted, agents tend to use that kind of explicit language in their configuration.

K, let's keep it like this, then. Anyway, I think this would be a really rarely used setting

> I suggested that change for clarity.

Yeah, that was a good one 👍

marquiz:
ping @ArangoGutierrez, seems like we're ok with this

ArangoGutierrez (approved):
/lgtm

PR #704, opened 2021-12-20
docs: add customization guide

marquiz:
Add a separate customization guide. Move documentation of the custom and
local sources there. Also, cover the new NodeFeatureRules custom
resource and the new expression-based label rule format.

This patch also simplifies the "Feature labels" page, describing
built-in labels. Reformat the tables describing feature labels.

### DIFF
--- a/docs/advanced/customization-guide.md
+++ b/docs/advanced/customization-guide.md
@@ -2,7 +2,6 @@
 title: "Customization guide"
 layout: default
 sort: 5
-published: false
 ---
 
 # Customization guide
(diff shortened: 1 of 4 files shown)


marquiz:
/assign @zvonkok  @ArangoGutierrez 
ping @mythi 

It's possibly easier to review by looking at the end result. This can be seen at gh-pages of my personal repo:
https://marquiz.github.io/node-feature-discovery/master/get-started/features.html
https://marquiz.github.io/node-feature-discovery/master/advanced/customization-guide.html

Some notes/questions:
- I simplified the page [describing built-in labels](https://marquiz.github.io/node-feature-discovery/master/get-started/features.html) quite a bit, also changed the formatting of the tables. WDYT?
- I tried to make the guide more example-driven, starting with a simple example and describing the details after that.
- Terminology is always hard, e.g. [feature types](https://marquiz.github.io/node-feature-discovery/master/advanced/customization-guide.html#feature-types)
- Customization guide is quite long page, could/should it be split into multiple sub-pages 🤔

zvonkok:
Looking good to me, but I would like to have someone else to take a look, just to be sure we're not missing anything. @ArangoGutierrez PTAL, thanks!

docs/advanced/customization-guide.md
```diff
@@ -16,8 +15,1007 @@ published: false
...
+```yaml
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: my-rule-object
+spec:
+  rules:
+    - name: "my sgx feature rule"
+      labels:
+        "my-sgx-feature": "true"
```
mythi, line 49: nit: even if this is just an example, to make it work in all cases add (or change `cpu.cpuid`):
```
     - feature: cpu.sgx
        matchExpressions:
          enabled: {op: IsTrue}
```

docs/advanced/customization-guide.md
```diff
@@ -16,8 +15,1007 @@ published: false
...
+```
+
+It specifies one rule which creates node label
+`feature.node.kubenernetes.io/my-sgx-feature=true` if both of the following
+conditions are true (`matchFeatures` implements a logical AND over the
+matchers):
+
+- the CPU has SGX capability
+- X86_SGX option has been enabled in the kernel
+
```
mythi, line 66: maybe add the basic `$ kubectl apply -f nfd-rule-sgx.yaml` to show how these are deployed, or maybe also the syntax how CRs can be added via `kustomization.yaml`.

docs/advanced/customization-guide.md
```diff
@@ -16,8 +15,1007 @@ published: false
...
+## Local feature source
+
+NFD-Worker has a special feature source named *local* which is an integration
+point for external feature detectors. It provides a mechanism for pluggable
+extensions, allowing the creation of new user-specific features and even
+overriding built-in labels.
+
+The *local* feature source has two methods for detecting features, hooks and
+feature files.
```
mythi, line 92: some readers might ask can `local` sources info be used with `matchFeatures` with the two other customization features?

docs/advanced/customization-guide.md
```diff
@@ -16,8 +15,1007 @@ published: false
...
+The standard NFD deployments contain `hostPath` mounts for
+`/etc/kubernetes/node-feature-discovery/source.d/` and
+`/etc/kubernetes/node-feature-discovery/features.d/`, making these directories
+from the host available inside the nfd-worker container.
+
+#### Injecting labels from other pods
+
+One use case for the hooks and/or feature files is detecting features in other
+Pods outside NFD, e.g. in Kubernetes device plugins. By using the same
+`hostPath` mounts for `/etc/kubernetes/node-feature-discovery/source.d/` and
```
mythi, line 183: it's not NFD specific as such but on SELinux enabled platforms it's more than just `hostPath`. Maybe add a note here that it needs to be taken into account too or, preferably, the suggestion what needs to be done to overcome that limitation.

docs/advanced/customization-guide.md
```diff
@@ -16,8 +15,1007 @@ published: false
...
+        - feature: cpu.cpuid
+          matchExpressions:
+            SGX: {op: Exists}
+        - feature: kernel.config
+          matchExpressions:
+            X86_SGX: {op: In, value: ["y"]}
+```
+
+It specifies one rule which creates node label
+`feature.node.kubenernetes.io/my-sgx-feature=true` if both of the following
```
mythi, line 222: `custom` rule adds it's own prefix:
`feature.node.kubenernetes.io/custom-my-sgx-feature=true`

docs/advanced/customization-guide.md
```diff
@@ -16,8 +15,1007 @@ published: false
...
+```yaml
+core:
+  labelSources: ["custom"]
+sources:
+  custom:
+    - name: "my sgx feature rule"
+      labels:
+        "my-sgx-feature": "true"
+      matchFeatures:
+        - feature: cpu.cpuid
```
mythi, line 213: the earlier comment about `cpu.sgx` applies here too.

docs/advanced/customization-guide.md
```diff
@@ -16,8 +15,1007 @@ published: false
...
+Feature labels have the following format:
+
+```plaintext
+<namespace>/<name> = <value>
+```
+
+The namespace part (i.e. prefix) of the labels is controlled by nfd:
+
+- All built-in labels use `feature.node.kubernetes.io`. This is also
+  the default for user defined features that don't specify any namespace.
```
mythi, line 272: and that `<name>` includes the source as prefix?

docs/get-started/features.md
```diff
@@ -137,519 +105,111 @@ See the full list in [github.com/klauspost/cpuid][klauspost-cpuid].
...
 
-The **network** feature source supports the following labels:
-
-| Feature | Attribute  | Description                                           |
-| ------- | ---------- | ----------------------------------------------------- |
-| sriov   | capable    | [Single Root Input/Output Virtualization][sriov] (SR-IOV) enabled Network Interface Card(s) present
-|         | configured | SR-IOV virtual functions have been configured
+| Feature     | Value | Description
+| ----------- | ----- | -----------
+| **`network-sriov.capable`**    | true | [Single Root Input/Output Virtualization][sriov] (SR-IOV) enabled Network Interface Card(s) present
```
mythi, line 153: did you add some new labels for `network`?

mythi (commented):
A couple of comments/questions. (I skipped the match syntax part)

docs/advanced/customization-guide.md
marquiz: Yeah, I know. The idea behind the example was to have a rule with more than one matcher. Maybe I'll change this to something completely different like `cpu.pststate.turbo` and `kernel.config.PREEMPT` or smth 🤔

docs/advanced/customization-guide.md
marquiz: Makes sense. Maybe we could even include the sample CR in the repo 🧐

docs/advanced/customization-guide.md
marquiz: Good point. Any suggestion how to fix this? *The features discovered by the local source can further be used in label rules of NodeFeatureRule objects and the custom source*(?)

docs/advanced/customization-guide.md
marquiz: It shouldn't anymore, after #697.

The legacy `matchOn` rules add the prefix, though. Maybe we should add a note about this 🤔

docs/advanced/customization-guide.md
marquiz: For built-in labels, yes. But I don't think it needs to be stated here(?)

docs/get-started/features.md
marquiz: Nope, this has been there for a while, already

docs/advanced/customization-guide.md
marquiz: TBH, I haven't tried this in an SELinux-enabled environment. Any suggestion what to add? *Note that if SELinux is enabled you need to correctly set `seLinuxOptions` in the security context of all pods mounting the same host path.*(?)

docs/advanced/customization-guide.md
mythi: Good to know!

docs/get-started/features.md
mythi: I was referring to #660 but after taking a closer look it did not change the existing labels.

docs/advanced/customization-guide.md
mythi: I was thinking @ArangoGutierrez might be able to help with that.

docs/advanced/customization-guide.md
mythi: The proposal looks good to me!

docs/advanced/customization-guide.md
marquiz: Changed this to
```
      matchFeatures:
          matchExpressions:
            dummy: {op: Exists}
        - feature: kernel.config
          matchExpressions:
            X86: {op: In, value: ["y"]}
```

docs/advanced/customization-guide.md
marquiz: Added a kubectl example and added a sample CR yaml in the repo

docs/advanced/customization-guide.md
marquiz: Changed along these lines

docs/advanced/customization-guide.md
marquiz: Changed the example ximilarly to the CR example, using `kernel.loadedmodule.dummy`

docs/advanced/customization-guide.md
marquiz: Did not change this

marquiz:
Thanks for the feedback @mythi! Addressed most of the feedback, hostpath mounts with selinux still open.

Review changes are now in a separate commit. Will squash after we're happy
/hold

docs/advanced/customization-guide.md
ArangoGutierrez: OpenShift has SELinux enabled, or you can enable it on MiniKube for testing.
Whit that said, yes, I usually get around that by running the 2 containers in the same pod, that way they share namespace and you can easily share mounts. (This for documentation comes handy we don't want people to think they can run pod's on other nodes or something, it has to be same node, and preferably same pod)

mythi: This doc talks about pods outside nfd. My comment was should we add note that they need to run with some specific selinux label to be able to write to nfd dirs?

docs/advanced/customization-guide.md
```diff
@@ -16,8 +15,1030 @@ published: false
...
+kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/node-feature-discovery/{{ site.release }}/examples/nodefeaturerule.yaml
+```
+
+Now, on X86 platforms the feature label appears after doing `modprobe dummy` on
+a system and correspondingly the label is removed after `rmmod dummy`. Note a
+re-labeling delay up to the sleep-interval of nfd-worker (1 minute by default).
+
+### NodeFeatureRule controller
+
+NFD-Master acts as the controller for NodeFeatureRule objects. It applies these
```
mythi, line 79:
```suggestion
NFD-Master acts as the controller for `NodeFeatureRule` objects. It applies these
```

ditto for other cases too

mythi (commented):
LGTM, one small suggestion

docs/advanced/customization-guide.md
marquiz: Yeah, maybe it makes it more readable. Changed as suggested. I also changed *local* and *custom* similarly

marquiz:
@mythi I addressed your comment in a separate patch. You check if you want.

@ArangoGutierrez anything to add? If not, I'll squash this and we could possibly merge

ArangoGutierrez:
> @mythi I addressed your comment in a separate patch. You check if you want.
> 
> @ArangoGutierrez anything to add? If not, I'll squash this and we could possibly merge

Squash it!

marquiz:
Rebased and squashed

mythi (approved):
+1

marquiz:
ping @zvonkok @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #677, opened 2021-11-26
source/iommu: deprecate and disable by default 

marquiz:
Deprecate the iommu source and disable it by default.

First step of #676

PR #705, opened 2021-12-21
source/pci: add iommu_group/type attribute

marquiz:
Add `iommu_group/type` to the list of PCI device attributes that are
discovered. The value is the raw value from sysfs (i.e DMA, DMA-FQ or
identity).

No built-in (automatic) labels are generated based on this, but, the
attribute is available for custom label rules to use. Examples of custom
rules:

```
  - name: "iommu enabled rule"
    labels:
      iommu.enabled: "true"
    matchFeatures:
      - feature: pci.device
        matchExpressions:
          "iommu_group/type": {op: NotIn, value: ["unknown"]}
```

```
  - name: "iommu passthrough rule"
    labels:
      iommu.passthrough: "true"
    matchFeatures:
      - feature: pci.device
        matchExpressions:
          "iommu_group/type": {op: In, value: ["identity"]}
```

### DIFF
--- a/docs/advanced/customization-guide.md
+++ b/docs/advanced/customization-guide.md
@@ -515,7 +515,7 @@ The following features are available for matching:
 | **`network.device`** | instance |          |            | Physical (non-virtual) network interfaces present in the system
 |                  |              | **`<sysfs-attribute>`** | string | Sysfs network interface attribute, available attributes: `name`, `operstate`, `speed`, `sriov_numvfs`, `sriov_totalvfs`
 | **`pci.device`** | instance     |          |            | PCI devices present in the system
-|                  |              | **`<sysfs-attribute>`** | string | Value of the sysfs device attribute, available attributes: `class`, `vendor`, `device`, `subsystem_vendor`, `subsystem_device`, `sriov_totalvfs`
+|                  |              | **`<sysfs-attribute>`** | string | Value of the sysfs device attribute, available attributes: `class`, `vendor`, `device`, `subsystem_vendor`, `subsystem_device`, `sriov_totalvfs`, `iommu_group/type`
 | **`storage.device`** | instance |          |            | Block storage devices present in the system
 |                  |              | **`<sysfs-attribute>`** | string | Sysfs network interface attribute, available attributes: `name`, `dax`, `rotational`, `nr_zones`, `zoned`
 | **`system.osrelease`** | attribute |          |            | System identification data from `/etc/os-release`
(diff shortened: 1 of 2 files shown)


marquiz:
Depends on #677 
/hold

marquiz:
/assign zvonkok
ping @mythi

mythi:
```
"iommu_group/type": {op: Exists}
```
I have a VM without an IOMMU and it gives `iommu_group/type` as well. 

```
matchFeatures:
      - feature: pci.device
```
does it mean there's also `net.device` with the same function after this PR?

Might be useful to have more capabilities supported too, e.g., amd vs intel IOMMU and IOMMU version

marquiz:
> I have a VM without an IOMMU and it gives `iommu_group/type` as well.

Hmm, interesting 🧐 Want to take a closer look at that, what is the value of the type then?
 
> does it mean there's also `net.device` with the same function after this PR?

Yup, those will have the same property.

> Might be useful to have more capabilities supported too, e.g., amd vs intel IOMMU and IOMMU version

We could add e.g. `iommu/intel-iommu/version` to address that. Do we have a use case for that, WDYT? I have no experience (or hw) on amd.

marquiz:
Maybe it the instructions to see if iommu is enabled should then be

```
  - name: "iommu enabled rule"
    labels:
      iommu.enabled: "true"
    matchFeatures:
      - feature: pci.device
        matchExpressions:
          "iommu_group/type": {op: NotIn, value: ["unknown"]}
```

marquiz:
Rebased. Changed the instructions in the first commit message

marquiz:
@mythi any comments to the above?

ping @zvonkok, any feedback?

This is virtually the last open bit for v0.10.0 😊

marquiz:
Rebased, updated docs

mythi:
`source/pci` is OK but I feel `source/network` is unnecessary. with `source/network` today it's not possible to say to which device the attribute is linked. therefore, for better usability checking "PCI is a NIC and PCI has IOMMU"  sounds more reasonable?

marquiz:
> I feel `source/network` is unnecessary. with `source/network` today it's not possible to say to which device the attribute is linked. therefore, for better usability checking "PCI is a NIC and PCI has IOMMU" sounds more reasonable?

This is a fair question to which I don't have an answer 🧐 I guess why I added this to network, too, was just following sr-iov. The sriov attributes are available for both pci and network devices. `network` can be easily dropped from this pr. Any thoughts @zvonkok ?

marquiz:
Dropped the patch for source/network. Now only concerns source/pci.

/retitle source/pci: add iommu_group/type attribute

marquiz:
Network part now submitted separately in #717

mythi:
/lgtm

thanks!

ArangoGutierrez:
> /lgtm
> 
> thanks!

/lgtm

PR #707, opened 2021-12-22
Dockerfile: build grpc_health-probe from source

ArangoGutierrez:
Current grpc-health-probe functionality is pulling a binary, hard coded
to amd64, both unsecure and only works for 1 arch, preparing to build
NFD for multiple Arch's require we build the health probe from source,
that way we get rid of the unsecure binary pull, and guarantee a proper
arch build for the grpc-health-probe


ArangoGutierrez:
cc @fcami

marquiz (approved):
I like this, thanks @ArangoGutierrez 
/lgtm

PR #708, opened 2021-12-22
fix probe check

bart0sh:
Due to recently changed way of building health probe binary its
name changed from grpc_health_probe to grpc-health-probe.
It broke readiness check causing non-functional deployments.

Changing binary name should fix this issue.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG BASE_IMAGE_MINIMAL
 # Build node feature discovery
 FROM golang:1.17.2-buster as builder
 
-# Download the grpc_health_probe bin
+# Download the grpc-health-probe bin
 RUN GRPC_HEALTH_PROBE_VERSION=v0.4.6 && \
 	go install github.com/grpc-ecosystem/grpc-health-probe@${GRPC_HEALTH_PROBE_VERSION}
 
--- a/deployment/base/master-worker-combined/master-worker-daemonset.yaml
+++ b/deployment/base/master-worker-combined/master-worker-daemonset.yaml
@@ -21,12 +21,12 @@ spec:
         imagePullPolicy: Always
         livenessProbe:
           exec:
-            command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
+            command: ["/usr/bin/grpc-health-probe", "-addr=:8080"]
           initialDelaySeconds: 10
           periodSeconds: 10
         readinessProbe:
           exec:
-            command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
+            command: ["/usr/bin/grpc-health-probe", "-addr=:8080"]
           initialDelaySeconds: 5
           periodSeconds: 10
           failureThreshold: 10
--- a/deployment/base/master/master-deployment.yaml
+++ b/deployment/base/master/master-deployment.yaml
@@ -22,12 +22,12 @@ spec:
           imagePullPolicy: Always
           livenessProbe:
             exec:
-              command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
+              command: ["/usr/bin/grpc-health-probe", "-addr=:8080"]
             initialDelaySeconds: 10
             periodSeconds: 10
           readinessProbe:
             exec:
-              command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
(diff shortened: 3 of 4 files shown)


bart0sh:
@marquiz @mythi PTAL

ArangoGutierrez (approved):
/lgtm

Dockerfile
```diff
@@ -47,4 +47,4 @@ USER 65534:65534
 ENV GRPC_GO_LOG_SEVERITY_LEVEL="INFO"
 
 COPY --from=builder /go/node-feature-discovery/deployment/components/worker-config/nfd-worker.conf.example /etc/kubernetes/node-feature-discovery/nfd-worker.conf
-COPY --from=builder /go/bin/* /usr/bin/
+COPY --from=builder /go/bin/grpc-health-probe /usr/bin/
```
marquiz, line 50: Hmm, we should copy the nfd binaries too, right?

```suggestion
COPY --from=builder /go/bin/* /usr/bin/
```

bart0sh: @marquiz thank you for the review. That's interesting as I've tested this change before submitting PR. /usr/bin/nfd-master and /usr/bin/nfd-worker binaries present in the image. Do they come there from the base image?

bart0sh: Here is what I can see in the running nfd-master container:
```
nobody@nfd-master-79bf6776fd-cn586:/$ ls -la /usr/bin/grpc*
-rwxr-xr-x 1 root root 13659507 Dec 22 14:33 /usr/bin/grpc-health-probe

nobody@nfd-master-79bf6776fd-cn586:/$ ls -la /usr/bin/nfd-*
-rwxr-xr-x 1 root root 34410496 Dec 22 14:37 /usr/bin/nfd-master
-rwxr-xr-x 1 root root 35565568 Dec 22 14:37 /usr/bin/nfd-topology-updater
-rwxr-xr-x 1 root root 18563072 Dec 22 14:37 /usr/bin/nfd-worker
```

Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG BASE_IMAGE_MINIMAL
 # Build node feature discovery
 FROM golang:1.17.2-buster as builder
 
-# Download the grpc_health_probe bin
+# Download the grpc-health-probe bin
```
mythi, line 7: since we are changing this file, maybe update that it's no longer "Download" but "Build and install"

mythi: it was Download until #707

bart0sh:
@marquiz [This documentation](https://github.com/grpc-ecosystem/grpc-health-probe) suggests that the binary name is grpc_health_probe. Should we follow their suggestion and don't change binary name? We can either rename it or just revert [recent change](https://github.com/kubernetes-sigs/node-feature-discovery/commit/bc88c91109c29328169ba353de0db17b5e3eb4e4) ?

marquiz:
> @marquiz [This documentation](https://github.com/grpc-ecosystem/grpc-health-probe) suggests that the binary name is grpc_health_probe. Should we follow their suggestion and don't change binary name? We can either rename it or just revert [recent change](https://github.com/kubernetes-sigs/node-feature-discovery/commit/bc88c91109c29328169ba353de0db17b5e3eb4e4). wdyt?

TBH, I'm pretty indifferent about the exact name of the binary. If you want you can rename, e.g. add `&& mv /go/bin/grpc-health-probe /go/bin/grpc_health_probe` to the dockerfile. We definitely don't want to revert that commit but want to do `go install` in order to support multiarch builds.

/ok-to-test

bart0sh:
@marquiz I'd prefer to keep the name: https://github.com/kubernetes-sigs/node-feature-discovery/pull/709

Dockerfile
bart0sh: @mythi thanks. I did it in the new PR: https://github.com/kubernetes-sigs/node-feature-discovery/pull/709

Dockerfile
marquiz: > @marquiz thank you for the review. That's interesting as I've tested this change before submitting PR. /usr/bin/nfd-master and /usr/bin/nfd-worker binaries present in the image. Do they come there from the base image?

We have "normal" and "minimal" image variants. You changed the minimal variant so the normal image worked fine

PR #709, opened 2021-12-23
Fix readiness and liveness checks

bart0sh:
Rename grpc-health-probe -> grpc_health_probe as
deployment yamls refer to it by this name.

This should fix broken NFD deployments.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,9 +4,12 @@ ARG BASE_IMAGE_MINIMAL
 # Build node feature discovery
 FROM golang:1.17.2-buster as builder
 
-# Download the grpc_health_probe bin
+# Build and install the grpc-health-probe binary
 RUN GRPC_HEALTH_PROBE_VERSION=v0.4.6 && \
-	go install github.com/grpc-ecosystem/grpc-health-probe@${GRPC_HEALTH_PROBE_VERSION}
+	go install github.com/grpc-ecosystem/grpc-health-probe@${GRPC_HEALTH_PROBE_VERSION} \
+        # Rename it as it's referenced as grpc_health_probe in the deployment yamls
+        # and in its own project https://github.com/grpc-ecosystem/grpc-health-probe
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,10 +4,13 @@ ARG BASE_IMAGE_MINIMAL
 # Build node feature discovery
 FROM golang:1.17.2-buster as builder
 
-# Download the grpc_health_probe bin
+# Build and install the grpc-health-probe binary
 RUN GRPC_HEALTH_PROBE_VERSION=v0.4.6 && \
 	go install github.com/grpc-ecosystem/grpc-health-probe@${GRPC_HEALTH_PROBE_VERSION}
 
+# Rename it as it's referenced as grpc_health_probe in the deployment yamls
+RUN mv /go/bin/grpc-health-probe /go/bin/grpc_health_probe
```
marquiz: nit: could you join this to the previous `RUN` with `&&`. Just a small optimization to avoid one excess layer. WDYT?

marquiz (commented):
Thanks @bart0sh, we definitely want to fix this asap.

One nit: could you change the [commit message to start with a capital letter](https://www.kubernetes.dev/docs/guide/pull-requests/#the-first-word-in-the-commit-message-subject-should-be-capitalized-unless-it-starts-with-a-lowercase-symbol-or-other-identifier)? 😊 Maybe
```
Fix readiness and liveness checks
```
?

/ok-to-test

Dockerfile
bart0sh: done

marquiz:
> done

Thx! How about the commit message subject?

bart0sh:
> Thx! How about the commit message subject?

I think I've updated it as well. Please, elaborate.

bart0sh:
ah, got it. fixed.

marquiz (approved):
Thanks for the help @bart0sh 
/lgtm

(we'd really need an integration test running for every pr... 🙄 )

PR #710, opened 2021-12-24
Fix kustomization template to work with cert-manager

dbaker-rh:
A few changes in this PR:
1/ Make use of cert-manager's selfSigned{} to bootstrap the actual CA creation (removes the need for a manual openssl step)
2/ Update grpc_health_check to use TLS (without this, the master pod won't start)
3/ Update the certs to include "localhost" in the SAN list, so grpc_health_check can work
4/ Update docs (and bump cert-manager referenced to 1.6.1)

Note - this is partly in response to #563

### DIFF
--- a/deployment/overlays/samples/cert-manager/probes.yaml
+++ b/deployment/overlays/samples/cert-manager/probes.yaml
@@ -1,0 +1,26 @@
+- op: add
+  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
+  value: "-tls"
+- op: add
+  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
+  value: "-tls-ca-cert=/etc/kubernetes/node-feature-discovery/certs/ca.crt"
+- op: add
+  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
+  value: "-tls-client-key=/etc/kubernetes/node-feature-discovery/certs/tls.key"
+- op: add
+  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
+  value: "-tls-client-cert=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
+
+- op: add
+  path: /spec/template/spec/containers/0/readinessProbe/exec/command/-
+  value: "-tls"
+- op: add
+  path: /spec/template/spec/containers/0/readinessProbe/exec/command/-
+  value: "-tls-ca-cert=/etc/kubernetes/node-feature-discovery/certs/ca.crt"
+- op: add
+  path: /spec/template/spec/containers/0/readinessProbe/exec/command/-
+  value: "-tls-client-key=/etc/kubernetes/node-feature-discovery/certs/tls.key"
+- op: add
+  path: /spec/template/spec/containers/0/readinessProbe/exec/command/-
+  value: "-tls-client-cert=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
+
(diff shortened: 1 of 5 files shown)


deployment/overlays/samples/cert-manager/probes.yaml
```diff
@@ -0,0 +1,8 @@
+- op: replace
+  path: /spec/template/spec/containers/0/livenessProbe/exec/command
+  value: ["/usr/bin/grpc_health_probe", "-addr=:8080", "-tls", "-tls-ca-cert=/etc/kubernetes/node-feature-discovery/certs/ca.crt", "-tls-client-key=/etc/kubernetes/node-feature-discovery/certs/tls.key", "-tls-client-cert=/etc/kubernetes/node-feature-discovery/certs/tls.crt"]
```
marquiz, line 3: nit: could we use `add` (instead of `replace`), similarly what we do in `args.yaml`? I think this "additive" method is probably less prone to problems e.g. if something in the base yaml would change.

marquiz (commented):
Nice work @dbaker-rh!

Now that the manual steps of creating the ca cert are eliminated maybe we could even make this a component under `deployment/components/`. WDYT?

/ok-to-test

deployment/overlays/samples/cert-manager/probes.yaml
dbaker-rh: Agree.  I wanted the more simpler patch to begin with, but I agree that to add the new args on the end would be more future proof.

marquiz: I think it would be easier to just change it in this PR. I think we could merge this after that

```suggestion
- op: add
  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
  value: "-tls"
- op: add
  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
  value: "-tls-ca-cert=/etc/kubernetes/node-feature-discovery/certs/ca.crt"
- op: add
  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
  value: "-tls-client-key=/etc/kubernetes/node-feature-discovery/certs/tls.key"
- op: add
  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
  value: "-tls-client-cert=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
```

deployment/overlays/samples/cert-manager/probes.yaml
```diff
@@ -0,0 +1,17 @@
...
+- op: add
+  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
+  value: "-tls-client-key=/etc/kubernetes/node-feature-discovery/certs/tls.key"
+- op: add
+  path: /spec/template/spec/containers/0/livenessProbe/exec/command/-
+  value: "-tls-client-cert=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
+
+- op: replace
+  path: /spec/template/spec/containers/0/readinessProbe/exec/command
+  value: ["/usr/bin/grpc_health_probe", "-addr=:8080", "-tls", "-tls-ca-cert=/etc/kubernetes/node-feature-discovery/certs/ca.crt", "-tls-client-key=/etc/kubernetes/node-feature-discovery/certs/tls.key", "-tls-client-cert=/etc/kubernetes/node-feature-discovery/certs/tls.crt"]
```
marquiz, line 16: Could you change this similarly to liveness probe command above?

dbaker-rh: done.

marquiz (commented):
Looks good 👍 If you could squash this into one commit I think we could merge this

dbaker-rh:
squished!

marquiz:
`probes.yaml` disappeared in the latest force push 🤓

marquiz (approved):
Very welcome improvement @dbaker-rh 😊 
/lgtm

PR #715, opened 2022-01-11
docs: small fix in block and net features in customization guide

marquiz:
In the list of available features, list 'name' separately as it's not
from a sysfs file like the other attributes.


ArangoGutierrez (approved):
/lgtm

PR #717, opened 2022-01-11
source/network: add iommu_group/type attribute

marquiz:
Detect "iommu_group/type" sysfs attribute for network devices. It is
available for custom label rules. The value is the raw value from sysfs
(i.e DMA, DMA-FQ or identity).

An example rule:

  - name: "nic iommu passthrough rule"
    labels:
      network-iommu.passthrough: "true"
    matchFeatures:
      - feature: network.device
        matchExpressions:
          "iommu_group/type": {op: In, value: ["identity"]}


marquiz:
/assign @zvonkok 
ping @mythi

marquiz:
Doesn't seem to be much of interest. Let's re-open if needed
/close

PR #718, opened 2022-01-11
source/network: detect intel-iommu/version

marquiz:
Discover "iommu/intel-iommu/version" sysfs attribute forc network
devices. This information is available for custom label rules.

An example custom rule:

```
  - name: "nic iommu version rule"
    labels:
      network-iommu.version_1: "true"
    matchFeatures:
      - feature: network.device
        matchExpressions:
          "iommu/intel-iommu/version": {op: In, value: ["1:0"]}
```

Builds on #717


marquiz:
Let's re-open if there's some interest in this in the future...
/close

PR #730, opened 2022-01-20
cloudbuild.yaml: set HOME to /root

marquiz:
As per
https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md
we need that for buildx.


marquiz:
Tests on arm seem flaky 🤔 

PR #731, opened 2022-01-20
cloudbuild.yaml: double the timeout

marquiz:
We hit the timeout with default 600s with multiarch builds enabled.


marquiz:
Don't have time to wait for image building in prow as I'm interested in gcloud staging build

PR #732, opened 2022-01-20
nfd-master: print gRPC server error correctly

marquiz:


zvonkok:
Looks like this works. Gonna do a verification run on an ARM system. Done the build several times with build and it worked. Let me sanitize this and approve.

PR #733, opened 2022-01-20
cloudbuild.yaml: increase timeout to 2400s

marquiz:
20 minutes was not clearly enough for multiarch buildx build.


marquiz:
Another short-circuit:

PR #734, opened 2022-01-20
cloudbuild.yaml: increase timeout to 1 hour

marquiz:
Even 40 minutes failed us for multi-arch builds.

Also, increase the e2e-test image polling timeout correspondingly.


marquiz:
This becomes a habit:

PR #735, opened 2022-01-20
cloudbuild.yaml: increase timeout to 1h 20min

marquiz:
Fourth timeout increase's the charm.


marquiz:
The usual:

Issue #737, opened 2022-01-20
Image builds failing with buildx


mnaser:
**What happened**:
When trying to build NFD using `buildx` on an environment where an architecture is emulated (i.e will be slower), tests will fail because the `WaitForReady` timeout for tests is one second only and that's not fast enough in an emulated environment.

**What you expected to happen**:
Test to pass.

**How to reproduce it (as minimally and precisely as possible)**:
```
docker buildx build \
  --cache-from type=local,src=.docker-cache/node-feature-discovery \
  --cache-to type=local,dest=.docker-cache/node-feature-discovery,mode=max \
  --file .build/node-feature-discovery/Dockerfile \
  --tag us-docker.pkg.dev/vexxhost-infra/openstack/node-feature-discovery:0.10.0 \
  --build-arg BASE_IMAGE_FULL=debian:buster-slim \
  --build-arg BASE_IMAGE_MINIMAL=gcr.io/distroless/base \
  --platform linux/amd64,linux/arm64 \
  --push \
  --target minimal \
  .build/node-feature-discovery
```

**Anything else we need to know?**:
Nope :)

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

PR #738, opened 2022-01-21
Increase timeout in test setups

mnaser:
This patch increases the timeout when setting up the NFD master
to 5 seconds instead of 1 second to allow for running tests in
slow environments.

Closes: #737

### DIFF
--- a/pkg/nfd-client/topology-updater/nfd-topology-updater_test.go
+++ b/pkg/nfd-client/topology-updater/nfd-topology-updater_test.go
@@ -53,7 +53,7 @@ func setupTest(args *nfdmaster.Args) testContext {
 		ctx.errs <- ctx.master.Run()
 		close(ctx.errs)
 	}()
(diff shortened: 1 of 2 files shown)


mnaser:
I've signed the CLA.

mnaser:
OK, looks like there is two CLAs or something?  I signed the second one too.

marquiz (approved):
Thanks @mnaser! Hope 5s is enough for everyone 🤞 
/lgtm

marquiz:
We need #732:

zvonkok:
5s are enough in slow environments when building natively, but it is not enough for virtualized builds. I had a 20s timeout and it sometimes failed. Running buildx in a slow environment with 5s can still fail ...

mnaser:
@zvonkok do you think we should bump it to 20 seconds?  It probably doesn't hurt but makes the builds wrk better.

Ideally, the whole timeout shouldn't be a thing but eh..

PR #748, opened 2022-01-21
docs: drop topology-updater cmdline help from developer guide

marquiz:
Similar to what we did for nfd-master and nfd-worker in
0d21b3d7200272157438b76a62f7aad1361fe31e.


ArangoGutierrez (approved):
/lgtm

PR #754, opened 2022-01-25
docs: use new custom rule format in worker config reference

marquiz:
Also, update the reference linking to the documentation of the custom
source to point to the customization guide.


ArangoGutierrez (approved):
/lgtm

PR #755, opened 2022-01-25
scripts/prepare-release: fix upating of readme

marquiz:
Fix the sed script to update the version of kustomize overlay.

Also, add a sed script to update the documentation link to point to the
correct version in gh-pages.

### DIFF
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -93,7 +93,9 @@ if [ -z "$assets_only" ]; then
 
     # Patch README
     echo Patching README.md to refer to $release
-    sed s"!node-feature-discovery/v.*/!node-feature-discovery/$release/!" -i README.md
+    sed -e s"!\(node-feature-discovery/deployment/.*\)=v.*!\1=$release!" \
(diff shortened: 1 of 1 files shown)


scripts/prepare-release.sh
```diff
@@ -93,7 +93,9 @@ if [ -z "$assets_only" ]; then
 
     # Patch README
     echo Patching README.md to refer to $release
-    sed s"!node-feature-discovery/v.*/!node-feature-discovery/$release/!" -i README.md
+    sed -e s"!\(node-feature-discovery/deployment/.*\)=v.*!\1=$release!" -i README.md \
+        -e s"!^\[documentation\]:.*![documentation]: https://kubernetes-sigs.github.io/node-feature-discovery/$docs_version!" \
+        -i README.md
```
ArangoGutierrez, line 98: I see README 2 times, is this desired?

ArangoGutierrez (changes requested):

scripts/prepare-release.sh
marquiz: Definitely not, good catch, thanks. Fixed

ArangoGutierrez (approved):
/lgtm

PR #783, opened 2022-03-15
topologyupdater: Prevent crash with incorrect node id

tkatila:
It's possible for device plugins to advertise non-existent
numa node ids that cause topology updater to crash.

Signed-off-by: Tuomas Katila <tuomas.katila@intel.com>

### DIFF
--- a/pkg/resourcemonitor/noderesourcesaggregator.go
+++ b/pkg/resourcemonitor/noderesourcesaggregator.go
@@ -246,6 +246,11 @@ func (noderesourceData *nodeResources) updateAvailable(numaData map[int]map[v1.R
 			klog.Infof("unknown resource %q: %q", resName, resID)
 			continue
 		}
+		if _, ok := numaData[nodeID]; !ok {
+			klog.Infof("unknown node id: %q", nodeID)
+			continue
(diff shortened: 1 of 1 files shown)


ArangoGutierrez:
Thanks for the patch @tkatila , could you post an issue reporting thing as a bug, and some example of when this can crash.

fromanirh:
Thanks @tkatila . The change per se looks correct and yes, we can be more defensive in this area. I just wonder if you seen an actual crash with this root cause, and if so I'd love to learn more details, or this is "just" defensive (and welcome) programming.

tkatila:
I noticed this while trying to artificially create a testing node with multiple numa nodes while the node actually only has one. To (try to) accomplish this, I modified our device plugin to split devices evenly on two nodes which then caused the topology updater to crash.

I haven't seen this in the wild without my device plugin changes.

marquiz (commented):
Probably better to not crash even if the input data does not make sense. 

/ok-to-test

tkatila:
I created an issue: #791

ArangoGutierrez (approved):
/lgtm

PR #787, opened 2022-03-17
apis/nfd: empty match expression set returns no features for templates

marquiz:
This patch changes a rare corner case of custom label rules with an
empty set of matchexpressions. The patch removes a special case where an
empty match expression set matched everything and returned all feature
elements for templates to consume. With this patch the match expression
set logically evaluates all expressions in the set and returns all
matches - if there are no expressions there are no matches and no
matched features are returned. However, the overall match result
(determining if "non-template" labels will be created) in this special
case will be "true" as before as none of the zero match expressions
failed.

The former behavior was somewhat illogical and counterintuitive: having
1 to N expressions matched and returned 1 to N features (at most), but,
having 0 expressions always matched everything and returned all
features. This was some leftover proof-of-concept functionality (for
some possible future extensions) that should have been removed before
merging.

### DIFF
--- a/docs/advanced/customization-guide.md
+++ b/docs/advanced/customization-guide.md
@@ -420,10 +420,6 @@ element whose name matches the `<key>`. However, for *instance* features all
 MatchExpressions are evaluated against the attributes of each instance
 separately.
 
-A special case of an empty `matchExpressions` field matches everything, i.e.
-matches/returns all elements of the feature. This makes it possible to write
-[templates](#templating) that run over all discovered features.
-
 #### MatchAny
 
 The `.matchAny` field is a list of of [`matchFeatures`](#matchfeatures)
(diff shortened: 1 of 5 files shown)


marquiz:
Still thinking about some special cases of this...
/hold

marquiz:
Updated, now I think it is better
/unhold
/retitle apis/nfd: empty match expression set returns no features for templates

ArangoGutierrez:
tested ✅ 
```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

images:
- name: '*'
  newName: quay.io/eduardoarango/node-feature-discovery
  newTag: v0.11.0-devel-69-g36341bf4

resources:
- deployment/overlays/default
```

ArangoGutierrez (approved):
lgtm

PR #788, opened 2022-03-17
apis/nfd: add matchName field in feature matcher terms

marquiz:
## Extend NodeFeatureRule API

Extend the format of feature matcher terms (the elements of the
arrayspecified under under matchFeatures field) with new matchName
field. The value of this field is an expression that is evaluated
against the names of feature elements instead of their values (values
are matched with the matchExpressions field, instead).

The matchName field is useful e.g. in template rules for creating
per-feature-element labels based on feature names (instead of values)
and in non-template rules for checking if (at least) one of certain
feature element names are present.

If both matchExpressions and matchName for certain feature matcher term
is specified, they both must match in order to get an overall match.
Also, in this case the list of matched features (used in templating) is
the union of the results from matchExpressions and matchName.

An example of creating an "avx512" label if any AVX512* CPUID feature is
present:

```
  - name: "avx wildcard rule"
    labels:
        avx512: "true"
    matchFeatures:
      - feature: cpu.cpuid
        matchName: {op: InRegexp, value: ["^AVX512"]}
```

An example of a template rule creating a dynamic set of labels  based on
the existence of certain kconfig options.

```
  - name: "kconfig template rule"
    labelsTemplate: |
      {{ range .kernel.config }}kconfig-{{ .Name }}={{ .Value }}
      {{ end }}
    matchFeatures:
      - feature: kernel.config
        matchName: {op: In, value: ["SWAP", "X86", "ARM"]}
```

## Sample rules

This PR contains also another patch that adds sample NodeFeatureRules that correspond the built-in
labels created by nfd-worker (with its default configuration).

The samples provide examples on how to utilize NodeFeatureRules and an
easy way to modify the labels being generated. In order to replace the
built-in node labeling of nfd-worker with these sample rules, all node
labeling from nfd-worker must be disabled by setting the
`core.labelSources` configuration option to an empty list (or specify
`-label-sources= ` on the command line). Then, with all nfd-worker side
labeling disabled, just apply the rules with

```sh
kubectl apply -f samples/
```

### DIFF
--- a/samples/nodefeaturerule-cpu.yaml
+++ b/samples/nodefeaturerule-cpu.yaml
@@ -1,0 +1,112 @@
+#
+# This NodeFeatureRule replicates all built-in cpu feature labels of NFD.
+#
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: nfd-builtin-cpu-features
+spec:
+  rules:
+    - name: "nfd built-in cpu-cpuid labels"
+      labelsTemplate: |
+        {{ range .cpu.cpuid }}cpu-cpuid.{{ .Name }}=true
+        {{ end }}
+      matchFeatures:
+        - feature: cpu.cpuid
+          matchName:
+            op: NotIn
+            value:
+              - "BMI1"
+              - "BMI2"
+              - "CLMUL"
+              - "CMOV"
+              - "CX16"
+              - "ERMS"
+              - "F16C"
+              - "HTT"
+              - "LZCNT"
+              - "MMX"
+              - "MMXEXT"
+              - "NX"
+              - "POPCNT"
+              - "RDRAND"
+              - "RDSEED"
+              - "RDTSCP"
+              - "SGX"
+              - "SGXLC"
+              - "SSE"
+              - "SSE2"
+              - "SSE3"
+              - "SSE4"
+              - "SSE42"
+              - "SSSE3"
+
+    - name: "nfd built-in cpu-hardware_multithreading label"
+      labelsTemplate: |
+        {{ range .cpu.topology }}cpu-{{ .Name }}={{ .Value }}
+        {{ end }}
+      matchFeatures:
+        - feature: cpu.topology
+          matchName:
+            op: In
+            value:
+              - "hardware_multithreading"
+
+    - name: "nfd built-in cpu-cstate and cpu-pstate labels"
+      labelsTemplate: |
+        {{ range .cpu.cstate }}cpu-cstate.{{ .Name }}={{ .Value }}
+        {{ end }}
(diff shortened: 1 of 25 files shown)


ArangoGutierrez:
still a hold?

marquiz:
> still a hold?

Heh, I'm still a bit unsure do we need this, if anybody wants this 😄 OTOH, it would make it possible to turn all "built-in" labels into `NodeFeatureRule`s 🧐 

In addition, docs are still missing.

marquiz:
Rebased

marquiz:
Rebased. Still RFC/POC

marquiz:
Still relevant
/remove-lifecycle stale

ArangoGutierrez:
Still POC @marquiz ?

marquiz:
> Still POC @marquiz ?

Yeah, I'm still a bit unsure about the naming, for example. I have one potential use case in mind but will keep this in hold for now...

marquiz:
I added a patch that adds sample NodeFeatureRules that correspond the built-in labels created by nfd-worker (with its default configuration).

These samples demonstrate how to utilize NodeFeatureRules. Maybe we could replace the whole nfd-worker side label generation with NodeFeatureRules at some point in the future...

ping @ArangoGutierrez

marquiz:
Update:
- rebased
- fixed a bug where the matchNames never failed
- added e2e tests

I think this would be ready for review, no reason to keep it hanging open forever. At least should not have any effect (cause problems) if not used 😉 
/unhold

marquiz:
Ach, docs is missing 🙈 (plus unit tests fail)
/hold

marquiz:
Rebased. Added documentation.
/unhold

samples/nodefeaturerule-cpu.yaml
```diff
@@ -0,0 +1,112 @@
+#
```
ArangoGutierrez, line 1: with this Samples, should we remove `examples` ?

marquiz: It is a good question where we should put these 🤔 The `examples/` are really supposed to be examples but these are "fully usable" samples that replicate the built-in labels. Maybe these should be put somewhere under `deployment/` instead, e.g. `deployment/nodefeaturerule/`. Any thoughts?

ArangoGutierrez: I like `deployment/nodefeaturerule/samples`

ArangoGutierrez (approved):
v0.15 here we gooo! 
/lgtm

samples/nodefeaturerule-cpu.yaml
marquiz: Let's do that, then. Better I think

marquiz:
Too slow 😿

marquiz:
OK, see #1504

PR #790, opened 2022-03-18
cpu: Discover IBM Secure Execution

Jakob-Naucke:
Set `cpu.se-enabled` to `true` when IBM Secure Execution for Linux (IBM Z & LinuxONE) is available and has been enabled.

Uses `/sys/firmware/uv/prot_virt_host`, which is available in kernels >=5.12 + backports. For simplicity, skip more complicated facility & kernel cmdline lookups.

### DIFF
--- a/source/cpu/se_s390x.go
+++ b/source/cpu/se_s390x.go
@@ -1,0 +1,40 @@
+//go:build s390x
+// +build s390x
+
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cpu
+
+import (
+	"os"
+
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+func discoverSE() map[string]string {
+	se := make(map[string]string)
+	// This file is available in kernels >=5.12 + backports. Skip specifically
+	// checking facilities and kernel command lines and just assume Secure
+	// Execution to be unavailable or disabled if the file is not present.
+	protVirtHost := source.SysfsDir.Path("firmware/uv/prot_virt_host")
+	if content, err := os.ReadFile(protVirtHost); err == nil {
+		if string(content) == "1\n" {
+			se["enabled"] = "true"
+		}
+	}
+	return se
+}
--- a/docs/advanced/customization-guide.md
+++ b/docs/advanced/customization-guide.md
@@ -492,6 +492,8 @@ The following features are available for matching:
 |                  |              | **`enabled`** | bool  | `true` if Intel SGX has been enabled, otherwise does not exist
 | **`cpu.sst`**    | attribute    |          |            | Intel SST (Speed Select Technology) capabilities
 |                  |              | **`bf.enabled`** | bool | `true` if Intel SST-BF (Intel Speed Select Technology - Base frequency) has been enabled, otherwise does not exist
+| **`cpu.se`**     | attribute    |          |            | IBM Secure Execution for Linux (IBM Z & LinuxONE)
+|                  |              | **`enabled`** | bool  | `true` if IBM Secure Execution for Linux is available and has been enabled, otherwise does not exist
 | **`cpu.topology`** | attribute  |          |            | CPU topology related features
 | | |          **`hardware_multithreading`** | bool       | Hardware multithreading, such as Intel HTT, is enabled
 | **`kernel.config`** | attribute |          |            | Kernel configuration options
--- a/docs/get-started/features.md
+++ b/docs/get-started/features.md
@@ -51,6 +51,7 @@ such as restricting discovered features with the -label-whitelist option.*
 | **`cpu-cstate.enabled`**          | bool   | Set to 'true' if cstates are set in the intel_idle driver, otherwise set to 'false'. Unset if intel_idle cpuidle driver is not active.
 | **`cpu-rdt.<rdt-flag>`**          | true   | [Intel RDT][intel-rdt] capability is supported. See [RDT flags](#intel-rdt-flags) for details.
 | **`cpu-sgx.enabled`**             | true   | Set to 'true' if Intel SGX is enabled in BIOS (based a non-zero sum value of SGX EPC section sizes).
+| **`cpu-se.enabled`**              | true   | Set to 'true' if IBM Secure Execution for Linux (IBM Z & LinuxONE) is available and enabled (requires `/sys/firmware/uv/prot_virt_host` facility)
 | **`cpu-model.vendor_id`**         | string | Comparable CPU vendor ID.
 | **`cpu-model.family`**            | int    | CPU family.
 | **`cpu-model.id`**                | int    | CPU model number.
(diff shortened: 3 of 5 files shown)


Jakob-Naucke:
I am in IBM's CNCF CLA group, but I'm currently having trouble selecting the appropriate organisation in EasyCLA. I'll get back to you.

zvonkok:
@Jakob-Naucke If we run NFD inside a guest lets say Kube on OpenStack, it might be also helpful to expose `prot_virt_guest`, is this something that could be read also from CPU flags? 

@marquiz klauspost/cpuid has limited CPU arch support, if we agree to use it for all things related to CPU feature detection we might need to add some PRs there as well, besides arm64 and x86_64 detection.

zvonkok:
@marquiz @Jakob-Naucke may be worthwhile to think about to package all the detection for confidential computing features into a separate source: AMD-SEV-{ES, SNP, ...}, Intel TDX, SGX, ARM CCA, IBM SE (Power, s390x)

Jakob-Naucke:
woah @zvonkok that's a bunch of cans of worms :)

For context: I'm coming from the [Confidential Containers](https://github.com/confidential-containers) project, we primarily utilise [Kata Containers](https://github.com/kata-containers) (where I'm also active) to run pods as minimal TEE-enabled VMs (AMD SEV, IBM Secure Execution, Intel TDX). NFD could provide useful information for scheduling in heterogeneous clusters there.

> If we run NFD inside a guest lets say Kube on OpenStack, it might be also helpful to expose `prot_virt_guest`

Sure. Philosophically, I'm somewhat hesitant to call this a CPU flag though -- where do you think does this belong?

> is this something that could be read also from CPU flags?

We're looking into making this information consumable through CPU flags/HW caps -- the current sysfs is already an upgrade from facility bits & kernel command line parsing. But since it's some time in the future to actually land in distros, I'd like to suggest the approach presented here for now.

> may be worthwhile to think about to package all the detection for confidential computing features into a separate source: AMD-SEV-{ES, SNP, ...}, Intel TDX, SGX, ARM CCA, IBM SE (Power, s390x)

In principle, yes. Do you think that we should have a generic field for that? I didn't do this for this PR because it would seem like this flag had support for all major TEEs when it hadn't. Let me loop in @bpradipt, also active in Confidential Containers, here -- and if it is required, I could loop in some more people from manufacturers to help out (although I can't recruit them obviously :)) In Kata, we're doing a [bunch](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/hypervisor_linux_amd64.go) of [such](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/hypervisor_ppc64le.go) checks [already](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/hypervisor_s390x.go) (that isn't consumable for scheduling though).

zvonkok:
> woah @zvonkok that's a bunch of cans of worms :)
> 
> For context: I'm coming from the [Confidential Containers](https://github.com/confidential-containers) project, we primarily utilise [Kata Containers](https://github.com/kata-containers) (where I'm also active) to run pods as minimal TEE-enabled VMs (AMD SEV, IBM Secure Execution, Intel TDX). NFD could provide useful information for scheduling in heterogeneous clusters there.

I know that is why I brought it up :), I am leading the effort on enabling GPUs (our next GPU: Hopper will have CC HW support built in) and other accelerators in Kata. I am new to the Kata community but some Issues and PRs are already open. 

Yes, that is my thought as well, if you have AMD and Intel machines you might want to schedule your workload to an SGX | TDX | SEV node. I mean we can start with the usual suspects and extend as needed. 

> > If we run NFD inside a guest lets say Kube on OpenStack, it might be also helpful to expose `prot_virt_guest`
> 
> Sure. Philosophically, I'm somewhat hesitant to call this a CPU flag though -- where do you think does this belong?

Just trying to sort out what we need to expose "any" of those TEEs. 

> 
> > is this something that could be read also from CPU flags?
> 
> We're looking into making this information consumable through CPU flags/HW caps -- the current sysfs is already an upgrade from facility bits & kernel command line parsing. But since it's some time in the future to actually land in distros, I'd like to suggest the approach presented here for now.
> 
> > may be worthwhile to think about to package all the detection for confidential computing features into a separate source: AMD-SEV-{ES, SNP, ...}, Intel TDX, SGX, ARM CCA, IBM SE (Power, s390x)
> 
> In principle, yes. Do you think that we should have a generic field for that? I didn't do this for this PR because it would seem like this flag had support for all major TEEs when it hadn't. Let me loop in @bpradipt, also active in Confidential Containers, here -- and if it is required, I could loop in some more people from manufacturers to help out (although I can't recruit them obviously :)) In Kata, we're doing a [bunch](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/hypervisor_linux_amd64.go) of [such](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/hypervisor_ppc64le.go) checks [already](https://github.com/kata-containers/kata-containers/blob/main/src/runtime/virtcontainers/hypervisor_s390x.go) (that isn't consumable for scheduling though).

Yep, I have seen that, and I am glad you're bringing this to NFD. I was anyway in the process to propose something like this to NFD. I think there are also other parts in Kata that could be exposed via NFD and have it not reimplemented another time in Kata or other sandboxed environments.

marquiz:
Good discussion above. I've also been trying to figure out (silently 😄) what to do with all these scattered pieces of tee/cc information. Going forward, I think there will be more...

As @zvonkok suggested above, maybe a separate feature source would make sense 🧐 I.e. put all trusted computing related stuff under `source/tee/` or smth instead of cramming everything under `source/cpu/`. This also means that the labels would be something like `feature.node.kubernetes.io/tee-sgx.enabled`.  But, I don't have any good vision or strong opinions on this yet.

For now, I think we could just merge this PR as we don't have any better alternatives (and we already have e.g. `cpu-sgx`. 

@mythi do you have any thoughts on this one?

marquiz (approved):
I would be ok to merge this.

I don't have objections to the new functionality. My only concern is the "general worry" that we need to do something about these security/tee features, going forward (before v1.0). But as I don't have any better concrete suggestions/ideas for now, I'm fine with this.

source/cpu/se_s390x.go
```diff
@@ -0,0 +1,38 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cpu
+
+import "os"
+
+// This assumes the `host-sys` mount, which provides access to the host sysfs.
+const protVirtHost = "/host-sys/firmware/uv/prot_virt_host"
```
Ethyling, line 25: Should we use `source.SysfsDir` here?

marquiz: Uh-oh, yes, definitely. Thanks @Ethyling!

marquiz (changes requested):
Per @Ethyling's comment use `source.SysfsDir`

source/cpu/se_s390x.go
Jakob-Naucke: We should. Tested & done.

mythi:
> @marquiz @Jakob-Naucke may be worthwhile to think about to package all the detection for confidential computing features into a separate source: AMD-SEV-{ES, SNP, ...}, Intel TDX, SGX, ARM CCA, IBM SE (Power, s390x)

+1 but not blocking this PR. Another thing to consider is whether we wanted to have some "standard" label(s) (via the new custom rules), e.g., kata could rely on (`feature.node.kubernetes.io/protected-guest.capable`etc.)

marquiz:
> Another thing to consider is whether we wanted to have some "standard" label(s) (via the new custom rules), e.g., kata could rely on (`feature.node.kubernetes.io/protected-guest.capable`etc.)

Yeah, this is what I've been pondering as well, I.e. maintain some "standard" `NodeFeatureRule` object(s) in the NFD repo. Would need some feedback/input from the downstream regarding what kind of labels would be useful 😊

marquiz (approved):
That was a quick fix @Jakob-Naucke 👍

@zvonkok @Ethyling feel free to lgtm if it looks good to you, now

PR #795, opened 2022-03-23
Rename lists

ArangoGutierrez:
Part of #562 


marquiz (commented):
Thanks @ArangoGutierrez for working on this.

My concern is that this changes the user interface. Thus, I think we should first deprecate the undesired options, introduce new parallel options/aliases, and then remove the deprecated ones after two releases or so.

This might be good time to ditch at least `worker.core.labelWhiteList` for good. I doubt that it is really being used by anybody.

PR #800, opened 2022-04-11
scripts/update-gh-pages: adjust commit message body

marquiz:
Align the version number inserted in the commit message with what we use
in the Makefile.


ArangoGutierrez (approved):
/lgtm

PR #822, opened 2022-06-07
github: small fix in new-release issue template

marquiz:


ArangoGutierrez (changes requested):
we will hopefully soon talk about 1.x, so I am not sold into hardcoding it like 0.major right?

marquiz:
`0.$MAJ` is used elsewhere in this template so I'd suggest to use it for now. Let's update the template if we ever get to 1.0 😎 Looking at the pace of improvements/contributions of the past few months we're not reaching 1.0 anytime soon...

ArangoGutierrez (approved):
/lgtm

PR #827, opened 2022-06-14
Drop the iommu source

### LINKED ISSUE #676: Drop the iommu feature source

I'd like to drop the iommu feature source. It has only one feature `enabled` that I doubt nobody uses. This could be replaced by per pci device attribute(s).

Plan:
- v0.10.0: mark as deprecated, disable by default
- drop after two releases (in v0.12.0)

If we get feedback that people are using this we need to re-evaluate the plan

**Why is this needed**:

Simplifying the code, dropping unused functionality.

marquiz:
It was deprecated in v0.10.0.

Fixes #676


e0ne (approved):
LGTM

marquiz:
Thanks @e0ne!
/assign @zvonkok @ArangoGutierrez

PR #829, opened 2022-06-15
Containerized auto-generation

marquiz:
Run code auto-generation inside a container instead of the host system.
Our auto-generation depends on specific versions of a multitude of tools
(like k8s code-generator, controller-gen, protoc, mockery etc). This
made it really awkward (and error-prone) to run in the host environment,
especially if/when you needed different versions of those tools for
other projects. Making it even more unwieldy, the required versions of
tools were not neatly documented anywhere (except for git commits,
perhaps).

With this patch we have a "fixed environment", as we build a special
auto-generate-builder container which has correct versions of all the
dependencies. Using the container makes auto-generation easy to run
anywhere, independent of the host system, giving reproducibility and
reliability. Also, the patch moves the auto-generation steps out from
the makefile into a separate script, making the makefile cleaner and the
script easier to maintain.


marquiz:
ping @mythi, would you have time to look at this? 😊 

A lot of files changed but basically only the first patch is interesting

ArangoGutierrez:
And what is the use case of this? maybe add some context to the PR description, about the auto-generator tool.

marquiz:
> And what is the use case of this? maybe add some context to the PR description, about the auto-generator tool.

Oh, I thought it was so obvious 😅 I added a longer commit message (copy-pasted to the PR description). I hope it is easier to understand now.

ArangoGutierrez (approved):
/lgtm

Issue #832, opened 2022-06-27
Re-organize security related features


marquiz:
**What would you like to be added**:

We start to have a lot of somewhat cluttered features/labels related to security and trusted-execution-environments (e.g. #647, #790, #830) and expecting more to come. I feel it would be good to group all of these more closely together, e.g. under a separate domain (say `security`) or under a single feature (say `cpu.security`).

From the users point of view this would mean renaming some of the existing labels but I think we could manage that by deprecation over a few releases.

With a separate `security` feature source (or domain) we could have e.g.

| Old label | New label
| ---------- | ------------
| cpu-sgx.enabled | security-cpu.sgx.enabled
| cpu-se.enabled | security-cpu.se.enabled
| kernel-selinux.enabled | security-kernel.selinux.enabled

With just a `security` feature under the `cpu` source we could have something like:

| Old label | New label
| ---------- | ------------
| cpu-sgx.enabled | cpu-security.sgx.enabled
| cpu-se.enabled | cpu-security.se.enabled

**Why is this needed**:

Better organization in terms of code and (hopefully) user interface and documentation.

fidencio:
@marquiz, on the kernel side what would be part of the new `security` feature? SELinux and AppArmor?

Maybe I'm not actually seeing the bigger picture here, but it seems that with a `security` feature under `cpu` things would be cleaner.  But then I also see this growing to having a `security` feature under each component that it may apply, which could only be `cpu` and `kernel`?

marquiz:
> on the kernel side what would be part of the new `security` feature? SELinux and AppArmor?

Currently we only have selinux but I think apparmor would be obvious possible future addition.
 
> Maybe I'm not actually seeing the bigger picture here

Buaha, I think I'm a bit lost here too 😸 

but it seems that with a `security` feature under `cpu` things would be cleaner. But then I also see this growing to having a `security` feature under each component that it may apply, which could only be `cpu` and `kernel`?

Yup, now after sleeping over it I'm inclined to agree with you. It would seem cleaner and more natural to have `security` feature under each "domain".  I think I'll draft a PR for comments.

mythi:
> I dropped a quick PR for comments #833 re-organizing the CPU security features. This version does not introduce a new feature source but consolidates all cpu security features under `cpu-security.*`.
> 
> ping @zvonkok @Jakob-Naucke @fidencio @mythi PTAL

[...]

PR #833, opened 2022-06-28
cpu: re-organize security features

marquiz:
Move existing security/trusted-execution related features (i.e. SGX and
SE) under the same `security` feature, deprecating the old features. The
motivation for the change is to keep the source code and user interface
more organized as we experience a constant inflow of similar security
related features. This change will affect the user interface so it is
less painful to do it early on.

New feature labels will be:

```
  feature.node.kubernetes.io/cpu-security.se.enabled
  feature.node.kubernetes.io/cpu-security.sgx.enabled
```

and correspondingly new  **`cpu.security`** feature with `se.enabled` and
`sgx.enabled` elements will be available for custom rules, for example:

```
      - name: "sample sgx rule"
        labels:
          sgx.sample.feature: "true"
        matchFeatures:
          - feature: cpu.security
            matchExpressions:
              "sgx.enabled": {op: IsTrue}
```

At the same time deprecate old labels `cpu-sgx.enabled` and
`cpu-se.enabled` feature labels and the corresponding features for
custom rules. These will be removed in the future causing an effective
change in NFDs user interface.


marquiz:
RFC
/hold

zvonkok:
@marquiz Where are we with the topic of having shorter feature labels? I think this is also still open. I think we should start thinking of doing that as well as soon as possible before we are adding features that are used by other projects and it creates to much confusion when switching.

marquiz:
Any thoughts on this? @mythi @zvonkok @ArangoGutierrez ?

marquiz:
> I added my thoughts in [#832 (comment)](https://github.com/kubernetes-sigs/node-feature-discovery/issues/832#issuecomment-1191202692) earlier

Ach yes, sorry I already forgot that comment 😊 Suggestions (or PRs) how to change the feature descriptions in docs are welcome.

Removing the RFC status of this PR
/unhold

ArangoGutierrez (approved):
/lgtm

PR #834, opened 2022-06-28
scripts/test-e2e: update aws-iam-authenticator to v0.5.7

marquiz:
We need the latest version to support client.authentication.k8s.io/v1.
The previous version we used only supported v1alpha1 which was dropped
in K8s v1.24.


ArangoGutierrez (approved):
/lgtm

PR #839, opened 2022-06-29
test/e2e: fix checking of nfd-master annotation

marquiz:
The test was broken twofold: Firstly, the annotation was not checked at
all because the name of the node where nfd-master is running was not
set. Secondly, the annotation prefix was used incorrectly.

### DIFF
--- a/test/e2e/node_feature_discovery.go
+++ b/test/e2e/node_feature_discovery.go
@@ -506,6 +506,15 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
 			By("Waiting for the nfd-master pod to be running")
 			Expect(e2epod.WaitTimeoutForPodRunningInNamespace(f.ClientSet, masterPod.Name, masterPod.Namespace, time.Minute)).NotTo(HaveOccurred())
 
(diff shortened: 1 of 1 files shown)


fromanirh (commented):
LGTM - makes sense to me and matches my admittedly limited experience.
In general, this seems more careful testing so it seems like a good improvement.

test/e2e/node_feature_discovery.go
```diff
@@ -506,6 +506,15 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
 			By("Waiting for the nfd-master pod to be running")
 			Expect(e2epod.WaitTimeoutForPodRunningInNamespace(f.ClientSet, masterPod.Name, masterPod.Namespace, time.Minute)).NotTo(HaveOccurred())
 
+			By("Verifying node where nfd-master is running")
```
ArangoGutierrez, line 509: Nit
```suggestion
			By("Verifying the node where nfd-master is running")
```

ArangoGutierrez (approved):
/lgtm

PR #845, opened 2022-07-07
Revert type hack in api

marquiz:
- Revert the hack that was a workaround for issues with k8s deepcopy-gen.
  New deepcopy-gen is able to generate code correctly without issues so
  this is not needed anymore.

  Also, removing this hack solves issues with object validation when
  creating NodeFeatureRules programmatically with nfd go-client. This is
  needed later with NodeFeatureRules e2e-tests.
- Update controller-gen and k8s code-generator to the latest releases.
- Refresh auto-generated files

Depends on #829


marquiz:
#829 was merged ->
/unhold

ArangoGutierrez (approved):
/lgtm

PR #846, opened 2022-07-08
Improvements to scripts/prepare-release.sh

marquiz:
Improve the release preparation script. The main motivation is to make the golang version fixed for release builds.

- option `-b` to not generate assets
  Implement a new command line option `-b` to prevent the script from
  creating release assets (Helm chart).
- option `-k` to specify the signing key
  Make the GPG signing an optional argument. It doesn't make much sense to
  require that when no assets are being generated (i.e. when `-b` is
  used).
- patch golang version in Makefile
  Make it mandatory to specify a golang version with `-g` when running the
  script (and patching repo content). The given version will be patched
  into the Makefile to fix the version used in the release build - it is
  meant to use a exact point release (like 1.18.3) to get reproducible
  builds.

Depends on #836

### DIFF
--- a/scripts/prepare-release.sh
+++ b/scripts/prepare-release.sh
@@ -5,12 +5,14 @@ this=`basename $0`
 
 usage () {
 cat << EOF
-Usage: $this [-h] [-a] [-b] [-k GPG_KEY] RELEASE_VERSION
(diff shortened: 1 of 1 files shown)


marquiz:
Depends on #836 

/hold

marquiz:
#836 was merged

/unhold
/assign @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

Issue #540, opened 2021-06-10
feature to allow optionally setting taints based on node properties


rptaylor:
**What would you like to be added**:

It would be nice if NFD could be configured with options to set node taints as well as labels, based on certain properties of nodes. Would you consider that in scope of NFD? 

**Why is this needed**:
Cluster operators may wish to automatically taint nodes with certain properties, for example tainting a node that has GPUs to prevent other pods from running on it if they don't actually need (tolerate) the GPUs.

marquiz:
Hi @rptaylor. Yes, this would be useful and I've been thinking this myself as part of the work I've done on #464 and particularly #468 which are very much on prototype level, still.

I've done some initial experiments and started to think whether it should be possible to taint only some of the nodes (a configurable proportion). WDYT? This complicates things (in implementation) quite a bit, though so maybe that would be a future enhancement.

rptaylor:
Okay, nice. It makes sense to me that NFD could have the flexibility and generalization to apply arbitrary properties (taints as well as labels) based on the features of nodes.

What would be the use case to only taint a portion of nodes with a given feature and configuration?

marquiz:
> What would be the use case to only taint a portion of nodes with a given feature and configuration?

Reserving some of the nodes for general usage or alternatively reserving only  a fraction of the nodes for special workloads. Dunno if that is useful in practice 🤔

zvonkok:
To taint only a subset of nodes in a cluster makes kinda sense for extended resources. If a Pod is allocating an extended resource, the ExtendedResourceAdmissionWebhook will automatically tolerate the extended resource taint. 

But we need to be careful when we're applying this taint. In some cases of hardware enablement usually adding the taint is when the extended resources are exposed. 

It also depends on how you want to partition your cluster. We used taints and tolerations for a "hard-partitioning" meaning no workloads allowed that are not tolerating the taint, repelling workloads.

Or using "soft-partitioning" e.g. with priority classes to have mixed workloads but special workloads could have higher priorities. 

Another use-case would be behavioural partitioning, let's say e.g. you have one cluster and want to do some AI/ML pipeline, one could imaging to taint some nodes as `inference` and others as `training` or `data-lake` resembling a pipeline in one cluster rather then having several clusters each for "one" specific feature.

rptaylor:
[...]

PR #848, opened 2022-07-08
test/e2e: add tests for NodeFeatureRules

marquiz:
Add tests covering the basic functionality of NodeFeatureRule objects,
covering different feature types ("flag features", "attribute features"
and "instance features") as well as backreferencing (using the output of
previously run rules) and templating. The test relies on the "fake"
feature source and its default configuration.

Depends on #845


marquiz:
Depends on #845

/hold

marquiz:
#845 was merged
/unhold

marquiz:
Need to re-verify this
/hold

marquiz:
Tests passed on manual run
/unhold

marquiz:
ping @ArangoGutierrez @zvonkok 

I think we should do smth about this 😉 Especially with #540 and #863 in the works we'd need to have e2e tests to cover them. At least some baseline on top of which to implement new tests

marquiz:
ping @ArangoGutierrez @zvonkok

test/e2e/node_feature_discovery.go
```diff
@@ -419,6 +423,103 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
...
+				crd, err = testutils.CreateNodeFeatureRulesCRD(extClient)
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			AfterEach(func() {
+				err := extClient.ApiextensionsV1().CustomResourceDefinitions().Delete(context.TODO(), crd.Name, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("custom labels should create labels should be created", func() {
```
fmuyassarov, line 451: nit, there seems to be some typo

test/e2e/node_feature_discovery.go
```diff
@@ -419,6 +423,103 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
...
+		return nil
+	}
+
+	// Simple and stupid re-try loop
+	var err error
+	for retry := 0; retry < 3; retry++ {
+		if err = poll(); err == nil {
+			return nil
+		}
+		time.Sleep(2 * time.Second)
```
fmuyassarov, line 509: If we are using default `sleepInterval` of 60 seconds, isn't the sleep interval of 2 seconds too short for the retry?

fmuyassarov (commented):
This is a great addition, especially for the https://github.com/kubernetes-sigs/node-feature-discovery/pull/910. 
LGTM

test/e2e/node_feature_discovery.go
marquiz: Thanks, fixed

test/e2e/node_feature_discovery.go
marquiz: The e2e tests (and this polling) is in no sync with the worker pods. There can basically be any random timeshift (deployment delays, image pulling etc) so we have have no idea when the labeling really happens. It might be just 1 second after we just polled and we don't want to wait for e.g. 60 seconds. I just picked some small number because we basically just want to see the desired state as fast as possible and continue the test sequence. Makes sense?

marquiz:
Rebased in order to incorporate the latest refactoring

test/e2e/node_feature_discovery.go
fmuyassarov: Thanks for the explanation. This makes sense indeed.

fmuyassarov:
LGTM

marquiz:
ping @zvonkok prow didn't understand the extra parameter to /lgtm 😸

PR #861, opened 2022-08-17
nfd-master: more fixes to log messages

marquiz:
Use correct name for the CR (NodeFeatureRule) object. Also, the resource
is cluster-scoped so don't print the namespace.


ArangoGutierrez (approved):
/lgtm

PR #865, opened 2022-08-21
Fix templates for NodeFeatureRule with MatchAny

### LINKED ISSUE #864: Templates don't work on NodeFeatureRule with MatchAny only
**What happened**:
I've stumbled into it while sketching a PR for adding annotations with nfd, so I'm opening this issue just for tracking purposes.

Run tests from submitted PR without the fix in submitted PR.

stek29:
When `MatchAny` is matched, template existence is checked with `r.labelsTemplate == nil`, and if it's nil - templating is skipped.

This is wrong because:
- there are `vars` templates too, not only `labels` templates
- `labelsTemplate` is initialized from `LabelsTemplate` when `executeLabelsTemplate` is called first time, so it would always be nil unless Rule has `MatchFeatures` in addition to `MatchAny`, and was executed already.

`LabelsTemplate` should be checked instead of `labelsTemplate` (same for vars) to see if there's template – but `executeLabelsTemplate` already does that.

So the check should be removed completely.

I've also added a test case which catches this issue.

Fixes #864.

### DIFF
--- a/pkg/apis/nfd/v1alpha1/rule.go
+++ b/pkg/apis/nfd/v1alpha1/rule.go
@@ -50,11 +50,13 @@ func (r *Rule) Execute(features feature.Features) (RuleOutput, error) {
 				matched = true
 				utils.KlogDump(4, "matches for matchAny "+r.Name, "  ", matches)
 
-				if r.labelsTemplate == nil {
-					// No templating so we stop here (further matches would just
-					// produce the same labels)
+				if r.LabelsTemplate == "" && r.VarsTemplate == "" {
+					// there's no need to evaluate other matchers in MatchAny
+					// if there are no templates to be executed on them - so
+					// short-circuit and stop on first match here
 					break
 				}
+
 				if err := r.executeLabelsTemplate(matches, labels); err != nil {
 					return RuleOutput{}, err
 				}
--- a/pkg/apis/nfd/v1alpha1/rule_test.go
+++ b/pkg/apis/nfd/v1alpha1/rule_test.go
@@ -297,6 +297,11 @@ var-2=
 		},
 	}
 
+	// test with empty MatchFeatures, but with MatchAny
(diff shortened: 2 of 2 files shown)


pkg/apis/nfd/v1alpha1/rule.go
```diff
@@ -50,11 +50,6 @@ func (r *Rule) Execute(features feature.Features) (RuleOutput, error) {
 				matched = true
 				utils.KlogDump(4, "matches for matchAny "+r.Name, "  ", matches)
 
-				if r.labelsTemplate == nil {
-					// No templating so we stop here (further matches would just
-					// produce the same labels)
-					break
-				}
```
marquiz, line 57: It is still useful to check if the templates have been defined. If there would be 10 items under `MatchAny` and the first matches (and no templates are specified) there is no use to evaluate the 9 remaining items.

My suggestion to change/fix this instead

```
                                if r.LabelsTemplate == "" && r.VarsTemplate == "" {
                                        // No templating so we stop here (further matches would just
                                        // produce the same labels / vars)
                                        break
                                }
```

marquiz (commented):
Wov, thanks for the patch @stek29! This is indeed broken. And great that you updated the unit tests, too 👍

I had one suggestion, though.

/ok-to-test

marquiz:
...and seems that somebody's really using the templating functionality 😊 

/cherry-pick release-0.11

ArangoGutierrez:
PING @stek29

pkg/apis/nfd/v1alpha1/rule.go
stek29: sure, will do

marquiz (approved):
Looks good to me now 👌

ArangoGutierrez (approved):
/lgtm

PR #866, opened 2022-08-22
Discover Node Features as annotations based on NodeFeatureRules

### LINKED ISSUE #863: Discover node features as annotations
Discover node features as annotations instead of labels

**Why is this needed**:
Some features don't feel like they belong to labels – some extra info about topology or internal info, which won't be used for scheduling or in selection, but is nice to have on Node.
Also, there are limits on label contents in contrast to annotations.

stek29:
This PR fixes #863, it's still WIP, but is ready for review.

Design:
- NFR now has two more properties: annotations and annotationsTemplate, which allow setting Node Features as annotations
- Default NS for such annotations is same as labels – `feature.node.kubernetes.io`, and more can be allowed with `extra-annotation-ns`
- Unlike labels, `profile.node.kubernetes.io` is not allowed by default
- Unlike labels, there's a "restricted" ns – feature annotations can't be in `nfd.node.kubernetes.io`, to avoid conflicts with nfd's internal annotations
- These annotations are tracked in new annotation - `feature-annotations`, just like `feature-labels` or `extended-resources`
- Unlike labels, Annotations can only be set with NodeFeatureRule, and aren't supported on worker side

Docs are yet to be updated, and e2e tests don't cover this new feature at all – partially because there are no e2e tests with NodeFeatureRules in them

It's also written on top of #865, to avoid having to fix conflicts later on, since I'm assuming #865 would be merged before this PR.

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -85,6 +85,8 @@ func initFlags(flagset *flag.FlagSet) *master.Args {
 		"Certificate used for authenticating connections")
 	flagset.Var(&args.ExtraLabelNs, "extra-label-ns",
 		"Comma separated list of allowed extra label namespaces")
+	flagset.Var(&args.ExtraAnnotationNs, "extra-annotation-ns",
+		"Comma separated list of allowed extra annotaion namespaces")
 	flagset.StringVar(&args.Instance, "instance", "",
 		"Instance name. Used to separate annotation namespaces for multiple parallel deployments.")
 	flagset.StringVar(&args.KeyFile, "key-file", "",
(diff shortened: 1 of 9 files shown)


marquiz:
ping @stek29 are you still planning to work on this?

stek29:
@marquiz I expected it to be reviewed to know if I should continue working on it.

Realistically -- I don't think I'll work on it in close future.

marquiz:
> I expected it to be reviewed to know if I should continue working on it.
> 
> Realistically -- I don't think I'll work on it in close future.

K, thanks for the update. Apologies for the unresponsiveness on this, but it was in draft state so mentally in the "I'll look at it the day after" pile...

I think this is a "nice-to-have" feature in the next release so maybe we (e.g. @fmuyassarov or me) could pick up this if we have time? E.g. take part of your work as a base and finalize it with a *co-authored-by* stamp, wdyt?

fmuyassarov:
> I think this is a "nice-to-have" feature in the next release so maybe we (e.g. @fmuyassarov or me) could pick up this if we have time? E.g. take part of your work as a base and finalize it with a _co-authored-by_ stamp, wdyt?

Happy to help on this.

stek29:
@marquiz sure, feel free to use it, it's signed off anyways and is under Apache-2.0 :)

PR #868, opened 2022-08-22
logging: do not use %w with klog.Errorf

marquiz:
It is not recognized (and does not work like with fmt.Errorf) so use %v
instead.


ArangoGutierrez (approved):
/lgtm

Issue #859, opened 2022-08-16
Config option to disable hooks


marquiz:
**What would you like to be added**:

Add a config file option to disable hooks of the `local` feature source, e.g.
```
sources:
  local:
    hooksEnabled: true  # true by default at this point
``` 

Also, add a disclaimer in docs and logs that hooks are now deprecated and will be disabled in a future release (and eventually removed alltogether).

**Why is this needed**:

First step of #856.

marquiz:
/cc @zvonkok
ping @mythi

fmuyassarov:
I'm interested to work on this issue
/assign

marquiz:
> I'm interested to work on this issue

That would be great! 😊

Issue #856, opened 2022-08-15
Drop support for hooks


marquiz:
**What would you like to be added**:

Deprecate and eventually remove support for hooks in the `local` feature source.

**Why is this needed**:

In #855 @eero-t had a merituos pondering why hooks should removed. Some rationale why hooks are bad and we want to disable them:
- container image maintenance and size
- feature files provide basically the same functionality with a more controllable/maintainable approach
- possible security issues
- resource consumption of buggy or bad-behaving hooks

**Plan**

- n (v0.12): declare hooks as deprecated, introduce config option to disable them but have them still enabled by default (#855)
- n+1: disable hooks by default but have still the option to enable them
- n+2: drop hooks support entirely, make minimal image the default image, drop the big image or e.g. rename it to -debug

eero-t:
> n+2: drop hooks support entirely, make minimal image the default image, drop the big image or e.g. rename it to -debug

I would propose NFD to drop the big image completely, and add an initcontainer that removes host source.d/ subdirectory content, if it exists ("os.removeAll(path)").  All this should of course have prominent warnings in NFD release notes and worker log.

I.e. if one wants to continue using hooks, one would need to restart things installing hooks, in addition to downgrading back to NFD version supporting them.

n+X release would eventually drop the hook removal initcontainer.

@mythi Any comments from the Intel device plugins (which are currently using hooks) maintenance side?

eero-t:
I would propose that:
* #858 is implemented before this (if accepted), as it's an alternative for labels that may need to be updated during NFD run-time
* #857 is implemented at the same time as this, so that features moved from hooks to feature files need to be updated only once

mythi:
> @mythi Any comments from the Intel device plugins (which are currently using hooks) maintenance side?

running the source binaries in our initContainers (instead of using the initContainers to install the binaries) and piping the output to, e.g., `tee /etc/kubernetes/node-feature-discovery/features.d/foo.feature` should work equally well.

ArangoGutierrez:
I have had sooooo many conversations recently, that I can't not support this issue moaaaarrr.
👍🏽 👍🏽

marquiz:
This will still take a few releases to completely drop the hooks, I think

/remove-lifecycle stale

marquiz:
[...]

Issue #855, opened 2022-08-12
Disable hook support by default


eero-t:
**What would you like to be added**:

Disable support for hooks by default, and make minimal image the default. Add NFD worker option to enable support for hooks, for backwards compatibility.  Deprecate and eventually remove support for dynamic binary + script hooks (to prepare for potential removal of also static binaries hook support).

Feature files can support all the same things while avoiding the issues plaguing hooks (listed below).

**Why is this needed**:

There are multiple issues with hook support: https://kubernetes-sigs.github.io/node-feature-discovery/v0.11/advanced/customization-guide.html#hooks

Bash, Perl and dynamic binaries support:
* Security: larger image + attack surface
* Maintenance: need to update these images to handle security issues in them / their dependencies (in addition to Go ones relevant for NFD itself), see e.g: #853
* Maintenance: unintentional(?) support also for dynamic executables which happen to link to same system libraries, that NFD users could start relying on
* Documentation: missing on what exactly is supported in given NFD version, which Bash and Perl version, which Perl modules, and which dependency libraries
  * In theory this could be auto-generated for each release image, and diff generated against previous release

Hook support in general:
* Security: impact of running "random" scripts / binaries in same context as NFD
* Functionality: Whomever provides the hook, not being able to impact what is exposed to it
* Documentation: missing on what is exposed for the hook from the system and network
* Reliability: not being able to tighten NFD worker resource limits as resource usage of hooks is unknown
* Reliability: additional validation needed for verifying that NFD handles + logs misbehaving hooks correctly (e.g. ones generating output forever or using too much of some other resources)
* Conformance: binaries in /etc/ being FHS spec violation: https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html

**External feature life-cycle management**

I think feature life-cycle management would also be easier if support for hooks were disabled/dropped.

Life-cycle issues:
* Documentation: what party is supposed to clean out the obsolete features and how?
* Functionality: what happens if e.g. Pod adds differently named hook binary or feature file on each run, or some test Pods run year ago were obsolete the next day, but their features still persist in /etc/?  NFD continues running them and slowing down with time?
* Documentation: what happens if multiple hooks or feature files update same labels, does NFD complain loudly, or just take the last value it gets?

Potential solutions:
* Feature files could include best-before timestamp to solve that, after which NFD could ignore or remove them.  With source hooks, another file and some mapping between them would be needed for that
[...]

PR #871, opened 2022-08-23
Config option to disable hooks

fmuyassarov:
Fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/859

### DIFF
--- a/docs/advanced/customization-guide.md
+++ b/docs/advanced/customization-guide.md
@@ -153,6 +153,16 @@ should be placed in a separate directory in order to avoid NFD unnecessarily
 trying to execute them. A subdirectory under the hooks directory can be used,
 for example `/etc/kubernetes/node-feature-discovery/source.d/conf/`.
 
+**NOTE:** Hooks are being DEPRECATED and will be removed in a future release.
+For backward compatibility, currently hooks are enabled by default and can be
+disabled via `sources.local.hooksEnabled` field in the worker configuration.
+
+```yaml
+sources:
+  local:
+    hooksEnabled: true  # true by default at this point
+```
+
 **NOTE:** NFD will blindly run any executables placed/mounted in the hooks
 directory. It is the user's responsibility to review the hooks for e.g.
 possible security implications.
--- a/docs/advanced/worker-configuration-reference.md
+++ b/docs/advanced/worker-configuration-reference.md
@@ -329,6 +329,29 @@ sources:
     configOpts: [NO_HZ, X86, DMI]
 ```
 
+### sources.local
+
+### sources.local.hooksEnabled
+
+Configuration option to disable/enable hooks execution. Enabled by default.
+Hooks are DEPRECATED since v0.12.0 release and support will be removed in a
+future release. Use [feature files](./customization-guide.md#feature-files)
+instead.
+
+Related tracking issues:
+
+1. Config option to disable hooks [#859](https://github.com/kubernetes-sigs/node-feature-discovery/issues/859).
+1. Disable hook support by default [#855](https://github.com/kubernetes-sigs/node-feature-discovery/issues/855).
+1. Drop support for hooks [#856](https://github.com/kubernetes-sigs/node-feature-discovery/issues/856).
+
+Example:
+
+```yaml
+sources:
+  local:
+    hooksEnabled: true  # true by default
+```
+
 ### soures.pci
 
 #### soures.pci.deviceClassWhitelist
--- a/source/local/local.go
+++ b/source/local/local.go
@@ -47,17 +47,39 @@ var (
 // localSource implements the FeatureSource and LabelSource interfaces.
 type localSource struct {
 	features *feature.DomainFeatures
+	config   *Config
+}
+
+type Config struct {
+	HooksEnabled bool `json:"hooksEnabled,omitempty"`
 }
 
 // Singleton source instance
 var (
-	src localSource
-	_   source.FeatureSource = &src
-	_   source.LabelSource   = &src
+	src                           = localSource{config: newDefaultConfig()}
+	_   source.FeatureSource      = &src
+	_   source.LabelSource        = &src
+	_   source.ConfigurableSource = &src
 )
 
 // Name method of the LabelSource interface
 func (s *localSource) Name() string { return Name }
+
+// NewConfig method of the LabelSource interface
+func (s *localSource) NewConfig() source.Config { return newDefaultConfig() }
+
+// GetConfig method of the LabelSource interface
+func (s *localSource) GetConfig() source.Config { return s.config }
+
+// SetConfig method of the LabelSource interface
+func (s *localSource) SetConfig(conf source.Config) {
+	switch v := conf.(type) {
+	case *Config:
+		s.config = v
+	default:
+		klog.Fatalf("invalid config type: %T", conf)
+	}
+}
 
 // Priority method of the LabelSource interface
 func (s *localSource) Priority() int { return 20 }
@@ -73,28 +95,41 @@ func (s *localSource) GetLabels() (source.FeatureLabels, error) {
 	return labels, nil
 }
 
+// newDefaultConfig returns a new config with pre-populated defaults
+func newDefaultConfig() *Config {
+	return &Config{
+		HooksEnabled: true,
+	}
+}
+
 // Discover method of the FeatureSource interface
 func (s *localSource) Discover() error {
 	s.features = feature.NewDomainFeatures()
-
-	featuresFromHooks, err := getFeaturesFromHooks()
-	if err != nil {
-		klog.Error(err)
-	}
 
 	featuresFromFiles, err := getFeaturesFromFiles()
 	if err != nil {
 		klog.Error(err)
 	}
 
-	// Merge features from hooks and files
-	for k, v := range featuresFromHooks {
-		if old, ok := featuresFromFiles[k]; ok {
-			klog.Warningf("overriding '%s': value changed from '%s' to '%s'",
-				k, old, v)
+	if s.config.HooksEnabled {
+
+		klog.Info("starting hooks...")
+
+		featuresFromHooks, err := getFeaturesFromHooks()
+		if err != nil {
+			klog.Error(err)
 		}
-		featuresFromFiles[k] = v
+
+		// Merge features from hooks and files
+		for k, v := range featuresFromHooks {
+			if old, ok := featuresFromFiles[k]; ok {
+				klog.Warningf("overriding '%s': value changed from '%s' to '%s'",
+					k, old, v)
+			}
+			featuresFromFiles[k] = v
+		}
 	}
+
 	s.features.Values[LabelFeature] = feature.NewValueFeatures(featuresFromFiles)
 
 	utils.KlogDump(3, "discovered local features:", "  ", s.features)
@@ -133,6 +168,7 @@ func parseFeatures(lines [][]byte) map[string]string {
 
 // Run all hooks and get features
 func getFeaturesFromHooks() (map[string]string, error) {
+
 	features := make(map[string]string)
 
 	files, err := ioutil.ReadDir(hookDir)
@@ -142,6 +178,9 @@ func getFeaturesFromHooks() (map[string]string, error) {
 			return features, nil
 		}
 		return features, fmt.Errorf("unable to access %v: %v", hookDir, err)
+	}
+	if len(files) > 0 {
+		klog.Warning("hooks are DEPRECATED since v0.12.0 and support will be removed in a future release; use feature files instead")
 	}
 
 	for _, file := range files {
--- a/deployment/components/worker-config/nfd-worker.conf.example
+++ b/deployment/components/worker-config/nfd-worker.conf.example
@@ -75,6 +75,8 @@
 #      - "class"
 #      - "vendor"
 #      - "device"
+#  local:
+#    hooksEnabled: true
 #  custom:
 #    # The following feature demonstrates the capabilities of the matchFeatures
 #    - name: "my custom rule"
--- a/deployment/helm/node-feature-discovery/values.yaml
+++ b/deployment/helm/node-feature-discovery/values.yaml
@@ -169,6 +169,8 @@ worker:
     #      - "class"
     #      - "vendor"
     #      - "device"
+    #  local:
+    #    hooksEnabled: true
     #  custom:
     #    # The following feature demonstrates the capabilities of the matchFeatures
     #    - name: "my custom rule"


source/local/local.go
```diff
@@ -133,6 +168,8 @@ func parseFeatures(lines [][]byte) map[string]string {
 
 // Run all hooks and get features
 func getFeaturesFromHooks() (map[string]string, error) {
+	klog.Warning("hooks are DEPRECATED since v0.12.0; use feature files instead")
```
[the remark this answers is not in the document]
marquiz: Yeah, it will be v0.12.0 so it's correct. We can keep it here or drop it whatever you like.

Unrelated: maybe you could supplement this with a mention that they will be dropped
```suggestion
	klog.Warning("hooks are DEPRECATED since v0.12.0 and support will be removed in a future release; use feature files instead")
```

marquiz:
ping @eero-t @mythi

source/local/local.go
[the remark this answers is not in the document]
fmuyassarov: Thanks, updated.

docs/advanced/customization-guide.md
```diff
@@ -153,6 +153,16 @@ should be placed in a separate directory in order to avoid NFD unnecessarily
 trying to execute them. A subdirectory under the hooks directory can be used,
 for example `/etc/kubernetes/node-feature-discovery/source.d/conf/`.
 
+**NOTE:** Hooks are being DEPRECATED and will be removed in the future releases.
```
eero-t, line 156: For consistency with other changes, use "in a future release."

docs/advanced/worker-configuration-reference.md
```diff
@@ -329,6 +329,23 @@ sources:
     configOpts: [NO_HZ, X86, DMI]
 ```
 
+### sources.local
+
+### sources.local.hooksEnabled
+
+Configuration option to disable/enable hooks execution. Enabled by default.
+Hooks are DEPRECATED since v0.12.0 release and support will be removed in a
+future release. Use [feature files](./customization-guide.md#feature-files)
```
eero-t, line 338: This could include links to the relevant tickets, so that interested parties can read the discussion:
* Step 1, add option: https://github.com/kubernetes-sigs/node-feature-discovery/issues/859
* Step 2, disable by default: https://github.com/kubernetes-sigs/node-feature-discovery/issues/855
* Step 3, remove hook support: https://github.com/kubernetes-sigs/node-feature-discovery/issues/856

eero-t (changes requested):

docs/advanced/customization-guide.md
fmuyassarov: done

docs/advanced/worker-configuration-reference.md
fmuyassarov: done

source/local/local.go
```diff
@@ -133,6 +168,8 @@ func parseFeatures(lines [][]byte) map[string]string {
 
 // Run all hooks and get features
 func getFeaturesFromHooks() (map[string]string, error) {
+	klog.Warning("hooks are DEPRECATED since v0.12.0 and support will be removed in a future release; use feature files instead")
```
eero-t, line 171: If this is logged every time hooks are called (which happens periodically), it could be too much.

IMHO it would be better to log the warning only at start, when reading local config, if HooksEnabled is set.

@marquiz?

(Otherwise looks OK to me.)

fmuyassarov: I agree on this point. Initially I added in https://github.com/kubernetes-sigs/node-feature-discovery/blob/3df81d1ace1f74b43a74f998f8b238df6a53ca6a/pkg/nfd-client/worker/nfd-worker.go#L156 but then moved to here. Would https://github.com/kubernetes-sigs/node-feature-discovery/blob/3df81d1ace1f74b43a74f998f8b238df6a53ca6a/pkg/nfd-client/worker/nfd-worker.go#L156 be okay ?

fmuyassarov: it would basically log once, as general warning regardless user enables/disabled the hooks.

eero-t: @marquiz should really be the one commenting on where to do the check...

(I filed the ticket for hooks disabling + few other things, so Markus CC'd me to this. However, this is first time I'm looking at NFD code, so I'm not at all familiar how it works.)

marquiz: Thanks @eero-t for the review.

@fmuyassarov I would probably put the log message later in this function if some hook files are really found and executed 🤔 Better not to print a warning for people who don't use hooks (most of them, I think). We could ratelimit into only one message e.g. with a boolean flag (that could be reseted e.g. in `SetConfig()` of the local source.

marquiz (changes requested):
Thanks @fmuyassarov for this. Looks good, almost ready for merging. In addition to addressing the comment on logging, please run `make templates` so that the example config in the Helm chart also gets updated.

/ok-to-test

source/local/local.go
fmuyassarov: Sure thing @marquiz . So we have two options I guess...
**1.** in `SetConfig()`, 

```
// SetConfig method of the LabelSource interface
func (s *localSource) SetConfig(conf source.Config) {
	switch v := conf.(type) {
	case *Config:
		s.config = v
		if s.config.HooksEnabled {
			klog.Warning("hooks are DEPRECATED since v0.12.0 and support will be removed in a future release; use feature files instead")
		}
	default:
		klog.Fatalf("invalid config type: %T", conf)
	}
}
```
log trace when hooks enabled
```
I0826 07:54:40.275016       1 nfd-worker.go:154] Node Feature Discovery Worker v0.12.0-devel-76-ga3100ce-dirty
I0826 07:54:40.275237       1 nfd-worker.go:155] NodeName: 'nfd-worker'
I0826 07:54:40.275465       1 nfd-worker.go:422] configuration file "/etc/kubernetes/node-feature-discovery/nfd-worker.conf" parsed
W0826 07:54:40.275535       1 local.go:80] hooks are DEPRECATED since v0.12.0 and support will be removed in a future release; use feature files instead
I0826 07:54:40.275547       1 nfd-worker.go:460] worker (re-)configuration successfully completed
I0826 07:54:40.275592       1 base.go:127] connecting to nfd-master at nfd-master:8080 ...
I0826 07:54:40.275630       1 component.go:36] [core]parsed scheme: ""
I0826 07:54:40.275640       1 component.go:36] [core]scheme "" not registered, fallback to default scheme

```
log trace when hooks enabled
```
I0826 07:58:39.551162       1 nfd-worker.go:154] Node Feature Discovery Worker v0.12.0-devel-76-ga3100ce-dirty
I0826 07:58:39.551321       1 nfd-worker.go:155] NodeName: 'nfd-worker'
I0826 07:58:39.552185       1 nfd-worker.go:422] configuration file "/etc/kubernetes/node-feature-discovery/nfd-worker.conf" parsed
I0826 07:58:39.552341       1 nfd-worker.go:460] worker (re-)configuration successfully completed
I0826 07:58:39.552453       1 base.go:127] connecting to nfd-master at nfd-master:8080 ...
I0826 07:58:39.552612       1 component.go:36] [core]parsed scheme: ""
I0826 07:58:39.552627       1 component.go:36] [core]scheme "" not registered, fallback to default scheme
```

**2.** in `getFeaturesFromHooks()`. When hooks are enabled, depreciation message gets printed on every reload
```
func getFeaturesFromHooks(s *localSource) (map[string]string, error) {
.....
	if s.config.HooksEnabled {
		klog.Warning("hooks are DEPRECATED since v0.12.0 and support will be removed in a future release; use feature files instead")
	}
}
```
log trace when hooks enabled
```

I0826 07:42:56.606694       1 local.go:116] starting hooks...
W0826 07:42:56.618914       1 local.go:203] hooks are DEPRECATED since v0.12.0 and support will be removed in a future release; use feature files instead
I0826 07:42:56.645574       1 nfd-worker.go:471] starting feature discovery...
I0826 07:42:56.646019       1 nfd-worker.go:483] feature discovery completed
I0826 07:42:56.646139       1 nfd-worker.go:564] sending labeling request to nfd-master
E0826 07:43:56.700575       1 kernel.go:132] failed to read kconfig: failed to read kernel config from [ /proc/config.gz /host-usr/src/linux-5.15.0-46-generic/.config /host-usr/src/linux/.config /host-usr/lib/modules/5.15.0-46-generic/config /host-usr/lib/ostree-boot/config-5.15.0-46-generic /host-usr/lib/kernel/config-5.15.0-46-generic /host-usr/src/linux-headers-5.15.0-46-generic/.config /lib/modules/5.15.0-46-generic/build/.config /host-boot/config-5.15.0-46-generic]
I0826 07:43:56.702207       1 local.go:116] starting hooks...
W0826 07:43:56.709296       1 local.go:203] hooks are DEPRECATED since v0.12.0 and support will be removed in a future release; use feature files instead
I0826 07:43:56.735445       1 nfd-worker.go:471] starting feature discovery...
I0826 07:43:56.735863       1 nfd-worker.go:483] feature discovery completed
I0826 07:43:56.735929       1 nfd-worker.go:564] sending labeling request to nfd-master
```
log trace when hooks disabled
```
I0826 07:44:05.110774       1 nfd-worker.go:211] reloading configuration
I0826 07:44:05.116216       1 nfd-worker.go:422] configuration file "/etc/kubernetes/node-feature-discovery/nfd-worker.conf" parsed
I0826 07:44:05.116438       1 nfd-worker.go:460] worker (re-)configuration successfully completed
E0826 07:44:05.119196       1 kernel.go:132] failed to read kconfig: failed to read kernel config from [ /proc/config.gz /host-usr/src/linux-5.15.0-46-generic/.config /host-usr/src/linux/.config /host-usr/lib/modules/5.15.0-46-generic/config /host-usr/lib/ostree-boot/config-5.15.0-46-generic /host-usr/lib/kernel/config-5.15.0-46-generic /host-usr/src/linux-headers-5.15.0-46-generic/.config /lib/modules/5.15.0-46-generic/build/.config /host-boot/config-5.15.0-46-generic]
I0826 07:44:05.145666       1 nfd-worker.go:471] starting feature discovery...
I0826 07:44:05.146036       1 nfd-worker.go:483] feature discovery completed
I0826 07:44:05.146066       1 nfd-worker.go:564] sending labeling request to nfd-master
```

 I personally prefer option 1 but fine with other opinions too. WDYT?

fmuyassarov: Perhaps it is better to keep it as it is since the logs get printed only when hooks are enabled? It is true that users might not be using it although enabled and still see the logs, but it seems fine to me because we warning them about the feature that they should not enable. WDYT @marquiz ?

fmuyassarov:
> Thanks @fmuyassarov for this. Looks good, almost ready for merging. In addition to addressing the comment on logging, please run `make templates` so that the example config in the Helm chart also gets updated.
> 
> /ok-to-test

Thanks, done.

source/local/local.go
marquiz: In my thinking this might create confusion if you get warning like this with the default configuration (and you're not using hooks). But maybe it's good to print anyway - but in this case I'd probably do some rate limiting as I suggested, e.g. once after every config change or smth. WDYT?

source/local/local.go
```diff
@@ -143,6 +183,9 @@ func getFeaturesFromHooks() (map[string]string, error) {
 		}
 		return features, fmt.Errorf("unable to access %v: %v", hookDir, err)
 	}
+	if !(len(files) > 0) {
+		printHooksWarning = false
+	}
```
marquiz, line 188: Hmm, now it looks that we only print the warning when NO hooks are found 🧐

I'd just print it here (without any ratelimiting)

```suggestion
	if !(len(files) > 0) {
		klog.Warning("hooks are DEPRECATED since v0.12.0 and support will be removed in a future release; use feature files instead")
		}
```

fmuyassarov: @marquiz No, I would say it is the opposite
`if !(len(files) > 0)` -> when hooks dir is empty, don't print. Perhaps you missed exclamation mark

eero-t: That should really be written without '!' for clarity: `if len(files) == 0`.

Also, above warning is about *use* of a deprecated API.

Because enabling hooks opens potential NFD security hole [1], I think security warning should be shown when hooks are enabled.  For example:
`klog.Warning("insecure hooks API enabled with 'HooksEnabled: true' (DEPRECATED since v0.12.0)")`

[1] of NFD executing random binaries from host, which may be  insecure, or even potentially malicious, and hooks require having extra components in NFD container which may later turn out to contain security holes.

fmuyassarov: I agree with your first point on simplifying condition. But for the warning message it seems we are overcomplicating it. There could be another issue apart from security. I think, if there are some drawbacks of using a feature it should be documented clearly but not necessarily warned in the log trace.

marquiz: > of NFD executing random binaries from host

I wouldn't say random binaries from host. System / cluster admin is responsible for those in a similar way that any "system" files, I'd say. If the administrator just randomly installs any scripts or runs random containers with root privileges nfd is one of the minor problems :)

eero-t: They are random from the **NFD** PoV, and may force keeping it less secure.  It does not e.g. know whether what they do matches NFD security context settings: can some of its security context settings tightened or would that break some random hook?

marquiz: I get that but nfd-worker (with the default deployments) runs with probably the tightest security context possible: fileystem and all mounts are ro, run-as-non-root, allow-privilege-escalation=false, all caps dropped...

eero-t: Ok. I concur that warning about deprecated API *use* is enough.  And while there's no rate-limiting, it's indeed better just to output that warning directly in getFeaturesFromHooks().   Adding package-global "printHooksWarning" flag just to do warning from calling function is redundant complication.

fmuyassarov: @eero-t No problem. @marquiz I think it is good to merge it now.

marquiz: Yeah, looks good now

marquiz (approved):
Thanks @fmuyassarov, good to merge from my point of view

marquiz:
> Because enabling hooks opens potential NFD security hole [1], I think warning should be shown (also) when hooks are enabled.

I think we could print another note (possibly just an info/note, not a warning at this stage when `hooksEnabled=true` by default) in `SetConfig`. This would be printed whenever config is changed. Can be done in a separate PR, too

zvonkok:
On PTO will take a look next week.

ArangoGutierrez:
I like this PR
but since it introduces a deprecating change, let's go for more eyes
cc @zvonkok @yevgeny-shnaidman

Issue #778, opened 2022-02-17
Use shorter label namespace


marquiz:
**What would you like to be added**:

I'd like to move to a shorter label namespace (e.g. `nfd.k8s.io/` or `feature.node.k8s.io/`) before nfd v1.0. I'm not sure what would be the least painful way there (or if it even is worth it).

One path could be:
1. Allow the shorter namespace. For custom labels only
2. Have an option to switch to the shorter namespace for all labels (the old long namespaces would still be the default)
3. Make the new namespace the default for built-in labels. Have an option to use the old one instead
4. Drop the compatibility mode (no hurry with this)

**Why is this needed**:

Usability

marquiz:
@zvonkok @ArangoGutierrez what do you think about this? Is it too disruptive?

OTOH we can start implementing this with the two first steps (allow `nfd.k8s.io` and create a flag for turning that on for all labels). This gives us time to experiment and get feedback based on which we can do the switchover of back off.

ArangoGutierrez:
Agree that this should be optional for some time. Many workloads might have hard coded the current schema.

Question: can we use k8s when becoming a 1.0 API?

fmuyassarov:
I think it is better we keep discussion notes here. So, I will copy paste my comments here from the PR. 

This is a breaking change, and usually in k8s for such a change it takes ~3-4 releases to remove a feature. Example: [rename "node-role.kubernetes.io/master" ](https://github.com/kubernetes/kubeadm/issues/2200) `node-role.kubernetes.io/master` label & taint renaming in Kubeadm, they had both labels in parallel that gave time to users to switch to the new format. And perhaps we can follow the same path?

For example, 

#### Release - v0.12.X
1. introduce a short label and a flag to switch between long and short labels
2. set the flag to use long label by default
3. announce to community/users to adapt to use the short label
4. get community/users feedback

#### Release - v0.13.X
1. turn on both short and long labels by default but still have an option to disable them
2. get community/users feedback

#### Release - v0.14.X
1. Turn off long label by default but still have an option to enable it

#### Release - v0.15.X
1. Remove the long label entirely from the codebase and keep only short label
2. Remove the flag to switch between short and long labels

I'm not sure about the estimated time between MINOR releases in NFD project, but if you think it is too short that we take above steps in every MINOR release, then of course we can extend it. And actually better if we get users feedback on that.

[...]

PR #881, opened 2022-09-01
Use shorter label namespace

fmuyassarov:
Fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/778

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -107,6 +107,8 @@ func initFlags(flagset *flag.FlagSet) *master.Args {
 	flagset.BoolVar(&args.VerifyNodeName, "verify-node-name", false,
 		"Verify worker node name against the worker's TLS certificate. "+
 			"Only takes effect when TLS authentication has been enabled.")
+	flagset.StringVar(&args.NsFormat, "label-ns", "legacy",
+		"Feature label namespace format. By default set to legacy.")
 
 	return args
 }
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -52,6 +52,24 @@ import (
 	"sigs.k8s.io/node-feature-discovery/pkg/version"
 )
 
+var (
+	// FeatureLabelNs is the namespace for feature labels
+	FeatureLabelNs = "feature.node.kubernetes.io"
+	// FeatureLabelSubNsSuffix is the suffix for allowed feature label sub-namespaces
+	FeatureLabelSubNsSuffix = "." + FeatureLabelNs
+)
+
+const (
+	// ProfileLabelNs is the namespace for profile labels
+	ProfileLabelNs = "profile.node.kubernetes.io"
+
+	// ProfileLabelSubNsSuffix is the suffix for allowed profile label sub-namespaces
+	ProfileLabelSubNsSuffix = "." + ProfileLabelNs
+
+	// AnnotationNsBase namespace for all NFD-related annotations
+	AnnotationNsBase = "nfd.node.kubernetes.io"
+)
+
 // Labels are a Kubernetes representation of discovered features.
 type Labels map[string]string
 
@@ -72,6 +90,7 @@ type Args struct {
 	LabelWhiteList         utils.RegexpVal
 	FeatureRulesController bool
 	NoPublish              bool
+	NsFormat               string
 	Port                   int
 	Prune                  bool
 	VerifyNodeName         bool
@@ -102,6 +121,13 @@ func NewNfdMaster(args *Args) (NfdMaster, error) {
 		nodeName: os.Getenv("NODE_NAME"),
 		ready:    make(chan bool, 1),
 		stop:     make(chan struct{}, 1),
+	}
+
+	if args.NsFormat != "legacy" {
+		// FeatureLabelNs is the namespace for feature labels
+		FeatureLabelNs = "nfd.k8s.io"
+		// FeatureLabelSubNsSuffix is the suffix for allowed feature label sub-namespaces
+		FeatureLabelSubNsSuffix = "." + FeatureLabelNs
 	}
 
 	if args.Instance != "" {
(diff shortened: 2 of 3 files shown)


marquiz:
> @marquiz Sorry to bother again :)

Sorry, haven't had time to concentrate on this. Just some quick thoughts below. The details and corner cases might get hairier than expected.

If we want to do this, I think we need to think about a transition plan from long to short labels. I.e. have a way to not break existing applications. This would probably mean the possibility to have both long and short labels turned on at the same time. Then, after apps have been migrated long labels could be turned off 🤔 Another consideration is that in the code cleaning up old labels we would need to take care about both the long and short labels (I think). And the (still experimental) extended resources should be handled, too.

/ok-to-test

fmuyassarov:
This is a breaking change, and usually in k8s for such a changes it takes ~3-4 releases to remove a feature. Example: [rename "node-role.kubernetes.io/master" ](https://github.com/kubernetes/kubeadm/issues/2200) `node-role.kubernetes.io/master` label & taint renaming in Kubeadm, they had both labels in parallel that gave time to users to switch to the new format. And perhaps we can follow the same path?

For example, 

#### Release - v0.12.X
1. introduce a short label and a flag to switch between long and short labels
2. set the flag to use long label by default
3. announce to community/users to adapt to use the short label
4. get community/users feedback

#### Release - v0.13.X
1. turn on both short and long labels by default but still have an option to disable them
2. get community/users feedback

#### Release - v0.14.X
1. Turn off long label by default but still have an option to enable it

#### Release - v0.15.X
1. Remove the long label entirely from the codebase and keep only short label
2. Remove the flag to switch between short and long labels

I'm not sure about the estimated time between MINOR releases in NFD project, but if you think it is too short that we take above steps in every MINOR release, then of course we can extend it. And actually better we get users feedback on that.

marquiz:
> #### Release - v0.12.X
> 1. introduce a short label and a flag to switch between long and short labels
> 2. set the flag to use long label by default
> 3. announce to community/users to adapt to use the short label
> 4. get community/users feedback
> 
> #### Release - v0.13.X
> 1. turn on both short and long labels by default but still have an option to disable them
> 2. get community/users feedback
> 
> #### Release - v0.14.X
> 1. Turn off long label by default but still have an option to enable it
> 
> #### Release - v0.15.X
> 1. Remove the long label entirely from the codebase and keep only short label
> 2. Remove the flag to switch between short and long labels
> 
> I'm not sure about the estimated time between MINOR releases in NFD project, but if you think it is too short that we take above steps in every MINOR release, then of course we can extend it. And actually better if we get users feedback on that.

Yeah, along those lines, I guess. I had something similar in my mind. Do transition over several releases and we can also back off if it turns out to be too much of an hassle 😅

fmuyassarov:
Rebased

cmd/nfd-master/main.go
```diff
@@ -107,6 +107,8 @@ func initFlags(flagset *flag.FlagSet) *master.Args {
 	flagset.BoolVar(&args.VerifyNodeName, "verify-node-name", false,
 		"Verify worker node name against the worker's TLS certificate. "+
 			"Only takes effect when TLS authentication has been enabled.")
+	flagset.StringVar(&args.NsFormat, "label-ns", "legacy",
+		"Feature label namespace format. By default set to legacy.")
```
marquiz, line 111: I think we need the possibility to have both short and long turned on at the same time 🤔 

One option is to make this flag tri-state, i.e. smth like `-label-ns={legacy,short,both}`. If keeping the flag as string parameter we need to do input validation of it.

Another way could be having separate bool flags, i.e. smth like `-label-ns-legacy` and `-label-ns-short`. In this case we probably want to check that at least one of them is turned on.

pkg/nfd-master/nfd-master.go
```diff
@@ -52,6 +52,24 @@ import (
...
+var (
+	// FeatureLabelNs is the namespace for feature labels
+	FeatureLabelNs = "feature.node.kubernetes.io"
+	// FeatureLabelSubNsSuffix is the suffix for allowed feature label sub-namespaces
+	FeatureLabelSubNsSuffix = "." + FeatureLabelNs
+)
+
+const (
+	// ProfileLabelNs is the namespace for profile labels
+	ProfileLabelNs = "profile.node.kubernetes.io"
```
marquiz, line 64: The profile ns should switch to a shorter version, too

marquiz (commented):
Thanks @fmuyassarov for working on this.

I think there's more things to do here. At least:

1. Deletion of labels: labels created by NFD are stored in the `feature-labels` annotation. When deleting old labels we must delete labels from both the long and short ns.
2. Allowed namespaces: both the long and short should be allowed independent of what is enabled for the default labels. People might have rules creating labels in the old namespace and we don't want to break those. Don't know if we should have some automation to convert/duplicate those into the shorter ns version 🤔
3. I think we need a bit more documentation, even at this phase. E.g. some notes in the docs where `feature.node.kubernetes.io` namespace is mentioned in the docs (like `docs/get-started/features.md`)

ArangoGutierrez:
PING this PR alive?

fmuyassarov:
> PING this PR alive?

Not at the moment. But I plan to update it hopefully this week.

ArangoGutierrez:
@fmuyassarov PING this PR alive?

ArangoGutierrez:
> * /remove-lifecycle stale

/remove-lifecycle stale

marquiz:
I've started to think if we really want/need this at all. I'm not sure it's worth all the hassle, especially wrt. to breaking the "user interface", i.e. changing all labels. After thinking it again, I don't think `nfd.k8s.io` is necessarily a good choice, especially as we already have two subnamespaces `feature.node.kubernetes.io` and `profile.node.kubernetes.io`. We could save a few characters with `feature.node.k8s.io` but is it really worth it

fmuyassarov:
> 

yeah, that's one of the reasons I didn't have a motivation to finish this work, because as we discussed earlier f2f, I was also not sure if we really need it.

ArangoGutierrez:
> > 
> 
> yeah, that's one of the reasons I didn't have a motivation to finish this work, because as we discussed earlier f2f, I was also not sure if we really need it.

Close?

fmuyassarov:
> > > 
> > 
> > 
> > yeah, that's one of the reasons I didn't have a motivation to finish this work, because as we discussed earlier f2f, I was also not sure if we really need it.
> 
> Close?

for now at least yes, 
/close

PR #882, opened 2022-09-02
Run local markdown tests inside an isolated container

fmuyassarov:
When working on a patch that touches docs/, there is no way to test changes against markdown checks before submitting a PR. In fact, we have a makefile target `mdlint` but it has a prerequisite of having mdl installed on the developer machine. What do you think of swapping the tasks of mdlint Makefile target with scripts/test-infra/mdlint.sh script with minor modifications so that developer can just run `make mdlint` which calls scripts/test-infra/mdlint.sh and the tests get executed within a container. Once execution is completed the temporary container gets removed so that we don't have garbage collected. I've chosen [ruby:slim](https://hub.docker.com/layers/library/ruby/slim/images/sha256-2e3b4a4628fead83109e77fa6732e430cf9881e2d8cb3fa6148935e87ea4ea70?context=explore) image (70.39 MB), which is almost five times smaller than [ruby:2.7](https://hub.docker.com/layers/library/ruby/2.7/images/sha256-5a32f6abc3c485231580c241047954c9d857e208a65392a8bca7fbb1888d2ed6?context=explore) (321.36 MB) that [we use in our](https://github.com/kubernetes/test-infra/blob/6403fa097aec8093550c398ded1022bbf180e770/config/jobs/kubernetes-sigs/node-feature-discovery/node-feature-discovery-presubmits.yaml#L49) Prow tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,8 +12,6 @@ BUILDER_IMAGE ?= golang:1.18-buster
 BASE_IMAGE_FULL ?= debian:buster-slim
 BASE_IMAGE_MINIMAL ?= gcr.io/distroless/base
 
-MDL ?= mdl
-
 # Docker base command for working with html documentation.
 # Use host networking because 'jekyll serve' is stupid enough to use the
 # same site url than the "host" it binds to. Thus, all the links will be
@@ -157,7 +155,12 @@ lint:
 	golint -set_exit_status ./...
 
 mdlint:
-	find docs/ -path docs/vendor -prune -false -o -name '*.md' | xargs $(MDL) -s docs/mdl-style.rb
+	${CONTAINER_RUN_CMD} \
+	--rm \
+	--volume "${PWD}:/workdir:ro,z" \
+	--workdir /workdir \
+	ruby:slim \
+	/workdir/scripts/test-infra/mdlint.sh
 
 helm-lint:
 	helm lint --strict deployment/helm/node-feature-discovery/
--- a/scripts/test-infra/mdlint.sh
+++ b/scripts/test-infra/mdlint.sh
@@ -2,6 +2,5 @@
 
 # Install mdl
 gem install mdl -v 0.11.0
-
 # Run verify steps
-make mdlint
+find docs/ -path docs/vendor -prune -false -o -name '*.md' | xargs mdl -s docs/mdl-style.rb


ArangoGutierrez (commented):
Test run on an isolated container at https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/node-feature-discovery/node-feature-discovery-presubmits.yaml#L38. 
If what you want is a devel `make` target, then you need to have in consideration the above.

fmuyassarov:
> Test run on an isolated container at https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/node-feature-discovery/node-feature-discovery-presubmits.yaml#L38. 
> 
> If what you want is a devel `make` target, then you need to have in consideration the above.

Yes, so my plan is run it in container when testing locally. If I am not wrong, this won't break Prow tests too, it is just that there will be an extra container created within prowJon container.

marquiz:
I think the idea makes sense. Makes life easier for most of the people

> If I am not wrong, this won't break Prow tests too, it is just that there will be an extra container created within prowJob container.

I think the idea behind he scripts under `scripts/test-infra` has been to keep them as simple and possible and only have strictly prow-specific stuff there. One possibility:
1. put the one-liner from the makefile into a separate script (say `scripts/mdlint.sh`)
2. call this^ script from `scripts/test-infra/mdlint.sh`
3. write another script for running mdlint from a container (say `hack/mdlint-containerized.sh` that runs `scripts/mdlint.sh` inside a container)
4. Makefile `mdlint:` target would call `hack/mdlint-containerized.sh`

WDYT?

fmuyassarov:
> I think the idea makes sense. Makes life easier for most of the people
> 
> > If I am not wrong, this won't break Prow tests too, it is just that there will be an extra container created within prowJob container.
> 
> I think the idea behind he scripts under `scripts/test-infra` has been to keep them as simple and possible and only have strictly prow-specific stuff there. One possibility:
> 
Got it.
> 1. put the one-liner from the makefile into a separate script (say `scripts/mdlint.sh`)
> 2. call this^ script from `scripts/test-infra/mdlint.sh`
> 3. write another script for running mdlint from a container (say `hack/mdlint-containerized.sh` that runs `scripts/mdlint.sh` inside a container)
> 4. Makefile `mdlint:` target would call `hack/mdlint-containerized.sh`
> 
> WDYT?

Yes this should work and I'm fine with this. I will just mention that, we can reuse the same scripts/test-infra/mdlint.sh for both local and Prow tests. When running on Prow, we just have to add variable in test infra with `IS_CONTAINER=true`. What I don't like with duplication is that we always have to take care two sets of the same things and we have to be careful that they don't diverge. But as said, it is not a big deal at all and I'm ok to decouple the piece for local runs.

marquiz:
> Got it.

And to put it in other words, the stuff under `scripts/test-infra` is only supposed to be called from prow 😄 

> > 1. put the one-liner from the makefile into a separate script (say `scripts/mdlint.sh`)
> > 2. call this^ script from `scripts/test-infra/mdlint.sh`
> > 3. write another script for running mdlint from a container (say `hack/mdlint-containerized.sh` that runs `scripts/mdlint.sh` inside a container)
> > 4. Makefile `mdlint:` target would call `hack/mdlint-containerized.sh`
> > 
> > WDYT?
> 
> Yes this should work and I'm fine with this. I will just mention that, we can reuse the same scripts/test-infra/mdlint.sh for both local and Prow tests. When running on Prow, we just have to add variable in test infra with `IS_CONTAINER=true`. What I don't like with duplication is that we always have to take care two sets of the same things and we have to be careful that they don't diverge. But as said, it is not a big deal at all and I'm ok to decouple the piece for local runs.

Yeah, I think this is fine. No need to have two separate scripts

marquiz:
> > Got it.
> 
> And to put it in other words, the stuff under `scripts/test-infra` is only supposed to be called from prow 😄
> 
> > > 1. put the one-liner from the makefile into a separate script (say `scripts/mdlint.sh`)
> > > 2. call this^ script from `scripts/test-infra/mdlint.sh`
> > > 3. write another script for running mdlint from a container (say `hack/mdlint-containerized.sh` that runs `scripts/mdlint.sh` inside a container)
> > > 4. Makefile `mdlint:` target would call `hack/mdlint-containerized.sh`
> > > 
> > > WDYT?
> > 
> > 
> > Yes this should work and I'm fine with this. I will just mention that, we can reuse the same scripts/test-infra/mdlint.sh for both local and Prow tests. When running on Prow, we just have to add variable in test infra with `IS_CONTAINER=true`. What I don't like with duplication is that we always have to take care two sets of the same things and we have to be careful that they don't diverge. But as said, it is not a big deal at all and I'm ok to decouple the piece for local runs.
> 
> Yeah, I think this is fine. No need to have two separate scripts

Hmm, actually what I don't like in this is the recursive nature of the script. With a small mistake end up with infinite loop of nested containers... There should be no duplication with the scripts: one script simply runs mdlint, another script runs the first one in a container environment, and test-infra script simply calls the first one as well. WDYT?

fmuyassarov:
> > > Got it.
> > 
> > 
> > And to put it in other words, the stuff under `scripts/test-infra` is only supposed to be called from prow smile
> > > > 1. put the one-liner from the makefile into a separate script (say `scripts/mdlint.sh`)
> > > > 2. call this^ script from `scripts/test-infra/mdlint.sh`
> > > > 3. write another script for running mdlint from a container (say `hack/mdlint-containerized.sh` that runs `scripts/mdlint.sh` inside a container)
> > > > 4. Makefile `mdlint:` target would call `hack/mdlint-containerized.sh`
> > > > 
> > > > WDYT?
> > > 
> > > 
> > > Yes this should work and I'm fine with this. I will just mention that, we can reuse the same scripts/test-infra/mdlint.sh for both local and Prow tests. When running on Prow, we just have to add variable in test infra with `IS_CONTAINER=true`. What I don't like with duplication is that we always have to take care two sets of the same things and we have to be careful that they don't diverge. But as said, it is not a big deal at all and I'm ok to decouple the piece for local runs.
> > 
> > 
> > Yeah, I think this is fine. No need to have two separate scripts
> 
> Hmm, actually what I don't like in this is the recursive nature of the script. With a small mistake end up with infinite loop of nested containers... There should be no duplication with the scripts: one script simply runs mdlint, another script runs the first one in a container environment, and test-infra script simply calls the first one as well. WDYT?

Sorry perhaps I confused you. But I'm not changing how we run tests in Prow, because in prow config, we will have  `IS_CONTAINER=true` which will execute
```
if [ "${IS_CONTAINER}" != "false" ]; then
    # Install mdl
    gem install mdl -v 0.11.0
    # Run verify steps
    find docs/ -path docs/vendor -prune -false -o -name '*.md' -exec mdl -s docs/mdl-style.rb {} \+ 
```
Basically nothing changes from the Prow's perspective and we are not having nested containers.

As a developer who wants to tests, we would just run `make lint` which will create a container and
run the tests inside, exactly the same as Prow (1:1). So, this patch is only an addition I would say.

fmuyassarov:
> 
> As a developer who wants to test locally, we would just run `make mdlint` which will create a container and run the tests inside, exactly the same as Prow (1:1). So, this patch is only an addition I would say.

It is the same as Prow, because ProwJob jobs also creates the same temporary container with the same container image that we are using here(ruby:slim)

fmuyassarov:
I've updated https://github.com/kubernetes/test-infra/pull/27376 to match the changes. Let know what you think. Thanks

scripts/test-infra/mdlint.sh
```diff
@@ -1,7 +1,19 @@
...
+    find docs/ -path docs/vendor -prune -false -o -name '*.md' -exec mdl -s docs/mdl-style.rb {} \+
+else
+  "${CONTAINER_RUNTIME}" run --rm \
+    --env IS_CONTAINER=TRUE \
+    --volume "${PWD}:/workdir:ro,z" \
+    --entrypoint sh \
+    --workdir /workdir \
+    ruby:slim \
+    /workdir/scripts/test-infra/mdlint.sh "${@}"
+fi;
```
marquiz, line 19: Somehow this feels too complicated. Should be super simple. The "philosophy" has been that `scripts/test-infra` are as simple as possible and basically only supposed to be called from prox.

```suggestion
# Install mdl
gem install mdl -v 0.11.0

find docs/ -path docs/vendor -prune -false -o -name '*.md' | xargs $(MDL) -s docs/mdl-style.rb
```

Makefile
```diff
@@ -157,7 +155,7 @@ lint:
 	golint -set_exit_status ./...
 
 mdlint:
-	find docs/ -path docs/vendor -prune -false -o -name '*.md' | xargs $(MDL) -s docs/mdl-style.rb
+	./scripts/test-infra/mdlint.sh
```
marquiz, line 158: I commented (below) that `scripts/test-infra` are supposed to be called by prow only but I think this could be an exception as the environment is supposed to be identical (container) to prow

```suggestion
	${CONTAINER_RUN_CMD} \
	    --rm \
	    --volume "${PWD}:/workdir:ro,z" \
	    --workdir /workdir \
	    ruby:slim \
	    /workdir/scripts/test-infra/mdlint.sh
```

fmuyassarov:
Thanks for reviews Markus. Updated as per suggestion.

marquiz (approved):
> Thanks for reviews Markus. Updated as per suggestion.

Thanks, looks quite simple and clean to me now 😎

ArangoGutierrez (approved):
/lgtm

fmuyassarov:
well, I will do it manually:)

PR #894, opened 2022-09-15
Run local markdown tests inside an isolated container

fmuyassarov:
When working on a patch that touches docs/, there is no way to test changes against markdown checks before submitting a PR. In fact, we have a makefile target mdlint but it has a prerequisite of having mdl installed on the developer machine. What do you think of swapping the tasks of mdlint Makefile target with scripts/test-infra/mdlint.sh script with minor modifications so that developer can just run make mdlint which calls scripts/test-infra/mdlint.sh and the tests get executed within a container. Once execution is completed the temporary container gets removed so that we don't have garbage collected. I've chosen [ruby:slim](https://hub.docker.com/layers/library/ruby/slim/images/sha256-2e3b4a4628fead83109e77fa6732e430cf9881e2d8cb3fa6148935e87ea4ea70?context=explore) image (70.39 MB), which is almost five times smaller than [ruby:2.7](https://hub.docker.com/layers/library/ruby/2.7/images/sha256-5a32f6abc3c485231580c241047954c9d857e208a65392a8bca7fbb1888d2ed6?context=explore) (321.36 MB) that [we use in our](https://github.com/kubernetes/test-infra/blob/6403fa097aec8093550c398ded1022bbf180e770/config/jobs/kubernetes-sigs/node-feature-discovery/node-feature-discovery-presubmits.yaml#L49) Prow tests.

Backport of https://github.com/kubernetes-sigs/node-feature-discovery/pull/882

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,8 +11,6 @@ CONTAINER_RUN_CMD ?= docker run
 BASE_IMAGE_FULL ?= debian:buster-slim
 BASE_IMAGE_MINIMAL ?= gcr.io/distroless/base
 
-MDL ?= mdl
-
 K8S_CODE_GENERATOR ?= ../code-generator
 
 # Docker base command for working with html documentation.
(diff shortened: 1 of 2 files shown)


fmuyassarov:
I had a chat with @marquiz and we decided that the correct path would be to actually have branch specific test entries in prow config. As such, I'm going to close this patch and extend our prow config.
/close

PR #897, opened 2022-09-20
nfd-master: drop cleanup of ancient incubator labels

marquiz:
Remove the cleanup code that removes ancient NFD labels with the node.alpha.kubernetes-incubator.io/ prefix. This label namespace was deprecated/dropped already in v0.4.0 so it should be safe to drop this code.


ArangoGutierrez (approved):
/lgtm

PR #903, opened 2022-09-29
docs: document NodeFeature API

marquiz:
Document the usage of the NodeFeature CRD API. Also re-organize the
documentation a bit, moving the description of NodeFeatureRule
controller from customization guide to nfd-master usage page.

PR #904, opened 2022-10-04
apis/nfd: move annotation and label consts from nfd-master

marquiz:
Move consts related to NFD annotations and labels from nfd-master to the api. Makes them more logically accessible for clients.

Split out of #903


fmuyassarov (approved):
This is better :+1:

ArangoGutierrez (approved):
/lgtm

PR #909, opened 2022-10-06
pkg/utils: move hostpath helpers from source to utils

marquiz:
Refactor the code, moving the hostpath helper functionality to new "pkg/utils/hostpath" package. This breaks odd-ish dependency "pkg/utils" -> "source".

No functional changes, preparing for #903


zvonkok (commented):
/lgtm sound like it

Issue #122, opened 2018-04-18
"Failure updating labels" sometimes happens, exits daemonset-mode pod


okartau:
When one-shot job is turned into daemon, few things should get double-checked: resource leaking and stability (i.e. exit points) as these are minor issues for single shot but more serious for daemon.
K8s is good in restarting pods, but that's not excuse to allow low quality.
So I keep NFD daemonset running to check for both.
There is weak sign of slow mem.growth pattern, but it may get leveled at longer run, so I planned to keep it up for long monitoring. But no luck there, as nfd pod exits and gets restarted about once per 24h. Next, I tried shortened cycle time from 60 seconds in hope I see mem.pattern more quickly, but that does not work as the exit rate follows: with 1-sec cycle rate I see one exit in about 1h.

I ran pod log in -f mode to capture the exit messages, and these are:

2018/04/18 11:00:27 can't update node: Operation cannot be fulfilled on nodes "k8-vm-2": the object has been modified; please apply your changes to the latest version and try again
2018/04/18 11:00:27 failed to advertise labels: Operation cannot be fulfilled on nodes "k8-vm-2": the object has been modified; please apply your changes to the latest version and try again
2018/04/18 11:00:27 error occurred while updating node with feature labels: Operation cannot be fulfilled on nodes "k8-vm-2": the object has been modified; please apply your changes to the latest version and try again

That issue has likely always been there, but probability hitting it was low, it has not surfaced before. 
Seems there is window during which node labels update is sentenced to fail?
Can that state be detected somehow so that we don't try to update?
Or, should we retry after some delay once such condition is hit?
Or, is this sign of some other problem somewhere else?
In any case, we should try to do better than exit, now that we can run as daemonset.

okartau:
Termination happens implicitly in stderrLogger.Fatalf(), 
which gets called from main loop in case of error.
Note that before "run as daemonset" change, this was last line of main(), means forced exit
did not change the outcome of one-shot run at all.
Now it makes big difference in daemonset mode.
One simple improvement approach would be: use less aggressive log level and avoid forced exit.
Then there will remain another area for study: why updating labels fails sometimes.

balajismaniam:
[...]

PR #910, opened 2022-10-06
Allow optionally setting node taints defined on the NodeFeatureRule CR

fmuyassarov:
This PR enables setting node taints defined on NodeFeatureRule CR. User can
define a list of taints under spec.rule of the CR in this example:

```yaml
kind: NodeFeatureRule
metadata:
  name: rule
spec:
  rules:
    - name: "rule"
      labels:
        "my-sample-feature": "true"
        "my-sample-feature2": "true"
      taints:
        - effect: NoSchedule
          key: foo
          value: abc
        - effect: NoSchedule
          key: bar
          value: abc
      matchFeatures:
          matchExpressions:
            dummy: {op: Exists}
```
The only allowed values for the effect are `NoExecute`, `NoSchedule`, `PrefereNoSchedule`
(following [k8s](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/)). By default this feature is disabled and enabling it requires explicitly setting
`--no-taint` flag false. Even if there are taints defined in the NodeFeatureRule CR and the
rule matches, taints will not be set on the node until flag is enabled. 

Fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/540

Currently missing unit tests.


fmuyassarov:
/cc @marquiz 
I think it is ready for the review and meanwhile I will add a commit with unit tests.

docs/advanced/customization-guide.md
```diff
@@ -47,6 +50,14 @@ spec:
     - name: "my sample rule"
       labels:
         "my-sample-feature": "true"
+# Using tainting feature requires the feature gate (--no-taint) to be enabled (set to false)
+      taints:
+        - effect: PreferNoSchedule
+          key: bar
+          value: true
```
marquiz, line 57: You need to put `true` inside quotes, otherwise the unmarshaller treats it as bool (instead of string which is what we want) and fails

```suggestion
          value: "true"
```

docs/advanced/customization-guide.md
```diff
@@ -30,8 +30,11 @@ labeling:
 ## NodeFeatureRule custom resource
 
 `NodeFeatureRule` objects provide an easy way to create vendor or application
-specific labels. It uses a flexible rule-based mechanism for creating labels
-based on node feature.
+specific labels and taints. It uses a flexible rule-based mechanism for creating
+labels/taints based on node features. Tainting feature is disabled by default and
+to enable it set `--no-taint` flag to `false`. If the feature gate `--no-taint`
+is `true` (i.e. disabled), taints defined in the NodeFeatureRule CRD have no effect
+and will be ignored by the NFD controller.
```
marquiz, line 37: Maybe we should add a note/warning that the taints will possibly cause `nfd-worker` pods to be evicted from the nodes that get tainted. And note that this should be negated by customizing the nfd-worker daemonset with corresponding tolerations. Or, even creating a generic toleration that tolerates all possible taints:

```
  tolerations:
  - operator: Exists
```

We could even move the taint-specific content here into a separate *Taints* sub-section (under the same *NodeFeatureRule custom resource* section)

WDYT?

Also, we should describe the new `taints` field in the *Label rule format* section below.

pkg/apis/nfd/v1alpha1/expression.go
```diff
@@ -30,17 +30,17 @@ import (
...
+	MatchIn:           {},
+	MatchNotIn:        {},
+	MatchInRegexp:     {},
+	MatchExists:       {},
+	MatchDoesNotExist: {},
+	MatchGt:           {},
+	MatchLt:           {},
+	MatchGtLt:         {},
+	MatchIsTrue:       {},
+	MatchIsFalse:      {},
```
marquiz, line 43: nit: again, all changes in this file make sense but are unrelated (and could be just submitted as a separate PR for speeding up reviews)

pkg/nfd-master/nfd-master-internal_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	"github.com/stretchr/testify/mock"
 	"github.com/vektra/errors"
 	"golang.org/x/net/context"
-	api "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
```
marquiz, line 30: nit: I'd suggest to submit these "optional" unrelated changes as a separate PR

marquiz (commented):
Thanks @fmuyassarov! I took a first quick pass (incomplete, not even going through all the changes, yet) and this looks really good already.

One suggestion/ask: could you submit the package renames

```diff
- whatever "k8s.io/api/core/v1"
+ corev1 "k8s.io/api/core/v1"
```

and other possible "beautifying" as separate PR(s). This will greatly reduce the "noise" in this PR and make review easier

marquiz:
BTW, now that #848 is in, we should also add some e2e coverage. But that can be done in a separate PR (and tracked in a separate issue), imo

fmuyassarov:
> BTW, now that #848 is in, we should also add some e2e coverage. But that can be done in a separate PR (and tracked in a separate issue), imo

Yep, I'm on it now :blush:

pkg/apis/nfd/v1alpha1/expression.go
fmuyassarov: yeah I agree with you. Will decouple those changes.

pkg/nfd-master/nfd-master.go
```diff
@@ -417,6 +421,85 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
 			klog.Errorf("failed to advertise labels: %v", err)
 			return &pb.SetLabelsReply{}, err
 		}
+		if !m.args.NoTaint {
+
+			// ensure we delete the old leftover taints
+			labelReply, err := m.deleteTaints(c, r)
+			if err != nil {
+				return labelReply, err
+			}
```
marquiz, line 430: Hey, just popped into my mind: maybe we should remove the old nfd-generated taints irrespective of the `--no-taint` setting. E.g. if one tries this (taints) out but then decides to switch the tainting off, this would ensure that old stale taints will be cleaned up.

WDYT?

docs/advanced/customization-guide.md
```diff
@@ -47,6 +50,14 @@ spec:
...
       labels:
         "my-sample-feature": "true"
+# Using tainting feature requires the feature gate (--no-taint) to be enabled (set to false)
+      taints:
+        - effect: PreferNoSchedule
+          key: bar
+          value: true
+        - effect: NoExecute
+          key: foo
+          value: true
```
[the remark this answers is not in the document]
fmuyassarov: yep, I will updated it.

docs/advanced/customization-guide.md
fmuyassarov: Good point about eviction, I actually ran into the same situation with Tilt while re-deploying the NFD. I will document them as per suggestion.

pkg/nfd-master/nfd-master.go
fmuyassarov: Ow, very good catch @marquiz. I totally missed that scenario. Yes, I definitely agree that we should take care of cleaning the taints  in that situation too. Thanks a lot. I will fix it.

docs/advanced/customization-guide.md
marquiz: Another thought: if one would add a generic toleration I suggested above (` tolerations: [operator: Exists]`) that will mean that nfd-worker would likely be scheduled on K8s control-plane node(s) too. And that in turn, might cause control plane node(s) to be tainted and nfd-master be evicted 🫤 So, if that kind of über-toleration would be used that should be set to nfd-master, too.

fmuyassarov: I've updated documentation commit, basically took it into a separate sub-section and added more info there. As for the toleration discussion, I added a note that user will need to add corresponding toleration on the Pod/DaemonSet. Should we add a note about generic toleration or leave it?

fmuyassarov:
@marquiz thank you for the feedback. Updated commits as following
- took out the unrelated changes
- docs update
- deletion function call update
PTAL.

fmuyassarov:
One thing I didn't like is the naming of the flag I have chosen. Because the to enable it we need to set the false and true to disable, which seems confusing. Maybe I should rename it to something like `--enable-taints`, and false means disabled while true means enabled.

marquiz:
> One thing I didn't like is the naming of the flag I have chosen. Because the to enable it we need to set the false and true to disable, which seems confusing. Maybe I should rename it to something like `--enable-taints`, and false means disabled while true means enabled.

I agree on this. Especially as this won't be enabled by default any time soon (I think). I'd rename this to `--enable-taints` (it is prolly better than just `--taints`)

docs/advanced/customization-guide.md
```diff
@@ -74,6 +74,57 @@ Now, on X86 platforms the feature label appears after doing `modprobe dummy` on
...
+  name: my-sample-rule-object
+spec:
+  rules:
+    - name: "my sample taint rule"
+      taints:
+        - effect: NoExecute
+          key: "feature.node.kubernetes.io/pci-0300_1d0f.present"
+          value: "true"
+        - effect: NoExecute
+          key: "feature.node.kubernetes.io/cpu-cpuid.ADX"
```
marquiz, line 115: Also this key looks a bit odd to me

```suggestion
        - effect: NoExecute
          key: "feature.node.kubernetes.io/dedicated-node"
```

fmuyassarov:
Updated the flag name.

docs/advanced/customization-guide.md
```diff
@@ -74,6 +74,57 @@ Now, on X86 platforms the feature label appears after doing `modprobe dummy` on
...
+one or more custom taints via the `taints` field of the NodeFeatureRule CR. The
+same rule-based mechanism is applied here and the NFD taints only rule matching nodes.
+
+*taints* is a list of taint entries and each entry can have `key`, `value` and `effect`,
+where the `value` is optional. Effect could be `NoSchedule`, `PreferNoSchedule`
+or `NoExecute`. To learn more about the meaning of these effects, check out k8s [documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/).
+
+To enable the tainting feature, `--enable-taints` flag needs to be set to `true`.
+If the flag `--enable-taints` is set to `false` (i.e. disabled), taints defined in
+the NodeFeatureRule CR have no effect and will be ignored by the NFD master.
```
marquiz, line 93: I think we should document the `taints` field in a separate subsection under [Fields](https://kubernetes-sigs.github.io/node-feature-discovery/master/advanced/customization-guide.html#fields) like the other fields, too. I think this paragraph could go there. And we need to state there that the `taints` field is ONLY available for `NodeFeatureRule` objects, not for the `custom` rules of nfd-worker.

docs/advanced/customization-guide.md
```diff
@@ -74,6 +74,57 @@ Now, on X86 platforms the feature label appears after doing `modprobe dummy` on
 a system and correspondingly the label is removed after `rmmod dummy`. Note a
 re-labeling delay up to the sleep-interval of nfd-worker (1 minute by default).
 
+### Taints
```
marquiz, line 77: We could declare this as *EXPERIMENTAL*, dunno 🤔 This would keep the door better open for future adjustments if needed 😅 Any thoughts?

marquiz (commented):
Some comments on the documentation side. Still need to go through the code...

docs/advanced/customization-guide.md
```diff
@@ -74,6 +74,57 @@ Now, on X86 platforms the feature label appears after doing `modprobe dummy` on
...
+and add corresponding toleration to pods that require the special hardware. NFD
+offers node tainting functionality which is disabled by default. User can define
+one or more custom taints via the `taints` field of the NodeFeatureRule CR. The
+same rule-based mechanism is applied here and the NFD taints only rule matching nodes.
+
+*taints* is a list of taint entries and each entry can have `key`, `value` and `effect`,
+where the `value` is optional. Effect could be `NoSchedule`, `PreferNoSchedule`
+or `NoExecute`. To learn more about the meaning of these effects, check out k8s [documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/).
+
+To enable the tainting feature, `--no-taint` flag needs to be set to `false`.
```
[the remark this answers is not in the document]
fmuyassarov: good catch !

docs/advanced/customization-guide.md
fmuyassarov: +1

docs/advanced/customization-guide.md
fmuyassarov: Ah sorry I misunderstood you earlier. Fixed now. Thanks

fmuyassarov:
Rebased

pkg/nfd-master/nfd-master.go
```diff
@@ -753,3 +844,70 @@ func (m *nfdMaster) instanceAnnotation(name string) string {
 	}
 	return m.args.Instance + "." + name
 }
+
+// deleteTaints deletes node taint(s) and taints annotation
+func (m *nfdMaster) deleteTaints(c context.Context, r *pb.SetLabelsRequest) (*pb.SetLabelsReply, error) {
```
marquiz, line 849: nodename should be enough

```suggestion
func (m *nfdMaster) deleteTaints(c context.Context, nodeName string) (*pb.SetLabelsReply, error) {
```

pkg/nfd-master/nfd-master.go
```diff
@@ -432,6 +436,85 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+
+		if m.args.EnableTaints {
+			// Set NodeFeatureRule provided taints
+			if mustTaint {
+				labelReply, err := m.setTaints(c, r)
+				if err != nil {
+					return labelReply, err
+				}
+			}
+		}
```
marquiz, line 454: Hmm, we need to think a bit more this taint handling 🧐 It would be nice to handle taints by patching in `updateNodeFeatures()` similarly to other changes we do on the Node object. I just think that patching isn't an option in this case as taints are an array and we cannot unambiguously and reliably specify elements that we want to delete (a classic race: the node object and the list of taints might've changed between we read the object and send the patch).

So, I think we're stuck with updating the object. But with this. two node updates is an overkill (first delete and then add) plus there is a small time window when nfd taints do not exist(!) So I think a separate one function to handle this would be needed. 

1. get node
2. see if taints need to be updated
3. if not, return
4. if an update is needed, update the taint array (delete, modify, add)
5. update annotations
6. update node (only if smth was changed)

That function could have taints as an arg so that the function could be called with empty taints if `m.args.EnableTaints==true`, deleting anything managed by NFD.

Also, doing node updates we re-introduce this race: #122. I think some re-try mechanism (of few tries) would be needed to mitigate that.

Thoughts?

BTW, we also need to delete taints in `prune()`, too

pkg/nfd-master/nfd-master.go
```diff
@@ -432,6 +436,85 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+		return &pb.SetLabelsReply{}, err
+	}
+
+	// fetch the node object
+	node, err := m.apihelper.GetNode(cli, r.NodeName)
+	if err != nil {
+		return &pb.SetLabelsReply{}, err
+	}
+
+	if !m.args.NoPublish {
```
marquiz, line 488: nit: I don't think we need to check this here as it's done already upper in the call chain (just makes the code a bit simpler)

pkg/nfd-master/nfd-master.go
```diff
@@ -432,6 +436,85 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+					return labelReply, err
+				}
+			}
+		}
+	}
+	return &pb.SetLabelsReply{}, nil
+}
+
+// setTaints sets taint(s) on the node matching rule(s) in the NodeFeatureRule CRD
+func (m *nfdMaster) setTaints(c context.Context, r *pb.SetLabelsRequest) (*pb.SetLabelsReply, error) {
```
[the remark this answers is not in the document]
fmuyassarov: Good point, thanks

docs/advanced/customization-guide.md
```diff
@@ -364,6 +364,62 @@ details.
...
+or `NoExecute`. To learn more about the meaning of these effects, check out k8s [documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/).
+
+To enable the tainting feature, `--enable-taints` flag needs to be set to `true`.
+If the flag `--enable-taints` is set to `false` (i.e. disabled), taints defined in
+the NodeFeatureRule CR have no effect and will be ignored by the NFD master.
+
+**NOTE**: After NFD sets a taint, already running pods that do not tolerate the taint
+are evicted immediately from the node including the nfd-worker pod. To allow scheduling
+nfd-worker pods onto the node where they were running before, customization of the
+nfd-worker daemonset with corresponding tolerations is required.
```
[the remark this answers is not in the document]
fmuyassarov: Sure

docs/advanced/customization-guide.md
```diff
@@ -364,6 +364,62 @@ details.
...
+offers node tainting functionality which is disabled by default. User can define
+one or more custom taints via the `taints` field of the NodeFeatureRule CR. The
+same rule-based mechanism is applied here and the NFD taints only rule matching nodes.
+
+*taints* is a list of taint entries and each entry can have `key`, `value` and `effect`,
+where the `value` is optional. Effect could be `NoSchedule`, `PreferNoSchedule`
+or `NoExecute`. To learn more about the meaning of these effects, check out k8s [documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/).
+
+To enable the tainting feature, `--enable-taints` flag needs to be set to `true`.
+If the flag `--enable-taints` is set to `false` (i.e. disabled), taints defined in
```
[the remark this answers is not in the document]
fmuyassarov: ops, good catch

pkg/nfd-master/nfd-master.go
fmuyassarov: that's right, fixing it. Thanks

pkg/nfd-master/nfd-master.go
fmuyassarov: Thanks for the feedback @marquiz . I've refactored the code on my latest commit https://github.com/kubernetes-sigs/node-feature-discovery/pull/910/commits/6331eadb96c8bf45e60e62b1f66bed295a49a84e as per suggestion. Please have a look, if looks good I will squash the commit with prior once.

pkg/apis/nfd/v1alpha1/rule.go
```diff
@@ -97,7 +101,7 @@ func (r *Rule) Execute(features Features) (RuleOutput, error) {
 	ret := RuleOutput{Labels: labels, Vars: vars}
 	utils.KlogDump(2, fmt.Sprintf("rule %q matched with: ", r.Name), "  ", ret)
 
-	return ret, nil
+	return anyMatched, featureMatched, ret, nil
```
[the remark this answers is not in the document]
fmuyassarov: Hmm, that will require `RuleOutput` struct to be updated for something that will not be used. Perhaps we can hook a warning somewhere else instead of sending the taints and then ignoring it?

docs/advanced/customization-guide.md
```diff
@@ -364,6 +364,62 @@ details.
 labels specified in the `labels` field will override anything
 originating from `labelsTemplate`.
 
+### Taints
```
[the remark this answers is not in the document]
fmuyassarov: Got it, should be fixed now.

pkg/apis/nfd/v1alpha1/rule.go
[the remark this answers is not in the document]
marquiz: Iirc the idea behind ruleoutput was that it would contain all changes applied to the node (we'll likely have annotations in the future, too). The taints part would be used by nfd-master but not nfd-worker. Of course we could have something like a  `taints bool` arg on the Execute() function and then hook a warning somewhere else in nfd-worker code path, but imo that would feel just useless complication. Currently we don't print any warning on nfd-worker side (if taints are specified), I think.

fmuyassarov:
Rebased

pkg/apis/nfd/v1alpha1/rule.go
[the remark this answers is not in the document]
fmuyassarov: Do we really need to add this warning in the nfd-worker? I feel like, as long as the feature field doesn't exist in the worker config then user doesn't have a wrong assumption about availability of the feature. I mean if config doesn't allow me to set something then I assume it is not supported. WDYT?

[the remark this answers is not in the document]
marquiz: It might be useful if people mix up things. The thing is that the yaml/json unmarshalling (of the config) does not have any spec validation. Unknown fields are just silently ignored. People can put taints field and nfd-worker is happy with that,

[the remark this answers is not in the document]
fmuyassarov: You have a point, agreed. Added a warning in https://github.com/kubernetes-sigs/node-feature-discovery/pull/910/commits/d451e5db712ddd133393fa842b206b1a70a5aa5e.

fmuyassarov:
@marquiz I've addressed the comments. PTAL

marquiz:
> @marquiz I've addressed the comments. PTAL

I will as soon as I find a suitable time slot 😅 Thanks @fmuyassarov for the update

marquiz (commented):
Thanks for the update @fmuyassarov. Just a few comments based on a quick pass through. Looks better but there seems to be some logical error in the code

pkg/nfd-master/nfd-master.go
```diff
@@ -417,22 +428,145 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
 	}
-	for k, v := range m.crLabels(r) {
+	crLabels, mustTaint := m.crLabels(r)
+
+	for k, v := range crLabels {
 		rawLabels[k] = v
 	}
 
-	labels, extendedResources := filterFeatureLabels(rawLabels, m.args.ExtraLabelNs, m.args.LabelWhiteList.Regexp, m.args.ResourceLabels)
+	labelsw, extendedResources := filterFeatureLabels(rawLabels, m.args.ExtraLabelNs, m.args.LabelWhiteList.Regexp, m.args.ResourceLabels)
```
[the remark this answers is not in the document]
fmuyassarov: Ops, unintentional

docs/advanced/customization-guide.md
```diff
@@ -364,6 +385,43 @@ details.
...
+
+*taints* is a list of taint entries and each entry can have `key`, `value` and `effect`,
+where the `value` is optional. Effect could be `NoSchedule`, `PreferNoSchedule`
+or `NoExecute`. To learn more about the meaning of these effects, check out k8s [documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/).
+
+**NOTE** taints field is not available for the custom rules of nfd-worker and only
+for NodeFeatureRule objects.
+
+Example NodeFeatureRule with custom taints:
```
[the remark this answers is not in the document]
fmuyassarov: Yes definitely, during the refactoring and moving the code around I forgot to move it there.

pkg/nfd-master/nfd-master.go
```diff
@@ -417,22 +428,145 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+					crt = append(crt, rule.Taints...)
+				}
+			}
+
+			if mustTaint || len(crt) == 0 {
+				labelReply, err := m.setTaints(c, crt, r.NodeName)
+				if err != nil {
+					return labelReply, err
+				}
+			}
```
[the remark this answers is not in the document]
fmuyassarov: Good point, I fixed this and the above comment. PTAL

pkg/nfd-master/nfd-master.go
```diff
@@ -514,11 +636,14 @@ func (m *nfdMaster) crLabels(r *pb.SetLabelsRequest) map[string]string {
...
 		for _, rule := range spec.Spec.Rules {
-			ruleOut, err := rule.Execute(features)
+			featureMatched, anyMatched, ruleOut, err := rule.Execute(features)
 			if err != nil {
 				klog.Errorf("failed to process Rule %q: %v", rule.Name, err)
 				continue
 			}
+			if featureMatched || anyMatched {
+				crt = append(crt, rule.Taints...)
+			}
```
marquiz, line 646: We shouldn't need these `featureMatched` and `anyMatched` arguments at all. We should have all taints in `ruleOut`, right (i.e. just unconditional `crt = append(crt, ruleOut.Taints`)?

fmuyassarov: That's right. I just realized that I misunderstood the matched/unmatched rule logic but thanks to your comment I understand it now. Please have a look once more.

pkg/nfd-master/nfd-master.go
```diff
@@ -432,10 +446,116 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+
+	err := authorizeClient(c, m.args.VerifyNodeName, nodeName)
+	if err != nil {
+		return &pb.SetLabelsReply{}, err
+	}
+
+	cli, err := m.apihelper.GetClient()
+	if err != nil {
+		return &pb.SetLabelsReply{}, err
+	}
```
marquiz, line 486: I think we could just take `cli` as an argument to this method and drop all this initialization

```suggestion
```

pkg/nfd-master/nfd-master.go
```diff
@@ -432,10 +446,116 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+			}
+		}
 	}
 	return &pb.SetLabelsReply{}, nil
 }
 
+// setTaints sets note taints and annotations based on the taints passed via
+// nodeFeatureRule custom resorce. If empty list of taints is passed, currently
+// NFD owned taints and annotations are removed from the node.
+func (m *nfdMaster) setTaints(c context.Context, taints []corev1.Taint, nodeName string) (*pb.SetLabelsReply, error) {
```
marquiz, line 473: Just returning an error should be enough.

```suggestion
func (m *nfdMaster) setTaints(c context.Context, taints []corev1.Taint, nodeName string) error {
```

pkg/nfd-master/nfd-master.go
```diff
@@ -432,10 +446,116 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+			taintStrs = append(taintStrs, taint.ToString())
+			val = strings.Join(taintStrs, ",")
+		}
+		annotation[nfdv1alpha1.NodeTaintsAnnotation] = val
+	case len(taints) == 0:
+		delete(annotation, nfdv1alpha1.NodeTaintsAnnotation)
+	}
+	node.SetAnnotations(annotation)
+
+	// delete stail taints
```
marquiz, line 509: stale?
```suggestion
	// delete stale taints
```

pkg/nfd-master/nfd-master.go
```diff
@@ -432,10 +446,116 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+	}
+
+	// fetch the node object
+	node, err := m.apihelper.GetNode(cli, nodeName)
+	if err != nil {
+		return &pb.SetLabelsReply{}, err
+	}
+
+	// update the taint annotation
+	annotation := node.GetAnnotations()
```
marquiz, line 495: super nit: many annotations in this var so rename to `annotations`(?)


```suggestion
	annotations := node.GetAnnotations()
```

pkg/nfd-master/nfd-master.go
```diff
@@ -432,10 +446,116 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+	// update the node taints
+	if len(taints) > 0 {
+		for _, taint := range taints {
+			if !taintutils.TaintExists(node.Spec.Taints, &taint) {
+				nodeTaints = append(nodeTaints, taint)
+			}
+			continue
+		}
+	}
+	node.Spec.Taints = append(node.Spec.Taints, nodeTaints...)
```
marquiz, line 545: This whole logic looks odd and partly broken. E.g. serializing the new taints into a string and the de-serializing back to see what we want to add looks a bit strange 🧐

What we should do imo is smth like:
1. Get the diff between old and new **nfd managed** taints, basically what was *deleted*
2. Go through the list of *deleted* and remove those from the node
3. Go through the list of all *new* taints and add them to the node if they don't exist

marquiz (changes requested):
Now a more thorough look at the core changes in nfd-master. Some work still there to do but I think we're approaching the finish line 😅

pkg/nfd-master/nfd-master.go
```diff
@@ -485,9 +605,11 @@ func (m *nfdMaster) UpdateNodeTopology(c context.Context, r *topologypb.NodeTopo
 	return &topologypb.NodeTopologyResponse{}, nil
 }
 
-func (m *nfdMaster) crLabels(r *pb.SetLabelsRequest) map[string]string {
+func (m *nfdMaster) crLabels(r *pb.SetLabelsRequest) (map[string]string, []corev1.Taint) {
```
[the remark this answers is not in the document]
fmuyassarov: I agree, I wanted to change this name when submitted the patch but didn't had chance to come back to this point. So, `processNodeFeatureRule` sounds good to me

pkg/nfd-master/nfd-master.go
fmuyassarov: > 3\. Go through the list of all _new_ tain

The reason for converting new taints into string is that we want to update taint annotation based on the latest taints passed and annotation is in string.  And we need to first ensure that annotation is up to date and in sync with the CR taints because when deleting a taint from the node we rely on the annotation to know if we are allowed to touch that taint or not. I think the complexity comes from the fact that we can not just compare node taints and CR taints and remove based on the diff of two. There will be taints not owned by NFD as well which makes deletion a bit complicated.
This is how it looks like right now:
```
1. Update the annotation so that we can really trust that it is up to date source of truth
   1.1 convert CR taints in to string
   1.2 compare CR taints string vs node taint annotation to get the diff
   1.3 add/remove diff in annotation

2. Delete stale taints from the node
   2.1 convert updated annotation in to taints array
   2.2 compare taints array vs node taints to get the diff
   2.3 add/remove diff in node taint
```
We can't get diff (2.2) without 1.3. Since annotation is source of truth, the logic first tries to update it and the next steps rely on that updated annotation.

marquiz: > And we need to first ensure that annotation is up to date and in sync with the CR taints

Really? Couldn't you just loop over the old taints and use `taintutils.TaintExists` to see if each of those exist in the new set ot taints? You shouldn't need any serialization/de-serialization step in this.

> when deleting a taint from the node we rely on the annotation to know if we are allowed to touch that taint or not

Exactly! That's why we should get oldTaints from the annotation and get the diff to newTaints and see if any of those need to be deleted.

The sources of truth are:
1. The (old) annotation for *old* annotations
2. The `taints` arg (of the method) for the new annotations

Basically we just should covert the `taints` arg into the updated annotation once. (We might actually want to do some filtering there e.g. disallow any taints with a key with the `node.kubernetes.io/` prefix, for example)

>   3.2 compare taints array vs node taints to get the diff
   2.3 add/remove diff in node taint

I don't think this works. You should not remove anything that was not listed in the oldAnnotation (i.e. marked as "managed by nfd") in the first place

pkg/nfd-master/nfd-master.go
fmuyassarov: Yes it makes sense, but as discussed offline, let's do it in a follow up since `SetLabels()` doesn't have a client initialized and I will need to touch some unrelated funcs to this PR. I will add a follow up PR.

pkg/nfd-master/nfd-master.go
```diff
@@ -432,10 +446,116 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+			}
+			// Call setTaints() with empty list of taints even though the feature
+			// flag is disabled. This ensures that we delete NFD owned stale taints
+			// when flag got turned off.
+		case false:
+			labelReply, err := m.setTaints(c, emptyTaint, r.NodeName)
+			if err != nil {
+				return labelReply, err
+			}
+		}
```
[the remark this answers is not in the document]
fmuyassarov: Good point

fmuyassarov:
Rebased and addressed comments. @marquiz PTAL and let me know please if there more things we need to fix before merging. Thanks.

pkg/nfd-master/nfd-master.go
fmuyassarov: here is the separate PR: https://github.com/kubernetes-sigs/node-feature-discovery/pull/952

fmuyassarov: Rebased on top of https://github.com/kubernetes-sigs/node-feature-discovery/pull/952 and updated setTaints() to accept client argument.

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,114 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+		node.Spec.Taints = append(node.Spec.Taints, nodeTaints...)
+	}
+
+	// Serialize the new taints into string and update the annotation
+	// with that string.
+	taintStrs := []string{}
+	oldAnnotation := annotations[nfdv1alpha1.NodeTaintsAnnotation]
+	if len(taints) > 0 {
+		for _, taint := range taints {
```
marquiz, line 541: nit: declare variables where they are used
```suggestion
	oldAnnotation := annotations[nfdv1alpha1.NodeTaintsAnnotation]
	if len(taints) > 0 {
		taintStrs := []string{}
		for _, taint := range taints {
```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,114 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+			taintStrs = append(taintStrs, taint.ToString())
+			val = strings.Join(taintStrs, ",")
+		}
+		annotations[nfdv1alpha1.NodeTaintsAnnotation] = val
```
marquiz, line 545: no need to do strings.join more than once
```suggestion
		}
		annotations[nfdv1alpha1.NodeTaintsAnnotation] = strings.Join(taintStrs, ",")
```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,114 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+				nodeTaints = append(nodeTaints, taint)
+			}
+			continue
+		}
+		node.Spec.Taints = append(node.Spec.Taints, nodeTaints...)
+	}
+
+	// Serialize the new taints into string and update the annotation
+	// with that string.
+	taintStrs := []string{}
```
marquiz, line 538: nit: you could initialize the slice to correct lenght (and then use `taintStrs[i]` instead of `append` in the for loop below)

```suggestion
	taintStrs := make([]string{}, len(taints))
```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,114 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+		}
+	}
+
+	// Add new taints.
+	var nodeTaints []corev1.Taint
+	if len(taints) > 0 {
+		for _, taint := range taints {
+			if !taintutils.TaintExists(node.Spec.Taints, &taint) {
+				nodeTaints = append(nodeTaints, taint)
+			}
```
marquiz, line 530: Hmm, I'm not sure this actually works as expected in all cases. E.g. what happens if the taint `value` is changed? I don't think we capture it correctly, i.e. we probably get `TaintExists==true` even though we should actually update the taint 🧐 Could you check this?

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,114 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+			val = strings.Join(taintStrs, ",")
+		}
+		annotations[nfdv1alpha1.NodeTaintsAnnotation] = val
+	} else {
+		delete(annotations, nfdv1alpha1.NodeTaintsAnnotation)
+	}
+	node.SetAnnotations(annotations)
+
+	// Check if node requires an update.
+	if oldAnnotation != annotations[nfdv1alpha1.NodeTaintsAnnotation] {
```
marquiz, line 552: It would probably be safer for corner cases to not rely on the annotation values alone but actually track in a boolean variable if something was changed in the node object (with the deletion and addition of taints). Imagine a situation that I go and manually delete an NFD-managed taint. Annotations are not changed at all but I'd assume that NFD would re-create that taint nevertheless. WDYT?

marquiz (commented):
Thanks for the update @fmuyassarov. Looks a lot cleaner now. I still came up with a few nits and comments regarding different corner cases.

pkg/nfd-master/nfd-master.go
marquiz: I took a quick look at the `taintutils` package and `AddOrUpdateTaint()` looks like something we might want to use (or mimic)

fmuyassarov: both `AddOrUpdateTaint` and `TaintExists` are using [MatchTaint](https://github.com/kubernetes/kubernetes/blob/2b7b5245eaff4ed60d84df67e4351b04ad9d45ae/staging/src/k8s.io/api/core/v1/taint.go#L23) under the hood to compare taints based on `key:effect`. So, even if we use AddOrUpdateTaint(), it is still going to be the same result for the scenario you mentioned.

marquiz: > So, even if we use AddOrUpdateTaint, it is still going to be the same result for the scenario you mentioned.

Actually, it's not. Take a closer look:
https://github.com/kubernetes/kubernetes/blob/2b7b5245eaff4ed60d84df67e4351b04ad9d45ae/pkg/util/taints/taints.go#L196

It uses `DeepEqual()` to if the taints match fully.

fmuyassarov: ow missed:)

fmuyassarov:
Thanks for the feedback @marquiz . Please take a look on the updated version.

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +452,120 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+		if err != nil {
+			return err
+		}
+		for _, ct := range corev1Taints {
+			oldTaints = append(oldTaints,
+				corev1.Taint{
+					Key:    ct.Key,
+					Value:  ct.Value,
+					Effect: ct.Effect})
+		}
```
marquiz, line 498: Wny do we copy the data, couldn't we use return value of `taintutils.ParseTaints()` directly?

pkg/nfd-master/nfd-master.go
```diff
@@ -294,6 +299,12 @@ func (m *nfdMaster) prune() error {
 			return fmt.Errorf("failed to prune labels from node %q: %v", node.Name, err)
 		}
 
+		// Prune taints
+		err = m.setTaints(c, cli, crt, node.Name)
```
marquiz, line 303: No need to declare any variables above

```suggestion
		err = m.setTaints(context.TODO(), cli, []corev1.Taint{}, node.Name)
```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +452,120 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+	// we first check if the taints still exist on the node and if yes
+	// remove it, otherwise move on.
+	for _, taintToRemove := range taintsToRemove {
+		if taintutils.TaintExists(node.Spec.Taints, &taintToRemove) {
+			newTaints, removed := taintutils.DeleteTaint(node.Spec.Taints, &taintToRemove)
+			if !removed {
+				klog.V(1).Infof("taint %q already deleted from node", taintToRemove.ToString())
+			}
+			node.Spec.Taints = newTaints
+		}
```
marquiz, line 521: Still, I don't think you need to pre-check the existence of the taint (moreover, you will never hit the log message below if you pre-check because `removed` will never be false)

```suggestion
		newTaints, removed := taintutils.DeleteTaint(node.Spec.Taints, &taintToRemove)
		if !removed {
			klog.V(1).Infof("taint %q already deleted from node", taintToRemove.ToString())
		}
		node.Spec.Taints = newTaints
```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +452,120 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+	node.SetAnnotations(annotations)
+
+	// Update the node only if there is stale tains to remove or new tains to add.
+	if len(taintsToRemove) > 0 || len(taints) > 0 {
+		klog.Infof("received tainting request for node %q", nodeName)
+		err = m.apihelper.UpdateNode(cli, node)
+		if err != nil {
+			return fmt.Errorf("failed to taint the node %q: %v", node.Name, err)
+		}
+	}
```
marquiz, line 565: Now that `PatchNodeTaints()` handles update of the node taints we only need to update the annotation here. We can simply compare the old and new annotation and patch node if a change is needed. We should not update e.g. taints anymore in this phase

In pseudo-code something like this

```
if oldAnnotation != newAnnotation {
    createJsonPatchForAnnotation()
    patchNode()
}

```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +452,120 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+	// We have taintsToRemove that we found from the annotation and here
+	// we first check if the taints still exist on the node and if yes
+	// remove it, otherwise move on.
+	for _, taintToRemove := range taintsToRemove {
+		if taintutils.TaintExists(node.Spec.Taints, &taintToRemove) {
+			newTaints, removed := taintutils.DeleteTaint(node.Spec.Taints, &taintToRemove)
+			if !removed {
+				klog.V(1).Infof("taint %q already deleted from node", taintToRemove.ToString())
+			}
+			node.Spec.Taints = newTaints
```
marquiz, line 520: I think we should start tracking changes in `newNode` already here. `node` would be the original node object and any changes we do in taints would be tracked in `newNode` that would then be given to `controller.PatchNodeTaints()` to update everything (i.e. all taints-related changes) in one go.

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +452,120 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+			newNode = curNewNode
+			oldNodeCopy = curNewNode
+			continue
+		}
+		err = controller.PatchNodeTaints(c, cli, nodeName, oldNode, newNode)
+		if err != nil {
+			return fmt.Errorf("failed to patch the node %v", node.Name)
+		}
+
+	}
```
marquiz, line 543: Good catch finding `controller.PatchNodeTaints`, looks handy and more reliable than trying to update the whole node object 👍

But AFAIU, we should delegate all taints update to this function, i.e. also just deleting old taints. No need to check for `if len(taints) > 0` and also `newNode` should've been declared already eariler.

```suggestion
	for _, taint := range taints {
		newNode, _, err := taintutils.AddOrUpdateTaint(oldNodeCopy, &taint)
		if err != nil {
			return fmt.Errorf("failed to add %q taint on node %v", taint, node.Name)
		}
	}
	err = controller.PatchNodeTaints(c, cli, nodeName, oldNode, newNode)
	if err != nil {
		return fmt.Errorf("failed to patch the node %v", node.Name)
	}
```

pkg/nfd-master/nfd-master.go
```diff
@@ -493,20 +615,21 @@ func (m *nfdMaster) UpdateNodeTopology(c context.Context, r *topologypb.NodeTopo
...
 
-func (m *nfdMaster) crLabels(r *pb.SetLabelsRequest) map[string]string {
+func (m *nfdMaster) processNodeFeatureRule(r *pb.SetLabelsRequest) (map[string]string, []corev1.Taint) {
 	if m.nfdController == nil {
-		return nil
+		return nil, nil
 	}
 
 	l := make(map[string]string)
+	var crt []corev1.Taint
```
marquiz, line 624: nit: just caught my eye now that the name `crt` looks a bit odd here (I'm not sure `l` is a good name either) 😊 How about `taints` or `t`?

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +452,120 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+			return &pb.SetLabelsReply{}, err
+		}
 	}
 	return &pb.SetLabelsReply{}, nil
 }
 
+// setTaints sets note taints and annotations based on the taints passed via
+// nodeFeatureRule custom resorce. If empty list of taints is passed, currently
+// NFD owned taints and annotations are removed from the node.
+func (m *nfdMaster) setTaints(c context.Context, cli *kubernetes.Clientset, taints []corev1.Taint, nodeName string) error {
```
marquiz, line 476: We haven't been using any context in other parts of the code where interacting with the k8s apiserver (see e.g. `pkg/apihelper/k8shelpers.go`) so I think it would be more consistent to drop it from here as well and use `context.TODO()` when calling `controller.PatchNodeTaints()` below. WDYT?

```suggestion
func (m *nfdMaster) setTaints(cli *kubernetes.Clientset, taints []corev1.Taint, nodeName string) error {
```

marquiz (commented):
Thanks for the update @fmuyassarov! Getting closer on each iteration 😅 I had a few more comments - I think there's still room to make this simpler, cleaner and more robust.

pkg/nfd-master/nfd-master.go
fmuyassarov: Ah right 🤦‍♂️

pkg/nfd-master/nfd-master.go
fmuyassarov: fine with me, but just a heads up that we have it at least in `SetLabels`.

pkg/nfd-master/nfd-master.go
fmuyassarov: let's use `taints` and I will update `l` to `labels`

pkg/nfd-master/nfd-master.go
fmuyassarov: > But AFAIU, we should delegate all taints update to this function, i.e. also just deleting old taints. No need to check for `if len(taints) > 0` and also `newNode` should've been declared already eariler.
Suggested changes makes sense. We can even make it a bit shorter like this...
```
	if len(taints) > 0 {
		for _, taint := range taints {
			curNewNode, _, err := taintutils.AddOrUpdateTaint(newNode, &taint)
			if err != nil {
				return fmt.Errorf("failed to add %q taint on node %v", taint, node.Name)
			}
			newNode = curNewNode
			continue
		}
		err = controller.PatchNodeTaints(context.TODO(), cli, nodeName, node, newNode)
		if err != nil {
			return fmt.Errorf("failed to patch the node %v", node.Name)
		}
	}
```
Basically, `oldNodeCopy` `oldNode` variables are not needed. Regarding  `if len(taints) > 0`, I think we will hit nil pointer if we don't use it.

marquiz: > Regarding if len(taints) > 0, I think we will hit nil pointer if we don't use it.

Why would that be? We don't have any pointers. If `taints==nil` the for loop just short-circuits and we don't try to reference `&taint`.

pkg/nfd-master/nfd-master.go
marquiz: Yeah, that comes from the gRPC server interface but we don't actually use the context for anything 🙄 Maybe we could do that but I think that would probably be better to do in one PR consistently accross all related functions

pkg/nfd-master/nfd-master.go
fmuyassarov: sorry I meant in the tests https://github.com/kubernetes-sigs/node-feature-discovery/blob/0ec98a57d2bca625ee949950a14aec64b5fa5d79/pkg/nfd-master/nfd-master-internal_test.go#L333

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,110 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+		return err
+	}
+
+	// De-serialize the taints annotation into Taint type for comparision below.
+	annotations := node.GetAnnotations()
+	// Keep the backup of old annotation.
+	oldAnnotation := make(map[string]string)
+	for key, value := range annotations {
+		oldAnnotation[key] = value
+	}
```
marquiz, line 488: I don't think we need to clone the whole set of annotations. Just save the one that we're interested in
```suggestion
	oldAnnotation := annotations[nfdv1alpha1.NodeTaintsAnnotation]
```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,110 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+	oldTaints := []corev1.Taint{}
+	if val, ok := annotations[nfdv1alpha1.NodeTaintsAnnotation]; ok {
+		sts := strings.Split(val, ",")
```
marquiz, line 491: ... and then we can simplify this pme

```suggestion
	if oldAnnotation != "" {
		sts := strings.Split(oldAnnotation, ",")
```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,110 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+	// we first check if the taints still exist on the node and if yes
+	// remove it, otherwise move on.
+	newNode := node
+	for _, taintToRemove := range taintsToRemove {
+		newTaints, removed := taintutils.DeleteTaint(node.Spec.Taints, &taintToRemove)
+		if !removed {
+			klog.V(1).Infof("taint %q already deleted from node", taintToRemove.ToString())
+		}
+		newNode.Spec.Taints = newTaints
+	}
```
marquiz, line 519: Hmm, maybe we could simplify and combine these two for loops

```suggestion
	newNode := node
	for _, taint := range oldTaints {
		if !taintutils.TaintExists(taints, &taint) {
			newTaints, removed := taintutils.DeleteTaint(node.Spec.Taints, &taint)
			if !removed {
				klog.V(1).Infof("taint %q already deleted from node", taintToRemove.ToString())
			}
			newNode.Spec.Taints = newTaints
		}
	}
```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,110 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+
+	if len(taints) > 0 {
```
marquiz, line 521: As we discussed, this length check needs to go away 😊 Otherwise `PatchNodeTaints()` doesn't get called with an empty list of taints

fmuyassarov: Sorry could not get back to this patch. But I will update my patch with fixes on your comments by the end of today.

ArangoGutierrez: PING

examples/nodefeaturerule.yaml
```diff
@@ -7,10 +7,18 @@ spec:
...
+          key: foo
+          value: "true"
       matchFeatures:
         - feature: kernel.loadedmodule
           matchExpressions:
             dummy: {op: Exists}
         - feature: kernel.config
           matchExpressions:
-            X86: {op: In, value: ["y"]}
+            X86: {op: In, value: ["y"]}
```
ArangoGutierrez, line 24: nit: no new line

docs/usage/customization-guide.md
```diff
@@ -74,6 +74,55 @@ Now, on X86 platforms the feature label appears after doing `modprobe dummy` on
 a system and correspondingly the label is removed after `rmmod dummy`. Note a
 re-labeling delay up to the sleep-interval of nfd-worker (1 minute by default).
 
+### NodeFeatureRule tainting feature
+
+This feature is experimental.
+
+In some circumstances it is desirable keep nodes with specialized hardware away from
+running general workload and instead leave them for pods that need the specialized
```
ArangoGutierrez, line 82:
```suggestion
running general workload and instead leave them for workloads that need the specialized
```
nit

docs/usage/customization-guide.md
```diff
@@ -74,6 +74,55 @@ Now, on X86 platforms the feature label appears after doing `modprobe dummy` on
...
+and add corresponding toleration to pods that require the special hardware. NFD
+offers node tainting functionality which is disabled by default. User can define
+one or more custom taints via the `taints` field of the NodeFeatureRule CR. The
+same rule-based mechanism is applied here and the NFD taints only rule matching nodes.
+
+To enable the tainting feature, `--enable-taints` flag needs to be set to `true`.
+If the flag `--enable-taints` is set to `false` (i.e. disabled), taints defined in
+the NodeFeatureRule CR have no effect and will be ignored by the NFD master.
+
+**NOTE**: After NFD sets a taint, already running pods that do not tolerate the taint
```
ArangoGutierrez, line 93: I would re-write this to be
" Before enabling the any taints, make sure you edit nfd-worker daemonset to tolerate the taints to be created, otherwise......" 
In documentation is better to recommend what to do in preparation for, not how to fix something after. (In my opinion)

docs/reference/master-commandline-reference.md
```diff
@@ -99,6 +99,23 @@ Example:
 nfd-master -cert-file=/opt/nfd/master.crt -key-file=/opt/nfd/master.key -ca-file=/opt/nfd/ca.crt
 ```
 
+### -enable-taints
+
+The `-enable-taints` flag enables/disables node tainting feature of NFD.
+
+Default: *false*
+
+NOTE: After NFD sets a taint, already running pods that do not tolerate the taint
```
ArangoGutierrez, line 108: I would re-write this to be
" Before enabling the any taints, make sure you edit nfd-worker daemonset to tolerate the taints to be created, otherwise......"
In documentation is better to recommend what to do in preparation for, not how to fix something after. (In my opinion)

fmuyassarov: Good point. Done.

docs/usage/customization-guide.md
fmuyassarov: Good point. Done.

docs/usage/customization-guide.md
fmuyassarov: Done.

pkg/nfd-master/nfd-master.go
fmuyassarov: finally got chance to finalize it. Updated as per suggestion

pkg/nfd-master/nfd-master.go
fmuyassarov: yep, done

pkg/nfd-master/nfd-master.go
fmuyassarov: done

pkg/nfd-master/nfd-master.go
fmuyassarov: I've modified the code, so we don't clone it at all now.

fmuyassarov:
@marquiz @ArangoGutierrez I've addressed latest comments, thanks a lot for your time on the review.
I've previously used patching for updating taints and annotations. For some mysterious reason, the logic didn't work, either only one taint gets updated or annotations are not patched properly, or annotations are updated but taints are not removed... 🤯. We even debugged it once with @marquiz and I thought we found the problem (with `if`) but then there were more :). So, at the end what we have now is we update taints, then annotations and then update the Node object once. I think it is better then sending patch requests to the API server and at least it works fine than with patching😂
Please have a look, if looks good to you, I will squash the last commit because that one is for addressing comments.

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,111 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+		// with that string.
+		taintStrs := make([]string, 0, len(taints))
+		for _, taint := range taints {
+			taintStrs = append(taintStrs, taint.ToString())
+		}
+		newAnnotations[nfdv1alpha1.NodeTaintsAnnotation] = strings.Join(taintStrs, ",")
+	}
+
+	if !(node.Annotations[nfdv1alpha1.NodeTaintsAnnotation] == newAnnotations[nfdv1alpha1.NodeTaintsAnnotation]) {
+		patches := createPatches([]string{nfdv1alpha1.NodeTaintsAnnotation}, node.Annotations, newAnnotations, "/metadata/annotations")
```
marquiz, line 547: This is really subtle nit, but looking at the values does not do exactly what we want in the case that the old annotation would be an empty string (shouldn't really happen but nevertheless).

```suggestion
	patches := createPatches([]string{nfdv1alpha1.NodeTaintsAnnotation}, node.Annotations, newAnnotations, "/metadata/annotations")
	if len(patches) > 0 {
```

pkg/nfd-master/nfd-master.go
```diff
@@ -440,10 +451,111 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
...
+	// We have taintsToRemove that we found from the annotation and here
+	// we first check if the taints still exist on the node and if yes
+	// remove it, otherwise move on.
+	taintsUpdated := false
+	newNode := node.DeepCopy()
+	for _, taintToRemove := range taintsToRemove {
+		newTaints, removed := taintutils.DeleteTaint(newNode.Spec.Taints, &taintToRemove)
+		if !removed {
+			klog.V(1).Infof("taint %q already deleted from node", taintToRemove.ToString())
+		}
```
marquiz, line 511: I think this could be simplified a bit. WDYT?

```suggestion
	// Delete old nfd-managed taints that are not found in the set of new taints.
	taintsUpdated := false
	newNode := node.DeepCopy()
	for _, taint := range oldTaints {
		if taintutils.TaintExists(taints, &taint) {
			continue
		}

		newTaints, removed := taintutils.DeleteTaint(newNode.Spec.Taints, &taint)
		if !removed {
			klog.V(1).Infof("taint %q already deleted from node", taint.ToString())
		}
```

marquiz (commented):
Two small comments but otherwise looks ready to me

pkg/nfd-master/nfd-master.go
fmuyassarov: makes sense indeed, thanks.

pkg/nfd-master/nfd-master.go
fmuyassarov: agreed. Done.

marquiz (approved):
Great work @fmuyassarov on this! I'd be ready to merge this cool new feature 🚀

marquiz:
ping @ArangoGutierrez

marquiz:
@fmuyassarov please rebase
ping @ArangoGutierrez

fmuyassarov:
> @fmuyassarov please rebase ping @ArangoGutierrez

done

marquiz (approved):
Thx @fmuyassarov

ArangoGutierrez (approved):
/lgtm

PR #912, opened 2022-10-12
apis/nfd: migrate pkg/api/feature

marquiz:
Move the previously-protobuf-only internal "feature api" over to the
public "nfd api" package. This is in preparation for introducing a new
CRD API for communicating features.

This patch carries no functional changes. Just moving code around.

Split out from #903


marquiz:
Rebased

fmuyassarov (commented):
/lgtm

Issue #913, opened 2022-10-13
Invalid feature name 'cpu-cpuid.' warning is shown in nfd-worker pod logs


Chandan-Abhyankar:
**What happened**:
Invalid feature name 'cpu-cpuid.' warning is shown in nfd-worker pod logs
**What you expected to happen**:
Invalid feature name 'cpu-cpuid.' warning should not be shown in nfd-worker pod logs
**How to reproduce it (as minimally and precisely as possible)**:
Install necessary pre-requisites
e.g. git, make, podman-docker, go etc

Take the latest code from master branch
git clone https://github.com/kubernetes-sigs/node-feature-discovery.git

Edit the makefile (so it doesn't use the git shasum)
$ cd node-feature-discovery
$ vi Makefile and save changes
Replace
VERSION := $(shell git describe --tags --dirty --always)
with
VERSION ?= '1.0.0'

$ make image

Push image to your preferred registry and deploy it

$ make deploy
You should see the following:
  namespace/node-feature-discovery created
  customresourcedefinition.apiextensions.k8s.io/nodefeaturerules.nfd.k8s-sigs.io created
  serviceaccount/nfd-master created
  clusterrole.rbac.authorization.k8s.io/nfd-master created
  clusterrolebinding.rbac.authorization.k8s.io/nfd-master created
  configmap/nfd-worker-conf created
  service/nfd-master created
  deployment.apps/nfd-master created
  daemonset.apps/nfd-worker created

Confirm master and worker pods are running fine on node-feature-discovery namespace. [Cross check recently created image is used in nfd-master deployment and nfd-worker deamonset. If not then change image to what you have recently built]

$ kubectl get pods -n node-feature-discovery
NAME                                                  READY   STATUS      RESTARTS   AGE
nfd-master-845b5bb867-ql65p                           1/1     Running     0          88m
nfd-worker-nt855                                      1/1     Running     0          88m

Check pod log for nfd-worker
$ kubectl logs nfd-worker-nt855
I1013 06:26:33.193306       1 component.go:36] [core][Channel #1] Channel Connectivity change to READY
I1013 06:26:33.196659       1 local.go:115] starting hooks...
I1013 06:26:33.198069       1 nfd-worker.go:470] starting feature discovery...
[...]

PR #914, opened 2022-10-13
cpu: ignore unknown cpuid flags on non-x86

marquiz:
Avoid trying to create empty "cpu-cpuid." labels for cpuid flags that we don't have a description for.

### DIFF
--- a/source/cpu/cpuid_arm64.go
+++ b/source/cpu/cpuid_arm64.go
@@ -185,15 +185,15 @@ func getCpuidFlags() []string {
 	hwcap2 := uint64(C.gethwcap2())
 	for i := uint(0); i < 64; i++ {
 		key := uint64(1 << i)
-		val := flagNames_arm64[key]
-		if hwcap&key != 0 {
+		val, ok := flagNames_arm64[key]
+		if hwcap&key != 0 && ok {
 			r = append(r, val)
 		}
 	}
 	for i := uint(0); i < 64; i++ {
 		key := uint64(1 << i)
-		val := flag2Names_arm64[key]
-		if hwcap2&key != 0 {
+		val, ok := flag2Names_arm64[key]
+		if hwcap2&key != 0 && ok {
 			r = append(r, val)
 		}
 	}
--- a/source/cpu/cpuid_arm.go
+++ b/source/cpu/cpuid_arm.go
@@ -98,8 +98,8 @@ func getCpuidFlags() []string {
 	hwcap := uint64(C.gethwcap())
 	for i := uint(0); i < 64; i++ {
 		key := uint64(1 << i)
-		val := flagNames_arm[key]
-		if hwcap&key != 0 {
+		val, ok := flagNames_arm[key]
+		if hwcap&key != 0 && ok {
 			r = append(r, val)
 		}
 	}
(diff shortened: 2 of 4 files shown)


marquiz:
Not verified
/hold

marquiz:
ping @Chandan-Abhyankar @yselkowitz

yselkowitz:
1. this should be done for all architectures' code with manually curated lists (arm64, ppc64le, and s390x)
2. let's also update the cpu flags for POWER 9 and 10, Z 14/15/16, and ARM v8.latest.

prb112:
> 1. this should be done for all architectures' code with manually curated lists (arm64, ppc64le, and s390x)
> 
>     2. let's also update the cpu flags for POWER 9 and 10, z14/15/16, and ARM v8.latest.

I think the defensive fix proposed by @marquiz 'future' proofs the code against changes. I do like the referenced PR from Yaakov. It's a great catch/help. Chandan and I can try this out on a p10.

marquiz:
/retitle cpu: ignore unknown cpuid flags on non-x86

marquiz:
> 1. this should be done for all architectures' code with manually curated lists (arm64, ppc64le, and s390x)

Ach, good point @yselkowitz. I updated the PR to handle other non-x86 architectures.

PTAL @Chandan-Abhyankar @yselkowitz @prb112

jschintag:
Thank you @marquiz , i don't have time to test it on s390x right now and i will be away next week. However the changes look good to me. Since they are the same for all arches, i'm fine with it if they work on the other arches.

One question unrelated to this PR: When i reviewed the loop, where does the capacity from the string array come from?
```
	r := make([]string, 0, 20)
	hwcap := uint64(C.gethwcap())
	for i := uint(0); i < 64; i++ {
```
I mean we have 23 Flags defined for s390x and we check 64 bits. Is there a reason the array only has a cap of 20?

marquiz:
Thanks @jschintag, I'll then be happy with an ack from the arm and ppc guys 😊

> One question unrelated to this PR: When i reviewed the loop, where does the capacity from the string array come from?
> 
> ```
> 	r := make([]string, 0, 20)
> 	hwcap := uint64(C.gethwcap())
> 	for i := uint(0); i < 64; i++ {
> ```
> 
> I mean we have 23 Flags defined for s390x and we check 64 bits. Is there a reason the array only has a cap of 20?

Good question, I'm not the original author of this code. I think it's just a (harmless) leftover from some eariler development stage. The slice capacity will get increased if more than 20 flags are present. This could be fixed in a separate PR though, I think (e.g. set cap to `len(flagNames_s390x)` or something more meaningful).

prb112:
@marquiz This works on ppc64le.  Chandan has tried it out and would resolve https://github.com/kubernetes-sigs/node-feature-discovery/issues/913

source/cpu/cpuid_arm64.go
```diff
@@ -85,8 +85,8 @@ func getCpuidFlags() []string {
 	hwcap := uint64(C.gethwcap())
 	for i := uint(0); i < 64; i++ {
 		key := uint64(1 << i)
-		val := flagNames_arm64[key]
-		if hwcap&key != 0 {
+		val, ok := flagNames_arm64[key]
+		if hwcap&key != 0 && ok {
```
yselkowitz, line 89: FYI this hunk overlaps #919 so whichever is merged first, the second will need to rebase and adapt.

marquiz: ack

marquiz:
Any arm64 confirmation? @zvonkok @ArangoGutierrez ?

marquiz:
Rebased after #919 was merged
/unhold

marquiz:
Verified that images build on arm64 and s390x

marquiz:
> /lgtm

ping @fmuyassarov @zvonkok

PR #919, opened 2022-10-13
Update CPU flags for ARM, POWER, and Z

yselkowitz:
These updates are based on kernel 5.19.4 and glibc 2.36.

- cpu: update ARM CPU flags
- cpu: update POWER CPU flags
- cpu: update S390 CPU flags

Related: #914
Closes: #913


source/cpu/cpuid_arm64.go
```diff
@@ -78,6 +87,15 @@ var flagNames_arm64 = map[uint64]string{
 	CPU_ARM64_FEATURE_ASIMDDP:  "ASIMDDP",
 	CPU_ARM64_FEATURE_SHA512:   "SHA512",
 	CPU_ARM64_FEATURE_SVE:      "SVE",
+	CPU_ARM64_FEATURE_ASIMDFHM: "ASIMDFHM",
+	CPU_ARM64_FEATURE_DIT:      "DIT",
```
jeffdyoung, line 91: @yselkowitz Where is the list of arm cpu features coming from? I see [this](https://cs.opensource.google/go/x/sys/+/master:cpu/cpu_arm64.go;l=16-39) in go, but I don't see all of the flags you're mentioning here.

yselkowitz: https://github.com/torvalds/linux/blob/master/arch/arm64/include/uapi/asm/hwcap.h
https://github.com/torvalds/linux/blob/master/arch/powerpc/include/uapi/asm/cputable.h
http://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/s390/bits/hwcap.h;hb=HEAD

marquiz:
Thanks @yselkowitz for the update! Looks good to me but I'd like to get thumbs-up from some non-x86 specialist (@mkumatag?) 😊
/ok-to-test

source/cpu/cpuid_arm64.go
jschintag: I think for future reference, https://github.com/torvalds/linux/blob/master/arch/s390/include/asm/elf.h is a better source for cpu flags for s390x. Though in the end glibc is probably getting it from there anyway.

jschintag:
s390x: lgtm

Thank you @yselkowitz

marquiz:
ping @pmundt arm

marquiz:
ping @lubinszARM

source/cpu/cpuid_ppc64le.go
```diff
@@ -124,6 +126,8 @@ var flag2Names_ppc64le = map[uint64]string{
 	PPC_FEATURE2_DARN:           "DARN",
 	PPC_FEATURE2_SCV:            "SCV",
 	PPC_FEATURE2_HTM_NO_SUSPEND: "HTM-NO-SUSPEND",
+	PPC_FEATURE2_ARCH_3_1:       "ARCH_3_1",
+	PPC_FEATURE2_MMA:            "MMA",
```
prb112, line 130: These look right for the P10, and I can confirm I see the expected existing labels, and the new labels. 

MMA/ARCH_3_1 show on the p10 I'm testing on. 

```# lscpu
Architecture:            ppc64le
  Byte Order:            Little Endian
CPU(s):                  16
  On-line CPU(s) list:   0-15
Model name:              POWER10 (architected), altivec supported```

prb112 (approved):
Confirmed ppc64le
Tested on P10

source/cpu/cpuid_ppc64le.go
mkumatag: based on this test result, for ppc64le platform
/lgtm

marquiz:
> Not to pollute the code base.... however. would a comment pointing back to the source be helpful in future maintenance ?

@yselkowitz this would make sense to me. WDYT

yselkowitz:
Rebased and added comments with links to the kernel/glibc sources.

marquiz (approved):
Nice work @yselkowitz. Looks ready from my pov. I'll let somebody else to lgtm 😊

lubinszARM:
arm: lgtm

Good job. @yselkowitz

marquiz:
Thanks for all the reviews
/lgtm

PR #921, opened 2022-10-14
Error strings should not be capitalized

fmuyassarov:
Error strings should not be capitalized [(ST1005)](https://staticcheck.io/docs/checks#ST1005) and remove
the redundancy from array, slice or map composite literals.


marquiz (approved):
Thanks @fmuyassarov, good linting
/lgtm

Issue #902, opened 2022-09-28
E2E infra migration


fmuyassarov:
**What would you like to be added**:
Run E2E tests on dedicated baremetal machines as a presubmit job for every future PR. 
 **Why is this needed**:
Current infra (an AWS cluster) doesn't allow us to run more than one instance of E2E test in parallel because they all would need to run on the same K8S cluster, which potentially might break tests due to the overlap. Because of that we only have a post submit E2E job, that we run after a patch lands on the repository. In practice this seems to work fine until now but there is always a risk that we might miss a bug while merging the patch. As such, we need an ephemeral environment to run E2E on and configure our Prow to start E2E as a pre-submit. 

Work items:
- [x] Apply for compute resources (to run pre-submits) in CNCF
       https://github.com/cncf/cluster/issues/219
- [x] Add a wrapper script to run E2E on ephemeral cluster which would be cleaned up at the end of the each run
       https://github.com/kubernetes-sigs/node-feature-discovery/pull/917
- [ ] Add presubmit job in Prow
       https://github.com/kubernetes/test-infra/pull/27755
- [ ] Update current post submit job to run as periodic (on AWS cluster)
      https://github.com/kubernetes/test-infra/pull/27877

This is a progress tracking issue and I will keep it open until I finish all the necessary steps.

marquiz:
This would be really cool.

fmuyassarov:
@marquiz do you have access to re-open the issue, because I still have some patches handing around ?

fmuyassarov:
thank you

fmuyassarov:
Some updates on this:
1. We have received Equinix metal servers from CNCF. I've create a server already there with Ubuntu 22.04 and installing all the requires bits to be able to run e2e tests.
2. I inspected the server with features that is has and based on that I'm going to update https://github.com/kubernetes/test-infra/pull/27755 PR with the corresponding labels that we expect to have on the node
3. I found using a single local container registry problematic in a sense that it is hard to share it among many kind clusters in parallel and as such using minikube with its in-cluster docker seemed better approach to run multiple k8s clusters where we built images separately and they don't overlap with each other. They are isolated and lifetime depends on the lifetime of the cluster they are in. PR is ready here: https://github.com/kubernetes-sigs/node-feature-discovery/pull/1088

[...]

PR #924, opened 2022-10-17
Retrieve E2E image vars from os

fmuyassarov:
Retrieve E2E image vars from `os` and keep those vars on its own file.
This is a preparation step towards having presubmit (on demand cluster) 
and periodic E2E tests.

Partially fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/902


fmuyassarov:
/test pull-node-feature-discovery-build-image-cross-generic
I'm seeing flakiness on this job more often. Will try to dig the reason

PR #925, opened 2022-10-18
apis/nfd: flatten the structure of features data type 

marquiz:
Flatten the data structure that stores features, dropping the "domain"
level from the data model. That extra level of hierarchy brought little
benefit but just caused some extra complexity, instead. The new
structure nicely matches what we have in the NodeFeatureRule object (the
matchFeatures field of uses the same flat structure with the "feature"
field having a value `<domain>.<feature>`, e.g. `kernel.version`).

This is pre-work for introducing a new "node feature" CRD that contains
the raw feature data. It makes the life of both users and developers
easier when both CRDs, plus our internal code, handle feature data in a
similar flat structure.

Split out from #903


ArangoGutierrez (approved):
/lgtm

PR #932, opened 2022-10-21
Add E2E test for NFD tainting feature

fmuyassarov:
This PR extends E2E to test NodeFeatureRule tainting functionality implemented in https://github.com/kubernetes-sigs/node-feature-discovery/pull/910.

/hold 
until https://github.com/kubernetes-sigs/node-feature-discovery/pull/910 is not merged.

### DIFF
--- a/test/e2e/node_feature_discovery.go
+++ b/test/e2e/node_feature_discovery.go
@@ -49,10 +49,34 @@ import (
 )
 
 var (
-	dockerRepo  = flag.String("nfd.repo", "gcr.io/k8s-staging-nfd/node-feature-discovery", "Docker repository to fetch image from")
-	dockerTag   = flag.String("nfd.tag", "master", "Docker tag to use")
-	dockerImage = fmt.Sprintf("%s:%s", *dockerRepo, *dockerTag)
+	dockerRepo      = flag.String("nfd.repo", "gcr.io/k8s-staging-nfd/node-feature-discovery", "Docker repository to fetch image from")
+	dockerTag       = flag.String("nfd.tag", "master", "Docker tag to use")
+	dockerImage     = fmt.Sprintf("%s:%s", *dockerRepo, *dockerTag)
+	testTolerations = []corev1.Toleration{
+		{
+			Key:    "nfd.node.kubernetes.io/fake-special-node",
+			Value:  "exists",
+			Effect: "NoExecute",
+		},
+		{
+			Key:    "nfd.node.kubernetes.io/fake-dedicated-node",
+			Value:  "true",
+			Effect: "NoExecute",
+		},
+		{
+			Key:    "nfd.node.kubernetes.io/performance-optimized-node",
+			Value:  "true",
+			Effect: "NoExecute",
+		},
+		{
+			Key:    "nfd.node.kubernetes.io/foo",
+			Value:  "true",
+			Effect: "NoExecute",
+		},
+	}
 )
+
+const TestTaintNs = "nfd.node.kubernetes.io"
 
 // cleanupNode deletes all NFD-related metadata from the Node object, i.e.
 // labels and annotations
@@ -84,11 +108,22 @@ func cleanupNode(cs clientset.Interface) {
 				}
 			}
 
+			// Remove taints
+			for _, taint := range node.Spec.Taints {
+				if strings.HasPrefix(taint.Key, TestTaintNs) {
+					newTaints, removed := taintutils.DeleteTaint(node.Spec.Taints, &taint)
+					if removed {
+						node.Spec.Taints = newTaints
+						update = true
+					}
+				}
+			}
+
 			if !update {
 				break
 			}
 
-			By("Deleting NFD labels and annotations from node " + node.Name)
+			By("Deleting NFD labels, annotations and taints from node " + node.Name)
 			_, err = cs.CoreV1().Nodes().Update(context.TODO(), node, metav1.UpdateOptions{})
 			if err != nil {
 				time.Sleep(100 * time.Millisecond)
@@ -162,8 +197,13 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
 
 			// Launch nfd-master
 			By("Creating nfd master pod and nfd-master service")
-			imageOpt := testpod.SpecWithContainerImage(dockerImage)
-			masterPod = e2epod.NewPodClient(f).CreateSync(testpod.NFDMaster(imageOpt))
+
+			imageOpt := []testpod.SpecOption{
+				testpod.SpecWithContainerImage(dockerImage),
+				testpod.SpecWithTolerations(testTolerations),
+				testpod.SpecWithContainerExtraArgs("-enable-taints"),
+			}
+			masterPod = e2epod.NewPodClient(f).CreateSync(testpod.NFDMaster(imageOpt...))
 
 			// Create nfd-master service
 			nfdSvc, err := testutils.CreateService(f.ClientSet, f.Namespace.Name)
@@ -210,6 +250,7 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
 					testpod.SpecWithRestartPolicy(corev1.RestartPolicyNever),
 					testpod.SpecWithContainerImage(dockerImage),
 					testpod.SpecWithContainerExtraArgs("-oneshot", "-label-sources=fake"),
+					testpod.SpecWithTolerations(testTolerations),
 				}
 				workerPod := testpod.NFDWorker(podSpecOpts...)
 				workerPod, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), workerPod, metav1.CreateOptions{})
@@ -257,7 +298,10 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
 				fConf := cfg.DefaultFeatures
 
 				By("Creating nfd-worker daemonset")
-				podSpecOpts := []testpod.SpecOption{testpod.SpecWithContainerImage(dockerImage)}
+				podSpecOpts := []testpod.SpecOption{
+					testpod.SpecWithContainerImage(dockerImage),
+					testpod.SpecWithTolerations(testTolerations),
+				}
 				workerDS := testds.NFDWorker(podSpecOpts...)
 				workerDS, err = f.ClientSet.AppsV1().DaemonSets(f.Namespace.Name).Create(context.TODO(), workerDS, metav1.CreateOptions{})
 				Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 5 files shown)


test/e2e/node_feature_discovery.go
```diff
@@ -50,6 +51,13 @@ var (
 	dockerTag  = flag.String("nfd.tag", "master", "Docker tag to use")
 )
 
+const (
+	// AnnotationNs namespace for all NFD-related annotations.
+	AnnotationNs = "nfd.node.kubernetes.io"
+	// NodeTaintsAnnotation is the annotation that holds the taints that nfd-master set on the node
+	NodeTaintsAnnotation = AnnotationNs + "/taints"
+)
+
```
fmuyassarov, line 60: I will remove these once https://github.com/kubernetes-sigs/node-feature-discovery/pull/910 landed.

fmuyassarov:
/cc @marquiz 
No rush to review it this week, can be after the KubeCon :blush:

test/e2e/node_feature_discovery.go
```diff
@@ -452,22 +472,126 @@ core:
...
+	var err error
+	for retry := 0; retry < 3; retry++ {
+		if err = poll(); err == nil {
+			return nil
+		}
+		time.Sleep(2 * time.Second)
+	}
+	return err
+}
+
```
fmuyassarov, line 553: I should admit that this is not perfect, because this func is tightly coupled with taints annotation and can not be re-used by generic annotation. However, we don't have such cases for now that requires generic function for comparing two set of annotations. Perhaps we can improve it in the future. Same with tainting below.

fmuyassarov:
I will still keep the hold until https://github.com/kubernetes-sigs/node-feature-discovery/pull/910 is not merged. But I think it is ready to review.
/cc @marquiz @ArangoGutierrez

test/e2e/node_feature_discovery.go
fmuyassarov: Rebased.

fmuyassarov:
https://github.com/kubernetes-sigs/node-feature-discovery/pull/910 has landed and this PR is rebased on top latest master.

fmuyassarov:
@marquiz I think we need to take this in also before the release.

marquiz:
> I think we need to take this in also before the release.

Yes, definitely!

test/e2e/utils/pod/pod.go
```diff
@@ -299,6 +299,28 @@ func nfdWorkerSpec(opts ...SpecOption) *corev1.PodSpec {
...
+				Key:    "nfd.node.kubernetes.io/performance-optimized-node",
+				Value:  "true",
+				Effect: "NoExecute",
+			},
+			{
+				Key:    "nfd.node.kubernetes.io/foo",
+				Value:  "true",
+				Effect: "NoExecute",
+			},
+		},
```
marquiz, line 323: Hmm, shouldn't we add these to nfd-master, too?

Also, I'd suggest to not hardcode these but add a `SpecWithTolerations()` "option function" similar to what we do with other podspec customization. WDYT?

test/e2e/node_feature_discovery.go
```diff
@@ -453,22 +466,96 @@ core:
...
+func waitForNfdNodeAnnotation(cli clientset.Interface, expectedAnnotation map[string]string) error {
+	poll := func() error {
+		nodeList, err := cli.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+		for _, node := range nodeList.Items {
+			if diff := cmp.Diff(expectedAnnotation[nfdv1alpha1.NodeTaintsAnnotation], node.Annotations[nfdv1alpha1.NodeTaintsAnnotation]); diff != "" {
+				return fmt.Errorf("node %q annotation does not match expected, diff (expected vs. received): %s", node.Name, diff)
+			}
```
marquiz, line 532: As you commented separately, this could be made more generic. I think very easily, i.e. smth like this should work (also renaming the arg `expectedAnnotation` -> `expectedAnnotation` as it would be a map containing all the expected annotations):

```suggestion
			for k, v := range expectedAnnotations {
				if diff := cmp.Diff(v, node.Annotations[k]); diff != "" {
					return fmt.Errorf("node %q annotation does not match expected, diff (expected vs. received): %s", node.Name, diff)
				}
			}
```

WDYT? This should be then readily re-usable when/if we add support for creating node annotations from NodeFeatureRule objects.

marquiz (commented):
Great work @fmuyassarov. Just two comments. Could you also rebase on the latest master (easier to test locally)?

test/e2e/node_feature_discovery.go
fmuyassarov: Good point, thanks. Done.

test/e2e/utils/pod/pod.go
fmuyassarov: Good point, thanks. Done.

test/e2e/node_feature_discovery.go
```diff
@@ -453,22 +491,98 @@ core:
...
+				By("Verifying updated node taints and annotation from NodeFeatureRules #3")
+				Expect(waitForNfdNodeTaints(f.ClientSet, expectedTaintsUpdated)).NotTo(HaveOccurred())
+				Expect(waitForNfdNodeAnnotation(f.ClientSet, expectedAnnotationUpdated)).NotTo(HaveOccurred())
 			})
 		})
 	})
 })
 
+// waitForNfdNodeAnnotation waits for node to be annotated as expected.
+func waitForNfdNodeAnnotation(cli clientset.Interface, expectedAnnotations map[string]string) error {
```
marquiz, line 548: tiny tiny nit about naming

```suggestion
func waitForNfdNodeAnnotations(cli clientset.Interface, expectedAnnotations map[string]string) error {
```

sorry 🙈

test/e2e/node_feature_discovery.go
```diff
@@ -485,6 +599,51 @@ func waitForNfdNodeLabels(cli clientset.Interface, expected map[string]string) e
...
+		nodeList, err := cli.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+		if err != nil {
+			return err
+		}
+		for _, node := range nodeList.Items {
+			taints, err := nfdTaints(node.Spec.Taints, node)
+			if err != nil {
+				return fmt.Errorf("failed to fetch nfd owned taints for node: %s", node.Name)
+			}
+			if diff := cmp.Diff(taints, expectedTaints); diff != "" {
```
marquiz, line 614: You probably want to diff the other way around (sync with the error msg below)
```suggestion
			if diff := cmp.Diff(expectedTaints, taints); diff != "" {
```

test/e2e/node_feature_discovery.go
fmuyassarov: do'h, my bad

marquiz:
@fmuyassarov please rebase

fmuyassarov:
/hold 
until I get successful results from local tests.

fmuyassarov:
/hold cancel
Tested locally. @marquiz PTAL.

test/e2e/node_feature_discovery.go
```diff
@@ -506,6 +627,51 @@ func waitForNfdNodeLabels(cli clientset.Interface, expected map[string]string) e
...
+	}
+	return err
+}
+
+// nfdTaints returns taints that are owned by the nfd.
+func nfdTaints(taints []corev1.Taint, node corev1.Node) ([]corev1.Taint, error) {
+	// De-serialize the taints annotation into corev1.Taint type for comparision below.
+	var err error
+	nfdTaints := []corev1.Taint{}
+	if val, ok := node.Annotations[nfdv1alpha1.NodeTaintsAnnotation]; ok {
```
marquiz, line 665: You don't actually use the `taints` arg for anything even if that should be the thing you're looking at. I suggest forget about the annotation altogether (as that might be wrong), just filter all taints whose key starts with `nfd.node.kubernetes.io`(we know we use that key for the test taints)

test/e2e/node_feature_discovery.go
```diff
@@ -84,11 +106,22 @@ func cleanupNode(cs clientset.Interface) {
 				}
 			}
 
+			// Remove taints
+			for _, taint := range node.Spec.Taints {
+				if strings.HasPrefix(taint.Key, nfdv1alpha1.AnnotationNs) {
```
marquiz, line 111: Using `nfdv1alpha1.AnnotationNs` here looks a bit strange. Maybe you could e.g. define a const here like `TestTaintNs` or smth?

marquiz (commented):
Thanks @fmuyassarov, e2e-tests now pass 🎉 Just two comments

test/e2e/node_feature_discovery.go
fmuyassarov: Good point. Fixed

test/e2e/node_feature_discovery.go
fmuyassarov: Hmm why do you think so? All the taints that we are setting have prefix of `nfdv1alpha1.AnnotationNs` (i.e. `nfd.node.kubernetes.io`).

marquiz: My point was to specify a separate const for the test taints as they have in principle nothing to do with annotations even if the string value is the same

fmuyassarov: we are using the same NS for the taints annotation too https://github.com/kubernetes-sigs/node-feature-discovery/blob/8a153c12cf77b06d2f8c80a31d66a90d0a7fc58e/pkg/apis/nfd/v1alpha1/annotations_labels.go#L48. But fine with me to replace it with a const.

fmuyassarov: Done.

marquiz: > we are using the same NS for the taints annotation too

Yes we are, but that's an annotation, not a taint. I don't know if I'm thinking it in too complicated way 😅 But I feel that as they're independent/separate, better to specify a separate const

test/e2e/node_feature_discovery.go
```diff
@@ -506,6 +629,48 @@ func waitForNfdNodeLabels(cli clientset.Interface, expected map[string]string) e
...
+		if err = poll(); err == nil {
+			return nil
+		}
+		time.Sleep(10 * time.Second)
+	}
+	return err
+}
+
+// nfdTaints returns taints that are owned by the nfd.
+func nfdTaints(taints []corev1.Taint, node corev1.Node) []corev1.Taint {
```
marquiz, line 663: unused arg?
```suggestion
func nfdTaints(taints []corev1.Taint) []corev1.Taint {
```

fmuyassarov: ah right, forgot to clean up

marquiz (approved):
Great work here @fmuyassarov! Good to get tainting covered by e2e 🚀
/lgtm

test/e2e/node_feature_discovery.go
fmuyassarov: Sure, you have a point that they are separate things.

PR #935, opened 2022-10-25
apis/nfd: fix NodeFeatureRule templating

### LINKED ISSUE #934: Postsubmit E2E is broken
Our postsubmit [E2E test](https://prow.k8s.io/job-history/gs/kubernetes-jenkins/logs/postsubmit-node-feature-discovery-e2e-test) is failing for the last two days. I don't see any potential breaking change that landed within last two days. We had [dependency upgrade](https://github.com/kubernetes-sigs/node-feature-discovery/pull/930), minor [bug fix](https://github.com/kubernetes-sigs/node-feature-discovery/pull/933) and [flatting the data type](https://github.com/kubernetes-sigs/node-feature-discovery/pull/925). To me none of them seems to be a potential cause but maybe I'm missing something obvious. I tried to reproduce the issue by running locally but ended up having some other issues related to Pod Security context. Even after fixing them, I still get another issue. This signals to me that we need E2E running as a presubmit ASAP and possibly make it easy to run locally too (you can just run `KUBECONFIG=$(echo /home/fmuyassarov/.kube/config) make e2e-test ` and expect E2E to run as it would in the Prow). On this topic, we are still waiting CNCF to offer as some hardware, ref https://github.com/cncf/cluster/issues/219.

I continue debugging but any help is appreciated.

marquiz:
Fix handling of templates that got broken in
b907d07d7e0e7bd47f988b22c8b9a380f8abf75d when "flattening" the internal data structure of features. That happened because the golang text/template format uses dots to reference fields of a struct / elements of a map (i.e. 'foo.bar' means that 'bar' must be a sub-element of foo). Thus, using dots in our feature names (e.g. 'cpu.cpuid') means that that hierarchy must be reflected in the data structure that is fed to the templating engine. Thus, for templates we're now stuck stuck with two level hierarchy. It doesn't really matter for now as all our features follow that naming patter. We might be able to overcome this limitation e.g.  by using reflect but that's left as a future exercise.

Fixes #934

### DIFF
--- a/pkg/apis/nfd/v1alpha1/rule.go
+++ b/pkg/apis/nfd/v1alpha1/rule.go
@@ -145,7 +145,9 @@ func (r *Rule) executeVarsTemplate(in matchedFeatures, out map[string]string) er
 	return nil
 }
 
-type matchedFeatures map[string]interface{}
+type matchedFeatures map[string]domainMatchedFeatures
+
+type domainMatchedFeatures map[string]interface{}
 
 func (e *MatchAnyElem) match(features *Features) (bool, matchedFeatures, error) {
 	return e.MatchFeatures.match(features)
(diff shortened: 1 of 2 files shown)


fmuyassarov (commented):
+1 for the reflection but for now this looks as a good fix to me. Thanks for the quick response @marquiz 
/lgtm

marquiz:
And now we got odd timeout in the image builder 😕
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-node-feature-discovery-push-images/1585181601254346752

We might want to increase the timeout(s) a bit more 🧐

fmuyassarov:
> And now we got odd timeout in the image builder confused https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-node-feature-discovery-push-images/1585181601254346752
> 
> We might want to increase the timeout(s) a bit more monocle_face

Yeah makes sense. I can do it

PR #937, opened 2022-10-27
Stop using the beta.kubernetes.io/os and arch labels

pacoxu:
/kind cleanup
See https://github.com/kubernetes/kubernetes/issues/89477, https://github.com/kubernetes/kubernetes/issues/89477#issuecomment-603911496.


fmuyassarov (commented):
Thank you for the heads up @pacoxu 
/lgtm

PR #950, opened 2022-11-02
docs: restructure docs

### LINKED ISSUE #723: Re-organize deployment docs

Re-organize the *Get started* major section of documentation. Probably spitting it into multiple separate major sections.

E.g. something like

- get-started/
  - introduction
  - quick-start
- deployment/
  - kustomize
  - helm
  - olm
  - tls
  - uninstallation
- usage/
  - features
  - nfd-master
  - nfd-worker
  - nfd-topology-updater
  - using node labels
  - examples and demos
 

**Why is this needed**:

Better structure, better readability for docs

marquiz:
Introduce two main sections "Deployment" and "Usage" and move "Developer
guide" to the top level, too. In particular, split the huge
deployment-and-usage file into multiple parts under the new main
sections. Move customization guide from "Advanced" to "Usage".
This patch also renames "Advanced" to "Reference" as only that is left
there is reference documentation.

Fixes #723


docs/usage/nfd-master.md
```diff
@@ -0,0 +1,35 @@
...
+# NFD-Master
+{: .no_toc}
+
+## Table of contents
+{: .no_toc .text-delta}
+
+1. TOC
+{:toc}
+
+---
```
fmuyassarov, line 16: probably worth to remove since the list is empty

docs/usage/using-labels.md
```diff
@@ -0,0 +1,37 @@
...
+# Using node labels
+{: .no_toc}
+
+## Table of contents
+{: .no_toc .text-delta}
+
+1. TOC
+{:toc}
+
+---
```
fmuyassarov, line 16: same here, perhaps we can drop it?

fmuyassarov (commented):
I like it, looks more organized to be honest and readers are immediately presented with that they can do whereas right now some bits and pieces are bit hidden :). Some nits, but otherwise LGTM
One comment, maybe we should move "Developer guide" page under its own paragraph as below ? 
```
- Get started
- Deployment
- Usage
- Advanced
- Developer Guide
- Contributing
```

marquiz:
> One comment, maybe we should move "Developer guide" page under its own paragraph as below ?

Exactly the kind of feedback I was after 😊 I've written and stared at the docs so long that I've lost perspective (and context) myself. I like this change. After this, it prolly makes sense to rename *Advanced* -> *Reference*.

docs/usage/nfd-master.md
marquiz: K, fixed

docs/usage/using-labels.md
marquiz: k, dropped

fmuyassarov:
LGTM, will let others to have a look too. Please let me know when you think it is time merge it and I will give lgtm

fmuyassarov:
> . Partly because this kind of restructuring is such a pita

+1
/lgtm

Issue #957, opened 2022-11-14
Add specific IBM zSystems model detection and label to s390x nodes


haubenr:
**What would you like to be added**:

Knowing which particular IBM zSystems model (e.g. z15, z16) a particular s390x cluster node is running on might be beneficial to IBM zSystems users, especially when it comes to particular HW features that are only present in a specific IBM zSystems machine generation.

**Why is this needed**:

See above.

PR #958, opened 2022-11-14
Enhancement: add IBM zSystems model detection

haubenr:
This PR implements the required changes to address the enhancement suggestion in #957.

With this PR, s390x cluster nodes will have an additional label set called 'feature.node.kubernetes.io/system-model.s390x' that contains the corresponding IBM zSystems model generation name, e.g. 'z15'.

Here's an example for a worker running on a z15 model:

```
  labels:
    beta.kubernetes.io/arch: s390x
    beta.kubernetes.io/os: linux
    feature.node.kubernetes.io/cpu-cpuid.DFLT: "true"
    feature.node.kubernetes.io/cpu-cpuid.DFP: "true"
    feature.node.kubernetes.io/cpu-cpuid.EDAT: "true"
    feature.node.kubernetes.io/cpu-cpuid.EIMM: "true"
    feature.node.kubernetes.io/cpu-cpuid.ESAN3: "true"
    feature.node.kubernetes.io/cpu-cpuid.ETF3EH: "true"
    feature.node.kubernetes.io/cpu-cpuid.GS: "true"
    feature.node.kubernetes.io/cpu-cpuid.HIGHGPRS: "true"
    feature.node.kubernetes.io/cpu-cpuid.LDISP: "true"
    feature.node.kubernetes.io/cpu-cpuid.MSA: "true"
    feature.node.kubernetes.io/cpu-cpuid.SORT: "true"
    feature.node.kubernetes.io/cpu-cpuid.STFLE: "true"
    feature.node.kubernetes.io/cpu-cpuid.TE: "true"
    feature.node.kubernetes.io/cpu-cpuid.VX: "true"
    feature.node.kubernetes.io/cpu-cpuid.VXD: "true"
    feature.node.kubernetes.io/cpu-cpuid.VXE: "true"
    feature.node.kubernetes.io/cpu-cpuid.VXE2: "true"
    feature.node.kubernetes.io/cpu-cpuid.VXP: "true"
    feature.node.kubernetes.io/cpu-cpuid.ZARCH: "true"
    feature.node.kubernetes.io/cpu-hardware_multithreading: "false"
    feature.node.kubernetes.io/cpu-model.family: "0"
    feature.node.kubernetes.io/cpu-model.id: "0"
    feature.node.kubernetes.io/cpu-model.vendor_id: VendorUnknown
    feature.node.kubernetes.io/kernel-version.full: 4.15.0-191-generic
    feature.node.kubernetes.io/kernel-version.major: "4"
    feature.node.kubernetes.io/kernel-version.minor: "15"
    feature.node.kubernetes.io/kernel-version.revision: "0"
    feature.node.kubernetes.io/system-model.s390x: z15
    feature.node.kubernetes.io/system-os_release.ID: ubuntu
    feature.node.kubernetes.io/system-os_release.VERSION_ID: "21.10"
    feature.node.kubernetes.io/system-os_release.VERSION_ID.major: "21"
    feature.node.kubernetes.io/system-os_release.VERSION_ID.minor: "10"
    kubernetes.io/arch: s390x
    kubernetes.io/hostname: kind-worker
    kubernetes.io/os: linux
```

### DIFF
--- a/source/system/s390x_model.go
+++ b/source/system/s390x_model.go
@@ -1,0 +1,66 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package system
+
+var s390xModelMap = map[string][]string{
+	"z9":   {"2094", "2096"},
+	"z10":  {"2097", "2098"},
+	"z196": {"2817"},
+	"z114": {"2818"},
+	"z12":  {"2827", "2828"},
+	"z13":  {"2964", "2965"},
+	"z14":  {"3906", "3907"},
+	"z15":  {"8561", "8562"},
+	"z16":  {"3931"},
+}
+
+// S390xModels ist the main struct for querying IBM z Systems hardware models
+type S390xModels struct {
+	HwMap map[string][]string
+}
+
+// NewS390xModels creates a new instance of S390xModels
+func NewS390xModels() S390xModels {
+	s390xmodels := S390xModels{}
+	s390xmodels.HwMap = s390xModelMap
+	return s390xmodels
+}
+
+// LookupModel looks for the IBM z Systems hardware model matching the given machine type
+func (s S390xModels) LookupModel(machineType string) (model string) {
+	found := false
+	for k, v := range s.HwMap {
+		if s.stringInSlice(machineType, v) {
+			model = k
+			found = true
+			break
+		}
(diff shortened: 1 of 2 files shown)


marquiz (commented):
Hi @haubenr, thanks for the PR. However, I somehow think that we should contribute this in the [cpuid](https://github.com/klauspost/cpuid) package instead. I don't think NFD whould become a database of all kind of hardware IDs (pci and usb devices, cpus etc..). I think that information should be maintained somewhere else which also other projects could more easily use and depend on. WDYT?

I know we already have some arm and s390 cpuid isa detection in the nfd codebase but also that is something that would be better to be moved to the cpuid package in the future, I think...

Another note is that we already have `cpu-model.X` features so eventually this information should be there.

haubenr:
@marquiz Thanks for the heads-up! I was thinking about that and I'm not sure this kind of info belongs with the CPU-specific features package. The thing about IBM zSystems models is that while each model generation does come with their own CPU model / revision the overall system changes significantly as well. I guess what I'm trying to say is that the model denominations like z14, z15, z16 etc. do not only indicate a change in CPU feature sets but also a complete system upgrade (including system interconnects, special HW adapters etc.) as well.

ArangoGutierrez:
Hi @haubenr any status on this? 
should we close or refresh lifecycle?

PR #960, opened 2022-11-17
test/e2e: fix topologu-updater cmdline args

marquiz:
The -kubelet-config-file flag got renamed to -kubelet-config-uri and the expected format changed from plain filepath to URI.


fromanirh (commented):
/lgtm
nit: typo in the commit title (topologu)
besides that I wonder why it didn't fail when the PR which changed the commandline was posted

marquiz:
> nit: typo in the commit title (topologu)

Woops, we missed that one 🙄 

> besides that I wonder why it didn't fail when the PR which changed the commandline was posted

That's because our e2e tests are run "after the fact". Not ideal, I know. But thanks to @fmuyassarov that's hopefully going to be addressed soon (https://github.com/kubernetes/test-infra/pull/27755)

PR #961, opened 2022-11-17
e2e: topologyupdater: fix and stabilize tests

Tal-or:
A couple of commits which are addressing several issues in the tests.
We explained the rationale for every change in the commits themselves 

### DIFF
--- a/test/e2e/node_feature_discovery.go
+++ b/test/e2e/node_feature_discovery.go
@@ -38,9 +38,8 @@ import (
 	e2elog "k8s.io/kubernetes/test/e2e/framework/log"
 	e2enetwork "k8s.io/kubernetes/test/e2e/framework/network"
 	e2epod "k8s.io/kubernetes/test/e2e/framework/pod"
-	nfdclient "sigs.k8s.io/node-feature-discovery/pkg/generated/clientset/versioned"
-
 	nfdv1alpha1 "sigs.k8s.io/node-feature-discovery/pkg/apis/nfd/v1alpha1"
+	nfdclient "sigs.k8s.io/node-feature-discovery/pkg/generated/clientset/versioned"
 	"sigs.k8s.io/node-feature-discovery/source/custom"
 	testutils "sigs.k8s.io/node-feature-discovery/test/e2e/utils"
 )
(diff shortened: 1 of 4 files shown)


test/e2e/node_feature_discovery.go
```diff
@@ -100,6 +101,7 @@ func cleanupNode(cs clientset.Interface) {
 // Actual test suite
 var _ = SIGDescribe("Node Feature Discovery", func() {
 	f := framework.NewDefaultFramework("node-feature-discovery")
+	f.NamespacePodSecurityEnforceLevel = admissionapi.LevelPrivileged
```
marquiz, line 104: Here, for nfd-master and nfd-worker we shouldn't need this, right?

marquiz (commented):
Just one note, otherwise looks good to go

test/e2e/node_feature_discovery.go
Tal-or: Here we have a choice, but we can't continue with the existing configuration since the test is falling right at the beginning. The choices are:
1. Adding to the pod spec all the necessary security constraints such as:
 `RunAsNonRoot: true`, `AllowPrivilegeEscalation: false`, etc.
2. Edit the namespace as we have here. It's the simplest solution and since this is only for testing purposes I assumed it's good enough.

Anyhow, we can go with option number 1 as well. Please let me know what is your preference.

marquiz: > 1. Adding to the pod spec all the necessary security constraints such as:
>    `RunAsNonRoot: true`, `AllowPrivilegeEscalation: false`, etc.

Ach, these are missing in the e2e pod spec 🤦 I think we definitely want to add them and this would be the preferable solution. This way the e2e will also more closely resemble the actual (default) deployment.

Tal-or: Done!

Please consider that some tests are still falling but are in the `node_feature_discovery.go` division (which I am unfortunately not too experienced with). 
Issue: https://github.com/kubernetes-sigs/node-feature-discovery/issues/962

Tal-or: But IMVHO I don't think it should stop us from moving on with this PR

marquiz (approved):
Thank you @Tal-or for improving the e2e tests!

PR #917, opened 2022-10-13
add ephemeral e2e test

fmuyassarov:
Partially fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/902

PR #966, opened 2022-11-24
e2e: add SecurityContext to master

Tal-or:
The master pod need these `SecurityContext` configurations In order to run inside a namespace with restricted policy

### DIFF
--- a/test/e2e/utils/pod.go
+++ b/test/e2e/utils/pod.go
@@ -115,6 +115,8 @@ type PodSpecOption func(spec *corev1.PodSpec)
 
 // NFDMasterPod provide NFD master pod definition
 func NFDMasterPod(opts ...PodSpecOption) *corev1.Pod {
+	yes := true
+	no := false
 	p := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: "nfd-master-",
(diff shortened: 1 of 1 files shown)


test/e2e/utils/pod.go
```diff
@@ -136,6 +138,18 @@ func NFDMasterPod(opts ...PodSpecOption) *corev1.Pod {
...
+						Capabilities: &corev1.Capabilities{
+							Drop: []corev1.Capability{"ALL"},
+						},
+						Privileged:               &no,
+						RunAsNonRoot:             &yes,
+						ReadOnlyRootFilesystem:   &yes,
+						AllowPrivilegeEscalation: &no,
+						SeccompProfile: &corev1.SeccompProfile{
+							Type: corev1.SeccompProfileTypeRuntimeDefault,
+						},
```
fmuyassarov, line 151: I remember adding the same in https://github.com/kubernetes-sigs/node-feature-discovery/pull/917#discussion_r994753825 but for some reason I excluded it at that time. So, when running e2e had similar issues with the security context. So, it makes sense to have this change.
/lgtm
/ok-to-test

marquiz (approved):
Thanks @Tal-or!

PR #964, opened 2022-11-22
test/e2e: more flexible pod spec generation

marquiz:
Change the pod spec generator functions to accept parameterization in
the form of more generic "mutator functions". This makes the addition of
new test specific pod spec customizations a lot cleaner. Plus, hopefully
makes the code a bit more readable as well.

Also, slightly simplify the SpecWithConfigMap() but dropping one
redundant argument.

Inspired by latest contributions by Talor Itzhak (titzhak@redhat.com).

PR #967, opened 2022-11-24
e2e: move pod utils to a seperate package

Tal-or:
By moving those utils into a separate package,
we can make the function names shorter and more apparent.

For example, instead of:
```
testutils.NFDWorkerPod(opts...)
testutils.NFDMasterPod(opts...)
testutils.SpecWithContainerImage(...)
```
we'll have the following:
```
testpod.NFDWorker(opts...)
testpod.NFDMaster(opts...)
testpod.SpecWithContainerImage(...)
```

It will also make the package more isolated and portable.

This work was initiated by the discussion here: https://github.com/kubernetes-sigs/node-feature-discovery/pull/964#issuecomment-1324093399

### DIFF
--- a/test/e2e/utils/daemonset/daemonset.go
+++ b/test/e2e/utils/daemonset/daemonset.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


test/e2e/utils/daemonset/daemonset.go
```diff
@@ -0,0 +1,42 @@
+package daemonset
```
ArangoGutierrez, line 1: Even if this is a test file, please add license header

ArangoGutierrez (changes requested):
Overall looks good, just one nit.
WDYT @marquiz ?

test/e2e/utils/daemonset/daemonset.go
Tal-or: Thank you for your review. 
Done!

marquiz (approved):
Thanks @Tal-or I think this nicely improves the organization of the test code

fmuyassarov (commented):
/lgtm
/ok-to-test

PR #968, opened 2022-11-25
nfd-worker: drop deprecated command line flags

marquiz:
Drop the following flags that were deprecated already in v0.8.0:

`-sleep-interval`  (replaced by `core.sleepInterval` config file option)
`-label-whitelist` (replaced by `core.labelWhiteList` config file option)
`-sources`         (replaced by `-label-sources flag`)


fmuyassarov (commented):
/lgtm

PR #969, opened 2022-11-28
docs: revise topology-updater helm chart rbac parameters

marquiz:


adrianchiris:
nice, seems like the same can be applied for 
```
[annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
```

as well :) although not related to this PR 

/lgtm

adrianchiris (approved):

marquiz:
> nice, seems like the same can be applied for
> 
> ```
> [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)
> ```

Yeah, I think there are a lot of references like this. Let's keep in mind as future enhancement

PR #971, opened 2022-11-29
test/e2e: remove dropped -sleep-interval arg

marquiz:
Replace with a ConfigMap-based configuration.


fmuyassarov (commented):
Thanks @marquiz , looks good
/lgtm

PR #972, opened 2022-11-29
deployment: drop stale nfd-api-crds.yaml

marquiz:
Remove a stale unused file that was accidentally committed from an experimental work.


fmuyassarov (commented):
yep, we have duplication indeed.
/lgtm

PR #976, opened 2022-12-01
docs: small update to customization guide

marquiz:
Add a reference to the label rule format in the NodeFeatureRule section. Also make it explicit in the beginning of Hooks section that hooks are deprecated.


ArangoGutierrez (approved):
/lgtm

marquiz:
mdlint wasn't happy about my emphasized text that looked like a heading -> had to adjust a bit

fmuyassarov (commented):
/lgtm

PR #978, opened 2022-12-02
helm: drop NodeFeatureRule CRD from templates

marquiz:
Helm 3 can manage CRDs in a more user friendly way. In fact, this now causes deployment failure as Helm automatically tries to install the CRD from the "crds/" subdir, too.


ArangoGutierrez (approved):
/lgtm

PR #981, opened 2022-12-06
nfd-master svc should select only nfd-master pods

tariq1890:
The nfd-workers don't have any exposed ports. It would be cleaner to have the service selector only point to the pods that actually expose a container port and host the grpc service


tariq1890:
Thanks a lot @Ethyling !

/assign @marquiz

marquiz (approved):
Thanks @tariq1890 for the enhancement.

PR #984, opened 2022-12-08
nfd-worker: detect the namespace it is running in

marquiz:
Implement detection of kubernetes namespace by reading file /var/run/secrets/kubernetes.io/serviceaccount/namespace

Aa a fallback (if the file is not accessible) we take namespace from KUBERNETES_NAMESPACE environment variable. This is useful for e.g. testing and development where you might run nfd-worker directly from the command line on a host system.

This PR is split out from #903

### DIFF
--- a/pkg/utils/kubernetes.go
+++ b/pkg/utils/kubernetes.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"os"
+	"strings"
+)
+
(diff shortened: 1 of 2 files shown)


pkg/utils/kubernetes.go
```diff
@@ -0,0 +1,35 @@
...
+// or an empty string if the namespace cannot be determined.
+func GetKubernetesNamespace() string {
+	const kubernetesNamespaceFilePath = "/var/run/secrets/kubernetes.io/serviceaccount/namespace"
+	if _, err := os.Stat(kubernetesNamespaceFilePath); err == nil {
+		data, err := os.ReadFile(kubernetesNamespaceFilePath)
+		if err == nil {
+			return strings.TrimSpace(string(data))
+		}
+	}
+	return os.Getenv("KUBERNETES_NAMESPACE")
```
fmuyassarov, line 34: What if the data is empty but `KUBERNETES_NAMESPACE` holds the actual namespace? Currently we would return the empty namespace even though there is a chance that `KUBERNETES_NAMESPACE` environment variable has it. Perhaps checking the length of `data` can be added before returning, if 
```go
// we don't return in this case, but move fallback to KUBERNETES_NAMESPACE
!len(strings.TrimSpace(string(data))) > 0
```

marquiz: Good question, but I'm not sure. If that file exists I think we should really use that (and get an error if the content is invalid e.g. the file is empty). The env var is really only meant for development and debugging

fmuyassarov: alright, let's get this in now since there are some other dependents patches expected. Perhaps, we will improve it in the future patches.
/lgtm

PR #990, opened 2022-12-14
nfd-master: implement ratelimiter for nfd api updates

marquiz:
Implement a naive ratelimiter for node update events originating from the nfd API. We might get a ton of events in short interval. The simplest example is startup when we get a separate Add event for every NodeFeature and NodeFeatureRule object. Without rate limiting we run "update all nodes" separately for each NodeFeatureRule object, plus, we would run "update node X" separately for each NodeFeature object targeting node X. This is a huge amount of wasted work because in principle just running "update all nodes" once should be enough.

Split out from #903


ArangoGutierrez (approved):
/lgtm

marquiz:
Rebased

PR #991, opened 2022-12-14
nfd-master: rename -featurerules-controller flag to -crd-controller

marquiz:
Deprecate the '-featurerules-controller' command line flag as the name does not describe the functionality anymore: in practice it controls the CRD controller handling both NodeFeature and NodeFeatureRule objects. The patch introduces a duplicate, more generally named, flag '-crd-controller'. A warning is printed in the log if '-featurerules-controller' flag is encountered.

Split out from #903


ArangoGutierrez (approved):
/lgtm

PR #993, opened 2022-12-14
test/e2e: fix creation of NFD CRDs

marquiz:
After introducing NodeFeatureRule we packed two CRD definitions in one yaml file. Our e2e-tests were not prepared to that and the file itself was also renamed so it couldn't even be read by the test suite.

With this change the e2e-tests start to create NodeFeatre CRD in the test cluster, preparing for the addition of e2e-tests for NodeFeature API.


fmuyassarov (commented):
/lgtm

PR #997, opened 2022-12-14
test/e2e: create CRDs once in the beginning of the tests

marquiz:
Only generate CRDs once in the beginning of the test run. Use the "Ordered" option for the test container so that we can utilize ginkgo.BeforeAll to only do stuff once before the first test. Changing from unordered to ordered shouldn't make a big difference here.

Add a cleanup function to remove stale NodeFeatureRule objects that are cluster-scoped and not deleted with the test namespace.


fmuyassarov (commented):
This is looks better than before and more efficient!
Thank you @marquiz

PR #998, opened 2022-12-14
nfd-master: update all nodes at startup when NodeFeature API enabled

marquiz:
We want to always update all nodes at startup. Without this patch we don't get any update event from the controller if no NodeFeature or NodeFeatureRule objects exist in the cluster. Thus all nodes would stay untouched whereas we really want to remove all labels from all nodes in this case.


fmuyassarov (commented):
/lgtm

PR #999, opened 2022-12-14
nfd-master: update node if no NodeFeature objects are present

marquiz:
Correctly handle the case where no NodeFeature objects exist for certain
node (and NodeFeature API has been enabled with
`-enable-nodefeature-api`). In this case all the labels should be removed.


fmuyassarov (commented):
I'm might be asking dump question, but is this patch doing what the title says?

marquiz:
> I'm might be asking dump question, but is this patch doing what the title says?

No it's not a dumb question. But yes it does, I might want to change the wording a bit, though 
/hold

marquiz:
Improved the commit message
/retitle nfd-master: update node if no NodeFeature objects are present
/unhold

marquiz:
@fmuyassarov PTAL. Makes sense(?)

fmuyassarov (commented):
Thanks @marquiz . Looks better :) 
/lgtm

PR #1001, opened 2022-12-14
test/e2e: add basic e2e-tests for NodeFeature API

marquiz:
Add an initial test set for the NodeFeature API. This is done simply by
running a second pass of the tests but with -enable-nodefeature-api
(i.e. NodeFeature API enabled and gRPC disabled). This should give basic
confidence that the API actually works and form a basis for further
imporovements on testing the new CRD API.

### DIFF
--- a/test/e2e/node_feature_discovery.go
+++ b/test/e2e/node_feature_discovery.go
@@ -154,13 +154,30 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
 	f := framework.NewDefaultFramework("node-feature-discovery")
 
 	nfdTestSuite := func(useNodeFeatureApi bool) {
+		createPodSpecOpts := func(opts ...testpod.SpecOption) []testpod.SpecOption {
+			if useNodeFeatureApi {
+				return append(opts, testpod.SpecWithContainerExtraArgs("-enable-nodefeature-api"))
+			}
+			return opts
+		}
+
 		Context("when deploying a single nfd-master pod", Ordered, func() {
 			var (
-				masterPod *corev1.Pod
 				crds      []*apiextensionsv1.CustomResourceDefinition
 				extClient *extclient.Clientset
 				nfdClient *nfdclient.Clientset
 			)
+
+			checkNodeFeatureObject := func(name string) {
+				_, err := nfdClient.NfdV1alpha1().NodeFeatures(f.Namespace.Name).Get(context.TODO(), name, metav1.GetOptions{})
+				if useNodeFeatureApi {
+					By(fmt.Sprintf("Check that NodeFeature object for the node %q was created", name))
+					Expect(err).NotTo(HaveOccurred())
+				} else {
+					By(fmt.Sprintf("Check that NodeFeature object for the node %q hasn't been created", name))
+					Expect(err).To(HaveOccurred())
+				}
+			}
 
 			BeforeAll(func() {
 				// Create clients for apiextensions and our CRD api
@@ -198,13 +215,12 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
 
 				// Launch nfd-master
 				By("Creating nfd master pod and nfd-master service")
-
-				imageOpt := []testpod.SpecOption{
+				podSpecOpts := createPodSpecOpts(
 					testpod.SpecWithContainerImage(dockerImage),
 					testpod.SpecWithTolerations(testTolerations),
 					testpod.SpecWithContainerExtraArgs("-enable-taints"),
-				}
-				masterPod = e2epod.NewPodClient(f).CreateSync(testpod.NFDMaster(imageOpt...))
+				)
+				masterPod := e2epod.NewPodClient(f).CreateSync(testpod.NFDMaster(podSpecOpts...))
 
 				// Create nfd-master service
 				nfdSvc, err := testutils.CreateService(f.ClientSet, f.Namespace.Name)
@@ -247,12 +263,12 @@ var _ = SIGDescribe("Node Feature Discovery", func() {
 
 					// Launch nfd-worker
 					By("Creating a nfd worker pod")
-					podSpecOpts := []testpod.SpecOption{
+					podSpecOpts := createPodSpecOpts(
 						testpod.SpecWithRestartPolicy(corev1.RestartPolicyNever),
 						testpod.SpecWithContainerImage(dockerImage),
 						testpod.SpecWithContainerExtraArgs("-oneshot", "-label-sources=fake"),
 						testpod.SpecWithTolerations(testTolerations),
-					}
+					)
 					workerPod := testpod.NFDWorker(podSpecOpts...)
 					workerPod, err := f.ClientSet.CoreV1().Pods(f.Namespace.Name).Create(context.TODO(), workerPod, metav1.CreateOptions{})
 					Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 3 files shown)


marquiz:
We need #998 and #999 before merging this
/hold

marquiz:
Actually, #998 and #999 shouldn't be mandatory
/unhold

marquiz:
Rebased

test/e2e/node_feature_discovery.go
```diff
@@ -118,253 +118,276 @@ func cleanupCRDs(cli *nfdclient.Clientset) {
...
-							Expect(fConf.LabelWhitelist).NotTo(HaveKey(k))
+							Expect(fakeFeatureLabels).Should(HaveKey(k))
 						}
 					}
 
-					// Check annotations
-					e2elog.Logf("verifying annotations of node %q...", node.Name)
-					for k, v := range nodeConf.ExpectedAnnotationValues {
-						Expect(node.Annotations).To(HaveKeyWithValue(k, v))
+					checkNodeFeatureObject(workerPod.Spec.NodeName)
```
fmuyassarov, line 295:
```suggestion
					checkNodeFeatureObject(node.Name)
```
We have the node already from L282, so I think better to get the name from the node itself directly.

marquiz:
Yet another rebase

test/e2e/node_feature_discovery.go
```diff
@@ -153,263 +153,285 @@ func cleanupCRDs(cli *nfdclient.Clientset) {
...
-					}
-				}
+				By("Waiting for the nfd-master pod to be running")
+				Expect(e2epod.WaitTimeoutForPodRunningInNamespace(f.ClientSet, masterPod.Name, masterPod.Namespace, time.Minute)).NotTo(HaveOccurred())
 
-				By("Deleting the node-feature-discovery worker pod")
-				err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), workerPod.ObjectMeta.Name, metav1.DeleteOptions{})
+				By("Verifying the node where nfd-master is running")
+				// Get updated masterPod object (we want to know where it was scheduled)
+				masterPod, err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Get(context.TODO(), masterPod.ObjectMeta.Name, metav1.GetOptions{})
```
fmuyassarov, line 234:
```suggestion
				masterPod, err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Get(context.TODO(), masterPod.Name, metav1.GetOptions{})
```
nit

test/e2e/node_feature_discovery.go
```diff
@@ -153,263 +153,285 @@ func cleanupCRDs(cli *nfdclient.Clientset) {
...
 
-				if cfg == nil {
-					Skip("no e2e-config was specified")
-				}
-				if cfg.DefaultFeatures == nil {
-					Skip("no 'defaultFeatures' specified in e2e-config")
-				}
-				fConf := cfg.DefaultFeatures
+				By("Waiting for the nfd-master service to be up")
+				Expect(e2enetwork.WaitForService(f.ClientSet, f.Namespace.Name, nfdSvc.ObjectMeta.Name, true, time.Second, 10*time.Second)).NotTo(HaveOccurred())
```
fmuyassarov, line 242:
```suggestion
				Expect(e2enetwork.WaitForService(f.ClientSet, f.Namespace.Name, nfdSvc.Name, true, time.Second, 10*time.Second)).NotTo(HaveOccurred())
```

test/e2e/node_feature_discovery.go
```diff
@@ -153,263 +153,285 @@ func cleanupCRDs(cli *nfdclient.Clientset) {
...
 					}
 
-					// Check annotations
-					e2elog.Logf("verifying annotations of node %q...", node.Name)
-					for k, v := range nodeConf.ExpectedAnnotationValues {
-						Expect(node.Annotations).To(HaveKeyWithValue(k, v))
+					checkNodeFeatureObject(workerPod.Spec.NodeName)
+
+					By("Deleting the node-feature-discovery worker pod")
+					err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), workerPod.ObjectMeta.Name, metav1.DeleteOptions{})
```
fmuyassarov, line 298:
```suggestion
					err = f.ClientSet.CoreV1().Pods(f.Namespace.Name).Delete(context.TODO(), workerPod.Name, metav1.DeleteOptions{})
```

test/e2e/node_feature_discovery.go
```diff
@@ -153,263 +153,285 @@ func cleanupCRDs(cli *nfdclient.Clientset) {
...
+			)
+
+			checkNodeFeatureObject := func(name string) {
+				_, err := nfdClient.NfdV1alpha1().NodeFeatures(f.Namespace.Name).Get(context.TODO(), name, metav1.GetOptions{})
+				if useNodeFeatureApi {
+					By(fmt.Sprintf("Check that NodeFeature object for the node %q was created", name))
+					Expect(err).NotTo(HaveOccurred())
+				} else {
+					By(fmt.Sprintf("Check that NodeFeature object for the node %q hasn't been created", name))
+					Expect(err).To(HaveOccurred())
```
fmuyassarov, line 178: IINW, NodeFeature CRD will still be applied even though we are using gRPC because its definition is in the same [file](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/deployment/base/nfd-crds/nfd-api-crds.yaml). I think we should extend the `else` statement to check that NodeFeature is empty.

fmuyassarov (commented):
Looks good @marquiz. Some inline comments.

test/e2e/node_feature_discovery.go
marquiz: I think that's fine, doesn't do any harm afaics. I mean I think it's ok to create both CRDs even if we run with NodeFeature disabled in the first pass.

marquiz: > My point was, only expecting err to have occurred is not enough, because we might have no error of getting NodeFeature and still get 0 NodeFeature, which is the case that falls under `else{}`.

Ach, sorry for reading this comment way too hastily. This doesn't create the CRD. It's about nfd-worker creating the NodeFeature object (for the node it's running on). We `Get` a NodeFeature object with a specified name in a specified  namespace so we get a NotExist error if that does not exist

test/e2e/node_feature_discovery.go
marquiz: Thx, fixed!

test/e2e/node_feature_discovery.go
marquiz: Thx! A bit unrelated to this PR as the first patch basically just indents the code by one tab (creating a function wrapper). All these ObjectMeta fixed separately in #1005

test/e2e/node_feature_discovery.go
marquiz: See #1005

test/e2e/node_feature_discovery.go
marquiz: See #1005

marquiz:
Thanks @fmuyassarov for the review. Feedback addressed/responded. WDYT, PTAL

test/e2e/node_feature_discovery.go
fmuyassarov: cool, thanks

fmuyassarov (commented):
Looks good. I think with this we are close to cut the release soon.
/lgtm

marquiz:
> I think with this we are close to cut the release soon.

👍

PR #1004, opened 2022-12-16
test/e2e: don't expect control-plane nodes to be labeled

marquiz:
Without this e2e tests may fail in clusters with control-plane-only
nodes.


marquiz:
Hmm, there's still something
/hold

marquiz:
/retitle test/e2e: don't expect control-plane nodes to be labeled

marquiz:
Re-scoped a bit. Needed a more comprehensive fix
@Ethyling PTAL

/unhold

marquiz:
Verified in latest kind and minikube clusters

PR #1017, opened 2022-12-22
apis/nfd: make all fields in NodeFeatureSpec optional

marquiz:
Don't require features to be specified. The creator possibly only wants
to create labels or only some types of features. No need to specify
empty structs for the unused fields.


fmuyassarov (commented):
/lgtm

PR #1018, opened 2022-12-23
Add common utility function for getting node name

marquiz:


fmuyassarov (commented):
Better organized!
/lgtm

Issue #928, opened 2022-10-20
topology-updater: NRT CRD resourses are not cleaned up, on node deletion(scale-in)


nmathew:
Node Resource Topology node specific CRD resources are not cleaned up on node deletion(scale-in).
I think this should be part of nfd worker/updater.

nmathew:
@fromanirh

fromanirh:
yes, this is a gap. Besides littering the cluster, there is any other impact?

nmathew:
No functional impact...

marquiz:
NFD-Master could have a garbage-collection loop to delete NRT objects for non-existing nodes
/help

PiotrProkop:
I have created PR to address this issue: https://github.com/kubernetes-sigs/node-feature-discovery/pull/1024

PR #1024, opened 2023-01-04
Add NRT garbage collector

PiotrProkop:
Fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/928

I am still missing RBAC configuration. Also I am wondering whether this garbage collector should only be started when some flag is set to `true`. Let me know what you think.


pkg/nfd-master/nfd-nrt-gc.go
```diff
@@ -0,0 +1,131 @@
...
+		return err
+	}
+
+	for _, nrt := range nrts.Items {
+		node := &corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: nrt.GetName(),
+			},
+		}
+		klog.V(2).Infof("adding node %s to inexer", node.GetName())
```
eero-t, line 97: inexer -> indexer

pkg/nfd-master/nfd-nrt-gc.go
```diff
@@ -0,0 +1,131 @@
...
+		return
+	}
+	klog.Infof("deleting NodeResourceTopology for node: %s", node.GetName())
+	if err := n.topoClient.TopologyV1alpha1().NodeResourceTopologies().Delete(context.TODO(), node.GetName(), metav1.DeleteOptions{}); err != nil {
+		if errors.IsNotFound(err) {
+			klog.V(2).Infof("NodeResourceTopology %s not found, ommiting deletion", object)
+			return
+		} else {
+			klog.ErrorS(err, "failed to delete NodeResourceTopology", "nodeName", node.GetName())
+			return
```
eero-t, line 79: Couldn't `NodeResourceTopology` disappear before this is called (e.g. due to some other cluster admin activity happening at the same time) / isn't this all async i.e. why this is an error instead of a warning?

fromanirh (commented):
Thanks for implementing this!
very initial pass, only shallow review.

Some open questions for discussion, I don't have strong opinions here yet

- should this be (yet another) separate tiny component or do we want to merge this on nfd-master?
- I concur the GC should be enabled only if a flag is given - note: if we move the GC to its own tiny server we get this for free
- and probably the resync loop interval as well should be tunable using an option
- we may want to add, perhaps in the future, a check if a NRT is orphaned, meaning there is no daemon (as in daemonset) for a given node. So even if we node is still here, we may want to delete that NRT (this may be governed by another option)

eero-t:
>     * I concur the GC should be enabled only if a flag is given - note: if we move the GC to its own tiny server we get this for free
>     * and probably the resync loop interval as well should be tunable using an option

Another option would be running GC only e.g. at NFD master startup & exit.
 
>     * we may want to add, perhaps in the future, a check if a NRT is orphaned, meaning there is no daemon (as in daemonset) for a given node. So even if we node is still here, we may want to delete that NRT (this may be governed by another option)

fromanirh:
> > ```
> > * I concur the GC should be enabled only if a flag is given - note: if we move the GC to its own tiny server we get this for free
> > * and probably the resync loop interval as well should be tunable using an option
> > ```
> 
> Another option would be running GC only e.g. at NFD master startup & exit.

Designwise I have no objections, but in practical terms AFAIU nfd-master is supposed to stay running for very long periods, possibly forever (bar bugs/upgrades), and in this case the actual GC will happen very seldomly.

fromanirh:
> Thanks for implementing this! very initial pass, only shallow review.
> 
> Some open questions for discussion, I don't have strong opinions here yet
> 
>     * should this be (yet another) separate tiny component or do we want to merge this on nfd-master?

Or, we can add this option to nfd-topology-updater, and add manifests to reuse this binary as GC. It can work like this:

1. we add code to handle GC, but instead of calling it into nfd-master, we call into nfd-topology-updater
2. by default, the GC code is disabled
3. nfd-topology-updater becomes EITHER updater (= like today) or does GC
4. we add new manifests to run nfd-topology-updater as GC agent

benefits:

1. all the NRT handling logic is neatly included into nfd-topology-updater, no spillover on nfd-master
2. very clear when GC is enabled: only if you run `nfd-topology-updater --gc-enabled --gc-period 5m`
3. very clear separation of concerns
4. no new images (but this is also true for nfd-master integration)

@PiotrProkop @marquiz WDYT?
**PLEASE NOTE** I don't really have anything against the current direction and use nfd-master, so I'm not against it. I feel the proposed approach is a little cleaner though.

pkg/nfd-master/nfd-nrt-gc.go
PiotrProkop: My bad, will change this to warning

pkg/nfd-master/nfd-nrt-gc.go
```diff
@@ -0,0 +1,131 @@
...
+limitations under the License.
+*/
+
+package nfdmaster
+
+import (
+	"time"
+
+	topologyclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
+	"golang.org/x/net/context"
```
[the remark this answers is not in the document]
PiotrProkop: hmm my `GoImports` did something weird here. Chaning to stdlib context now.

pkg/nfd-master/nfd-nrt-gc_test.go
```diff
@@ -0,0 +1,174 @@
...
+
+package nfdmaster
+
+import (
+	"testing"
+	"time"
+
+	nrtapi "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha1"
+	faketopologyv1alpha1 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned/fake"
+	"golang.org/x/net/context"
```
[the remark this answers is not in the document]
PiotrProkop: same as above changing

eero-t:
I'm also in favor of moving this to `nfd-topology-updater`.

> > Another option would be running GC only e.g. at NFD master startup & exit.
> 
> Designwise I have no objections, but in practical terms AFAIU nfd-master is supposed to stay running for very long periods, possibly forever (bar bugs/upgrades), and in this case the actual GC will happen very seldomly.

One more option would have been doing GC only when notified, but to my amazement, k8s still does not have an API for that, despite 6 years of discussions:
* https://github.com/kubernetes/kubernetes/issues/24957
* https://github.com/kubernetes/enhancements/issues/1977
* https://github.com/kubernetes/enhancements/pull/1995

PiotrProkop:
Ok, I'll move `nrtGC` to `nfd-topology-updater`, it can be enabled when started with:
```
$ nfd-topology-updater --gc-enabled
```
I think we should also expose the informer's `defaultResync` period but I can' think of a good name for it `gc-period` will be misleading, because removing old NRTs happens on startup and then `nrtGC` is listening for `Node` deletion events and removes corresponding NRT object.

pkg/nfd-master/nfd-nrt-gc.go
```diff
@@ -0,0 +1,131 @@
...
+
+	node, ok := obj.(*corev1.Node)
+	if !ok {
+		klog.Errorf("cannot convert %v to v1.Node", object)
+		return
+	}
+	klog.Infof("deleting NodeResourceTopology for node: %s", node.GetName())
+	if err := n.topoClient.TopologyV1alpha1().NodeResourceTopologies().Delete(context.TODO(), node.GetName(), metav1.DeleteOptions{}); err != nil {
+		if errors.IsNotFound(err) {
+			klog.V(2).Infof("NodeResourceTopology %s not found, omitting deletion", object)
```
marquiz, line 75: I suggest to not print the whole node object, just the node name should be enough. I also suggest to change the wording slightly.

```suggestion
			klog.V(2).Infof("NodeResourceTopology for node %q not found", object.Name)
```

pkg/nfd-master/nfd-nrt-gc.go
```diff
@@ -0,0 +1,131 @@
...
+		klog.V(2).Infof("found stale NodeResourceTopology for node: %s ", object)
+		obj = deletedFinalStateUnknown.Obj
+	}
+
+	node, ok := obj.(*corev1.Node)
+	if !ok {
+		klog.Errorf("cannot convert %v to v1.Node", object)
+		return
+	}
+	klog.Infof("deleting NodeResourceTopology for node: %s", node.GetName())
```
marquiz, line 72: I think we should not print this unless we're really deleting an object. In most setups topology-updater is not even enabled and these log messages are probably confusing as no NRT objects have ever been created for the cluster.

I suggest to fix this either by:
1. Print the info message only after an object was really deleted
1. First try to `Get` the NRT object and if it exists, then print a message that we're about to delete it

pkg/nfd-master/nfd-nrt-gc.go
marquiz: But isn't that race (i.e. `NodeResourceTopology` disappearing by some parallel process) caught by the `if errors.IsNotFound(err)` above? So logging an error would make sense(?)

marquiz (commented):
Thanks @PiotrProkop for PR number 2^10 😄 I left some code comments. Thanks also @fromanirh and @eero-t for the reviews
/ok-to-test

About @fromanirh's suggestionm

> Or, we can add this option to nfd-topology-updater, and add manifests to reuse this binary as GC. It can work like this:
> 1. we add code to handle GC, but instead of calling it into nfd-master, we call into nfd-topology-updater
> 1. by default, the GC code is disabled
> 1. nfd-topology-updater becomes EITHER updater (= like today) or does GC
> 1. we add new manifests to run nfd-topology-updater as GC agent

> benefits:
> 1. all the NRT handling logic is neatly included into nfd-topology-updater, no spillover on nfd-master
> 1. very clear when GC is enabled: only if you run nfd-topology-updater --gc-enabled --gc-period 5m
> 1. very clear separation of concerns
> 1. no new images (but this is also true for nfd-master integration)

I kinda like this idea 👍 For the reasons (benefits) you mentioned. I think the nrt-gc could live in nfd-master but your suggestion would make it cleaner and also simpler/more understandable to deploy e.g. if one wants to deploy only topology-updater, without the nfd labeler at all.

Then there's the question whether it should be a separate tiny binary (as @fromanirh suggested in some other comment above) or bolt this into the current topology-updater binary under some separate cmdline flag. I don't have strong opinion about that but maybe a separate super simple binary would be cleaner. Any though? We need separate deployment files (deployment yamls, rbac rules etc) anyway in either case.

pkg/nfd-master/nfd-nrt-gc.go
PiotrProkop: fixed

pkg/nfd-master/nfd-nrt-gc.go
PiotrProkop: reworded

PiotrProkop:
@marquiz Thanks for the review! For now I have move this GC to `nfd-topology-updater` and have it enabled via `gc-enabled` flag(I have also added proper yaml manifest to deploy it). I can move it to seperate binary if necessary. 
I have also added triggering `gc` based on `gc-period` flag. In my testing on kind cluster when I removed node by just simply running:
```
$ kubect delete node worker`
```
`topology-updater` was still alive for some period of time and recreated NRT custom resources, so this periodic trigger should mitigate this in real clusters 😄

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -94,6 +96,28 @@ func NewTopologyUpdater(args Args, resourcemonitorArgs resourcemonitor.Args, pol
 // one request if OneShot is set to 'true' in the updater args.
 func (w *nfdTopologyUpdater) Run() error {
 	klog.Infof("Node Feature Discovery Topology Updater %s", version.Get())
+	if w.args.GCEnabled {
```
fromanirh, line 99: I'd dare more. I'd go as far as renaming this function `RunUpdater` and add `RunGarbageCollector` which includes only the code you added here. Then, in the calling site, we can make logic on `w.args.GCEnabled`.

fromanirh (commented):
Only glanced over the latest changes. Looks great, another (relatively minor) design/arch question inside. I'll have a proper review later, but I really like the direction here.
Thanks!

pkg/nfd-topology-updater/nfd-topology-updater.go
marquiz: Looking at this now I's suggest to make this a totally separate executable, `cmd/nfd-topology-updater-gc` or whatever. The gc functionality is basically a totally separate entrypoint and most of the cmdline flags are ignored for the gc (only kubeconfig is common, afaiu). WDYT?

fromanirh: seems like the best approach (and simpler overall)

PiotrProkop: agreed, I will move this to standalone binary.

marquiz:
One more note @PiotrProkop, we need to add Helm deployment files, too, before merging this

PiotrProkop:
@marquiz no problem, I'll try to do it early next week.

marquiz:
> @marquiz no problem, I'll try to do it early next week.

Sounds good.

Oh, and I think we want to have minima documentation for that, too. We could have it as just a separate sub-section under topology-updater in `docs/get-started/introduction.md` and `docs/usage/nfd-topology-updater.md`. For completeness, also cmdline reference would be nice (copy-paste-edit `docs/reference/topology-updater-commandline-reference.md` should be easy enough)

PiotrProkop:
@marquiz @ffromani I have move `nrtGC` to standalone binary, modified kustomize and helm. I am only missing docs but will add them after review 😄

ffromani:
> @marquiz @ffromani I have move `nrtGC` to standalone binary, modified kustomize and helm. I am only missing docs but will add them after review smile

sounds good to me. This work is high on my review list, will get to this shortly.

marquiz (commented):
Thanks @PiotrProkop, looks already really good to me.

I left some code comments. Also, could you be slightly more verbose in the commit message. I.e. add some body text, e.g. spelling out *NRT* and describing that we're adding a new daemon or whatever you think is appropriate.

pkg/nfd-topology-gc/nfd-nrt-gc.go
```diff
@@ -0,0 +1,219 @@
...
+)
+
+// Args are the command line arguments
+type Args struct {
+	GCPeriod time.Duration
+
+	Kubeconfig string
+}
+
+type NRTGarbageCollector interface {
```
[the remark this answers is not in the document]
PiotrProkop: Good idea, I have moved this code between packages 3 times already and my naming can be a bit outdated now 😄

deployment/helm/node-feature-discovery/templates/topology-gc.yaml
```diff
@@ -0,0 +1,66 @@
...
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        command:
+          - "nfd-topology-gc"
+        args:
+          {{- if .Values.topologyGC.interval | empty | not }}
+          - "-gc-interval={{ .Values.topologyGC.interval }}"
+          {{- else }}
+          - "-sleep-interval=1h"
```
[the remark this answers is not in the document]
PiotrProkop: typo, good catch!

deployment/base/topologyupdater-gc/topologyupdater-gc.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      dnsPolicy: ClusterFirstWithHostNet
+      serviceAccount: nfd-topology-gc
+      containers:
+        - name: nfd-topology-gc
+          image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
+          imagePullPolicy: Always
+          command:
+            - "nfd-topology-gc"
+            - "-gc-interval=10s"
```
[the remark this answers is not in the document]
PiotrProkop: Yeah, it's wrong. I was using this value for testing. Will change it.

pkg/nfd-topology-gc/nfd-nrt-gc.go
```diff
@@ -0,0 +1,219 @@
...
+		} else {
+			klog.Warningf("failed to delete NodeResourceTopology for node %s: %s", node.GetName(), err.Error())
+			return
+		}
+	}
+	klog.Infof("NodeResourceTopology for node %s has been deleted", node.GetName())
+}
+
+// populateNodeIndexer populates cache with NRTs so we know which one to delete on first synchronization
+func (n *nrtGarbageCollector) populateNodeIndexer(indexer cache.Indexer) error {
```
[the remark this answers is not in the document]
ffromani: > Because we want to remove obsolete objects during first synchronization of indexer. The indexer will list current objects, compare it with what it have in store and run `DeleteHandler` on objects that are not present in k8s. We don't have to list `Nodes` and we can entire rely on indexer/lister wrt to Nodes. It's just an optimization. If it's more clear I can just list Nodes and compare it with NRTs and remove obsolete objects without populating cache.

Thanks for the explanation. I for myself appreciate the effort and the optimization, but in this case, especially considering the GC binary goals, I'd go for the obvious, trivial solution, albeit less efficient.
In general, if the GC process becomes a bottleneck or consumes a nontrivial amount of resources, some horribly broken already happened,  so I'm not worried about performances just yet.

deployment/base/topologyupdater-gc/topologyupdater-gc.yaml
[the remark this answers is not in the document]
ffromani: agreed. 1h everywhere seems a good starting value

pkg/nfd-topology-gc/nfd-nrt-gc.go
```diff
@@ -0,0 +1,219 @@
...
+// periodicGC runs garbage collector at every gcPeriod to make sure we haven't missed any node
+func (n *nrtGarbageCollector) periodicGC(gcPeriod time.Duration) {
+	indexer := n.factory.Core().V1().Nodes().Informer().GetIndexer()
+	gcTrigger := time.NewTicker(gcPeriod)
+	for {
+		select {
+		case <-gcTrigger.C:
+			klog.Infof("Running periodic GC")
+			objects := indexer.List()
+			nodes := map[string]struct{}{}
```
ffromani, line 143: (minor) use `k8s.io/apimachinery/pkg/util/sets` sets.String? we vendor` apimachinery/pkg/util anyway already

pkg/nfd-topology-gc/nfd-nrt-gc.go
```diff
@@ -0,0 +1,219 @@
...
+				klog.Warningf("cannot list NRTs %s", err.Error())
+				break
+			}
+			for _, nrt := range nrts.Items {
+				key, err := cache.MetaNamespaceKeyFunc(&nrt)
+				if err != nil {
+					klog.Warningf("cannot create key for %v: %s", nrt, err.Error())
+					continue
+				}
+				if _, ok := nodes[key]; !ok {
```
ffromani, line 163: (not-so-minor but not blocking either) in this case I'd try to extract a helper to delete the NRT given a nodeName, to be reused both here and in `deleteNodeHandler`

ffromani (commented):
nice job @PiotrProkop ! just a handful minorish comments in addition to what @marquiz already highlighted.
I'd also recommend to add a bit more context in the commit messages, spending few lines to record the why and the rationale

pkg/nfd-topology-gc/nfd-nrt-gc.go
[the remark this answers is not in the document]
marquiz: I'm inclined to agree with @ffromani here. I'd also prefer keeping the GC simple, just for the sake of ease of future maintainability and readability. Especially the startup which is done only once (or very rarely) doing a small amount of more work shouldn't really be a problem.

marquiz:
One question of the deployment. Do we want to have `topology-gc` as a separate overlay in kustomize (and an independent setting in the Helm chart) or should it be always deployed with topology-updater (with possible opt-out setting in Helm? Just asking, you know better @PiotrProkop and @ffromani 😊

ffromani:
> One question of the deployment. Do we want to have `topology-gc` as a separate overlay in kustomize (and an independent setting in the Helm chart) or should it be always deployed with topology-updater (with possible opt-out setting in Helm? Just asking, you know better @PiotrProkop and @ffromani blush

regarding deployment, I'm still gathering thoughts, but for the sake of the UX I think it's slightly better to always deploy the GC and let the users opt-out. This is because this seems the best default configuration, the opposite variant seems a bit less friendly. Im open for discussion though.

PiotrProkop:
@marquiz and @ffromani thanks for review. I tried to address all comments. As for deployment I agree that bundling `topology-gc` with `topology-updater` is more friendly for end users, so I have implemented that in the newest patch.

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -422,6 +422,44 @@ topologyUpdater:
...
+topologyGC:
+  enable: false
+  replicaCount: 1
+
+  serviceAccount:
+    create: false
+    annotations: {}
+    name:
+  rbac:
+    create: false
```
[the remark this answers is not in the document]
PiotrProkop: added

pkg/nfd-topology-gc/nfd-nrt-gc.go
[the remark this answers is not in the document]
PiotrProkop: simplified.

pkg/nfd-topology-gc/nfd-nrt-gc.go
PiotrProkop: Good idea!

pkg/nfd-topology-gc/nfd-nrt-gc.go
PiotrProkop: done.

deployment/base/topologyupdater-gc/topologyupdater-gc.yaml
```diff
@@ -0,0 +1,24 @@
...
+    spec:
+      dnsPolicy: ClusterFirstWithHostNet
+      serviceAccount: nfd-topology-gc
+      containers:
+        - name: nfd-topology-gc
+          image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
+          imagePullPolicy: Always
+          command:
+            - "nfd-topology-gc"
+            - "-gc-interval=1h"
```
marquiz, line 24: I'd suggest to drop this as it matches with the default interval. If we later want to change the default we only need to patch it in `main.go`. WDYT?

deployment/helm/node-feature-discovery/templates/clusterrolebinding.yaml
```diff
@@ -32,3 +32,21 @@ subjects:
   name: {{ include "node-feature-discovery.topologyUpdater.serviceAccountName" . }}
   namespace: {{ include "node-feature-discovery.namespace" .  }}
 {{- end }}
+
+---
+{{- if and (eq .Values.topologyGC.enable true) (eq .Values.topologyGC.serviceAccount.create true) }}
```
marquiz, line 37: should be `rbac.create` instead of `serviceAccount.create`, right?

deployment/helm/node-feature-discovery/templates/clusterrole.yaml
```diff
@@ -66,3 +66,34 @@ rules:
   - get
   - update
 {{- end }}
+
+---
+{{- if and (eq .Values.topologyGC.enable true) (eq .Values.topologyGC.serviceAccount.create true) }}
```
marquiz, line 71: should be `rbac.create` instead of `serviceAccount.create`, right?

deployment/helm/node-feature-discovery/templates/topology-gc.yaml
```diff
@@ -0,0 +1,66 @@
...
+            fieldRef:
+              fieldPath: spec.nodeName
+        command:
+          - "nfd-topology-gc"
+        args:
+          {{- if .Values.topologyGC.interval | empty | not }}
+          - "-gc-interval={{ .Values.topologyGC.interval }}"
+          {{- else }}
+          - "-gc-interval=1h"
+          {{- end }}
```
marquiz, line 48: Here also we could just drop this and resort to the default from `main.go`, WDYT?

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -422,6 +422,44 @@ topologyUpdater:
...
+
+  interval: 1h
+
+  podSecurityContext: {}
+  securityContext:
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop: [ "ALL" ]
+    readOnlyRootFilesystem: true
+    runAsUser: 0
```
marquiz, line 444: There should be no need to run GC as root

```suggestion
    runAsNonRoot: true
```

marquiz (commented):
Good work @PiotrProkop, looking good! I had a few more comments on the code.

Also, for consistency I'd suggest to rename
- `deployment/base/topologyupdater-gc/` -> `deployment/base/topology-gc/`
- `deployment/base/rbac-topologyupdater-gc/` -> `deployment/base/rbac-topology-gc/`
- and also the files within the directories^, accordingly

deployment/base/topologyupdater-gc/topologyupdater-gc.yaml
PiotrProkop: 👍

deployment/helm/node-feature-discovery/templates/clusterrolebinding.yaml
PiotrProkop: good catch

PiotrProkop:
@marquiz all comments resolved. Thanks!

deployment/helm/node-feature-discovery/templates/clusterrolebinding.yaml
```diff
@@ -32,3 +32,21 @@ subjects:
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ include "node-feature-discovery.fullname" . }}-topology-gc
+  labels:
+    {{- include "node-feature-discovery.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ include "node-feature-discovery.fullname" . }}-topology-updater
```
marquiz, line 47:
```suggestion
  name: {{ include "node-feature-discovery.fullname" . }}-topology-gc
```

PiotrProkop: :+1:

cmd/nfd-topology-gc/main_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"time"
+
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestArgsParse(t *testing.T) {
+	Convey("When parsing command line arguments", t, func() {
+		flags := flag.NewFlagSet(ProgramName, flag.ExitOnError)
+
+		Convey("When valid -gc-inteval is specified", func() {
```
ffromani, line 31: (only if you resubmit): typo "inteval"

cmd/nfd-topology-gc/main_test.go
```diff
@@ -0,0 +1,41 @@
...
+			args := parseArgs(flags,
+				"-gc-interval=30s")
+
+			Convey("args.sources is set to appropriate values", func() {
```
ffromani, line 35: ((only if you resubmit) either generalize the text or replace "sources" with "GCPeriod"

pkg/nfd-topology-gc/nfd-nrt-gc.go
```diff
@@ -0,0 +1,195 @@
...
+		DeleteFunc: n.deleteNodeHandler,
+	}); err != nil {
+		return err
+	}
+
+	// start informers
+	n.factory.Start(n.stopChan)
+	n.factory.WaitForCacheSync(n.stopChan)
+
+	klog.Infof("Running initial GC")
```
ffromani, line 173: (only if you resubmit) drop "initial", so it reads "Running GC"

ffromani (commented):
/lgtm
Thanks! Only minor nits, please consider them only if you resubmit for other reasons. Once @marquiz is happy we can merge.

marquiz:
We still need the docs 😅

And one more thing I started to ponder about the Helm deployment. It would be nice for the Helm to also deploy topology-gc by default (if topology-updater is enabled). Probably the nicest/easiest way to do this would be to simply deploy topology-gc only if topology-updater is enabled, too. That is, deploy if `topologyUpdater.enabled==true && topologyGC.enabled==true`. This would prevent stand-alone deployment of topology-gc (i.e. without topology-updater) but I don't think that would be an interesting use case in any practical scenario. I think otherwise it gets a bit hairier with the `{{ if }}` statements without much  practical benefit. WDYT?

PiotrProkop:
@marquiz I remember about docs 😄 I think it is a sane default to deploy `topologyGC` only if `topologyUpdater` is enabled. We can always change it in the future to something more sophisticated if someone raise issue about that. I'll change helm chart, add docs and apply @ffromani review right now.

PiotrProkop:
@marquiz @ffromani docs added and review applied.

deployment/helm/node-feature-discovery/templates/clusterrole.yaml
```diff
@@ -66,3 +66,34 @@ rules:
   - get
   - update
 {{- end }}
+
+---
+{{- if and (eq .Values.topologyGC.enable true) (eq .Values.topologyGC.rbac.create true) ( eq .Values.topologyUpdater.enable true ) }}
```
marquiz, line 71: nit: could be simplified

```suggestion
{{- if and .Values.topologyGC.enable .Values.topologyGC.rbac.create .Values.topologyUpdater.enable }}
```

deployment/helm/node-feature-discovery/templates/clusterrolebinding.yaml
```diff
@@ -32,3 +32,21 @@ subjects:
   name: {{ include "node-feature-discovery.topologyUpdater.serviceAccountName" . }}
   namespace: {{ include "node-feature-discovery.namespace" .  }}
 {{- end }}
+
+---
+{{- if and (eq .Values.topologyGC.enable true) (eq .Values.topologyGC.rbac.create true) ( eq .Values.topologyUpdater.enable true ) }}
```
marquiz, line 37: nit: could be simplified

```suggestion
{{- if and .Values.topologyGC.enable .Values.topologyGC.rbac.create .Values.topologyUpdater.enable }}
```

deployment/helm/node-feature-discovery/templates/serviceaccount.yaml
```diff
@@ -27,6 +27,21 @@ metadata:
   {{- end }}
 {{- end }}
 
+---
+{{- if and (eq .Values.topologyGC.enable true) (eq .Values.topologyGC.serviceAccount.create true) (eq .Values.topologyUpdater.enable true) }}
```
marquiz, line 31: nit: could be simplified

```suggestion
{{- if and .Values.topologyGC.enable .Values.topologyGC.serviceAccount.create .Values.topologyUpdater.enable }}
```

deployment/helm/node-feature-discovery/templates/topology-gc.yaml
```diff
@@ -0,0 +1,64 @@
+{{- if and ( eq .Values.topologyGC.enable true ) ( eq .Values.topologyUpdater.enable true ) -}}
```
marquiz, line 1: nit: could be simplified
```suggestion
{{- if and .Values.topologyGC.enable .Values.topologyUpdater.enable -}}
```

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -422,6 +422,44 @@ topologyUpdater:
   annotations: {}
   affinity: {}
 
+topologyGC:
+  enable: false
```
marquiz, line 426: In the new setup we want to enable topology-gc by default

```suggestion
  enable: true
```

docs/deployment/helm.md
```diff
@@ -173,5 +173,24 @@ We have introduced the following Chart parameters.
...
 
+### Topology garbage collector parameters
+
+| Name                                          | Type   | Default | description                                                                                                                                                                                           |
+|-----------------------------------------------|--------|---------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `topologyGC.*`                                | dict   |         | NFD Topology Garbage Collector configuration                                                                                                                                                          |
+| `topologyGC.enable`                           | bool   | false   | Specifies whether the NFD Topology Garbage Collector should be created                                                                                                                                |
```
marquiz, line 181: Enable by default. Also mention topologyupdater

```suggestion
| `topologyGC.enable`                           | bool   | true   | Specifies whether the NFD Topology Garbage Collector should be created, only takes effect when `topologyUpdater.enable` is `true`                                                                                                                               |
```

docs/reference/topology-gc-commandline-reference.md
```diff
@@ -0,0 +1,46 @@
+---
+title: "Topology Garbage Collector Cmdline Reference"
+layout: default
+sort: 4
```
marquiz, line 4: Manual sorting 🙄, let's put this as the last one
```suggestion
sort: 6
```

marquiz (commented):
Thanks for the update @PiotrProkop. I had a few more comments but after those I think we should get this merged

deployment/helm/node-feature-discovery/values.yaml
PiotrProkop: Is it make sense to enable `topologyGC` by default when `topologyUpdater` is disabled by default(as well as creating NRT CRDs)?

PiotrProkop: My bad :) forgot that we only deploy `GC` when `updater` is enabled

deployment/helm/node-feature-discovery/templates/clusterrole.yaml
PiotrProkop: done

deployment/helm/node-feature-discovery/templates/clusterrolebinding.yaml
PiotrProkop: done

deployment/helm/node-feature-discovery/templates/topology-gc.yaml
PiotrProkop: done

docs/deployment/helm.md
PiotrProkop: done

docs/usage/nfd-topology-gc.md
```diff
@@ -0,0 +1,29 @@
+---
+title: "NFD-Topology-Garbage-Collector"
+layout: default
+sort: 5
```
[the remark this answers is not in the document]
PiotrProkop: done

deployment/helm/node-feature-discovery/templates/serviceaccount.yaml
PiotrProkop: done

docs/reference/topology-gc-commandline-reference.md
PiotrProkop: done

docs/usage/nfd-topology-gc.md
```diff
@@ -0,0 +1,29 @@
...
+This service watches for Node deletion events and removes NodeResourceTopology
+objects upon them. It is also running periodically to make sure no event was
+missed or NodeResourceTopology object was created without corresponding worker
+node. The default garbage collector interval is set to 1h which is the value
+when no -gc-interval is specified.
+
+## Topology-Garbage-Collector Configuration
+
+In Helm deployments,
+[Topology Garbage Collector parameters](../deployment/helm.md#topology-garbage-collector-parameters)
```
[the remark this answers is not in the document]
PiotrProkop: done

marquiz:
From https://github.com/kubernetes-sigs/node-feature-discovery/pull/1024#discussion_r1066741607

> Is it make sense to enable topologyGC by default when topologyUpdater is disabled by default(as well as creating NRT CRDs)?

A side note, talking about CRDs, how do you guys normally deploy the NRT CRD? Now this CRD is not properly/normally managed by the NFD Helm chart. It is not deployed by default (only created when `topologyUpdater.createCRDs==true`). Is this fine or should we move it under `deployment/helm/node-feature-discovery/crds/` so that it would  be managed by Helm and always created (unless `--skip-crds` is specified). WDYT @ffromani @PiotrProkop ?  Anyway, this is a separate issue and should be addressed in a separate PR.

PiotrProkop:
@marquiz To be honest I don't use helm to deploy NFD and CRDs but kustomize 😄 but I agree that `NRT` CRDs should be probably moved to `crds` directory and installed even when `topologyUpdater.enable` is set to false. But we should create seperate issue/PR for that.

ffromani:
> From [#1024 (comment)](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1024#discussion_r1066741607)
> 
> > Is it make sense to enable topologyGC by default when topologyUpdater is disabled by default(as well as creating NRT CRDs)?
> 
> A side note, talking about CRDs, how do you guys normally deploy the NRT CRD? Now this CRD is not properly/normally managed by the NFD Helm chart. It is not deployed by default (only created when `topologyUpdater.createCRDs==true`). Is this fine or should we move it under `deployment/helm/node-feature-discovery/crds/` so that it would be managed by Helm and always created (unless `--skip-crds` is specified). WDYT @ffromani @PiotrProkop ? Anyway, this is a separate issue and should be addressed in a separate PR.

I fully agree deploying the CRD is a separate topic. For the record:
- I kinda assumed nfd-topology-updater assumes the CRD is already in the cluster. I think at this stage is fair game to have this assumption yet, possibly we need to make this explicit in the docs (I'll check later)
- in general we still need to figure out who owns the CRD (it's a cross-component/cross-subproject topic), meaning which component is in charge of deploying the CRD and making sure it's the right version. I don't have strong feelings not a clear idea yet and I would surely not mind if NFD agrees to take ownership here, we can discuss separately.

marquiz:
> * I kinda assumed nfd-topology-updater assumes the CRD is already in the cluster. I think at this stage is fair game to have this assumption yet, possibly we need to make this explicit in the docs (I'll check later)

K, that's fine. Just started to think about that 😊 And yes, we should document clearly what is the expected method of getting the CRD deployed

> * in general we still need to figure out who owns the CRD (it's a cross-component/cross-subproject topic), meaning which component is in charge of deploying the CRD and making sure it's the right version. I don't have strong feelings not a clear idea yet and I would surely not mind if NFD agrees to take ownership here, we can discuss separately.

Yeah, I agree.

But let's take these separately, not belonging into this PR

marquiz (approved):
Thanks @PiotrProkop for working on this. I think this PR is ready to get in

ffromani:
> Thanks @PiotrProkop for working on this. I think this PR is ready to get in

agreed, thanks @PiotrProkop !

ffromani:
/lgtm
all my comments were addressed

PR #1028, opened 2023-01-11
e2e: init docker image

Tal-or:
The docker image that used during e2e test
composed of repo and tag flags that are
passed to the test itself.

The problem is that the docker image initialized
before the flags are parsed. Hence, it will always contains the default flags value.
Initialize the flag in later stage, fixing the issue.

Also, moving the global variables to `e2e_test.go` since it commonly used by all tests.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -18,6 +18,7 @@ package e2e
 
 import (
 	"flag"
+	"fmt"
 	"math/rand"
 	"os"
 	"testing"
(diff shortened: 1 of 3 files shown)


Tal-or:
@marquiz I would also append the `test` suffix for the test files:
`
node_feature_discovery_test.go
`
`
topology_updater_test.go
`
so the global variables would be recognized in modern IDE's such as Goland.

test/e2e/e2e_test.go
```diff
@@ -37,9 +44,15 @@ func handleFlags() {
 	flag.Parse()
 }
 
+// must be called after flags are parsed
+func initDockerImage() {
+	dockerImage = fmt.Sprintf("%s:%s", *dockerRepo, *dockerTag)
+}
```
marquiz, line 50: Maybe we should just make `dockerImage` a function, would drop the kinda useless variable. WDYT?

```suggestion
// must be called after flags are parsed
func dockerImage() string {
	return fmt.Sprintf("%s:%s", *dockerRepo, *dockerTag)
}
```

marquiz (commented):
Uh-oh, thanks @Tal-or for catching this one. I had just one suggestion to drop the "caching" variable for good

ping @fmuyassarov

marquiz:
> I would also append the test suffix for the test files:
node_feature_discovery_test.go
topology_updater_test.go
so the global variables would be recognized in modern IDE's such as Goland.

Sure, go ahead, I don't have anything against that. But probably in a separate PR (?)

test/e2e/e2e_test.go
Tal-or: Sure. let me just fix all the reference in the code

Tal-or:
> > I would also append the test suffix for the test files:
> > node_feature_discovery_test.go
> > topology_updater_test.go
> > so the global variables would be recognized in modern IDE's such as Goland.
> 
> Sure, go ahead, I don't have anything against that. But probably in a separate PR (?)

Done https://github.com/kubernetes-sigs/node-feature-discovery/pull/1029

fmuyassarov (commented):
/ok-to-test
/lgtm
Thanks for the fix.

marquiz:
Ach, needs rebase 🙃 Otherwise LGTM

Tal-or:
> Ach, needs rebase upside_down_face Otherwise LGTM

Done

marquiz (approved):
Thx @Tal-or!
/lgtm

PR #1039, opened 2023-01-17
[release-0.12] e2e: init docker image

marquiz:
The docker image that used during e2e test
composed of repo and tag flags that are
passed to the test itself.

The problem is that the docker image initialized
before the flags are parsed. Hence, it will always contains the default flags value.
Moving the variable into a separate function, fixing the issue.

Also, moving the global variables to `e2e_test.go` since it commonly used by all tests.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -18,6 +18,7 @@ package e2e
 
 import (
 	"flag"
+	"fmt"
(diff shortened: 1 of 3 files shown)


marquiz:
Manually backported from #1028

fmuyassarov (commented):
/lgtm

PR #1045, opened 2023-01-27
deployment: disable service links in NFD master pod

### LINKED ISSUE #883: Disable enableServiceLinks feature in NFD master pod

`spec.enableServiceLinks: false` to _nfd-master_ service pod spec.

**Why is this needed**:

See: https://github.com/kubernetes/kubernetes/issues/92615

(Kubernetes itself is not going to default that to false as that would break its core API compatibility, so each service pod should set that option when service is accessed by its "<service>[.<namespace>]" DNS name instead of env var IP address.)

AhmedGrati:
Resolves #883.
Signed-off-by: AhmedGrati <ahmedgrati1999@gmail.com>

### DIFF
--- a/deployment/base/master-worker-combined/master-worker-daemonset.yaml
+++ b/deployment/base/master-worker-combined/master-worker-daemonset.yaml
@@ -14,6 +14,7 @@ spec:
         app: nfd
     spec:
       serviceAccount: nfd-master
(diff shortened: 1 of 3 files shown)


marquiz (approved):
Thanks @AhmedGrati for the PR!
/ok-to-test

PR #1046, opened 2023-01-30
Replace gRPC health probe utility with k8s built-in health probe

fmuyassarov:
Kubernetes [1.23](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-grpc-liveness-probe) has introduced native health probes for gRPC which can replace
`grpc_health_probe` utility. This commit removes baking in `grpc_health_probe`
binary into the image and updates related health checks to use k8s native gRPC.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,14 +5,6 @@ ARG BASE_IMAGE_MINIMAL
 # Build node feature discovery
 FROM ${BUILDER_IMAGE} as builder
 
-# Build and install the grpc-health-probe binary
-RUN GRPC_HEALTH_PROBE_VERSION=v0.4.19 && \
-    go install -tags osusergo,netgo -ldflags -extldflags=-static \
-        github.com/grpc-ecosystem/grpc-health-probe@${GRPC_HEALTH_PROBE_VERSION} \
-        # Rename it as it's referenced as grpc_health_probe in the deployment yamls
-        # and in its own project https://github.com/grpc-ecosystem/grpc-health-probe
-        && mv /go/bin/grpc-health-probe /go/bin/grpc_health_probe
-
 # Get (cache) deps in a separate layer
 COPY go.mod go.sum /go/node-feature-discovery/
 
--- a/deployment/base/master-worker-combined/master-worker-daemonset.yaml
+++ b/deployment/base/master-worker-combined/master-worker-daemonset.yaml
@@ -21,13 +21,13 @@ spec:
         image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
         imagePullPolicy: Always
         livenessProbe:
-          exec:
-            command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
+          grpc:
+            port: 8080
           initialDelaySeconds: 10
           periodSeconds: 10
         readinessProbe:
-          exec:
-            command: ["/usr/bin/grpc_health_probe", "-addr=:8080"]
+          grpc:
+            port: 8080
           initialDelaySeconds: 5
           periodSeconds: 10
           failureThreshold: 10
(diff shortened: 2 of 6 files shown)


ArangoGutierrez (approved):
We do register the native gRPC protocol for health checks see: 
- https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/pkg/nfd-master/nfd-master.go#L236

/lgtm

marquiz:
Thanks @fmuyassarov for the patch. I think we're in no hurry with this one. There are probably many pre v1.23 clusters out there and I think e.g. AWS EKS still officially supports v1.22 for some time

/hold

fmuyassarov:
> 

Yep good point. Let's keep it here for some time.

ArangoGutierrez:
> Thanks @fmuyassarov for the patch. I think we're in no hurry with this one. There are probably many pre v1.23 clusters out there and I think e.g. AWS EKS still officially supports v1.22 for some time
> 
> /hold

What?  isn't everyone running latest:nightly  in prod?  what a world we live in.....

ArangoGutierrez:
[fmuyassarov](https://github.com/fmuyassarov) let's fix the merge conflict and I'll review again

fmuyassarov:
@ArangoGutierrez sorry for late response on this, rebase is done, but perhaps we can wait a bit more to not breaking possible users of k8s version below v1.23?

ArangoGutierrez:
@marquiz @fmuyassarov should we mark this as milestone for v0.15?

marquiz:
> @marquiz @fmuyassarov should we mark this as milestone for v0.15?

v0.15 might be, K8s v1.22 was EOL'd last autumn

fmuyassarov:
> @marquiz @fmuyassarov should we mark this as milestone for v0.15?

yes, 0.15 sounds good to me.
/milestone v0.15

ArangoGutierrez:
Let's make this part of https://github.com/kubernetes-sigs/node-feature-discovery/issues/1187

ArangoGutierrez:
> > Let's make this part of #1187
> 
> What do you mean? They're not related

If we are moving away from GRPC into the API (IOW Kubernetes native objects),  moving forward with this makes also more sense. 
But 1187 doesn't depend on this.

marquiz:
> If we are moving away from GRPC into the API (IOW Kubernetes native objects), moving forward with this makes also more sense.

But this still leverages gRPC. Just uses the K8s internal support for gRPC probes instead of an external tool

ArangoGutierrez:
> > If we are moving away from GRPC into the API (IOW Kubernetes native objects), moving forward with this makes also more sense.
> 
> But this still leverages gRPC. Just uses the K8s internal support for gRPC probes instead of an external tool

Agree, I was thinking just on the Kubernetes native part, and forgot the implementation

marquiz:
@fmuyassarov @ArangoGutierrez I think we could merge this one. K8s v1.22 was EOL'd nearly a year ago and at least all the bigger CSPs (like GKE, EKS, AKS) have officially stopped supporting it, also OpenShift 4.

ArangoGutierrez:
> @fmuyassarov @ArangoGutierrez I think we could merge this one. K8s v1.22 was EOL'd nearly a year ago and at least all the bigger CSPs (like GKE, EKS, AKS) have officially stopped supporting it, also OpenShift 4.

Needs rebase ;P 
but I support moving forward here

fmuyassarov:
> @fmuyassarov @ArangoGutierrez I think we could merge this one. K8s v1.22 was EOL'd nearly a year ago and at least all the bigger CSPs (like GKE, EKS, AKS) have officially stopped supporting it, also OpenShift 4.

Sounds good. Just rebased.

ArangoGutierrez (approved):
/lgtm

PR #1047, opened 2023-02-01
test: move out unit testing from Dockerfile

fmuyassarov:
Move out running unit tests as part of container image building and 
instead add it into verify.sh script which runs linter, formatter and
other checks. As a future improvement, we could have separate
Prow jobs for fmt, golangci-lint, helm-linter and unit tests separately.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -27,8 +27,6 @@ ARG HOSTMOUNT_PREFIX
 
 RUN make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
 
-RUN make test
-
 # Create full variant of the production image
 FROM ${BASE_IMAGE_FULL} as full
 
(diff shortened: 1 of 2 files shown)


ArangoGutierrez (approved):
Thanks!
Could you create an issue for 
"As a future improvement, we could have separate
Prow jobs for fmt, golangci-lint, helm-linter and unit tests separately."
Otherwise, looks good to me

marquiz (approved):
Thanks @fmuyassarov 👍 Makes sense (and the future improvements too)

fmuyassarov:
> Thanks! Could you create an issue for "As a future improvement, we could have separate Prow jobs for fmt, golangci-lint, helm-linter and unit tests separately." Otherwise, looks good to me

@ArangoGutierrez Sure and sorry, I somehow missed your comment :).

Issue #1048, opened 2023-02-03
Add option to compute the pod set fingerprint to improve NUMA aware scheduling 


jlojosnegros:
**What would you like to be added**:
Add an option to compute the fingerprint of the current pod set,
and report it as a top-level attribute in NRT objects. This is meant to
enable better caching in the scheduler side.
To use such an attribute we first need to update NRT API version to `v1alpha2`.

**Why is this needed**:
To minimize, or even eliminate, the occurrence of `TopologyAffinityError`s TopologyManager and Scheduler should have a synched view of node resources.
However it is hard for the NUMA aware scheduler plugins to track the resource allocation with NUMA zone granularity.
A new reserve plugin implementing a scheduler-side cache will improve this situation. [More info here](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/cache/docs/NUMA-aware-scheduler-side-reserve-plugin.md)
A prerequisite for this plugin to work is that topology updater agents, like NFD, should compute and provide a "node state" in NRT objects along with the resource information. 
A pod set fingerprint is a compact representation of this "node state" so this enhancement will allow NFD to work nicely with this new NUMA-aware scheduler-side cache and improve the NUMA-aware scheduling process.

**Additional Info and Links**
- [RTE PR supporting this same feature](https://github.com/k8stopologyawareschedwg/resource-topology-exporter/pull/109)
- [Numa-aware scheduler-side reserve plugin doc](https://github.com/kubernetes-sigs/scheduler-plugins/blob/master/pkg/noderesourcetopology/cache/docs/NUMA-aware-scheduler-side-reserve-plugin.md)
- [new NRT API version release](https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/issues/30)

ffromani:
thanks for filing this issue!
the "fingerprint" is a opaque string which can be checked for equality or not - no other operation is defined.
We are prepping a [new version of the NRT API](https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/issues/24), the fingerprint should be represented as [top-level attribute](https://github.com/k8stopologyawareschedwg/podfingerprint/pull/7).

jlojosnegros:
[...]

PR #1049, opened 2023-02-03
topology-updater:compute pod set fingerprint

jlojosnegros:
Add an option to compute the fingerprint of the current pod set on each node.
Report this new fingerprint using an attribute in new released `v1alpha2` NRT object.

needs: #1053 
address: #1048 
see: [this issue](https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/issues/30) for more info about new NRT api version

### DIFF
--- a/cmd/nfd-topology-updater/main.go
+++ b/cmd/nfd-topology-updater/main.go
@@ -139,6 +139,7 @@ func initFlags(flagset *flag.FlagSet) (*topology.Args, *resourcemonitor.Args) {
 		"Pod Resource Socket path to use.")
 	flagset.StringVar(&args.ConfigFile, "config", "/etc/kubernetes/node-feature-discovery/nfd-topology-updater.conf",
 		"Config file to use.")
+	flagset.BoolVar(&resourcemonitorArgs.PodSetFingerprint, "pods-fingerprint", false, "Compute and report the pod set fingerprint")
 
 	klog.InitFlags(flagset)
 
--- a/deployment/helm/node-feature-discovery/templates/topologyupdater.yaml
+++ b/deployment/helm/node-feature-discovery/templates/topologyupdater.yaml
@@ -55,6 +55,9 @@ spec:
           - "-key-file=/etc/kubernetes/node-feature-discovery/certs/tls.key"
           - "-cert-file=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
           {{- end }}
+          {{- if .Values.topologyUpdater.podSetFingerprint }}
+          - "-pods-fingerprint"
+          {{- end }}
         volumeMounts:
         - name: kubelet-config
           mountPath: /host-var/lib/kubelet/config.yaml
--- a/go.mod
+++ b/go.mod
@@ -9,6 +9,7 @@ require (
 	github.com/google/go-cmp v0.5.9
 	github.com/jaypipes/ghw v0.8.1-0.20210827132705-c7224150a17e
 	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.1.0
+	github.com/k8stopologyawareschedwg/podfingerprint v0.1.2
 	github.com/klauspost/cpuid/v2 v2.2.4
 	github.com/onsi/ginkgo/v2 v2.4.0
 	github.com/onsi/gomega v1.23.0
@@ -49,6 +50,7 @@ require (
 	github.com/Microsoft/go-winio v0.4.17 // indirect
 	github.com/Microsoft/hcsshim v0.8.22 // indirect
 	github.com/NYTimes/gziphandler v1.1.1 // indirect
+	github.com/OneOfOne/xxhash v1.2.8 // indirect
 	github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d // indirect
 	github.com/antlr/antlr4/runtime/Go/antlr v1.4.10 // indirect
 	github.com/armon/circbuf v0.0.0-20150827004946-bbbad097214e // indirect
--- a/pkg/nfd-topology-updater/nfd-topology-updater.go
+++ b/pkg/nfd-topology-updater/nfd-topology-updater.go
@@ -154,13 +154,13 @@ func (w *nfdTopologyUpdater) Run() error {
 		select {
 		case <-crTrigger.C:
 			klog.Infof("Scanning")
-			podResources, err := resScan.Scan()
-			utils.KlogDump(1, "podResources are", "  ", podResources)
+			scanResponse, err := resScan.Scan()
+			utils.KlogDump(1, "podResources are", "  ", scanResponse.PodResources)
 			if err != nil {
 				klog.Warningf("Scan failed: %v", err)
 				continue
 			}
-			zones = resAggr.Aggregate(podResources)
+			zones = resAggr.Aggregate(scanResponse.PodResources)
 			utils.KlogDump(1, "After aggregating resources identified zones are", "  ", zones)
 			if !w.args.NoPublish {
 				if err = w.updateNodeResourceTopology(zones); err != nil {
--- a/pkg/nfd-topology-updater/nfd-topology-updater.go
+++ b/pkg/nfd-topology-updater/nfd-topology-updater.go
@@ -130,7 +130,7 @@ func (w *nfdTopologyUpdater) Run() error {
 
 	var resScan resourcemonitor.ResourcesScanner
 
-	resScan, err = resourcemonitor.NewPodResourcesScanner(w.resourcemonitorArgs.Namespace, podResClient, w.apihelper)
+	resScan, err = resourcemonitor.NewPodResourcesScanner(w.resourcemonitorArgs.Namespace, podResClient, w.apihelper, w.resourcemonitorArgs.PodSetFingerprint)
 	if err != nil {
 		return fmt.Errorf("failed to initialize ResourceMonitor instance: %w", err)
 	}
@@ -163,7 +163,7 @@ func (w *nfdTopologyUpdater) Run() error {
 			zones = resAggr.Aggregate(scanResponse.PodResources)
 			utils.KlogDump(1, "After aggregating resources identified zones are", "  ", zones)
 			if !w.args.NoPublish {
-				if err = w.updateNodeResourceTopology(zones); err != nil {
+				if err = w.updateNodeResourceTopology(zones, scanResponse); err != nil {
 					return err
 				}
 			}
@@ -188,7 +188,7 @@ func (w *nfdTopologyUpdater) Stop() {
 	}
 }
 
-func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneList) error {
+func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneList, scanResponse resourcemonitor.ScanResponse) error {
 	cli, err := w.apihelper.GetTopologyClient()
 	if err != nil {
 		return err
@@ -205,6 +205,8 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 			Attributes:       createTopologyAttributes(w.nodeInfo.tmPolicy, w.nodeInfo.tmScope),
 		}
 
+		updateAttributes(&nrtNew.Attributes, scanResponse.Attributes)
+
 		_, err := cli.TopologyV1alpha2().NodeResourceTopologies().Create(context.TODO(), &nrtNew, metav1.CreateOptions{})
 		if err != nil {
 			return fmt.Errorf("failed to create NodeResourceTopology: %w", err)
@@ -216,6 +218,7 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 
 	nrtMutated := nrt.DeepCopy()
 	nrtMutated.Zones = zoneInfo
+	updateAttributes(&nrtMutated.Attributes, scanResponse.Attributes)
 
 	nrtUpdated, err := cli.TopologyV1alpha2().NodeResourceTopologies().Update(context.TODO(), nrtMutated, metav1.UpdateOptions{})
 	if err != nil {
@@ -259,5 +262,24 @@ func createTopologyAttributes(policy string, scope string) v1alpha2.AttributeLis
 			Name:  TopologyManagerScopeAttributeName,
 			Value: scope,
 		},
+	}
+}
+
+func updateAttribute(attrList *v1alpha2.AttributeList, attrInfo v1alpha2.AttributeInfo) {
+	if attrList == nil {
+		return
+	}
+
+	for idx := range *attrList {
+		if (*attrList)[idx].Name == attrInfo.Name {
+			(*attrList)[idx].Value = attrInfo.Value
+			return
+		}
+	}
+	*attrList = append(*attrList, attrInfo)
+}
+func updateAttributes(lhs *v1alpha2.AttributeList, rhs v1alpha2.AttributeList) {
+	for _, attr := range rhs {
+		updateAttribute(lhs, attr)
 	}
 }
--- a/pkg/nfd-topology-updater/nfd-topology-updater_test.go
+++ b/pkg/nfd-topology-updater/nfd-topology-updater_test.go
@@ -1,0 +1,112 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nfdtopologyupdater
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+	. "github.com/smartystreets/goconvey/convey"
+)
+
+func TestTopologyUpdater(t *testing.T) {
+
+	Convey("Given a list of Attributes", t, func() {
+
+		attr_two := v1alpha2.AttributeInfo{
+			Name:  "attr_two_name",
+			Value: "attr_two_value",
+		}
+
+		attrList := v1alpha2.AttributeList{
+			v1alpha2.AttributeInfo{
+				Name:  "attr_one_name",
+				Value: "attr_one_value",
+			},
+			attr_two,
+			v1alpha2.AttributeInfo{
+				Name:  "attr_three_name",
+				Value: "attr_three_value",
+			},
+		}
+		attrListLen := len(attrList)
+		attrNames := getListOfNames(attrList)
+
+		Convey("When an existing attribute is updated", func() {
+
+			updatedAttribute := v1alpha2.AttributeInfo{
+				Name:  attr_two.Name,
+				Value: attr_two.Value + "_new",
+			}
+			updateAttribute(&attrList, updatedAttribute)
+
+			Convey("Then list should have the same number of elements", func() {
+				So(attrList, ShouldHaveLength, attrListLen)
+			})
+			Convey("Then the order of the elemens should be the same", func() {
+				So(attrNames, ShouldResemble, getListOfNames(attrList))
+			})
+			Convey("Then Attribute value in the list should be updated", func() {
+				attr, err := findAttributeByName(attrList, attr_two.Name)
+				So(err, ShouldBeNil)
+				So(attr.Value, ShouldEqual, updatedAttribute.Value)
+			})
+		})
+
+		Convey("When a non existing attribute is updated", func() {
+			completelyNewAttribute := v1alpha2.AttributeInfo{
+				Name:  "NonExistingAttribute_Name",
+				Value: "NonExistingAttribute_Value",
+			}
+			_, err := findAttributeByName(attrList, completelyNewAttribute.Name)
+			So(err, ShouldNotBeNil)
+
+			updateAttribute(&attrList, completelyNewAttribute)
+
+			Convey("Then list should have the one more element", func() {
+				So(attrList, ShouldHaveLength, attrListLen+1)
+			})
+
+			Convey("Then new Attribute should be added at the end of the list", func() {
+				So(attrList[len(attrList)-1], ShouldResemble, completelyNewAttribute)
+			})
+
+			Convey("Then the order of the elemens should be the same", func() {
+				So(attrNames, ShouldResemble, getListOfNames(attrList[:len(attrList)-1]))
+			})
+		})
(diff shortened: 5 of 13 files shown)


pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -138,7 +138,7 @@ func (w *nfdTopologyUpdater) Run() error {
 		select {
 		case <-crTrigger.C:
 			klog.Infof("Scanning")
-			podResources, err := resScan.Scan()
+			podResources, annotations, err := resScan.Scan()
```
ffromani, line 141: I think we should add some preparatory refactoring letting `Scan()` here return a `struct` like `ScanResponse` which will hold `[]PodResources` initially and then also `map[string]string`

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -172,7 +172,7 @@ func (w *nfdTopologyUpdater) Stop() {
 	}
 }
 
-func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha1.ZoneList) error {
+func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha1.ZoneList, annotations map[string]string) error {
```
ffromani, line 175: we shoudl evaluate here to wrap the args in a `struct`

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -27,20 +27,24 @@ import (
...
 type PodResourcesScanner struct {
 	namespace         string
 	podResourceClient podresourcesapi.PodResourcesListerClient
 	apihelper         apihelper.APIHelpers
+	podFingerprint    bool
 }
 
 // NewPodResourcesScanner creates a new ResourcesScanner instance
-func NewPodResourcesScanner(namespace string, podResourceClient podresourcesapi.PodResourcesListerClient, kubeApihelper apihelper.APIHelpers) (ResourcesScanner, error) {
+func NewPodResourcesScanner(namespace string, podResourceClient podresourcesapi.PodResourcesListerClient, kubeApihelper apihelper.APIHelpers, podFingerprint bool) (ResourcesScanner, error) {
```
ffromani, line 42: also here we should consider to move args in a struct, but this can be deferred in a later PR

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -198,7 +202,22 @@ func (resMon *PodResourcesScanner) Scan() ([]PodResources, error) {
 
 	}
 
-	return podResData, nil
+	if resMon.podFingerprint && len(podResData) > 0 {
+		var status podfingerprint.Status
+		podFingerprintSign, err := computePodFingerprint(podResData, &status)
+		if err != nil {
+			klog.Errorf("podFingerprint: Unable to compute fingerprint %v", err)
```
ffromani, line 209: should we fail badly or should we update without the annotation? If it's an annotation, it seems excessive to fail because of this. I'll think more about this.

pkg/resourcemonitor/podresourcesscanner_test.go
```diff
@@ -499,14 +525,20 @@ func TestPodScanner(t *testing.T) {
...
 			Convey("Error is nil", func() {
 				So(err, ShouldBeNil)
 			})
 			Convey("Return PodResources should have values", func() {
 				So(len(res), ShouldBeGreaterThan, 0)
 			})
+			Convey("Return Annotations should have pod fingerprint annotation with proper value", func() {
+				So(len(annotations), ShouldEqual, 1)
+
+				expectedFingerprint := "pfp0v001fe53c4dbd2c5f4a0" // precomputed manually
```
ffromani, line 539: hard-earned lesson from RTE: we should document which pods (namespace/name) we expect which lead to this fingerprint. Or, if possible, compute it in the tests.

ffromani: (here and also in all the instances below)

ffromani (commented):
thanks for the PR! initial review.
Do we have plans to add integration/e2e tests?

pkg/resourcemonitor/podresourcesscanner.go
jlojosnegros: Yeah ... it somehow seems too much to fail due to an annotation but in the end I went for the failure just because the user has to explicitly activate this feature and I slightly prefer to fail ... but I had the same doubt

pkg/resourcemonitor/podresourcesscanner_test.go
jlojosnegros: I agree on the documentation part but how could we compute the result if we are testing the compute algorithm itself?

go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/google/go-cmp v0.5.9
 	github.com/jaypipes/ghw v0.8.1-0.20210827132705-c7224150a17e
 	github.com/k8stopologyawareschedwg/noderesourcetopology-api v0.1.0
+	github.com/k8stopologyawareschedwg/podfingerprint v0.1.1
```
ffromani, line 13: you want to pull 0.1.2 here to use `podfingerprint.Attribute`

pkg/resourcemonitor/podresourcesscanner_test.go
ffromani: the compute algorithm itself is encapsulated in the `podfingerprint` package, isn't it?
In order to make the tests debuggable is (much much) better to document/show the expectations in terms of pod running on nodes. How this set translate in which fingerprint value is the `podfingerprint` package job indeed.

pkg/nfd-topology-updater/nfd-topology-updater.go
jlojosnegros: I have changed it to use `ScanResponse` 
I could have created a new struct including all the params ( right now only ZoneList) but as this is a "private" method that is called only once it does not seems worthy.

pkg/resourcemonitor/podresourcesscanner_test.go
jlojosnegros: I have changed to compute the expected fingerprint using the list of expected pods. 
Also I have documented that fingerprint only depends on namespace and name of the running pods in a node.

ffromani:
> I think we should create a seperate PR that upgrades NRT api to `v1alpha2` and not bundle it in this one(you are missing changes to kustomize and helm deployment to install new CRDs).

good point! I'm fine with both approaches (and yes your suggestion is cleaner indeed)

PiotrProkop:
> > I think we should create a seperate PR that upgrades NRT api to `v1alpha2` and not bundle it in this one(you are missing changes to kustomize and helm deployment to install new CRDs).
> 
> good point! I'm fine with both approaches (and yes your suggestion is cleaner indeed)

I just feel like we can easily miss important things we need to do with an upgrade like advertising Policies and Scope with Attributes 😄

ffromani:
> > > I think we should create a seperate PR that upgrades NRT api to `v1alpha2` and not bundle it in this one(you are missing changes to kustomize and helm deployment to install new CRDs).
> > 
> > 
> > good point! I'm fine with both approaches (and yes your suggestion is cleaner indeed)
> 
> I just feel like we can easily miss important things we need to do with an upgrade like advertising Policies and Scope with Attributes smile

Yes, thinking about it you're right. Let's split the update to its own PR

jlojosnegros:
Let's split this PR and create a new one only for NRT API update to `v1alpha2`
/hold

jlojosnegros:
New PR created to update NRT API version: #1053

ffromani (commented):
/lgtm
thanks @jlojosnegros for addressing the comments, nice work here

jlojosnegros:
just a rebase....

ArangoGutierrez:
> /lgtm

/Approve
/lgtm

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -199,6 +200,7 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 
 	nrtMutated := nrt.DeepCopy()
 	nrtMutated.Zones = zoneInfo
+	nrtMutated.Attributes = append(nrt.Attributes, scanResponse.Attributes...)
```
marquiz, line 203: Shouldn't we check existing attributes and replace old attributes with the same name instead of blindly appending?

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -198,7 +203,27 @@ func (resMon *PodResourcesScanner) Scan() ([]PodResources, error) {
...
+		Attributes:   v1alpha2.AttributeList{},
+	}
+
+	if resMon.podFingerprint && len(retVal.PodResources) > 0 {
+		var status podfingerprint.Status
+		podFingerprintSign, err := computePodFingerprint(retVal.PodResources, &status)
+		if err != nil {
+			klog.Errorf("podFingerprint: Unable to compute fingerprint %v", err)
+		} else {
+			klog.Infof("podFingerprint: " + status.Repr())
```
marquiz, line 217: nit: no string formatting is used here so either

```suggestion
			klog.Info("podFingerprint: " + status.Repr())
```

or 
```suggestion
			klog.Infof("podFingerprint: %s", status.Repr())
```

marquiz (commented):
I happened to try to run the e2e tests but failed on two different clusters with

```
    Feb 15 11:52:06.023: INFO: podFingerprint attributte "nodeTopologyPodsFingerprint" not found attribute "nodeTopologyPodsFingerprint" not found
    Feb 15 11:52:15.519: FAIL: Timed out after 60.003s.
```

jlojosnegros:
rebase after #1054 was merged.

pkg/resourcemonitor/podresourcesscanner.go
jlojosnegros: Thanks for the comment and the suggestions.

pkg/nfd-topology-updater/nfd-topology-updater.go
jlojosnegros: Good point ...  I have just created a new function although it could be a better way to merge them..

jlojosnegros:
Address comments.

pkg/nfd-topology-updater/nfd-topology-updater.go
jlojosnegros: > I happened to try to run the e2e tests but failed on two different clusters with
> 
> ```
>     Feb 15 11:52:06.023: INFO: podFingerprint attributte "nodeTopologyPodsFingerprint" not found attribute "nodeTopologyPodsFingerprint" not found
>     Feb 15 11:52:15.519: FAIL: Timed out after 60.003s.
> ```

it seems like the Attribute is not being created in the NRT object ... 
I have checked and nothing pretty obvious pop up, so I need to look a little deeper. 
I will take a look next monday

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -261,3 +264,21 @@ func createTopologyAttributes(policy string, scope string) v1alpha2.AttributeLis
...
+	}
+
+	for _, rattr := range rhs {
+		check[rattr.Name] = rattr.Value
+	}
+
+	ret := v1alpha2.AttributeList{}
+	for k, v := range check {
+		ret = append(ret, v1alpha2.AttributeInfo{Name: k, Value: v})
+	}
```
marquiz, line 282: I'm a bit concerned that with this any time any attribute is updated you basically get the list/slice re-shuffled in random order. Possibly ideally, we'd preserve the order of lhs with new attributes just appended.

Also, the current implementation is merely like `mergeAttributes()`. But should it be possible to delete attributes with this?

Any thoughts @ffromani and @PiotrProkop ?

Would be good to get this utility function into the NRT api...

marquiz: > So the question is do we want to make it work for more than one attribute from the start or just update podfingerprint for now.

Good question – I'm open to either approach. But if we have a generic function like this it should do the "right thing"

ffromani: my take is this code should only manage the attributes @PiotrProkop mentioned above, leaving everything else alone (= ignoring anything else). `nfd-topology-updater` should be the source of truth in the cluster for these attributes, so it should be allowed to add/update/overwrite/delete them, but just them.

marquiz: @ffromani sounds legit to me 👍

marquiz: I would say it is, when thinking about patching the object for example. But wdyt @ffromani @PiotrProkop ?

ffromani: I'm not against keeping attributes ordered if we can help it, because this minimizes the changes. IOW, we should unnecessarily change objects or their fields. However well-behaving, robust clients should NOT rely on any ordering and treat the sequence as unordered.

jlojosnegros: > my take is this code should only manage the attributes @PiotrProkop mentioned above, leaving everything else alone (= ignoring anything else). `nfd-topology-updater` should be the source of truth in the cluster for these attributes, so it should be allowed to add/update/overwrite/delete them, but just them.

Right now this function only add the `ScanResponse` attributes a newly created NRT or the the readed one. 
`ScanResponse` only manage `podfingerprint.Attribute.
Sorry, I think I do not fully understand what the new function is suppose to do ... could you please elaborate a little bit? Thanks

marquiz: Clients should survive, I agree. But changing the order on every update breaks diffing, patching etc

ffromani: > Clients should survive, I agree. But changing the order on every update breaks diffing, patching etc

I think we are in agreement then. `nfd-topology-updater` should preserve the existing order of attributes and be smart about add/update/delete (note: I think delete is very unlikely). This is also easy to unit-test

ffromani: @marquiz I hear you about adding utilities to NRT API. I'll work on a PR, but I'm not sure I can make it quick enough to be used in this PR

PiotrProkop: The simplest case is that this function iterates over all Attributes, looking for `podfingerprint.Attribute` if it's there check if checksum should be updated if so, update it in place. If it's not there append it to the end of the list.

jlojosnegros: > The simplest case is that this function iterates over all Attributes, looking for `podfingerprint.Attribute` if it's there -> check if checksum should be updated if so, update it in place. If it's not there append it to the end of the list.

I can implement this approach here and change it to use @ffromani utility functions when available.

ffromani: works for me

marquiz: Sounds good

jlojosnegros: Addressed. Let me know what you think

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -113,24 +118,43 @@ func hasIntegralCPUs(pod *corev1.Pod, container *corev1.Container) bool {
...
+	respPodResources := resp.GetPodResources()
+	retVal := ScanResponse{
+		Attributes: v1alpha2.AttributeList{},
 	}
 
+	if resMon.podFingerprint && len(respPodResources) > 0 {
+		var status podfingerprint.Status
+		podFingerprintSign, err := computePodFingerprint(respPodResources, &status)
+		if err != nil {
+			klog.Errorf("podFingerprint: Unable to compute fingerprint %v", err)
```
marquiz, line 140: Should we in this case remove the existing pod fingerprint attribute?

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -261,3 +264,25 @@ func createTopologyAttributes(policy string, scope string) v1alpha2.AttributeLis
...
 	}
 }
+
+func updateAttribute(attrList *v1alpha2.AttributeList, attrInfo v1alpha2.AttributeInfo) {
+	if nil == attrList {
+		return
+	}
+	l := len(*attrList)
+	idx := 0
+	for idx < l {
```
marquiz, line 274:
```suggestion
	for idx, info := range *attrList {
```

pkg/nfd-topology-updater/nfd-topology-updater.go
jlojosnegros: I have changed the moment when podfingerprint is computed. 
I was doing if AFTER the resources were filtered (in podresourcescanner.go:152) and if no resource pass the filter the Attribute was not added at all. 
Now it is computed using all the resources which seems better because podfingerprint should resemble node resource status

pkg/resourcemonitor/podresourcesscanner.go
ffromani: > Should we in this case remove the existing pod fingerprint attribute?

I'm fine both ways. From the scheduler plugin perspective, clients consuming this attribute must tolerate either attribute disappearing (a previous update added, a later update removes) values and obsolete values.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -261,3 +264,25 @@ func createTopologyAttributes(policy string, scope string) v1alpha2.AttributeLis
 		},
 	}
 }
+
+func updateAttribute(attrList *v1alpha2.AttributeList, attrInfo v1alpha2.AttributeInfo) {
+	if nil == attrList {
```
PiotrProkop, line 269:
```suggestion
	if attrList == nil {
```
I know it is correct but it looks odd to have nil first.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -261,3 +264,25 @@ func createTopologyAttributes(policy string, scope string) v1alpha2.AttributeLis
...
+
+		if (*attrList)[idx].Name == attrInfo.Name {
+			(*attrList)[idx].Value = attrInfo.Value
+			return
+		}
+		idx = idx + 1
+	}
+	*attrList = append(*attrList, attrInfo)
+}
+func updateAttributes(lhs *v1alpha2.AttributeList, rhs v1alpha2.AttributeList) {
```
PiotrProkop, line 284: can you add unit test for this function?

pkg/resourcemonitor/podresourcesscanner.go
jlojosnegros: Here we are computing the ScanResponse. Podfingerprint attribute will not be added to the ScanResponse in case of an error.

Removing the existing attribute in the NTR should be done in the nfd-topology-updater's `updateNodeResourceTopology` function,  when updating the attribute list. 
That can be done but will go agains this (https://github.com/kubernetes-sigs/node-feature-discovery/pull/1049#discussion_r1109505690) and will require some changes in the `updateAttribute` function.
I understood that those changes will wait for the new utility functions in @ffromani 's PR

pkg/nfd-topology-updater/nfd-topology-updater.go
ffromani: In general I'd have squashed the fixing commit here. Codewise looks reasonnable.

pkg/nfd-topology-updater/nfd-topology-updater_test.go
```diff
@@ -0,0 +1,117 @@
...
+				Name:  "attr_three_name",
+				Value: "attr_three_value",
+			},
+		}
+		attrListLen := len(attrList)
+		attrNames := getListOfNames(attrList)
+
+		Convey("When an existing attribute is updated", func() {
+
+			existingAttribute := v1alpha2.AttributeInfo{
```
ffromani, line 52: nit: this looks more like `updatedAttribute` (I'd not resubmit just for this)

pkg/nfd-topology-updater/nfd-topology-updater_test.go
```diff
@@ -0,0 +1,117 @@
...
+		})
+	})
+}
+
+func getListOfNames(attrList v1alpha2.AttributeList) []string {
+	return Map(attrList, func(attr v1alpha2.AttributeInfo) string {
+		return attr.Name
+	})
+}
+func Map(attrList v1alpha2.AttributeList, f func(v1alpha2.AttributeInfo) string) []string {
```
ffromani, line 101: nit: I'd inline in the caller (I'd not resubmit just for this)

pkg/nfd-topology-updater/nfd-topology-updater_test.go
jlojosnegros: Done while squashing the fixing commit

pkg/nfd-topology-updater/nfd-topology-updater_test.go
jlojosnegros: Done while squashing the fixing commit

ffromani (commented):
/lgtm

cmd/nfd-topology-updater/main.go
```diff
@@ -139,6 +139,7 @@ func initFlags(flagset *flag.FlagSet) (*topology.Args, *resourcemonitor.Args) {
 		"Pod Resource Socket path to use.")
 	flagset.StringVar(&args.ConfigFile, "config", "/etc/kubernetes/node-feature-discovery/nfd-topology-updater.conf",
 		"Config file to use.")
+	flagset.BoolVar(&resourcemonitorArgs.PodSetFingerprint, "pods-fingerprint", false, "If enable, compute and report the pod set fingerprint")
```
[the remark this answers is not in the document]
jlojosnegros: > Thanks for working on this @jlojosnegros. I think the code is basically ready to be merged (I had one nit about the cmdline flag description).
> 
> However, I probably forgot to ask this before, but we need to update the documentation (`docs/reference/topology-updater-commandline-reference.md`) to describe the new command line flag.
> 
> Also, could you add support for the new flag in the help chart:
> 
> * add smth like `topologyUpdater.podSetFingerprint` into `deployment/helm/node-feature-discovery/values.yaml`
> * use it in `deployment/helm/node-feature-discovery/templates/topologyupdater.yaml`
> * document the new helm parameter in `docs/deployment/helm.md`

Thanks for the comment. 
Sorry, I did not see this lack of documentation before.

deployment/helm/node-feature-discovery/templates/topologyupdater.yaml
```diff
@@ -55,6 +55,9 @@ spec:
           - "-key-file=/etc/kubernetes/node-feature-discovery/certs/tls.key"
           - "-cert-file=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
           {{- end }}
+          {{-if .Values.topologyUpdater.podSetFingerprint }}
```
ffromani, line 58: typo here

jlojosnegros: thanks

ffromani (commented):
/lgtm

updates to docs and helm chart seem good

marquiz (approved):
Thanks @jlojosnegros for the quick update. Looks good to me now 👍 If something's missing let's fix it in subsequent PRs

Issue #899, opened 2022-09-22
Improve "core.sleepInterval" handling


eero-t:
**Why is this needed**:

Currently the time interval NFD takes between feature checking / labeling could differ significantly from the interval specified by the user, because it's used as an extra delay between activity, not an interval between them.  If some hook takes 50% of the interval time, the reporting interval is >= 150% of the specified interval value.

**What would you like to be added**:

NFD implements interval option as an actual interval.

Optionally it could also:
* When feature discovery phase exceeds the interval, disable feature/hook taking most time + log error about that, as it's something that cannot be supported within specified interval
* When feature discovery exceeds e.g. 50% of the interval, log a warning (because on more loaded, more complex, or less beefy system interval could be exceed).  Warning should include the used interval, and what discovery features are used, including listing each called hook, as those could differ between nodes (see: #857)

eero-t:
@marquiz Any comments on this proposal?

VillePihlava:
I can work on this

PR #1050, opened 2023-02-06
Change nfd-worker to use Ticker instead of After.

VillePihlava:
This pull request changes the nfd-worker to use `time.Ticker` instead of `time.After` to implement  `core.sleepInterval`. An infinite interval is represented by stopping the Ticker.

Issue: https://github.com/kubernetes-sigs/node-feature-discovery/issues/899

### DIFF
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -131,6 +131,11 @@ type duration struct {
 	time.Duration
 }
 
+// This ticker can represent infinite and normal intervals.
+type infiniteTicker struct {
+	*time.Ticker
+}
+
 // NewNfdWorker creates new NfdWorker instance.
 func NewNfdWorker(args *Args) (NfdWorker, error) {
 	nfd := &nfdWorker{
@@ -172,6 +177,37 @@ func newDefaultConfig() *NFDConfig {
 	}
 }
 
+func (i *infiniteTicker) Reset(d time.Duration) {
+	switch {
+	case d > 0:
+		i.Ticker.Reset(d)
+	default:
+		// If the sleep interval is not a positive number the ticker will act
+		// as if it was set to an infinite duration by not ticking.
+		i.Ticker.Stop()
+	}
+}
+
+// Run feature discovery.
+func (w *nfdWorker) runFeatureDiscovery() error {
+	for _, s := range w.featureSources {
+		klog.V(2).Infof("running discovery for %q source", s.Name())
+		if err := s.Discover(); err != nil {
+			klog.Errorf("feature discovery of %q source failed: %v", s.Name(), err)
+		}
+	}
+
+	// Get the set of feature labels.
+	labels := createFeatureLabels(w.labelSources, w.config.Core.LabelWhiteList.Regexp)
+
+	// Update the node with the feature labels.
+	if !w.config.Core.NoPublish {
+		return w.advertiseFeatures(labels)
+	}
+
+	return nil
+}
+
 // Run NfdWorker client. Returns if a fatal error is encountered, or, after
 // one request if OneShot is set to 'true' in the worker args.
 func (w *nfdWorker) Run() error {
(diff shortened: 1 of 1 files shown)


pkg/nfd-worker/nfd-worker.go
```diff
@@ -172,6 +172,39 @@ func newDefaultConfig() *NFDConfig {
 	}
 }
 
+func (w *nfdWorker) createTicker() *time.Ticker {
+	if w.config.Core.SleepInterval.Duration > 0 {
+		return time.NewTicker(w.config.Core.SleepInterval.Duration)
+	} else {
```
ArangoGutierrez, line 178: This Else is not needed

ArangoGutierrez (changes requested):
Patch looks good, thanks for this @VillePihlava !!
Just a small comment
/ok-to-test
The rest I'll leave it to 
/assign @marquiz

pkg/nfd-worker/nfd-worker.go
```diff
@@ -238,7 +263,12 @@ func (w *nfdWorker) Run() error {
 
 			// Always re-label after a re-config event. This way the new config
 			// comes into effect even if the sleep interval is long (or infinite)
-			labelTrigger = time.After(0)
+			labelTrigger.Stop()
+			labelTrigger = w.createTicker()
+			err = w.runFeatureDiscovery()
```
PiotrProkop, line 268: Wouldn't it be simpler just to run `w.runFeatureDiscovery` here instead of reinitializing  the Ticker?

pkg/nfd-worker/nfd-worker.go
VillePihlava: The else is now removed

pkg/nfd-worker/nfd-worker.go
```diff
@@ -172,6 +172,39 @@ func newDefaultConfig() *NFDConfig {
...
+			klog.Errorf("feature discovery of %q source failed: %v", s.Name(), err)
+		}
+	}
+
+	// Get the set of feature labels.
+	labels := createFeatureLabels(w.labelSources, w.config.Core.LabelWhiteList.Regexp)
+
+	// Update the node with the feature labels.
+	if !w.config.Core.NoPublish {
+		if err := w.advertiseFeatures(labels); err != nil {
```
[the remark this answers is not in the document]
VillePihlava: Its now changed to this

pkg/nfd-worker/nfd-worker.go
VillePihlava: `core.sleepInterval` can change if the config changes. I restart the Ticker for the possible new interval

PiotrProkop: It makes sense, I haven't thought about this.

fmuyassarov (commented):
Looks good. Can you perhaps squash the commits?

pkg/nfd-worker/nfd-worker.go
```diff
@@ -172,6 +172,36 @@ func newDefaultConfig() *NFDConfig {
...
 }
 
+func (w *nfdWorker) createTicker() *time.Ticker {
+	if w.config.Core.SleepInterval.Duration > 0 {
+		return time.NewTicker(w.config.Core.SleepInterval.Duration)
+	}
+	// If the sleep interval is not a positive number, the ticker will act as if it was set to an infinite duration by not ticking.
+	t := time.NewTicker(time.Hour)
+	t.Stop()
+	return t
```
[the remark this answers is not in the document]
VillePihlava: I'm not entirely sure what happens when calling `Stop()` with an uninitialized ticker, so I changed the implementation a little to not call it on an uninitialized one and incorporated this change in a sense

VillePihlava:
> Looks good. Can you perhaps squash the commits?

Done

pkg/nfd-worker/nfd-worker.go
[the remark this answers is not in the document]
marquiz: Mm, that's a fair point. Even if it works now it's not clearly specified (from what I can tell) if calling Stop() on uninitialized Ticker is legit which might cause a breakage later.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -172,6 +178,41 @@ func newDefaultConfig() *NFDConfig {
...
+	}
+}
+
+func (infiniteTicker *infiniteTicker) startTicker(w *nfdWorker) {
+	// If the sleep interval is not a positive number, the ticker will act as if it was set to an infinite duration by not ticking.
+	if w.config.Core.SleepInterval.Duration > 0 {
+		infiniteTicker.T = time.NewTicker(w.config.Core.SleepInterval.Duration)
+		infiniteTicker.initialized = true
+	}
+}
```
marquiz, line 194: Thanks for the update @VillePihlava. Quite a bike-shedding on this PR, but somehow this latest version feels just a tad too complicated (in terms of lines-of-code) for such a simple thing 😇

How about using struct composition to wrap the Ticker type, with something like

```golang
type infiniteTicker struct {
        *time.Ticker
}

func (i *infiniteTicker) Reset(d time.Duration) {
        if d > 0 {
                i.Ticker.Reset(d)
        } else {
                i.Ticker.Stop()
        }
}
```

...then you'd initialize with smth like (you could specify a dedicated create function if this looks too bad):

```golang
labelTrigger := infiniteTicker{Ticker: time.NewTicker(1)} // just any non-zero duration here, we'll do the actual configuration below
```

use the channel directly:

```golang
case <-labelTrigger.C:
```
and simply use reset for re-configuring (instead of creating a new Ticker instance)

```golang
```

WDYT? Haven't tested but I think that should work out

VillePihlava: Yeah this makes sense, I made these changes now. Also tested manually that making changes to the sleepinterval in the config worked

marquiz (approved):
Thanks @VillePihlava! Looks good to me 🤓

pkg/nfd-worker/nfd-worker.go
ArangoGutierrez: TO me this looks better like a  https://gobyexample.com/switch since the func doesn't have a `return`  a Switch will be more elegant here

pkg/nfd-worker/nfd-worker.go
```diff
@@ -172,6 +177,35 @@ func newDefaultConfig() *NFDConfig {
 	}
 }
 
+func (i *infiniteTicker) Reset(d time.Duration) {
+	if d > 0 {
+		i.Ticker.Reset(d)
+	} else {
+		// If the sleep interval is not a positive number the ticker will act as if it was set to an infinite duration by not ticking.
+		i.Ticker.Stop()
```
ArangoGutierrez, line 185: Could you format this comment to fit in 80 columns, see how other comments when too long they are several lines. makes it easier to read

ArangoGutierrez (changes requested):

pkg/nfd-worker/nfd-worker.go
VillePihlava: I changed the if-else to a switch and made the comment fit into 80 columns

ArangoGutierrez (approved):
Thanks for the effort @VillePihlava ! 

/lgtm

PR #1051, opened 2023-02-06
feat: add deny-label-ns flag which supports wildcard

### LINKED ISSUE #1022: Allow wildcard or regrx in "extra label ns"
The ability to pass a wildcard or regex etc. to the label namespace whitelist i.e.
```
# Command line
--extra-label-ns='*'

# and

# In helm chart
extraLabelNs: ["*"]
```

Another possibility could be to disable whitelist altogether because my use case doesn't need to prevent any label namespaces.

**Why is this needed**:
- I'm looking to retrofit a huge list of custom labels for a lot of different nodes to be managed by NFD. 
- The whitelist makes it rather tedious/error-prone to add every single namespace to the whitelist on the operator when I don't need to restrict any to begin with. 
- When creating a new label to the NodeFeatureRule that have a new label namespace, I don't want to have to remember to go and add it to the whitelist in the helm chart as well.
- Having this will also allow me to avoid having to update all of the workloads across a ton of systems to use the NFD default label namespaces which would be a large, error-prone undertaking.

Thanks so much in advance for your consideration! Happy to help if needed!

AhmedGrati:
Resolves #1022.

Signed-off-by: AhmedGrati <ahmedgrati1999@gmail.com>

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -85,12 +85,15 @@ func main() {
 func initFlags(flagset *flag.FlagSet) *master.Args {
 	args := &master.Args{
 		LabelWhiteList: utils.RegexpVal{Regexp: *regexp.MustCompile("")},
+		DenyLabelNs:    map[string]struct{}{"*.kubernetes.io": {}},
 	}
 
 	flagset.StringVar(&args.CaFile, "ca-file", "",
 		"Root certificate for verifying connections")
 	flagset.StringVar(&args.CertFile, "cert-file", "",
 		"Certificate used for authenticating connections")
+	flagset.Var(&args.DenyLabelNs, "deny-label-ns",
+		"Comma separated list of denied label namespaces")
 	flagset.Var(&args.ExtraLabelNs, "extra-label-ns",
 		"Comma separated list of allowed extra label namespaces")
 	flagset.StringVar(&args.Instance, "instance", "",
--- a/docs/reference/master-commandline-reference.md
+++ b/docs/reference/master-commandline-reference.md
@@ -216,12 +216,10 @@ nfd-master -label-whitelist='.*cpuid\.'
 ### -extra-label-ns
 
 The `-extra-label-ns` flag specifies a comma-separated list of allowed feature
-label namespaces. By default, nfd-master only allows creating labels in the
-default `feature.node.kubernetes.io` and `profile.node.kubernetes.io` label
-namespaces and their sub-namespaces (e.g. `vendor.feature.node.kubernetes.io`
-and `sub.ns.profile.node.kubernetes.io`). This option can be used to allow
+label namespaces. This option can be used to allow
 other vendor or application specific namespaces for custom labels from the
-local and custom feature sources.
+local and custom feature sources, even though these labels were denied using
+the `deny-label-ns` flag.
 
 The same namespace control and this flag applies Extended Resources (created
 with `-resource-labels`), too.
@@ -232,6 +230,28 @@ Example:
 
 ```bash
 nfd-master -extra-label-ns=vendor-1.com,vendor-2.io
+```
+
+### -deny-label-ns
+
+The `-deny-label-ns` flag specifies a comma-separated list of excluded
+label namespaces. By default, nfd-master allows creating labels in all
+namespaces, excluding `kubernetes.io` namespace and its sub-namespaces
+(i.e. `*.kubernetes.io`). However, you should note that
+`kubernetes.io` and its sub-namespaces are always denied.
+For example, `nfd-master -deny-label-ns=""` would still disallow
+`kubernetes.io` and `*.kubernetes.io`.
+This option can be used to exclude some vendors or application specific
+namespaces.
+Note that the namespaces `feature.node.kubernetes.io` and `profile.node.kubernetes.io`
+and their sub-namespaces are always allowed and cannot be denied.
+
+Default: *empty*
+
+Example:
+
+```bash
+nfd-master -deny-label-ns=*.vendor.com,vendor-2.io
 ```
 
 ### -resource-labels
--- a/pkg/nfd-master/nfd-master-internal_test.go
+++ b/pkg/nfd-master/nfd-master-internal_test.go
@@ -352,16 +352,18 @@ func TestSetLabels(t *testing.T) {
 			})
 		})
 
-		Convey("When -extra-label-ns and -instance are specified", func() {
+		Convey("When -extra-label-ns, -deny-label-ns and -instance are specified", func() {
 			// In the gRPC request the label names may omit the default ns
 			instance := "foo"
 			vendorFeatureLabel := "vendor." + nfdv1alpha1.FeatureLabelNs + "/feature-4"
 			vendorProfileLabel := "vendor." + nfdv1alpha1.ProfileLabelNs + "/feature-5"
 			mockLabels := map[string]string{"feature-1": "val-1",
-				"valid.ns/feature-2":   "val-2",
-				"invalid.ns/feature-3": "val-3",
-				vendorFeatureLabel:     " val-4",
-				vendorProfileLabel:     " val-5"}
+				"valid.ns/feature-2":             "val-2",
+				"random.denied.ns/feature-3":     "val-3",
+				"kubernetes.io/feature-4":        "val-4",
+				"sub.ns.kubernetes.io/feature-5": "val-5",
+				vendorFeatureLabel:               " val-6",
+				vendorProfileLabel:               " val-7"}
 			expectedPatches := []apihelper.JsonPatch{
 				apihelper.NewJsonPatch("add", "/metadata/annotations", instance+"."+nfdv1alpha1.WorkerVersionAnnotation, workerVer),
 				apihelper.NewJsonPatch("add", "/metadata/annotations",
@@ -374,6 +376,8 @@ func TestSetLabels(t *testing.T) {
 				apihelper.NewJsonPatch("add", "/metadata/labels", vendorProfileLabel, mockLabels[vendorProfileLabel]),
 			}
 
+			mockMaster.deniedNs.normal = map[string]struct{}{"random.denied.ns": {}}
+			mockMaster.deniedNs.wildcard = map[string]struct{}{"kubernetes.io": {}}
 			mockMaster.args.ExtraLabelNs = map[string]struct{}{"valid.ns": {}}
 			mockMaster.args.Instance = instance
 			mockHelper.On("GetClient").Return(mockClient, nil)
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -63,6 +63,7 @@ type Annotations map[string]string
 type Args struct {
 	CaFile               string
 	CertFile             string
+	DenyLabelNs          utils.StringSetVal
 	ExtraLabelNs         utils.StringSetVal
 	Instance             string
 	KeyFile              string
@@ -76,6 +77,11 @@ type Args struct {
 	Prune                bool
 	VerifyNodeName       bool
 	ResourceLabels       utils.StringSetVal
+}
+
+type deniedNs struct {
+	normal   utils.StringSetVal
+	wildcard utils.StringSetVal
 }
 
 type NfdMaster interface {
@@ -95,6 +101,7 @@ type nfdMaster struct {
 	ready      chan bool
 	apihelper  apihelper.APIHelpers
 	kubeconfig *restclient.Config
+	deniedNs
 }
 
 // NewNfdMaster creates a new NfdMaster server instance.
@@ -126,6 +133,16 @@ func NewNfdMaster(args *Args) (NfdMaster, error) {
 			return nfd, fmt.Errorf("-ca-file needs to be specified alongside -cert-file and -key-file")
 		}
 	}
+	if args.DenyLabelNs == nil {
+		args.DenyLabelNs = make(utils.StringSetVal)
+	}
+	// Pre-process DenyLabelNS into 2 lists: one for normal ns, and the other for wildcard ns
+	normalDeniedNs, wildcardDeniedNs := preProcessDeniedNamespaces(args.DenyLabelNs)
+	nfd.deniedNs.normal = normalDeniedNs
+	nfd.deniedNs.wildcard = wildcardDeniedNs
+	// We forcibly deny kubernetes.io
+	nfd.deniedNs.normal["kubernetes.io"] = struct{}{}
+	nfd.deniedNs.wildcard[".kubernetes.io"] = struct{}{}
 
 	// Initialize Kubernetes API helpers
 	if !args.NoPublish {
@@ -392,7 +409,7 @@ func (m *nfdMaster) updateMasterNode() error {
 // into extended resources. This function also handles proper namespacing of
 // labels and ERs, i.e. adds the possibly missing default namespace for labels
 // arriving through the gRPC API.
-func filterFeatureLabels(labels Labels, extraLabelNs map[string]struct{}, labelWhiteList regexp.Regexp, extendedResourceNames map[string]struct{}) (Labels, ExtendedResources) {
+func (m *nfdMaster) filterFeatureLabels(labels Labels) (Labels, ExtendedResources) {
 	outLabels := Labels{}
 
 	for label, value := range labels {
@@ -404,15 +421,18 @@ func filterFeatureLabels(labels Labels, extraLabelNs map[string]struct{}, labelW
 		// Check label namespace, filter out if ns is not whitelisted
 		if ns != nfdv1alpha1.FeatureLabelNs && ns != nfdv1alpha1.ProfileLabelNs &&
 			!strings.HasSuffix(ns, nfdv1alpha1.FeatureLabelSubNsSuffix) && !strings.HasSuffix(ns, nfdv1alpha1.ProfileLabelSubNsSuffix) {
-			if _, ok := extraLabelNs[ns]; !ok {
-				klog.Errorf("Namespace %q is not allowed. Ignoring label %q\n", ns, label)
-				continue
+			// If the namespace is denied, and not present in the extraLabelNs, label will be ignored
+			if isNamespaceDenied(ns, m.deniedNs.wildcard, m.deniedNs.normal) {
+				if _, ok := m.args.ExtraLabelNs[ns]; !ok {
+					klog.Errorf("Namespace %q is not allowed. Ignoring label %q\n", ns, label)
+					continue
+				}
 			}
 		}
 
 		// Skip if label doesn't match labelWhiteList
-		if !labelWhiteList.MatchString(name) {
-			klog.Errorf("%s (%s) does not match the whitelist (%s) and will not be published.", name, label, labelWhiteList.String())
+		if !m.args.LabelWhiteList.Regexp.MatchString(name) {
+			klog.Errorf("%s (%s) does not match the whitelist (%s) and will not be published.", name, label, m.args.LabelWhiteList.Regexp.String())
 			continue
 		}
 		outLabels[label] = value
@@ -420,7 +440,7 @@ func filterFeatureLabels(labels Labels, extraLabelNs map[string]struct{}, labelW
 
 	// Remove labels which are intended to be extended resources
 	extendedResources := ExtendedResources{}
-	for extendedResourceName := range extendedResourceNames {
+	for extendedResourceName := range m.args.ResourceLabels {
 		// Add possibly missing default ns
 		extendedResourceName = addNs(extendedResourceName, nfdv1alpha1.FeatureLabelNs)
 		if value, ok := outLabels[extendedResourceName]; ok {
@@ -447,6 +467,20 @@ func verifyNodeName(cert *x509.Certificate, nodeName string) error {
 		return fmt.Errorf("certificate %q not valid for node %q: %v", cert.Subject.CommonName, nodeName, err)
 	}
 	return nil
+}
+
+func isNamespaceDenied(labelNs string, wildcardDeniedNs map[string]struct{}, normalDeniedNs map[string]struct{}) bool {
+	for deniedNs := range normalDeniedNs {
+		if labelNs == deniedNs {
+			return true
+		}
+	}
+	for deniedNs := range wildcardDeniedNs {
+		if strings.HasSuffix(labelNs, deniedNs) {
+			return true
+		}
+	}
+	return false
 }
 
 // SetLabels implements LabelerServer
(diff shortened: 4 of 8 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -449,6 +451,17 @@ func verifyNodeName(cert *x509.Certificate, nodeName string) error {
 	return nil
 }
 
+func isNamespaceDenied(labelNs string, deniedNs map[string]struct{}) bool {
+	for ns := range deniedNs {
+		deniedNsWildcard := fmt.Sprintf("*.%v", ns)
```
marquiz, line 456: I don't think we want to forcibly make all items a glob. Otherwise you couldn't do something like "deny `kubernetes.io` but allow all `*.kubernetes.io`"

pkg/nfd-master/nfd-master.go
```diff
@@ -449,6 +451,17 @@ func verifyNodeName(cert *x509.Certificate, nodeName string) error {
...
+		g := glob.MustCompile(deniedNsWildcard)
+		if g.Match(labelNs) {
```
marquiz, line 458: I'm not entirely sure we need to support full-fledged globs (performance, extra external dependency), we could probably do away with something like

```golang
if strings.HasPrefix(ns, "*") {
    deniedSuffix := strings.TrimLeft(ns, "*")
    if strings.HasSuffix(labelNs, deniedSuffix) {
        return true
    }
} else if labelNs == ns {
    return true
}
return false
```

And just document that we support asterisk ("*") in front of the name to deny all sub-namespaces. WDYT?

Also, we probably want to pre-process the user input at startup to repeat same string (or glob) operations over and over again.

marquiz (commented):
Thanks @AhmedGrati for working on this!

I don't think we want to replace `-extra-label-ns`, just add `-deny-label-ns` on top of that. This will allow strict namespace control (if a user wants that) e.g. with
```
-deny-label-ns="*" -extra-label-ns=example.com
```

you could deny everything but example.com, kind of mimicking what we currently do

/ok-to-test

pkg/nfd-master/nfd-master.go
AhmedGrati: Yes true, but I just followed what was said [here.](https://github.com/kubernetes-sigs/node-feature-discovery/issues/1022#issuecomment-1412215582) Which means basically denying all namespaces. In fact, the user will input a namespace (without wildcard) and we will deny all sub-namespaces. 
However, If you see that we need to give users the flexibility to deny all sub-namespaces using wildcards, I think we should change this implementation above.

pkg/nfd-master/nfd-master.go
AhmedGrati: Yes got it, and it's a valid point I think. I just didn't understand the last point. Can you explain it to me further, please?

pkg/nfd-master/nfd-master.go
marquiz: It can be that what I suggested above is not the "right thing to do". NOT forcing a glob (i.e. not automatically denying sub-namespaces) would give us the most flexibility. I now feel that if we forcibly deny all sub-namespaces it will bit us later (could be countered by supporting wildcards in `-extra-label-ns` though)

Let's see what others think. Any thoughts on this @cedricmckinnie @ArangoGutierrez @zvonkok @fmuyassarov ?

AhmedGrati: Yes, I agree with you!

pkg/nfd-master/nfd-master.go
marquiz: > Can you explain it to me further, please?

My point is that in the current implementation for each label we create, we run through the deny list and do `glob.MustCompile` for each deny item. That is for each time a label request arrives, for each label, for each item in deny list do compile. Even though that information is basically static, doing compile once at the startup would be enough and save a lot of cpu cycles

AhmedGrati: @marquiz what about having working with such wildcards: `prefix.*.suffix`. The aesterik-based implementation won't work in the expected way. isn't it?

marquiz: It's just me, but I wouldn't be concerned about complex regexs or globs at the moment. Think about how domain names and IP addresses generally work, for example, you don't generally specify IP ranges like 192.168.*.3. For sub-namespaces I see a clear use case, though. More advanced globbing can be added later if it is clearly needed

cmd/nfd-master/main.go
```diff
@@ -85,14 +85,15 @@ func main() {
...
 	args := &master.Args{
 		LabelWhiteList: utils.RegexpVal{Regexp: *regexp.MustCompile("")},
+		DenyLabelNs:    map[string]struct{}{"kubernetes.io": {}},
 	}
 
 	flagset.StringVar(&args.CaFile, "ca-file", "",
 		"Root certificate for verifying connections")
 	flagset.StringVar(&args.CertFile, "cert-file", "",
 		"Certificate used for authenticating connections")
-	flagset.Var(&args.ExtraLabelNs, "extra-label-ns",
```
ArangoGutierrez, line 94: @marquiz should we have both flags at least for a release, having `ExtraLabelNS` to print a deprecation warning?  I feel that removing the flag all together could break some users deployments if they try a roll-update and the pod get's called with old flags. 
WDYT?

ArangoGutierrez (changes requested):

cmd/nfd-master/main.go
AhmedGrati: @marquiz suggested that we keep the `extra-label-ns` flag because that would give flexibility to our users.

marquiz: @ArangoGutierrez yes, I don't see any reason removing the existing flag. It will also give flexibility in enabling scenarios like "deny all except for example.com/ and acme.io/"

ArangoGutierrez (approved):
PR looks good now
Leaving the lgtm to @marquiz

cmd/nfd-master/main.go
```diff
@@ -122,5 +126,24 @@ func initFlags(flagset *flag.FlagSet) *master.Args {
 		"Verify worker node name against the worker's TLS certificate. "+
 			"Only takes effect when TLS authentication has been enabled.")
 
+	normalDeniedNs, wildcardDeniedNs := preProcessDeniedNamespaces(args.DenyLabelNs)
+	args.NormalDenyLabelNs = normalDeniedNs
+	args.WildcardDenyLabelNs = wildcardDeniedNs
```
marquiz, line 131: Hmm, I don't think this processing actually does anything as it's inside `initFlags()`. We need to do this after the args are parsed.

I'd suggest to move this processing to `pkg/nfd-master/`. Anyway, don't carry the same data in two (potentially conflicting) ways i.e. args.DenyLabelNs plus args.NormalDenyLabelNs and args.WildcardDenyLabelNs

pkg/nfd-master/nfd-master.go
```diff
@@ -408,6 +411,10 @@ func filterFeatureLabels(labels Labels, extraLabelNs map[string]struct{}, labelW
 				klog.Errorf("Namespace %q is not allowed. Ignoring label %q\n", ns, label)
 				continue
 			}
+			if isNamespaceDenied(ns, wildcardDeniedLabelNs, normalDeniedLabelsNs) {
+				klog.Errorf("Namespace %q is not allowed. Ignoring label %q\n", ns, label)
+				continue
+			}
```
marquiz, line 417: I think this logic is now broken because we don't change the handling of extraLabelNs above. We basically still deny everything. The logic should be so that extraLabelNs can be used to override DeniedLabelNs. I.e. label is denied if namespace is in DeniedLabelNs, unless specifically allowed with extraLabelNs. Makes sense?

marquiz (commented):
Thanks for the update @AhmedGrati

I think we need to adjust the docs a bit more, probably removing all references to `-extra-label-ns` in `docs/usage/customization-guide.md` and also adjust description of `-extra-label-ns` in `docs/reference/master-commandline-reference.md` to match the new behavior.

pkg/nfd-master/nfd-master.go
AhmedGrati: Yes makes sense, good point. I was actually thinking of the old logic. But when I reviewed our conversation, I see that the logic is broken. I'll update it

cmd/nfd-master/main.go
AhmedGrati: Do you have any suggestions on where to store the data, and parse flags? I think that storing data in separate variables can do the job, and for the processing of the DenyLabelNS, do you see that it's appropriate to do it under `pkg/nfd-master/nfd-master.go`, in the function `NewNfdMaster()`?

AhmedGrati:
@marquiz why should we remove all references to `-extra-label-ns` in `docs/usage/customization-guide.md`?

cmd/nfd-master/main.go
marquiz: Yeah, I think that would be the logical place. I.e. pass the "raw" arg via `args.DenyLabelNs` and then store the processed data in separate variables in `nfdMaster` struct for example

marquiz:
> @marquiz why should we remove all references to `-extra-label-ns` in `docs/usage/customization-guide.md`?

Heh, my comment wasn't very clear 😇 I should've written that review all references to `-extra-label-ns` in the doc. IIRC it's mentioned in few places, saying that by default all namespaces except for '*.feature.node.kubernetes.io` are denied... which obviously isn't true anymore after this PR so we could just remove those mentions

AhmedGrati:
> > @marquiz why should we remove all references to `-extra-label-ns` in `docs/usage/customization-guide.md`?
> 
> Heh, my comment wasn't very clear 😇 I should've written that review all references to `-extra-label-ns` in the doc. IIRC it's mentioned in few places, saying that by default all namespaces except for '*.feature.node.kubernetes.io` are denied... which obviously isn't true anymore after this PR so we could just remove those mentions

Oh yes got it now, thank you 😊

AhmedGrati:
@marquiz I added the fact that we'll forcibly deny the `*.kubernetes.io` namespace.

AhmedGrati:
@marquiz since the pre-processing of the DeniedLabelNs happens whenever we create a new nfd master, the unit tests of the `-extra-label-ns` and `deny-label-ns` flags won't work ( because we are mocking the nfd master instance, then we won't have the DenyLabelNs processed). WDYT?

marquiz:
> @marquiz since the pre-processing of the DeniedLabelNs happens whenever we create a new nfd master, the unit tests of the `-extra-label-ns` and `deny-label-ns` flags won't work ( because we are mocking the nfd master instance, then we won't have the DenyLabelNs processed).

The unit tests have their history and are quite awkward to maintain 🙄 But your problem should be fairly easy to fix by just setting `mockMaster.deniedNs` in the unit test (instead of `mockMaster.args.DenyLabelNs` which doesn't have any effect here, as you explained).

> I suggest testing that on the e2e level. WDYT?

That would be a great addition if you can add it! We're lacking on the e2e front on the cmdline flags.

AhmedGrati:
> The unit tests have their history and are quite awkward to maintain 🙄 But your problem should be fairly easy to fix by just setting `mockMaster.deniedNs` in the unit test (instead of `mockMaster.args.DenyLabelNs` which doesn't have any effect here, as you explained).

True, but setting directly the `mockMaster.deniedNs`, won't cover the test of the preProcessing part of the `DeniedLabelNs`.

docs/reference/master-commandline-reference.md
```diff
@@ -234,6 +232,24 @@ Example:
...
 
+### -deny-label-ns
+
+The `-deny-label-ns` flag specifies a comma-separated list of excluded
+label namespaces. By default, nfd-master allows creating labels in all
+namespaces, excluding `*.kubernetes.io` namespace and its sub-namespaces
+(e.g. `sub.ns.kubernetes.io`). However, you should note that the
+`*.kubernetes.io` label is forcibly denied.
+e.g: `nfd-master -deny-label-ns=""` would still deny `*.kubernetes.io`
+This option can be used to exclude some vendors or application specific namespaces.
```
ArangoGutierrez, line 243: Format to 80 columns 

```suggestion
This option can be used to exclude some vendors or application specific 
namespaces.
```

pkg/nfd-master/nfd-master.go
```diff
@@ -126,6 +133,15 @@ func NewNfdMaster(args *Args) (NfdMaster, error) {
 			return nfd, fmt.Errorf("-ca-file needs to be specified alongside -cert-file and -key-file")
 		}
 	}
+	if args.DenyLabelNs == nil {
+		args.DenyLabelNs = make(utils.StringSetVal)
+	}
+	// We should forcibly deny *.kubernetes.io
```
ArangoGutierrez, line 139: Remove `Should` bc we do 
```suggestion
	// We forcibly deny *.kubernetes.io
```

ArangoGutierrez (changes requested):
Just a couple minor nits, but PR looks good now

docs/reference/master-commandline-reference.md
```diff
@@ -234,6 +232,24 @@ Example:
...
+
+Default: *\*.kubernetes.io*
+
+Example:
+
+```bash
+nfd-master -deny-label-ns=kubernetes.io,vendor-2.io
```
marquiz, line 250: A bit misleading example as kubernetes.io should be always forcibly denied anyway

```suggestion
nfd-master -deny-label-ns=*.vendor.com,vendor-2.io
```

marquiz:
> but setting directly the `mockMaster.deniedNs`, won't cover the test of the preProcessing part of the `DeniedLabelNs`. For the moment, I think it should be done that way.

That is very much true. D'ya have plans for adding e2e-tests? 😉

pkg/nfd-master/nfd-master.go
```diff
@@ -449,6 +468,21 @@ func verifyNodeName(cert *x509.Certificate, nodeName string) error {
...
 }
 
+func isNamespaceDenied(labelNs string, wildcardDeniedNs map[string]struct{}, normalDeniedNs map[string]struct{}) bool {
+	for deniedNs := range normalDeniedNs {
+		if labelNs == deniedNs {
+			return true
+		}
+	}
+	for deniedNs := range wildcardDeniedNs {
+		deniedSuffix := strings.TrimLeft(deniedNs, "*")
```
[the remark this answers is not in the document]
AhmedGrati: Makes sense, will do it 👌

AhmedGrati:
@marquiz I'll work on adding the e2e tests, but it'll take me some time.

pkg/nfd-master/nfd-master.go
```diff
@@ -126,6 +133,15 @@ func NewNfdMaster(args *Args) (NfdMaster, error) {
 			return nfd, fmt.Errorf("-ca-file needs to be specified alongside -cert-file and -key-file")
 		}
 	}
+	if args.DenyLabelNs == nil {
+		args.DenyLabelNs = make(utils.StringSetVal)
+	}
+	// We should forcibly deny *.kubernetes.io
+	args.DenyLabelNs["*.kubernetes.io"] = struct{}{}
```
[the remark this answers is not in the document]
AhmedGrati: Good point, that's true.

marquiz:
> @marquiz I'll work on adding the e2e tests, but it'll take me some time.

No worries, that would be really cool. I think we can take that as a separate PR

AhmedGrati:
Okay then, I'll finish the requested changes and I'll create a separate issue and PR for the e2e tests.

pkg/nfd-master/nfd-master-internal_test.go
```diff
@@ -352,16 +352,16 @@ func TestSetLabels(t *testing.T) {
...
 			instance := "foo"
 			vendorFeatureLabel := "vendor." + nfdv1alpha1.FeatureLabelNs + "/feature-4"
 			vendorProfileLabel := "vendor." + nfdv1alpha1.ProfileLabelNs + "/feature-5"
 			mockLabels := map[string]string{"feature-1": "val-1",
-				"valid.ns/feature-2":   "val-2",
-				"invalid.ns/feature-3": "val-3",
-				vendorFeatureLabel:     " val-4",
-				vendorProfileLabel:     " val-5"}
+				"valid.ns/feature-2":         "val-2",
+				"random.denied.ns/feature-3": "val-3",
```
marquiz, line 362: We could actually add specific tests for `kubernetes.io` and `sub.ns.kubernetes.io` here as they should always be denied.

pkg/utils/flags.go
```diff
@@ -76,8 +76,7 @@ func (a *StringSetVal) String() string {
 	if *a == nil {
 		return ""
 	}
-
-	vals := make([]string, len(*a), 0)
+	vals := make([]string, 0, len(*a))
```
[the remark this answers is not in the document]
AhmedGrati: The old implementation caused an error for me. In fact, it makes sense because, in the `make` documentation, I found this `A second integer argument may be provided to specify a different capacity; it must be no smaller than the length`.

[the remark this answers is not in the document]
marquiz: Oh, apologies, I looked at this too hastily and didn't realize the order of args was swapped.

This is a legit change 👍 How come the old implementation even worked 🧐

docs/reference/master-commandline-reference.md
```diff
@@ -234,6 +232,24 @@ Example:
...
+
+The `-deny-label-ns` flag specifies a comma-separated list of excluded
+label namespaces. By default, nfd-master allows creating labels in all
+namespaces, excluding `*.kubernetes.io` namespace and its sub-namespaces
+(e.g. `sub.ns.kubernetes.io`). However, you should note that the
+`*.kubernetes.io` label is forcibly denied.
+e.g: `nfd-master -deny-label-ns=""` would still deny `*.kubernetes.io`
+This option can be used to exclude some vendors or application specific namespaces.
+
+Default: *\*.kubernetes.io*
```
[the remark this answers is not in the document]
marquiz: With this particular comment I meant

```suggestion
Default: *empty*
```

...as `Default: *\*.kubernetes.io*` would suggest (to me) that you can remove '*kubernetes.io` e.g.with `-deny-label-ns=foo.bar`. *kubernetes.io is always denied irrespective what is specified with this flag

pkg/nfd-master/nfd-master.go
```diff
@@ -126,6 +133,16 @@ func NewNfdMaster(args *Args) (NfdMaster, error) {
...
+	if args.DenyLabelNs == nil {
+		args.DenyLabelNs = make(utils.StringSetVal)
+	}
+	// Pre-process DenyLabelNS into 2 lists: one for normal ns, and the other for wildcard ns
+	normalDeniedNs, wildcardDeniedNs := preProcessDeniedNamespaces(args.DenyLabelNs)
+	nfd.deniedNs.normal = normalDeniedNs
+	nfd.deniedNs.wildcard = wildcardDeniedNs
+	// We forcibly deny kubernetes.io
+	nfd.deniedNs.normal["kubernetes.io"] = struct{}{}
+	nfd.deniedNs.wildcard["kubernetes.io"] = struct{}{}
```
marquiz, line 145: To be correct i think we want to deny `*.kubernetes.io `

```suggestion
	nfd.deniedNs.wildcard[".kubernetes.io"] = struct{}{}
```

docs/reference/master-commandline-reference.md
```diff
@@ -234,6 +232,27 @@ Example:
...
 ```
 
+### -deny-label-ns
+
+The `-deny-label-ns` flag specifies a comma-separated list of excluded
+label namespaces. By default, nfd-master allows creating labels in all
+namespaces, excluding `kubernetes.io` namespace and its sub-namespaces
+(e.g. `*.kubernetes.io`). However, you should note that the
+`kubernetes.io` namespace is forcibly denied.
+e.g: `nfd-master -deny-label-ns=""` would still deny `*.kubernetes.io`
```
marquiz, line 242: Still bike-shedding on this, I'd suggest to adjust the wording a bit

```suggestion
(i.e. `*.kubernetes.io`). However, you should note that
`kubernetes.io` and its sub-namespaces are always denied.
For example, `nfd-master -deny-label-ns=""` would still disallow
`kubernetes.io` and `*.kubernetes.io`.
```

pkg/nfd-master/nfd-master.go
```diff
@@ -78,6 +79,11 @@ type Args struct {
 	ResourceLabels       utils.StringSetVal
 }
 
+type DeniedNs struct {
```
marquiz, line 82: nitty nit: we don't need to export this type

```suggestion
type deniedNs struct {
```

pkg/nfd-master/nfd-master.go
```diff
@@ -392,7 +409,7 @@ func (m *nfdMaster) updateMasterNode() error {
 // into extended resources. This function also handles proper namespacing of
 // labels and ERs, i.e. adds the possibly missing default namespace for labels
 // arriving through the gRPC API.
-func filterFeatureLabels(labels Labels, extraLabelNs map[string]struct{}, labelWhiteList regexp.Regexp, extendedResourceNames map[string]struct{}) (Labels, ExtendedResources) {
+func filterFeatureLabels(labels Labels, extraLabelNs map[string]struct{}, wildcardDeniedLabelNs map[string]struct{}, normalDeniedLabelsNs map[string]struct{}, labelWhiteList regexp.Regexp, extendedResourceNames map[string]struct{}) (Labels, ExtendedResources) {
```
marquiz, line 412: This function prototype starts to be on the longer side, I'd suggest to shorten it a bit by dropping duplication of identical types

```suggestion
filterFeatureLabels(labels Labels, extraLabelNs, wildcardDeniedLabelNs, normalDeniedLabelsNs map[string]struct{}, labelWhiteList regexp.Regexp, extendedResourceNames map[string]struct{}) (Labels, ExtendedResources)
```

TBH, it looks like we should make this just a method of the nfdMaster type, like

```suggestion
func (m *nfdMaster) filterFeatureLabels() (Labels, ExtendedResources) {
...
```

But we can leave that as a future enhancement/refactor, in a separate PR

marquiz (commented):
Thanks @AhmedGrati for the update. Looks like we're almost there, just a few small comments/suggestions

pkg/nfd-master/nfd-master.go
```diff
@@ -392,7 +409,7 @@ func (m *nfdMaster) updateMasterNode() error {
 // into extended resources. This function also handles proper namespacing of
 // labels and ERs, i.e. adds the possibly missing default namespace for labels
 // arriving through the gRPC API.
-func filterFeatureLabels(labels Labels, extraLabelNs map[string]struct{}, labelWhiteList regexp.Regexp, extendedResourceNames map[string]struct{}) (Labels, ExtendedResources) {
+func (m *nfdMaster) filterFeatureLabels(labels Labels) (Labels, ExtendedResources) {
```
marquiz, line 412: Thanks, this looks so much cleaner 👍

marquiz (approved):
Thanks @AhmedGrati for your efforts on this! From my perspective this looks good to go 👍

AhmedGrati:
Many thanks for your review @marquiz 👌.

ArangoGutierrez (approved):
/lgtm

PR #1053, opened 2023-02-09
topology-updater: nrt-api Update to v1alpha2

jlojosnegros:
As part of #1048 we need to update NRT API to `v1alpha2` 
where new top-level attributes have been added to NRT objects.

We will use this top-level attributes to report new pod fingerprint values. see: #1049 

address: #1048 
see: [this issue](https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/issues/30) for more info about new NRT api version

### DIFF
--- a/deployment/base/noderesourcetopologies-crd/noderesourcetopologies.yaml
+++ b/deployment/base/noderesourcetopologies-crd/noderesourcetopologies.yaml
@@ -3,7 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes/enhancements/pull/1870
-    controller-gen.kubebuilder.io/version: v0.7.0
+    controller-gen.kubebuilder.io/version: v0.11.2
   creationTimestamp: null
   name: noderesourcetopologies.topology.node.k8s.io
 spec:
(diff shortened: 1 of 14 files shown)


PiotrProkop:
@jlojosnegros Thanks for the PR! I think we need to propagate `TopologyPolicies` and `Scope` as part of new top-level attributes as part of the effort to move to `v1alpha2`, we can do this as part of this PR or I can create a new one to do this. The logic for that is described here:
```

NodeResourceTopology producers should add top-level `Attributes` in the following format
- For `Name` and `Value` of attributes, words should be `snakeCase`
- The `Name` of each attribute should be **the same of the corresponding kubelet configuration option**.
  - example: `--topology-manager-scope` becomes `topologyManagerScope`
  - example: `topologyManagerPolicy` becomes `topologyMaanagerPolicy`
- The `Value` of each attribute should be **one of the value of the corresponding kubelet configuration option, VERBATIM**.
  - example: `single-numa-node` becomes `single-numa-node`
- Should `topologyManagerOptions` be exposed:
  - they should be expanded in key-value pairs, using the `String()` representation
  - each key-value pair should be preceded by the `topologyManagerOption` prefix
  - every other provision described above applies
  - example: the `prefer-closest-numa-nodes` option becomes `topologyManagerOptionPreferClosestNumaNodes`, accepting exactly one of either `true` and `false`.
  ```

marquiz (approved):
Looks simple and unsurprising to me 👍

ArangoGutierrez (approved):
/lgtm

ffromani:
> @jlojosnegros Thanks for the PR! I think we need to propagate `TopologyPolicies` and `Scope` as part of new top-level attributes as part of the effort to move to `v1alpha2`, we can do this as part of this PR or I can create a new one to do this. The logic for that is described here:
> 
> ```
> 
> NodeResourceTopology producers should add top-level `Attributes` in the following format
> - For `Name` and `Value` of attributes, words should be `snakeCase`
> - The `Name` of each attribute should be **the same of the corresponding kubelet configuration option**.
>   - example: `--topology-manager-scope` becomes `topologyManagerScope`
>   - example: `topologyManagerPolicy` becomes `topologyMaanagerPolicy`
> - The `Value` of each attribute should be **one of the value of the corresponding kubelet configuration option, VERBATIM**.
>   - example: `single-numa-node` becomes `single-numa-node`
> - Should `topologyManagerOptions` be exposed:
>   - they should be expanded in key-value pairs, using the `String()` representation
>   - each key-value pair should be preceded by the `topologyManagerOption` prefix
>   - every other provision described above applies
>   - example: the `prefer-closest-numa-nodes` option becomes `topologyManagerOptionPreferClosestNumaNodes`, accepting exactly one of either `true` and `false`.
> ```

apologies for late answer. I think the cleanest option would be to file a separate PR depending on this one and in parallel to #1049 to add the attributes reporting. Now that this is merged we don't have many other options anyway though :) I need to check if was already filed, if not I'd be happy to work on that.

PiotrProkop:
> > @jlojosnegros Thanks for the PR! I think we need to propagate `TopologyPolicies` and `Scope` as part of new top-level attributes as part of the effort to move to `v1alpha2`, we can do this as part of this PR or I can create a new one to do this. The logic for that is described here:
> > ```
> > 
> > NodeResourceTopology producers should add top-level `Attributes` in the following format
> > - For `Name` and `Value` of attributes, words should be `snakeCase`
> > - The `Name` of each attribute should be **the same of the corresponding kubelet configuration option**.
> >   - example: `--topology-manager-scope` becomes `topologyManagerScope`
> >   - example: `topologyManagerPolicy` becomes `topologyMaanagerPolicy`
> > - The `Value` of each attribute should be **one of the value of the corresponding kubelet configuration option, VERBATIM**.
> >   - example: `single-numa-node` becomes `single-numa-node`
> > - Should `topologyManagerOptions` be exposed:
> >   - they should be expanded in key-value pairs, using the `String()` representation
> >   - each key-value pair should be preceded by the `topologyManagerOption` prefix
> >   - every other provision described above applies
> >   - example: the `prefer-closest-numa-nodes` option becomes `topologyManagerOptionPreferClosestNumaNodes`, accepting exactly one of either `true` and `false`.
> > ```
> 
> apologies for late answer. I think the cleanest option would be to file a separate PR depending on this one and in parallel to #1049 to add the attributes reporting. Now that this is merged we don't have many other options anyway though :) I need to check if was already filed, if not I'd be happy to work on that.

No worries. I have already started working on that :) should have PR ready to review by tomorrow.

PR #1054, opened 2023-02-09
Advertise TopologyManger policy and scope as Attributes in NRT api v1alpha2

PiotrProkop:
We want to advertise TopologyManager policy and scope as top level attributes in noderesourcetopology-api `v1alpha2`.
Based on this PR https://github.com/kubernetes-sigs/scheduler-plugins/pull/488.

### DIFF
--- a/cmd/nfd-topology-updater/main.go
+++ b/cmd/nfd-topology-updater/main.go
@@ -28,7 +28,6 @@ import (
 
 	topology "sigs.k8s.io/node-feature-discovery/pkg/nfd-topology-updater"
 	"sigs.k8s.io/node-feature-discovery/pkg/resourcemonitor"
-	"sigs.k8s.io/node-feature-discovery/pkg/topologypolicy"
 	"sigs.k8s.io/node-feature-discovery/pkg/utils"
 	"sigs.k8s.io/node-feature-discovery/pkg/utils/hostpath"
 	"sigs.k8s.io/node-feature-discovery/pkg/utils/kubeconf"
@@ -88,11 +87,8 @@ func main() {
 		klog.Exitf("unsupported URI scheme: %v", u.Scheme)
 	}
 
-	tmPolicy := string(topologypolicy.DetectTopologyPolicy(klConfig.TopologyManagerPolicy, klConfig.TopologyManagerScope))
-	klog.Infof("detected kubelet Topology Manager policy %q", tmPolicy)
-
 	// Get new TopologyUpdater instance
-	instance := topology.NewTopologyUpdater(*args, *resourcemonitorArgs, tmPolicy)
+	instance := topology.NewTopologyUpdater(*args, *resourcemonitorArgs, klConfig.TopologyManagerPolicy, klConfig.TopologyManagerScope)
 
 	if err = instance.Run(); err != nil {
 		klog.Exit(err)
--- a/pkg/nfd-topology-updater/nfd-topology-updater.go
+++ b/pkg/nfd-topology-updater/nfd-topology-updater.go
@@ -32,9 +32,17 @@ import (
 	"sigs.k8s.io/node-feature-discovery/pkg/apihelper"
 	"sigs.k8s.io/node-feature-discovery/pkg/podres"
 	"sigs.k8s.io/node-feature-discovery/pkg/resourcemonitor"
+	"sigs.k8s.io/node-feature-discovery/pkg/topologypolicy"
 	"sigs.k8s.io/node-feature-discovery/pkg/utils"
 	"sigs.k8s.io/node-feature-discovery/pkg/version"
 	"sigs.k8s.io/yaml"
+)
+
+const (
+	// TopologyManagerPolicyAttributeName represents an attribute which defines Topology Manager Policy
+	TopologyManagerPolicyAttributeName = "topologyManagerPolicy"
+	// TopologyManagerScopeAttributeName represents an attribute which defines Topology Manager Policy Scope
+	TopologyManagerScopeAttributeName = "topologyManagerScope"
 )
 
 // Args are the command line arguments
@@ -60,10 +68,21 @@ type NfdTopologyUpdater interface {
 type staticNodeInfo struct {
 	nodeName string
 	tmPolicy string
+	tmScope  string
+}
+
+func newStaticNodeInfo(policy, scope string) staticNodeInfo {
+	nodeName := utils.NodeName()
+	klog.InfoS("detected kubelet Topology Manager configuration", "policy", policy, "scope", scope, "nodeName", nodeName)
+	return staticNodeInfo{
+		nodeName: nodeName,
+		tmPolicy: policy,
+		tmScope:  scope,
+	}
 }
 
 type nfdTopologyUpdater struct {
-	nodeInfo            *staticNodeInfo
+	nodeInfo            staticNodeInfo
 	args                Args
 	apihelper           apihelper.APIHelpers
 	resourcemonitorArgs resourcemonitor.Args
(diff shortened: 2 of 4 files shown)


PiotrProkop:
I'll add unit tests for this tomorrow. @ffromani @swatisehgal please review.

ffromani (commented):
thanks for the quick PR! looks straightforward, suggest possible improvements inside.

pkg/topologypolicy/topology-policy.go
```diff
@@ -63,3 +70,16 @@ func detectPolicyContainerScope(policy string) v1alpha2.TopologyManagerPolicy {
 		return v1alpha2.None
 	}
 }
+
+func DetectTopologyAttributes(policy string, scope string) v1alpha2.AttributeList {
```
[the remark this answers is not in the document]
PiotrProkop: will do.

cmd/nfd-topology-updater/main.go
```diff
@@ -88,11 +87,13 @@ func main() {
 		klog.Exitf("unsupported URI scheme: %v", u.Scheme)
 	}
 
-	tmPolicy := string(topologypolicy.DetectTopologyPolicy(klConfig.TopologyManagerPolicy, klConfig.TopologyManagerScope))
+	tmPolicy := string(klConfig.TopologyManagerPolicy)
 	klog.Infof("detected kubelet Topology Manager policy %q", tmPolicy)
+	tmScope := string(klConfig.TopologyManagerScope)
+	klog.Infof("detected kubelet Topology Manager scope %q", tmScope)
```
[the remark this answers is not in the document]
PiotrProkop: Makes sense, changing it now.

pkg/topologypolicy/topology-policy.go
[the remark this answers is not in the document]
ffromani: Thanks! if you move, please also rename this function! The code is just fine, but it is not detecting, rather constructing/creating

cmd/nfd-topology-updater/main.go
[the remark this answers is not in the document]
PiotrProkop: I have put logging policy, scope and nodeName to new func called `newStaticNodeInfo`.

pkg/topologypolicy/topology-policy.go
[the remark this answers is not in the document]
PiotrProkop: moved and changed, and got rid of `AttributeName` type alias which was unused :)

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -60,6 +68,17 @@ type NfdTopologyUpdater interface {
 type staticNodeInfo struct {
 	nodeName string
 	tmPolicy string
+	tmScope  string
+}
+
+func newStaticNodeInfo(policy, scope string) *staticNodeInfo {
```
ffromani, line 74: nit/nonblocking: I'd just return a struct (not a pointer to a struct) and trust the compiler.
In general, returning pointers *may* allow faster copying of objects (but it should actually be measured) but also contributes to the GC pressure (which, OTOH, should also be measured).
What I'm trying to say is that returning pointers is not necessarily the faster option,

ffromani (commented):
looks fine. I'd like to see some form of test coverage (proably integration/e2e) before LGTM

pkg/nfd-topology-updater/nfd-topology-updater.go
ffromani: note: I'd NOT resubmit just for this.

PiotrProkop: I want to add unit tests, so I'll change it when I submit them.

PiotrProkop:
@ffromani I had to change the way how we load Kubelet config when read from file to make e2e test working, previously we were just reading config from file and unamarshalling it to struct. This approach didn't fill out default values but were skipping it based on `omitempty` tag. In e2e tests we are reading Kubelet config via http, which is serving Kubeletconfig with all values filled 😄

test/e2e/utils/noderesourcetopology.go
```diff
@@ -259,3 +279,12 @@ func isValidResourceList(zoneName string, resources v1alpha2.ResourceInfoList) b
 	}
 	return foundCpu
 }
+
+func containsAttribute(attributes v1alpha2.AttributeList, attribute v1alpha2.AttributeInfo) bool {
+	for _, attr := range attributes {
+		if reflect.DeepEqual(attr, attribute) {
```
ffromani, line 285: minor: reflect is no bother, but can't we just use '=='?

pkg/utils/kubeconf/kubelet_config_file.go
```diff
@@ -17,23 +17,40 @@ limitations under the License.
...
 
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
+	kubeletconfigscheme "k8s.io/kubernetes/pkg/kubelet/apis/config/scheme"
+	"k8s.io/kubernetes/pkg/kubelet/kubeletconfig/configfiles"
+	utilfs "k8s.io/kubernetes/pkg/util/filesystem"
 )
 
 // GetKubeletConfigFromLocalFile returns KubeletConfiguration loaded from the node local config
+// based on https://github.com/kubernetes/kubernetes/blob/master/cmd/kubelet/app/server.go#L337
+// it fills empty fields with default values
```
ffromani, line 30: this looks indeed more correct, thanks for improving this function

ffromani (commented):
/lgtm
very nice work @PiotrProkop !

test/e2e/utils/noderesourcetopology.go
PiotrProkop: I was thinking about doing:
```go
if attr.Name == attribute.Name && attr.Value == attribute.Value {
    return true
}
```
but settle on shorter one.

pkg/utils/kubeconf/kubelet_config_file.go
PiotrProkop: 🎸

marquiz:
Thanks @PiotrProkop for the PR. I'll check this next week

test/e2e/utils/noderesourcetopology.go
ffromani: I would have used the explicit == version, but I don't have strong feelings so works for me.

marquiz: My thinking also that `attr == attribute` should've done the job but especially in the e2e tests I don't have that strong feelings either so let's go with this...

pkg/utils/kubeconf/kubelet_config_file.go
```diff
@@ -17,23 +17,40 @@ limitations under the License.
...
 func GetKubeletConfigFromLocalFile(kubeletConfigPath string) (*kubeletconfigv1beta1.KubeletConfiguration, error) {
-	kubeletBytes, err := os.ReadFile(kubeletConfigPath)
+	const errFmt = "failed to load Kubelet config file %s, error %v"
```
marquiz, line 32: nit: wrap errors with `%w`

```suggestion
	const errFmt = "failed to load Kubelet config file %s: %w"
```

PiotrProkop: fixed

marquiz (approved):
Thanks @PiotrProkop, looks good to me now

PiotrProkop:
@ffromani can you LGTM once more? I had to fix a small nit.

PR #1055, opened 2023-02-10
Add discovery duration logging

VillePihlava:
This pull request adds logging of discovery durations. There are some optional things mentioned in https://github.com/kubernetes-sigs/node-feature-discovery/issues/899 by @eero-t. I had a quick discussion with @marquiz and if I understood correctly something like what is implemented here could also be useful. Depending on what exactly is wanted, this could close https://github.com/kubernetes-sigs/node-feature-discovery/issues/899. This is related to work done in https://github.com/kubernetes-sigs/node-feature-discovery/pull/1050.

The logging output can look like this:
```console
<output from other sources>
I0213 12:42:32.317317 2204667 nfd-worker.go:199] discovery duration for "system" source: 472.156µs
I0213 12:42:32.318310 2204667 usb.go:150] discovered usb features:
I0213 10:51:09.882921 2197292 usb.go:150] <many lines of output>
I0213 12:42:32.318726 2204667 nfd-worker.go:199] discovery duration for "usb" source: 1.397359ms
I0213 12:42:32.318736 2204667 nfd-worker.go:203] feature discovery for 9 sources lasted for: 6s
W0213 12:42:32.318745 2204667 nfd-worker.go:205] feature discovery sources took over half (6s) of sleep interval (10s)
I0213 12:42:32.318755 2204667 nfd-worker.go:561] starting feature discovery...
```
Fixes https://github.com/kubernetes-sigs/node-feature-discovery/issues/899

### DIFF
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -190,11 +190,19 @@ func (i *infiniteTicker) Reset(d time.Duration) {
 
 // Run feature discovery.
 func (w *nfdWorker) runFeatureDiscovery() error {
+	discoveryStart := time.Now()
 	for _, s := range w.featureSources {
-		klog.V(2).Infof("running discovery for %q source", s.Name())
+		currentSourceStart := time.Now()
 		if err := s.Discover(); err != nil {
 			klog.Errorf("feature discovery of %q source failed: %v", s.Name(), err)
 		}
+		klog.V(3).Infof("discovery duration for %q source: %v", s.Name(), time.Since(currentSourceStart))
+	}
+
+	discoveryDuration := time.Since(discoveryStart)
+	klog.V(2).Infof("feature discovery for %d sources lasted for: %v", len(w.featureSources), discoveryDuration)
+	if w.config.Core.SleepInterval.Duration > 0 && discoveryDuration > w.config.Core.SleepInterval.Duration/2 {
+		klog.Warningf("feature discovery sources took over half (%v) of sleep interval (%v)", discoveryDuration, w.config.Core.SleepInterval.Duration)
 	}
 
 	// Get the set of feature labels.
@@ -359,7 +367,7 @@ func (w *nfdWorker) grpcDisconnect() {
 }
 func (c *coreConfig) sanitize() {
 	if c.SleepInterval.Duration > 0 && c.SleepInterval.Duration < time.Second {
-		klog.Warningf("too short sleep-intervall specified (%s), forcing to 1s",
+		klog.Warningf("too short sleep interval specified (%s), forcing to 1s",
 			c.SleepInterval.Duration.String())
 		c.SleepInterval = duration{time.Second}
 	}


marquiz (approved):
Thanks @VillePihlava, this is a nice improvement. The KlogDump prints sources in deterministic order which is nice.

What we'd like is to have similar information for hooks run in `getFeaturesFromGooks`( in `source/loca/local.go`) because those are the most likely source of "surprises" (even if they are deprecated). This can be addressed in a separate PR, though

pkg/nfd-worker/nfd-worker.go
```diff
@@ -190,12 +190,16 @@ func (i *infiniteTicker) Reset(d time.Duration) {
...
+	durations := make(map[string]string, len(w.featureSources))
 	for _, s := range w.featureSources {
 		klog.V(2).Infof("running discovery for %q source", s.Name())
+		discoverStart := time.Now()
 		if err := s.Discover(); err != nil {
 			klog.Errorf("feature discovery of %q source failed: %v", s.Name(), err)
 		}
+		durations[s.Name()] = time.Since(discoverStart).String()
 	}
+	utils.KlogDump(3, "discovery durations for feature sources:", "  ", durations)
```
eero-t, line 202: What's the advantage of using intermediate map & printing things only after whole loop, instead of just printing header before loop, moving Infof() printing after Discover(), and adding time to that log item?

eero-t:
> Depending on what exactly is wanted, this could close https://github.com/kubernetes-sigs/node-feature-discovery/issues/899

@VillePihlava While this could be a step into direction requested in #899, it does not yet do *any* of the suggested 3 items.

#899 requires measuring total time used for given discovery round, and deducting that from the sleep interval used between discovery rounds (or using time ticker for the interval as suggested by @marquiz in #899).

eero-t:
To finish #899 (now that https://github.com/kubernetes-sigs/node-feature-discovery/commit/2101cb20e47599d1ba11c78547ea2ef3508dadc8 is merged), this should check (also) the total time used for a discovery round, and e.g. log an error for discovery rounds that exceed sleep interval, and/or a warning if total (+ any individual discoveries?)  is (say) >50% of it.

I'm not sure which one is more important; an error telling that discoveries, or discovery interval, is not working, or an early warning of approaching the limit.   Because that is easy to fix just by increasing `core.sleepInterval`, I do not think these messages would need rate-limiting.

VillePihlava:
> To finish #899 (now that [2101cb2](https://github.com/kubernetes-sigs/node-feature-discovery/commit/2101cb20e47599d1ba11c78547ea2ef3508dadc8) is merged), this should check (also) the total time used for a discovery round, and e.g. log an error for discovery rounds that exceed sleep interval, and/or a warning if total (+ any individual discoveries?) is (say) >50% of it.
> 
> I'm not sure which one is more important; an error telling that discoveries, or discovery interval, is not working, or an early warning of approaching the limit. Because that is easy to fix just by increasing `core.sleepInterval`, I do not think these messages would need rate-limiting.

I'll look into this. Also added a link to the other pull request to avoid confusion in the original message.

marquiz:
> To finish #899 (now that [2101cb2](https://github.com/kubernetes-sigs/node-feature-discovery/commit/2101cb20e47599d1ba11c78547ea2ef3508dadc8) is merged), this should check (also) the total time used for a discovery round, and e.g. log an error for discovery rounds that exceed sleep interval, and/or a warning if total (+ any individual discoveries?) is (say) >50% of it.

We could log a warning. I wouldn't tie this to the sleep interval but put some reasonable hard limit, e.g. the discovery really shouldn't take more than 1 or 2 seconds in any scenario, I think. If you put sleep interval to 1 hour and the discovery takes 20 minutes something is really badly wrong (IMHO)

eero-t:
> If you put sleep interval to 1 hour and the discovery takes 20 minutes something is really badly wrong (IMHO)

Why? That would be a deliberate user decision.  It could just be something naturally slow, but not very taxing on resources (maybe not RFC 2549, but something).

ArangoGutierrez:
> > If you put sleep interval to 1 hour and the discovery takes 20 minutes something is really badly wrong (IMHO)
> 
> Why? That would be a deliberate user decision. It could just be something naturally slow, but not very taxing on resources (maybe not RFC 2549, but something).

I agree with @marquiz , discovery usually takes seconds, we should aim for lowering the foot print of NFD, allowing minutes won't move us in that direction 

Aside from that, this PR looks good to me

eero-t:
> I agree with @marquiz , discovery usually takes seconds, we should aim for lowering the foot print of NFD, allowing minutes won't move us in that direction

I think short enough default `sleepInterval`, and giving the warning much earlier (e.g. 10% of interval), would work better than using a fixed time limit for when warning is given.

If user increases the interval from default, he's likely to have a good reason for it.

ArangoGutierrez:
> > I agree with @marquiz , discovery usually takes seconds, we should aim for lowering the foot print of NFD, allowing minutes won't move us in that direction
> 
> I think short enough default `sleepInterval`, and giving the warning much earlier (e.g. 10% of interval), would work better than using a fixed time limit for when warning is given.
> 
> If user increases the interval from default, he's likely to have a good reason for it.

Umm Maybe for custom Hooks, but we are deprecating hooks, if user wants to discover something that they know takes time, is better for them to use a side car that publishes to the NodeFeature CR, than to increasing the discovery time toleration on NFD core

marquiz:
> If user increases the interval from default, he's likely to have a good reason for it.

I think the usual reason is something else than preparing for that the nfd-worker is churning feature discovery for ages. As @ArangoGutierrez noted hooks are being deprecated anyway so I wouldn't spend too much time on this. The built-in discovery should be in the order of milliseconds (sub-second anyway) so anything longer than a few seconds is raising concnerns

marquiz (approved):
Thanks @VillePihlava, looks good to me.

ping @eero-t

pkg/nfd-worker/nfd-worker.go
VillePihlava: I removed the `Infof` before `Discover` and added logging after which includes the duration.

ArangoGutierrez (approved):
/lgtm 
Unhold at will @marquiz 
/hold

VillePihlava:
Noticed that the log levels were different, I changed both to 2 now

eero-t:
> As @ArangoGutierrez noted hooks are being deprecated anyway so I wouldn't spend too much time on this. The built-in discovery should be in the order of milliseconds (sub-second anyway) so anything longer than a few seconds is raising concnerns

@marquiz unfortunately you both missed the point. If interval is **1** second, doing warning only when discovery actually takes **>2** seconds is not really useful.  Warning should be logged *before* the whole interval is used.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -190,11 +190,19 @@ func (i *infiniteTicker) Reset(d time.Duration) {
...
 		if err := s.Discover(); err != nil {
 			klog.Errorf("feature discovery of %q source failed: %v", s.Name(), err)
 		}
+		klog.V(2).Infof("discovery duration for %q source: %v", s.Name(), time.Since(currentSourceStart))
+	}
+
+	discoveryDuration := time.Since(discoveryStart)
+	klog.V(2).Infof("feature discovery for sources lasted for: %v", discoveryDuration)
+	if discoveryDuration > 5*time.Second {
+		klog.Warningf("feature discovery for sources lasted for more than 5s (%v)", discoveryDuration)
```
eero-t, line 205: Warning at 5s is not useful when the discovery interval is smaller.  I would go for something like this (untested!):
```suggestion
	if discoveryDuration > core.sleepInterval/2 {
		klog.Warningf("Feature discovery sources took over half (%v) of sleep interval (%v)", discoveryDuration, core.sleepInterval)
```

(Other alternative would be having measurement and warning like this where the sleep is done instead.)

PS. commit message is missing "Fixes:" tag for https://github.com/kubernetes-sigs/node-feature-discovery/issues/899.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -190,11 +190,19 @@ func (i *infiniteTicker) Reset(d time.Duration) {
...
-		klog.V(2).Infof("running discovery for %q source", s.Name())
+		currentSourceStart := time.Now()
 		if err := s.Discover(); err != nil {
 			klog.Errorf("feature discovery of %q source failed: %v", s.Name(), err)
 		}
+		klog.V(2).Infof("discovery duration for %q source: %v", s.Name(), time.Since(currentSourceStart))
+	}
+
+	discoveryDuration := time.Since(discoveryStart)
+	klog.V(2).Infof("feature discovery for sources lasted for: %v", discoveryDuration)
```
eero-t, line 203: For improved log output amount control, per-source logging could be V(3), and this V(2), with count included to it:
```suggestion
	klog.V(2).Infof("feature discovery for %d sources lasted for: %v", len(w.featureSources), discoveryDuration)
```
(Just a suggestion, feel free to ignore.)

ArangoGutierrez: @VillePihlava please address, I agree with @eero-t

pkg/nfd-worker/nfd-worker.go
ArangoGutierrez: @VillePihlava please address, I agree with @eero-t

ArangoGutierrez: @eero-t PTAL

pkg/nfd-worker/nfd-worker.go
ArangoGutierrez: @eero-t PTAL

VillePihlava: This has now been changed

ArangoGutierrez: @eero-t PTAL

pkg/nfd-worker/nfd-worker.go
VillePihlava: Also added a check to see if the duration is positive (negative durations should not cause a warning)

ArangoGutierrez (approved):
/lgtm

pkg/nfd-worker/nfd-worker.go
eero-t: Looks fine now!

pkg/nfd-worker/nfd-worker.go
eero-t: Good point, looks fine now!

marquiz:
Looks like everybody consent on merging
/unhold
/ok-to-test

PR #1061, opened 2023-02-16
scripts/test-infra: bump golangci-lint to v1.51.1

marquiz:
Update to the latest release.

### DIFF
--- a/scripts/test-infra/verify.sh
+++ b/scripts/test-infra/verify.sh
@@ -2,7 +2,7 @@
 
 # Install deps
 gobinpath="$(go env GOPATH)/bin"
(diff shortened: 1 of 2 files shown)


ArangoGutierrez (approved):
/lgtm

ArangoGutierrez:
the error is... 
```bash
test/e2e/e2e_test.go:66:2: SA1019: rand.Seed has been deprecated since Go 1.20 and an alternative has been available since Go 1.0: Programs that call Seed and then expect a specific sequence of results from the global random source (using functions such as Int) can be broken when a dependency changes how much it consumes from the global random source. To avoid such breakages, programs that need a specific result sequence should use NewRand(NewSource(seed)) to obtain a random generator that other packages cannot access. (staticcheck)
	rand.Seed(time.Now().UnixNano())
```

ArangoGutierrez (approved):
/lgtm

PR #1063, opened 2023-02-17
source/cpu: fix build flags of cpuid detection

marquiz:
Non-x86_64 platforms are virtually only buildable on Linux because the getauxval() glibc function is missing on many platforms.


marquiz:
Quick Friday night hack. Need to verify/think twice this doesn't break anything
/hold

marquiz:
Works for x86_64 (linux and darwin) and arm64 darwin
/unhold
/assign @ArangoGutierrez @zvonkok

marquiz:
ping @yselkowitz @lubinszARM

ArangoGutierrez (approved):
/lgtm

Issue #1033, opened 2023-01-16
Enable codecov reporting


marquiz:
**What would you like to be added**:

Now that https://github.com/kubernetes/org/issues/3084 has been solved, we could enable codecov reporing in NFD. We should probably separate unit tests from the image build (Dockerfile) into a distinct prow job. An outline of sub-tasks (might be missing some)
1. Enable coverage output in Makefile `test:` target
2. Add new `presubmit-test` Prow job
   - A new test script in NFD under `scripts/test-infra/`
   - Job config in https://github.com/kubernetes/test-infra
   - Possibly remove `make test` from Dockerfile(?)
3.  Get codecov token added as a secret in the k8s test-infra cluster
4. Add codecov reporting in the NFD test-infra script (``scripts/test-infra/`)

**Why is this needed**:

Improve development process

ArangoGutierrez:
+1

fmuyassarov:
@marquiz do you plan to work on this ?

marquiz:
> @marquiz do you plan to work on this ?

Feel free to help out here 😉 TBH, I don't know if I have time to take care of this in the near future

AhmedGrati:
@marquiz I can work on it once I finish my current PR.

fmuyassarov:
Hi @AhmedGrati. Thanks for offering help. But I had it halfway. If you don't mind let me finish it and if I don't for some reason I will let you know here.
/assign

marquiz:
I don't think we're quite ready yet
/reopen

ArangoGutierrez:
I think we are ready here, right @fmuyassarov

PR #1069, opened 2023-03-01
test: add code coverage reporting

fmuyassarov:
Add code coverage reporting so that the report is then available to the PR author and reviewers via codecov GitHub app.

@marquiz for uploading/reporting the coverage results, we need to run `make test`. We already run the `make test` with one of the Prow job and I didn't want to add another one in the form of GitHub action. But GA is the most straightforward way of adding codecov and uploading it. There is an uploader available which can be executed from shell as well, so that can be added into Prow job run section, but for that codecov requires the token. We could store the token as a secret on GitHub but I don't know if there is a way actually for the Prow to ready/fetch that token. At least I have not seen anything like that. 

With GitHub actions it is not a problem, because we just pass the `${{ secrets.CODECOV_TOKEN }}` and voilà.

Do you have any other alternatives in mind or if not shall I go for adding another make test as GitHub action? I also thought that maybe we drop running unit tests from Prow to avoid duplication but Prow runs on much faster environment than what GitHub provides for free.

Fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/1033

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,7 +178,7 @@ helm-lint:
 	helm lint --strict deployment/helm/node-feature-discovery/
 
 test:
-	$(GO_CMD) test ./cmd/... ./pkg/... ./source/...
+	$(GO_CMD) test -covermode=atomic -coverprofile=coverage.out ./cmd/... ./pkg/... ./source/...
 
 e2e-test:
 	@if [ -z ${KUBECONFIG} ]; then echo "[ERR] KUBECONFIG missing, must be defined"; exit 1; fi


fmuyassarov:
@marquiz  Actually from https://github.com/kubernetes-sigs/node-feature-discovery/issues/1033, I see that you have already mentioned that we can add codecov token as a secret in the k8s test-infra cluster. Do you know what's the procedure for that?

marquiz (approved):
Thanks @fmuyassarov for working on this.

I think integrating it into prow/test-infra (instead of Github actions) would be the way to go, as you pondered

fmuyassarov:
/cc @ArangoGutierrez

Just general update: we need to add a secret into the test infra k8s cluster and I'm in progress of reaching out the team responsible for that. Once done, I will submit another patch in the test-infra to add covecov report uploading steps.

marquiz:
> Just general update: we need to add a secret into the test infra k8s cluster and I'm in progress of reaching out the team responsible for that. Once done, I will submit another patch in the test-infra to add codecov report uploading steps.

But this PR is not dependent on that

fmuyassarov:
> > Just general update: we need to add a secret into the test infra k8s cluster and I'm in progress of reaching out the team responsible for that. Once done, I will submit another patch in the test-infra to add codecov report uploading steps.
> 
> But this PR is not dependent on that

yes yes. It was just a heads up for Eduardo, because we discussed with you the plan in slack but not here.

ArangoGutierrez:
> > > Just general update: we need to add a secret into the test infra k8s cluster and I'm in progress of reaching out the team responsible for that. Once done, I will submit another patch in the test-infra to add codecov report uploading steps.
> > 
> > 
> > But this PR is not dependent on that
> 
> yes yes. It was just a heads up for Eduardo, because we discussed with you the plan in slack but not here.

So... Can I lgtm?  or wait?

ArangoGutierrez (approved):
/lgtm

PR #1070, opened 2023-03-02
test: implement e2e test of the deny-label-ns flag

### LINKED ISSUE #1060: Enhancement: Implement e2e tests of the -deny-label-ns flag
Add the e2e tests of the newly added flag `-deny-label-ns`.

**Why is this needed**:
Ensures our product stability, and increases confidence in this new feature.

AhmedGrati:
Resolves #1060.


marquiz (commented):
Thanks @AhmedGrati for the PR. And, really good to get some coverage for the extra-label-ns flag, too. Just a few comments but generally looks good

test/e2e/data/nodefeature-3.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeature
+metadata:
+  name: zzz-e2e-features-3
+spec:
+  labels:
+    random.denied.ns/e2e-nodefeature-test-1: "denied-ns"
+    random.unwanted.ns/e2e-nodefeature-test-2: "unwanted-ns"
+    custom.vendor.io/e2e-nodefeature-test-3: "vendor-ns"
+    e2e-nodefeature-test-4: "obj-4"
```
marquiz, line 10: You could add `kubernetes.io/denied-label` and `subns.kubernetes.io/denied-label` here to verify that these really are denied by default

test/e2e/node_feature_discovery_test.go
```diff
@@ -215,6 +226,8 @@ var _ = SIGDescribe("NFD master and worker", func() {
 					testpod.SpecWithContainerImage(dockerImage()),
 					testpod.SpecWithTolerations(testTolerations),
 					testpod.SpecWithContainerExtraArgs("-enable-taints"),
+					addDenyLabelNsFlag([]string{"*.denied.ns", "random.unwanted.ns", "*.vendor.io"}),
```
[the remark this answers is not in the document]
AhmedGrati: Yes makes sense!

test/e2e/node_feature_discovery_test.go
```diff
@@ -725,25 +804,36 @@ func waitForNfdNodeAnnotations(cli clientset.Interface, expected map[string]stri
 
 type k8sLabels map[string]string
 
-// waitForNfdNodeLabels waits for node to be labeled as expected.
-func waitForNfdNodeLabels(cli clientset.Interface, expected map[string]k8sLabels) error {
+// checkForNfdNodeLabels waits and checks that node is labeled as expected.
+func checkForNodeLabels(cli clientset.Interface, expected map[string]k8sLabels, denied map[string]k8sLabels, namespaces []string) error {
```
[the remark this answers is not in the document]
AhmedGrati: What do you mean exactly by "original" labels?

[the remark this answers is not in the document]
marquiz: I mean whatever labels the node had before the test case. I.e. all `node.Lables`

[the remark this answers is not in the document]
marquiz: So you have as arguments `allLabels`, `allOldLabels` and `expectedAddedLabels` and then basically check that

```
allLabels == allOldLabels + expectedAddedLabels
```

Makes sense?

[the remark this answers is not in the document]
AhmedGrati: Yes, sure that's great!

test/e2e/node_feature_discovery_test.go
```diff
@@ -215,6 +215,18 @@ var _ = SIGDescribe("NFD master and worker", func() {
...
+							"-deny-label-ns=%v",
+							strings.Join([]string{"*.denied.ns", "random.unwanted.ns", "*.vendor.io"}[:], ","),
+						),
+					),
+					testpod.SpecWithContainerExtraArgs(
+						fmt.Sprintf(
+							"-extra-label-ns=%v",
+							strings.Join([]string{"custom.vendor.io"}[:], ","),
+						),
+					),
```
marquiz, line 229: Is there a reason you want to use Sprintf instead of simple one-liner

```suggestion
					testpod.SpecWithContainerExtraArgs("-extra-label-ns=custom.vendor.io"),
```

Also applies to the `-deny-label-ns` above...

test/e2e/node_feature_discovery_test.go
```diff
@@ -518,14 +532,17 @@ var _ = SIGDescribe("NFD master and worker", func() {
...
+						expectedLabels, oldLabels,
+					)).NotTo(HaveOccurred())
 					By("Deleting NodeFeature object")
 					err = nfdClient.NfdV1alpha1().NodeFeatures(f.Namespace.Name).Delete(context.TODO(), nodeFeatures[0], metav1.DeleteOptions{})
 					Expect(err).NotTo(HaveOccurred())
 
 					By("Verifying node labels from NodeFeature object were removed")
-					Expect(waitForNfdNodeLabels(f.ClientSet, nil)).NotTo(HaveOccurred())
+					Expect(checkForNodeLabels(f.ClientSet,
+						map[string]k8sLabels{}, oldLabels,
```
marquiz, line 544: You could probably use just nil here?

```suggestion
						nil, oldLabels,
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -725,24 +804,26 @@ func waitForNfdNodeAnnotations(cli clientset.Interface, expected map[string]stri
 
 type k8sLabels map[string]string
 
-// waitForNfdNodeLabels waits for node to be labeled as expected.
-func waitForNfdNodeLabels(cli clientset.Interface, expected map[string]k8sLabels) error {
+// checkForNfdNodeLabels waits and checks that node is labeled as expected.
+func checkForNodeLabels(cli clientset.Interface, expectedNewLabels map[string]k8sLabels, oldLabels map[string]string) error {
```
marquiz, line 808: Hmm, this doesn't work as the `oldLabels` might be different between nodes. Similar to `expectedNewLabels` you should have per-node old-labels like `oldLabels map[string]k8sLabels`.

Actually, probably the easiest would be to have just `oldNodes []corev1.Node` so you could just give the full nodelist as an arg when calling `checkForNodeLabels`, with no need to construct anything like `map[string]k8sLabels`...

AhmedGrati: @marquiz  I see that constructing the `map[string]k8sLabels` would be better because inside the `checkForNodeLabels` function, we can access the labels of a specific node in `O(1)` time complexity, but without the construction, we need to access the labels of a node in `O(n)` time complexity where `n is the number of nodes. please correct me if I'm wrong.

marquiz: Basically true, but it is insignificant, especially in this case of e2e tests. I think less lines-of-code and better readability is preferable.

test/e2e/node_feature_discovery_test.go
```diff
@@ -828,15 +897,28 @@ func getNonControlPlaneNodes(cli clientset.Interface) ([]corev1.Node, error) {
...
+}
 
+func mergeMaps(a map[string]string, b map[string]string) map[string]string {
+	result := map[string]string{}
+	if a == nil {
+		return b
+	}
+	if b == nil {
+		return a
+	}
```
marquiz, line 916: nit: i'd suggest this to work consistently and always return a new copy, i.e. never return the original maps

```suggestion
	if a == nil {
		return b
	}
	if b == nil {
		return a
	}
```

AhmedGrati: Nil checks are useless actually in this case, I can iterate directly on these maps, it won't cause any issue.

marquiz:
I tried to run these and hit some issues. We need #1074 before merging this one

marquiz:
@AhmedGrati please rebase (and address the one open comment). After that we should be good to go

marquiz:
> which comment?

Woops, my review was pending, not submitted 🤦

marquiz (approved):
Great work @AhmedGrati on this! I'm ready to get this in

test/e2e/node_feature_discovery_test.go
```diff
@@ -747,24 +814,27 @@ func waitForNfdNodeAnnotations(cli clientset.Interface, expected map[string]stri
...
-				if defaultExpected, ok := expected["*"]; ok {
+				if defaultExpected, ok := expectedNewLabels["*"]; ok {
 					nodeExpected = defaultExpected
 				}
 			}
-			if !cmp.Equal(nodeExpected, labels) {
-				return fmt.Errorf("node %q labels do not match expected, diff (expected vs. received): %s", node.Name, cmp.Diff(nodeExpected, labels))
+
+			oldLabels := getNodeLabels(oldNodes, node.Name)
+			newNodeLabels := mergeMaps(oldLabels, nodeExpected)
```
fmuyassarov, line 835: not a blocker, but I wonder if we could instead just use `Copy()` from [golang.org/x/exp/maps](golang.org/x/exp/maps) ?

fmuyassarov: @marquiz @AhmedGrati any thoughts ?

marquiz: Good point @fmuyassarov, I think we should start leveraging generics. If we do this, in addition to copy we probably want to clone too to be on the safe side

```golang
newNodeLabels := maps.Copy(maps.Clone(oldLabels), nodeExpected)
```

But not a blocker in my opinion either. It's up to you folks. @AhmedGrati if you want to do this, go ahead. If not, we can do this later – and optimize (in terms of LOC) other maps usage in NFD at the same time.

marquiz (commented):
Thanks @AhmedGrati! Looks good 👍 You just need to update go.mod to include the new dep e.g. by running `go mod tidy`

test/e2e/node_feature_discovery_test.go
```diff
@@ -747,24 +815,28 @@ func waitForNfdNodeAnnotations(cli clientset.Interface, expected map[string]stri
...
+				if defaultExpected, ok := expectedNewLabels["*"]; ok {
 					nodeExpected = defaultExpected
 				}
 			}
-			if !cmp.Equal(nodeExpected, labels) {
-				return fmt.Errorf("node %q labels do not match expected, diff (expected vs. received): %s", node.Name, cmp.Diff(nodeExpected, labels))
+
+			oldLabels := getNodeLabels(oldNodes, node.Name)
+
+			maps.Copy(maps.Clone(oldLabels), nodeExpected)
```
marquiz, line 837: This doesn't look correct now. Shouldn't it be smth like 

```suggestion
			expectedNewLabels := maps.Copy(maps.Clone(oldLabels), nodeExpected)
```

AhmedGrati:
```golang
func maps.Copy(dst map[string]string, src k8sLabels)
```
Actually, Copy function doesn't return anything, it copies from src into a destination.

test/e2e/node_feature_discovery_test.go
```diff
@@ -747,24 +815,28 @@ func waitForNfdNodeAnnotations(cli clientset.Interface, expected map[string]stri
...
+				if defaultExpected, ok := expectedNewLabels["*"]; ok {
 					nodeExpected = defaultExpected
 				}
 			}
-			if !cmp.Equal(nodeExpected, labels) {
-				return fmt.Errorf("node %q labels do not match expected, diff (expected vs. received): %s", node.Name, cmp.Diff(nodeExpected, labels))
+
+			oldLabels := getNodeLabels(oldNodes, node.Name)
+
+			maps.Copy(nodeExpected, maps.Clone(oldLabels))
```
marquiz, line 837: You cannot do it this way, overwriting stuff in expected with stuff from old labels

```suggestion
			expectedNewLabels := maps.Clone(oldLabels)
			maps.Copy(expectedNewLabels, nodeExpected)
```

test/e2e/node_feature_discovery_test.go
marquiz: True, my bad

marquiz (approved):
Looks good to me now 👍

ping @fmuyassarov

fmuyassarov (commented):
/lgtm
Thank you.

PR #1073, opened 2023-03-07
test/e2e: reduce worker wait-for-ready period to 2s

marquiz:
Reduce the wait time of nfd-worker pods to be in ready-state (before proceeding with tests) from five to two seconds. Make tests faster to run. Two seconds should be enough for nfd-workers to do their job and get nodes labeled.


fmuyassarov (approved):
/lgtm

PR #1074, opened 2023-03-07
test/e2e: cleanup NodeFeature objects before/after tests

marquiz:
Make sure that stale NodeFeature objects from previous test case are not interfering the next one.

### DIFF
--- a/test/e2e/node_feature_discovery_test.go
+++ b/test/e2e/node_feature_discovery_test.go
@@ -133,7 +133,7 @@ func cleanupNode(cs clientset.Interface) {
 
 }
 
-func cleanupCRDs(cli *nfdclient.Clientset) {
+func cleanupCRs(cli *nfdclient.Clientset, namespace string) {
 	// Drop NodeFeatureRule objects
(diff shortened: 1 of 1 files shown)


test/e2e/node_feature_discovery_test.go
```diff
@@ -133,7 +133,7 @@ func cleanupNode(cs clientset.Interface) {
 
 }
 
-func cleanupCRDs(cli *nfdclient.Clientset) {
+func cleanupCRDs(cli *nfdclient.Clientset, namespace string) {
```
PiotrProkop, line 136: sorry for nitpicking but i think this func name should be `CRs` instead of `CRDs` especially when you pass `namespace` arg 😄 
```suggestion
func cleanupCRs(cli *nfdclient.Clientset, namespace string) {
```

marquiz: Yeah, that's more correct naming. Fixed

PiotrProkop: :+1:

fmuyassarov (approved):
/lgtm

PR #1078, opened 2023-03-09
helm: fix topology-updater rbac

marquiz:
- fix topology-updater rbac clusterrole 
  Access to nodes/proxy resource was accidentally given to nfd-master
  (which really doesn't need it), not topology-updater.
- create topology-updater RBAC rules by default 
  Create RBAC rules if topology-updater is enabled. Previously installing
  with topologyUpdater.enable=true (without
  topologyUpdater.rbac.create=true) resulted in a crashloogbackoff as RBAC
  was missing.

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/clusterrole.yaml
+++ b/deployment/helm/node-feature-discovery/templates/clusterrole.yaml
@@ -19,12 +19,6 @@ rules:
   - update
   - list
 - apiGroups:
-    - ""
-  resources:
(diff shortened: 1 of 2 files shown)


marquiz:
ping @PiotrProkop @Tal-or @jlojosnegros @ffromani

PiotrProkop:
LGTM!

ArangoGutierrez (approved):
/lgtm

PR #1079, opened 2023-03-09
cpu: Expose the total number of keys for TDX

fidencio:
The total amount of keys that can be used on a specific TDX system is exposed via the cgroups misc.capacity. See:

```
$ cat /sys/fs/cgroup/misc.capacity
tdx 31
```

The first step to properly manage the amount of keys present in a node is exposing it via the NFD, and that's exactly what this commit does.

An example of how it ends up being exposed via the NFD:

```
$ kubectl get node 984fee00befb.jf.intel.com -o jsonpath='{.metadata.labels}'  | jq | grep tdx.total_keys
  "feature.node.kubernetes.io/cpu-security.tdx.total_keys": "31",
```

### DIFF
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -607,6 +607,7 @@ The following features are available for matching:
 |                  |              | **`sgx.enabled`** | bool | `true` if Intel SGX (Software Guard Extensions) has been enabled, otherwise does not exist
 |                  |              | **`se.enabled`** | bool  | `true` if IBM Secure Execution for Linux is available and has been enabled, otherwise does not exist
 |                  |              | **`tdx.enabled`** | bool | `true` if Intel TDX (Trusted Domain Extensions) is available on the host and has been enabled, otherwise does not exist
+|                  |              | **`tdx.total_keys`** | int | The total amount of keys an Intel TDX (Trusted Domain Extensions) host can provide.  It's only present if `tdx.enabled` is `true`.
 |                  |              | **`sev.enabled`** | bool | `true` if AMD SEV (Secure Encrypted Virtualization) is available on the host and has been enabled, otherwise does not exist
 |                  |              | **`sev.es.enabled`** | bool | `true` if AMD SEV-ES (Encrypted State supported) is available on the host and has been enabled, otherwise does not exist
 |                  |              | **`sev.snp.enabled`** | bool | `true` if AMD SEV-SNP (Secure Nested Paging supported) is available on the host and has been enabled, otherwise does not exist
--- a/docs/usage/features.md
+++ b/docs/usage/features.md
@@ -58,6 +58,7 @@ option of nfd-worker.
 | **`cpu-security.sgx.enabled`**    | true   | Set to 'true' if Intel SGX is enabled in BIOS (based a non-zero sum value of SGX EPC section sizes).
 | **`cpu-security.se.enabled`**     | true   | Set to 'true' if IBM Secure Execution for Linux (IBM Z & LinuxONE) is available and enabled (requires `/sys/firmware/uv/prot_virt_host` facility)
 | **`cpu-security.tdx.enabled`**    | true   | Set to 'true' if Intel TDX is available on the host and has been enabled (requires `/sys/module/kvm_intel/parameters/tdx`).
+| **`cpu-security.tdx.total_keys`** | int    | The total amount of keys an Intel TDX enabled host can provide, based on the `/sys/fs/cgroup/misc.capacity` information.
 | **`cpu-security.sev.enabled`**    | true   | Set to 'true' if ADM SEV is available on the host and has been enabled (requires `/sys/module/kvm_intel/parameters/sev`).
 | **`cpu-security.sev.es.enabled`** | true   | Set to 'true' if ADM SEV-ES is available on the host and has been enabled (requires `/sys/module/kvm_intel/parameters/sev_es`).
 | **`cpu-security.sev.snp.enabled`**| true   | Set to 'true' if ADM SEV-SNP is available on the host and has been enabled (requires `/sys/module/kvm_intel/parameters/sev_snp`).
--- a/source/cpu/security_amd64.go
+++ b/source/cpu/security_amd64.go
@@ -20,7 +20,11 @@ limitations under the License.
 package cpu
 
 import (
+	"bufio"
+	"io"
 	"os"
+	"strconv"
+	"strings"
 
 	"github.com/klauspost/cpuid/v2"
 
@@ -36,6 +40,11 @@ func discoverSecurity() map[string]string {
 
 	if tdxEnabled() {
 		elems["tdx.enabled"] = "true"
+
+		tdxTotalKeys := getCgroupMiscCapacity("tdx")
+		if tdxTotalKeys > -1 {
+			elems["tdx.total_keys"] = strconv.FormatInt(int64(tdxTotalKeys), 10)
+		}
 	}
 
 	if sevParameterEnabled("sev") {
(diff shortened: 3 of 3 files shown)


source/cpu/security_amd64.go
```diff
@@ -73,3 +82,44 @@ func tdxEnabled() bool {
...
+		line, _, err := r.ReadLine()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+
+			return totalKeys
+		}
+
+		if !strings.Contains(string(line), "tdx") {
```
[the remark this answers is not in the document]
marquiz: +1 would possibly make sense to make this a bit more generic

Also, using `strings.Contains` seems very easy to provide unexpected results (if new caps with string `tdx` are added) -> shouldn't we use  e.g.`strings.HasPrefix` instead?

```suggestion
		if !strings.HasPrefix(string(line), "tdx ") {
```

marquiz (commented):
Thanks @fidencio for the PR.

How are you planning to utilize this label in practice? So is the label actually useful for anything by itself or are you planning to turn that into an extended resource or smth? What I'm trying to understand if this is smth that we want to have that as a built-in feature label, enabled by default for all users.

Also, we need to document this new feature in `docs/usage/features.md` 
 and `docs/usage/customization-guide.md`

fidencio:
> How are you planning to utilize this label in practice? So is the label actually useful for anything by itself or are you planning to turn that into an extended resource or smth?

That's a very good question, and sorry for my lack of explanation on this.
I wat to turn that into an exteded resource, so I'm not totally sure if this needs to be exposed as it's right now or not.
What would you suggest, @marquiz?

marquiz:
> That's a very good question, and sorry for my lack of explanation on this.

No problem, that's what the reviews are for 

> I wat to turn that into an exteded resource, so I'm not totally sure if this needs to be exposed as it's right now or not.
> What would you suggest, @marquiz?

I think my kinda (so far hidden) idea for filling these kind of usage scenarios involving extended resources would be leverage NodeFeatureRule (and possibly NodeFeature) CRD. I.e. to have support for specifying ExtendedResources in the NodeFeatureRule object, basically following the lines of Labels, Taints (#540) and Annotations (#863).

But we don't have that capability, yet. Just realized that there's not even an issue tracking that, yet -> will create one.

How would NodeFeatureRule serve your use case? One plan could be to merge this PR without the labeling part and  try to squeeze in support for ExtendedResources (in NodeFeatureRule CRD) for the next v0.13.0 release.

source/cpu/security_amd64.go
```diff
@@ -73,3 +82,41 @@ func tdxEnabled() bool {
...
+	defer f.Close()
+
+	r := bufio.NewReader(f)
+	for {
+		line, _, err := r.ReadLine()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+
```
ArangoGutierrez, line 103: nit: extra line not needed

ArangoGutierrez (commented):
just one nit

zvonkok:
@ArangoGutierrez We need that for SNP as well, the total number of ASIDs

marquiz:
> @ArangoGutierrez We need that for SNP as well, the total number of ASIDs

You mean extended resources (#1081)?

zvonkok:
@marquiz Yes and No. (1) We need the extended resource via Rule (2) cpu: Expose the total number of keys for SNP

fidencio:
@marquiz, I've updated the PR and I'll test it together with @ArangoGutierrez's work and report back very soon.

docs/usage/customization-guide.md
```diff
@@ -607,6 +607,7 @@ The following features are available for matching:
 |                  |              | **`sgx.enabled`** | bool | `true` if Intel SGX (Software Guard Extensions) has been enabled, otherwise does not exist
 |                  |              | **`se.enabled`** | bool  | `true` if IBM Secure Execution for Linux is available and has been enabled, otherwise does not exist
 |                  |              | **`tdx.enabled`** | bool | `true` if Intel TDX (Trusted Domain Extensions) is available on the host and has been enabled, otherwise does not exist
+|                  |              | **`tdx.total_keys`** | int | The total amount of keys an Intel TDX (Trusted Domain Extensions) host can provide.  It's only present if `tdx.enabled` is `true`.
```
ArangoGutierrez, line 610:
```suggestion
|                  |              | **`tdx.total_keys`** | int | The total amount of keys an Intel TDX (Trusted Domain Extensions) host can provide.  It's only present if `tdx.enabled` is `true`.
```

ArangoGutierrez (changes requested):
rm trailing white space

docs/usage/customization-guide.md
fidencio: Fixed, thanks!

fidencio:
> There's still one bit missing, though, that's getting the amount of used keys. I'm working on that right now.

No, nevermind.

I've done some tests here, let me share the results:

I'm using the follow Kata Containers runtime class definition:
```yaml
---
kind: RuntimeClass
apiVersion: node.k8s.io/v1
metadata:
    name: kata
handler: kata
overhead:
    podFixed:
        memory: "160Mi"
        cpu: "250m"
        feature.node.kubernetes.io/cpu-security.tdx.total_keys: "1"
```

Doing a `kubectl node describe` I can see:
```
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource                                                Requests    Limits
  --------                                                --------    ------
  cpu                                                     950m (0%)   0 (0%)
  memory                                                  290Mi (0%)  340Mi (0%)
  ephemeral-storage                                       0 (0%)      0 (0%)
  hugepages-1Gi                                           0 (0%)      0 (0%)
  hugepages-2Mi                                           0 (0%)      0 (0%)
  feature.node.kubernetes.io/cpu-security.tdx.total_keys  0           0
```

Then when I start a pod, which has the follow definition:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-qemu-tdx
spec:
  runtimeClassName: kata
  containers:
  - name: nginx-qemu-tdx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
```

I can then see, as a result of `kubectl node describe`:
```yaml
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted.)
  Resource                                                Requests    Limits
  --------                                                --------    ------
  cpu                                                     1200m (0%)  0 (0%)
  memory                                                  450Mi (0%)  340Mi (0%)
  ephemeral-storage                                       0 (0%)      0 (0%)
  hugepages-1Gi                                           0 (0%)      0 (0%)
  hugepages-2Mi                                           0 (0%)      0 (0%)
  feature.node.kubernetes.io/cpu-security.tdx.total_keys  1           0
```

If I try to start a second pod, now requesting 30 keys (+1 coming from the podOverhead definition), using the definition shown below:
```
apiVersion: v1
kind: Pod
metadata:
  name: nginx-qemu-tdx-2
spec:
  runtimeClassName: kata
  containers:
  - name: nginx-qemu-tdx-2
    image: nginx:1.14.2
    resources:
      limits:
        feature.node.kubernetes.io/cpu-security.tdx.total_keys: 30
    ports:
    - containerPort: 80
```

Then I can see it doesn't start:
```
Events:
  Type     Reason            Age   From               Message
  ----     ------            ----  ----               -------
  Warning  FailedScheduling  50s   default-scheduler  0/1 nodes are available: 1 Insufficient feature.node.kubernetes.io/cpu-security.tdx.total_keys.
```

It means the book keeping is correctly done.

fidencio:
One thing that we may figure out **later** why it happens is that although the bookkeeping is correctly done those values (coming from `kubectl describe node`) are never updated:
```
Capacity:
  cpu:                                                     128
  ephemeral-storage:                                       383957772Ki
  feature.node.kubernetes.io/cpu-security.tdx.total_keys:  31
  hugepages-1Gi:                                           0
  hugepages-2Mi:                                           0
  memory:                                                  131192364Ki
  pods:                                                    110
Allocatable:
  cpu:                                                     127
  ephemeral-storage:                                       353855482090
  feature.node.kubernetes.io/cpu-security.tdx.total_keys:  31
  hugepages-1Gi:                                           0
  hugepages-2Mi:                                           0
  memory:                                                  130565676Ki
  pods:                                                    110
```

zvonkok:
/lgtm 

Last famous words @marquiz

docs/usage/features.md
```diff
@@ -58,6 +58,7 @@ option of nfd-worker.
 | **`cpu-security.sgx.enabled`**    | true   | Set to 'true' if Intel SGX is enabled in BIOS (based a non-zero sum value of SGX EPC section sizes).
 | **`cpu-security.se.enabled`**     | true   | Set to 'true' if IBM Secure Execution for Linux (IBM Z & LinuxONE) is available and enabled (requires `/sys/firmware/uv/prot_virt_host` facility)
 | **`cpu-security.tdx.enabled`**    | true   | Set to 'true' if Intel TDX is available on the host and has been enabled (requires `/sys/module/kvm_intel/parameters/tdx`).
+| **`cpu-security.tdx.total_keys`** | true   | The total amount of keys an Intel TDX enabled host can provide, based on the `/sys/fs/cgroup/misc.capacity` information.
```
ArangoGutierrez, line 61:
```suggestion
| **`cpu-security.tdx.total_keys`** | int   | The total amount of keys an Intel TDX enabled host can provide, based on the `/sys/fs/cgroup/misc.capacity` information.
```

ArangoGutierrez (changes requested):

docs/usage/features.md
fidencio: Done!

marquiz (approved):
Thanks @fidencio for the enhancement, looks good to me 👍

PR #1085, opened 2023-03-13
gitignore: ignore codecov coverage report

fmuyassarov:
We don't necessarily need to keep the codecov coverage report on the git. As such, adding it to the gitignore to avoid it from accidental commiting.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ bin/
 demo/helper-scripts/*.pdf
 demo/helper-scripts/*.log
 /kustomization.yaml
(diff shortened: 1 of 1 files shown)


marquiz (approved):
Thanks @fmuyassarov, good catch!
/retest

marquiz (approved):

fmuyassarov:
Prow is currently broken (not by us only) and it might take for it to get resolved.
Slack thread: https://kubernetes.slack.com/archives/C7J9RP96G/p1678696744216069

PR #1086, opened 2023-03-13
feat: support builtin kernel mods

### LINKED ISSUE #1025: `kernel.loadedmodule` doesn't work on all OS
Currently, `kernel.loadedmodule` feature only check for the content of `/proc/modules` for modules compiled as module and not built in tree module. Maybe there's a way to check for built-in kernel that is in `/lib/modules/<kernel-version>/modules.builtin`.
I think this can be done by either improve the function `getLoadedModules` or allow the user to choose what kind of test to get the information.

**Why is this needed**:

Currently [Talos Linux](https://www.talos.dev/) will just silently not creating `nodeFeatureRule` when I put `matchFeatures` that check for `kernel.loadedmodule` even if the module is loaded.

AhmedGrati:
Resolves #1025.
The solution parsed the `/lib/modules/<KERNEL_VERSION>/modules.builtin` and combined it with the `/proc/modules` file to get most kernel modules.


source/kernel/modules.go
```diff
@@ -25,20 +26,39 @@ import (
...
-		}
-		// append loaded module
-		loadedMods = append(loadedMods, strings.Fields(line)[0])
-	}
+  kVersion, err := getVersion()
+  var filesToLoad []string
+  if err != nil {
+    filesToLoad = []string{kmodProcfsPath}
+  } else {
+    kBuiltinModPath := "/lib/modules/" + kVersion + "/modules.builtin"
```
marquiz, line 34: We need to specify new `LibDir` in `pkg/utils/hostpath/hostpath.go` and use it here.

And, add corresponding mounts to all deployment files (kustomize and Helm)

source/kernel/modules.go
```diff
@@ -25,20 +26,39 @@ import (
...
+    filesToLoad = []string{kmodProcfsPath, kBuiltinModPath}
+  }
+  loadedMods, err := loadModules(filesToLoad)
+  if err != nil {
+    return nil, errors.New("error while loading modules")
+  }
 	return loadedMods, nil
 }
+
+func loadModules(kModFsPath []string) ([] string, error) {
```
marquiz, line 44: The format of `/proc/modules` and  `modules.builtin` is different so you need to have separate parser functions for them

marquiz (commented):
Thanks @AhmedGrati for working on this.

I'm not entirely sure we want to change the existing `kernel.loadedmodule` feature. If not, we want to document that it lists loaded dynamic kernel modules and omits builtin modules.

Instead this PR could add a new feature, say `kernel.enabledmodule` which would combine loaded dynamic modules and the builtins.

OTOH, if `kernel.loadedmodule` is seen totally useless/redundant maybe we just change that (even if the naming is a bit off).

Any comments @zvonkok @mythi @ArangoGutierrez?

ArangoGutierrez:
PING @AhmedGrati

AhmedGrati:
@ArangoGutierrez Just discussed with @marquiz today about what should be done, since we were waiting for inputs from others. I'll start working on it.

ArangoGutierrez:
Still a WIP
`make test e2e-test` passes

source/kernel/modules.go
```diff
@@ -42,3 +44,35 @@ func getLoadedModules() ([]string, error) {
...
+		// append loaded module
+		loadedMods = append(loadedMods, getBuiltinModName(line))
+	}
+	return loadedMods, nil
+}
+
+func getBuiltinModName(modPath string) string {
+	splittedModPath := strings.Split(modPath, "/")
+	splittedModFileName := strings.Split(splittedModPath[len(splittedModPath)-1], ".")
+	return strings.Join(splittedModFileName[:len(splittedModFileName)-1], ".")
```
marquiz, line 77: You could just use strings.TrimSuffix() to drop the `.ko` suffix

source/kernel/kernel.go
```diff
@@ -141,6 +142,12 @@ func (s *kernelSource) Discover() error {
 		s.features.Flags[LoadedModuleFeature] = nfdv1alpha1.NewFlagFeatures(kmods...)
 	}
 
+	if builtinMods, err := getBuiltinModules(); err != nil {
+		klog.Errorf("failed to get builtin kernel modules: %v", err)
+	} else {
+		s.features.Flags[EnabledModuleFeature] = nfdv1alpha1.NewFlagFeatures(builtinMods...)
```
marquiz, line 148: I think this feature should combine loadedmodules and builtinmodules, as discussed in #1025

source/kernel/modules.go
```diff
@@ -42,3 +44,35 @@ func getLoadedModules() ([]string, error) {
...
+	}
+
+	kBuiltinModPath := hostpath.LibDir.Path("modules/" + kVersion + "/modules.builtin")
+	out, err := os.ReadFile(kBuiltinModPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read file %s: %s", kBuiltinModPath, err.Error())
+	}
+
+	lines := strings.Split(string(out), "\n")
+	loadedMods := make([]string, len(lines))
```
marquiz, line 61: nit: they're not "loaded" (like dynamically) so I'd rename this just `mods` or then `builtinMods`

marquiz (commented):
Thanks @AhmedGrati, looks really good. I wanna test this myself 😅

We also need to document the new feature in the table of available features in `docs/usage/customization-guide.md`

source/kernel/modules.go
```diff
@@ -42,3 +45,29 @@ func getLoadedModules() ([]string, error) {
...
+		return nil, fmt.Errorf("failed to read file %s: %s", kBuiltinModPath, err.Error())
+	}
+
+	lines := strings.Split(string(out), "\n")
+	builtinMods := make([]string, len(lines))
+	for _, line := range lines {
+		// skip empty lines
+		if len(line) == 0 {
+			continue
+		}
```
marquiz, line 67: One suggestion: instead of checking for empty lines it would prolly be more robust to first trim all white space and then verify that the string ends with .ko (all the modules do AFAIU)

```suggestion
		line = strings.TrimSpace(line)
		if !strings.HasSuffix(line, ".ko") {
			continue
		}
```

marquiz (commented):
Thx! Just one comment but basically looks ready for me (after that) 👍

Would be nice to get some other eyes on this, too.
ping @budimanjojo @mythi @zvonkok, any comments?

deployment/helm/node-feature-discovery/templates/worker.yaml
```diff
@@ -99,6 +99,9 @@ spec:
         - name: host-usr-lib
           hostPath:
             path: "/usr/lib"
+        - name: host-lib
```
marquiz, line 102: Hey, you need to add  `host-lib` to the mounts of the container above, too

deployment/components/common/worker-mounts.yaml
```diff
@@ -13,6 +13,9 @@
   - name: host-usr-lib
     hostPath:
       path: "/usr/lib"
+  - name: host-lib
```
marquiz, line 16: Hey, you need to add  `host-lib` to the mounts of the container above, too

budimanjojo:
Looks good to me. This means if I use `kernel.enabledModules` it will detect both the old and new implementations while `kernel.loadedModules` will still behave like the old implementation?

deployment/components/common/worker-mounts-combined.yaml
```diff
@@ -38,6 +41,9 @@
   - name: host-usr-lib
     mountPath: "/host-usr/lib"
     readOnly: true
+  - name: host-lib
+    hostPath:
+      path: "/lib"
```
marquiz, line 46: The field in volumeMounts is `mountPath`. Also, you need to mount it with the `host-` prefix


```suggestion
    mountPath: "/host-lib"
    readOnly: true
```

...and ditto volumemounts you added into other files

marquiz:
> This means if I use `kernel.enabledModules` it will detect both the old and new implementations while `kernel.loadedModules` will still behave like the old implementation?

Yup, that's the way it works

deployment/helm/node-feature-discovery/templates/worker.yaml
```diff
@@ -67,6 +67,9 @@ spec:
         - name: host-usr-lib
           mountPath: "/host-usr/lib"
           readOnly: true
+        - name: host-lib
+          mountPath: "/lib"
```
marquiz, line 71: here, too

source/kernel/modules.go
```diff
@@ -42,3 +45,30 @@ func getLoadedModules() ([]string, error) {
...
+	}
+
+	kBuiltinModPath := hostpath.LibDir.Path("modules/" + kVersion + "/modules.builtin")
+	out, err := os.ReadFile(kBuiltinModPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read file %s: %s", kBuiltinModPath, err.Error())
+	}
+
+	lines := strings.Split(string(out), "\n")
+	builtinMods := make([]string, len(lines))
```
marquiz, line 62: You need to set the lenght of the array to zero, otherwise it will have no use here (it will have `len(lines)` number of empty strings on which you will append the new values, making the final slice 2 x `len(lines)` long)

```suggestion
	builtinMods := make([]string, 0, len(lines))
```

marquiz (commented):
One more issue I spotted.

Also, I'm again nitting about the commit message 😇 Could you add some flesh there, describing what is being added (and possibly also why)

docs/usage/customization-guide.md
```diff
@@ -620,6 +620,7 @@ The following features are available for matching:
 | **`kernel.config`** | attribute |          |            | Kernel configuration options
 |                  |              | **`<config-flag>`** | string | Value of the kconfig option
 | **`kernel.loadedmodule`** | flag |         |            | Loaded kernel modules
+| **`kernel.enabledmodule`** | flag |        |            | Builtin and dynamic kernel modules
```
mythi, line 623: this documentation does not explain it very well that `enabledmodule` is a superset of `loadedmodule` and what "buildin modules" are

marquiz: Yeah, now that you pointed this out, I agree. Any suggestions how to improve this @mythi? _All enabled kernel modules, i.e. union of loaded dynamic modules and modules built into the kernel_

marquiz: We could also clarify `kernel.loadedmodule` above i.e. with _Loaded dynamic kernel modules_(?)

source/kernel/modules.go
AhmedGrati: @marquiz should we change it also for the `loadedMods := make([]string, len(lines))` (It's in the function above)?

AhmedGrati: No worries, I'll include it here.

marquiz:
ping @mythi

docs/usage/customization-guide.md
```diff
@@ -619,7 +619,8 @@ The following features are available for matching:
 | | |          **`nx_gzip`**                 | bool       | Nest Accelerator GZIP support is enabled
 | **`kernel.config`** | attribute |          |            | Kernel configuration options
 |                  |              | **`<config-flag>`** | string | Value of the kconfig option
-| **`kernel.loadedmodule`** | flag |         |            | Loaded kernel modules
+| **`kernel.loadedmodule`** | flag |         |            | Loaded dynamic kernel modules
+| **`kernel.enabledmodule`** | flag |        |            | All enabled kernel modules, i.e. union of loaded dynamic modules and modules built into the kernel
```
mythi, line 623:
```suggestion
| **`kernel.loadedmodule`** | flag |         |            | Kernel modules loaded on the node as reported by `/proc/modules`
| **`kernel.enabledmodule`** | flag |        |            | Kernel modules loaded on the node and available as built-ins as reported by `modules.builtin`. Does not include any out-of-tree modules that are not loaded.
```

mythi:
> @mythi WDYT about the new wording(s) in the docs?

I added a suggestion but I'm also OK with how it is currently.

docs/usage/customization-guide.md
marquiz: Out-of-tree modules are no special in this regard. Also, we report **loaded** and **enabled-by-builtin** modules so it should be kinda evident that we don't list any **disabled** modules so I'd suggest just to drop this _Does not include any out-of-tree modules that are not loaded._ Or then reword it to _Does not include any modules that are not loaded or built into the kernel._ which sounds a bit dumb after the first sentence

mythi: > I'd suggest just to drop this

works for me

marquiz: K, so @mythi's suggestion would become

```suggestion
| **`kernel.loadedmodule`** | flag |         |            | Kernel modules loaded on the node as reported by `/proc/modules`
| **`kernel.enabledmodule`** | flag |        |            | Kernel modules loaded on the node and available as built-ins as reported by `modules.builtin`
```

I think this is good.

@AhmedGrati could I ask for this one more change? 😇

AhmedGrati: Sure, will do it 👍

marquiz (approved):
Thanks @AhmedGrati!

mythi (approved):

marquiz:
CI is failing, both pre-submit and e2e 😭 Let's fix those first
/hold

marquiz:
e2e is green again
/unhold

PR #1087, opened 2023-03-14
pkg/utils: add UnmarshalJSON method to StringSetVal

marquiz:
Make it possible to specify values in yaml as an array like

```yaml
  conf:
    - foo
    - bar
````

Instead of unwieldy map like

```yaml
  conf:
    foo:
    bar:
```


AhmedGrati (approved):
That would be better thanks!

ArangoGutierrez (approved):
/lgtm

PR #1088, opened 2023-03-15
e2e: re-write pre-submit test to use minikube as an ephemeral cluster

fmuyassarov:
Using minikube with docker being the `vm-driver` simplifies our pre-submit
e2e tests with image registry handling. One problem with using kind and
local registry container is that we need to have mounted the same port
(5000) to all the kind clusters that might be created at the same time. And
that's technically impossible, because we can't mount the same port to all
the clusters. Minikube on the other hand, allows pushing directly to the
in-cluster Docker daemon. It comes with its registry and we don't have to
take care of it ourselves. Every minikube cluster will have it is own registry
that will be created & destroyed as the cluster itself. [Ref](https://minikube.sigs.k8s.io/docs/handbook/pushing/#1-pushing-directly-to-the-in-cluster-docker-daemon-docker-env)

Also add a self hosted runner based GitHub Action to run E2E tests on.

### DIFF
--- a/scripts/test-infra/test-e2e-presubmit.sh
+++ b/scripts/test-infra/test-e2e-presubmit.sh
@@ -1,67 +1,25 @@
 #!/bin/bash -e
+
 set -o pipefail
 
-# Configure environment
-KIND_IMAGE="kindest/node:v1.25.3"
-export IMAGE_REGISTRY="localhost:5001"
 export CLUSTER_NAME=$(git describe --tags --dirty --always)
 export KUBECONFIG="/tmp/kubeconfig_$CLUSTER_NAME"
 
-# create registry container unless it already exists
-reg_name='kind-registry'
-reg_port='5001'
-if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
-  docker run \
-    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --name "${reg_name}" \
-    registry:2
-fi
+minikube start --bootstrapper=kubeadm \
+  --vm-driver=docker \
+  --memory 2048 \
+  --cpus 2 \
+  --profile $CLUSTER_NAME
 
-# create a cluster with the local registry enabled in containerd
-cat <<EOF | kind create cluster --kubeconfig $KUBECONFIG --config=-
-kind: Cluster
-apiVersion: kind.x-k8s.io/v1alpha4
-name: $CLUSTER_NAME
-nodes:
-- role: control-plane
-  image: $KIND_IMAGE
-- role: worker
-  image: $KIND_IMAGE
-- role: worker
-  image: $KIND_IMAGE
-containerdConfigPatches:
-- |-
-  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
-    endpoint = ["http://${reg_name}:5000"]
-EOF
+eval $(minikube --profile $CLUSTER_NAME docker-env)
 
-# connect the registry to the cluster network if not already connected
-if [ "$(docker inspect -f='{{json .NetworkSettings.Networks.kind}}' "${reg_name}")" = 'null' ]; then
-  docker network connect "kind" "${reg_name}"
-fi
-
-cat <<EOF | kubectl apply -f -
-apiVersion: v1
-kind: ConfigMap
-metadata:
-  name: local-registry-hosting
-  namespace: kube-public
-data:
-  localRegistryHosting.v1: |
-    host: "localhost:${reg_port}"
-    help: "https://kind.sigs.k8s.io/docs/user/local-registry/"
-EOF
-
-# build the local image
+# Build the local container image
 make image
 
-# push the image to the local registry
-make push
-
-# run the tests
+# Use IfNotPresent image pull policy to use locally build image
+export E2E_PULL_IF_NOT_PRESENT=true
 make e2e-test || rc=$?
 
-# clean up the environment
-kind delete cluster --kubeconfig $KUBECONFIG --name $CLUSTER_NAME
-echo "Deleting ${reg_name} container ..."
-docker rm -f "${reg_name}"
+minikube delete --profile $CLUSTER_NAME
+
 exit $rc
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -1,0 +1,24 @@
+name: End to end verification
+on:
+  pull_request:
+    types: [opened, edited, synchronize, reopened]
+    paths-ignore:
+      - 'docs/**'
+      - 'demo/**'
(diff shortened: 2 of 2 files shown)


scripts/test-infra/test-e2e-presubmit.sh
```diff
@@ -1,67 +1,25 @@
...
-reg_port='5001'
-if [ "$(docker inspect -f '{{.State.Running}}' "${reg_name}" 2>/dev/null || true)" != 'true' ]; then
-  docker run \
-    -d --restart=always -p "127.0.0.1:${reg_port}:5000" --name "${reg_name}" \
-    registry:2
-fi
+minikube start --bootstrapper=kubeadm \
+  --vm-driver=docker \
+  --memory 8192 \
+  --cpus 4 \
```
ArangoGutierrez, line 10: maybe we want to check `nproc`

ArangoGutierrez (commented):
Interesting concept 
Although we don't have an e2e for pre, only post, see https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/node-feature-discovery

fmuyassarov:
> > Interesting concept Although we don't have an e2e for pre, only post, see https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/node-feature-discovery
> 
> true, and we aim to add one in [kubernetes/test-infra#27755](https://github.com/kubernetes/test-infra/pull/27755)

@ArangoGutierrez more context on this: https://github.com/kubernetes-sigs/node-feature-discovery/issues/902

scripts/test-infra/test-e2e-presubmit.sh
fmuyassarov: I will check it. Never used this utility before.
But actually, these resources is something I wanted to ask. IINW, minikube defaults to 2 CPU and 2GB memory, but I used 4 CPUs here and 8 GB memory for each cluster, which might cause resource saturation on the host when there are many of those clusters in parallel. So, I should admit that I have not verified if we really need 8 GB, I can perhaps do tests with 4 GB as well.

ArangoGutierrez:
> > > Interesting concept Although we don't have an e2e for pre, only post, see https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/node-feature-discovery
> > 
> > 
> > true, and we aim to add one in [kubernetes/test-infra#27755](https://github.com/kubernetes/test-infra/pull/27755)
> 
> @ArangoGutierrez more context on this: #902

so we need this to  unblock the test-infra PR right?

fmuyassarov:
> so we need this to unblock the test-infra PR right?

Right. What I expect is
1. Merge this PR so that we have a script to spin up clusters
2. Update the infra PR with server (which we got from CNCF) characteristics (labels) and then merge it. I will ping you on the PR once it is ready.

scripts/test-infra/test-e2e-presubmit.sh
ArangoGutierrez: agree, for an e2e pre test, 4 cpu's is a bit too much

marquiz: Check this and I think we could merge this one

ArangoGutierrez: Agree, let's go with default resources, the rest of the PR looks ok

fmuyassarov: based on the tests results 2GB, 2CPU is enough. But there are some other bits I would like to add on this PR. As such, will hold it and ping you folks once ready again.
/hold

scripts/test-infra/test-e2e-presubmit.sh
```diff
@@ -1,67 +1,25 @@
...
-- role: worker
-  image: $KIND_IMAGE
-- role: worker
-  image: $KIND_IMAGE
-containerdConfigPatches:
-- |-
-  [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localhost:${reg_port}"]
-    endpoint = ["http://${reg_name}:5000"]
-EOF
+# kubectl config use-context $CLUSTER_NAME
```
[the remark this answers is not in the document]
fmuyassarov: yep

fmuyassarov:
Since there was a lot of going back and force and you might have lost where we are:

1. We have tried to add another Prow job to run E2E test as a presubmit on our managed servers. Running on externally (to Prow) managed servers is highly demotivated and there were issues with using ssh&scp from within the ProwJob created container to access our managed servers where we wanted our tests to be executed. As such, we decided not to rely on Prow (which is quite unfortunate, it would have been nice to have all the tests to be managed by a single tool) and use GitHub action.
2. Instead of ProwJob we are now adding GitHub Action. GitHub Actions allow adding your servers to host running jobs. Thus, this PR is adding a new job to run e2e test on our server. Before we take this patch in, we need to register our server to the GitHub as a self hosted runner. For that, we need someone with admin rights on the NFD repo.

I've tried to limit running e2e and avoid it for unnecessary changes (markdown only, OWNERS, etc). Currently there is only one self hosted runner (one server) but we have one more in our CNCF budget. Perhaps I will add another one once confirmed that we are happy with GitHub Action way of running E2E.

fmuyassarov:
> Since there was a lot of going back and force and you might have lost where we are:
> 
> 1. We have tried to add another Prow job to run E2E test as a presubmit on our managed servers. Running on externally (to Prow) managed servers is highly demotivated and there were issues with using ssh&scp from within the ProwJob created container to access our managed servers where we wanted our tests to be executed. As such, we decided not to rely on Prow (which is quite unfortunate, it would have been nice to have all the tests to be managed by a single tool) and use GitHub action.

https://github.com/kubernetes/test-infra/pull/29456 is for ^

fmuyassarov:
/cc @marquiz @ArangoGutierrez 
This is ready. PTAL.

fmuyassarov:
ping @marquiz @ArangoGutierrez do you guys think we can move forward or shall we move e2e presubmit to backlog ?

ArangoGutierrez:
> ping @marquiz @ArangoGutierrez do you guys think we can move forward or shall we move e2e presubmit to backlog ?

Is now PROW based but it works, I am up for it, @marquiz ?

fmuyassarov:
> > ping @marquiz @ArangoGutierrez do you guys think we can move forward or shall we move e2e presubmit to backlog ?
> 
> Is now PROW based but it works, I am up for it, @marquiz ?

No, it is now GitHub action based.

ArangoGutierrez:
> > > ping @marquiz @ArangoGutierrez do you guys think we can move forward or shall we move e2e presubmit to backlog ?
> > 
> > 
> > Is now PROW based but it works, I am up for it, @marquiz ?
> 
> No, it is now GitHub action based.

My bad I meant "is not" ;P ... getting back from a 10 days OOO has my head in wrong places

PR #1090, opened 2023-03-15
kustomize: trim prune overlay

ArangoGutierrez:


ArangoGutierrez:
Inspired on https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/188#discussion_r1137226151

deployment/overlays/prune/master-clusterrole.yaml
```diff
@@ -1,3 +1,5 @@
-- op: add
-  path: "/rules/0/verbs/-"
-  value: "list"
+- op: remove
+  path: "/rules/0/verbs/3"
+
```
marquiz, line 3: We need to be able to list (but we don't need to have it in the rule two times

```suggestion
```

Simplest thing would be to just drop this patch (prune/master-clusterrole.yaml) altogether. It would give access to NFD CRDs but that doesn't really do any harm. But your call

marquiz (commented):
Thanks @ArangoGutierrez for the fix!

One nit about the commit message the topic "Prune" is just a bit odd to me as this is about deployment, I'd suggest something like

```
deployment: trim kustomize prune overlay
```

or 

```
kustomize: trim prune overlay
```

marquiz (approved):
Looks good to me now

PR #1084, opened 2023-03-12
feat: add master config file

AhmedGrati:
Resolves #485.
Tasks done:
- [x] Support config file for the master, with `fsnotify`, `config parsing`, and `overrides from command line flags`
- [x] Unit Tests
- [x] Deployment (Kustomize, Helm, Update `make templates`) with the default configuration
- [x] Documentation (command line reference, config reference, nfd-master usage page)
- [x] E2E Tests

PR #1092, opened 2023-03-17
chore: add debug dump of nfd worker configuration

AhmedGrati:
As discussed in #1084, we need to add the dump of the nfd worker configuration, which can be really useful when debugging.


marquiz (approved):
Thanks @AhmedGrati, to me this looks like a good addition 👍
/ok-to-test

fmuyassarov (commented):
/lgtm

PR #1099, opened 2023-03-28
Create extended resources with NodeFeatureRule

### LINKED ISSUE #1081: Create extended resources with NodeFeatureRule

In addition to labels, taints (#540) and annotations (#863), support also management of extended resources with the NodeFeatureRule CRD API.

**Why is this needed**:

There are usage scenarios where people want to advertise features as extended resources instead of labels (or annotations), e.g. #1079. There has also been use cases like accounting number of network bridge interfaces (or ports, I don't remember) and GPU memory for example. We have long had experimental support for extended resources via the `-resource-labels` command line flag in nfd-master but that is very unwieldy for users (but obviously has worked as we haven't basically got any bug reports or complaints about that).

ArangoGutierrez:
Fixes #1081 

Add support for management of extended resources with the
`NodeFeatureRule` CRD API.

There are usage scenarios where users want to advertise features
as extended resources instead of labels (or annotations).

This patch enables the discovery of extended resources, via annotation
and patch of `node.status.capacity` and `node.status.allocatable` By using
the `NodeFeatureRule` API.

Co-authored-by: Carlos Eduardo Arango Gutierrez <eduardoa@nvidia.com>
Co-authored-by: Fabiano Fidêncio <fabiano.fidencio@intel.com>

### DIFF
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -485,7 +485,7 @@ details.
 labels specified in the `labels` field will override anything
 originating from `labelsTemplate`.
 
-### Taints
+#### Taints
 
 *taints* is a list of taint entries and each entry can have `key`, `value` and `effect`,
 where the `value` is optional. Effect could be `NoSchedule`, `PreferNoSchedule`
@@ -500,6 +500,65 @@ The `.vars` field is a map of values (key-value pairs) to store for subsequent
 rules to use. In other words, these are variables that are not advertised as
 node labels. See [backreferences](#backreferences) for more details on the
 usage of vars.
+
+#### Extended resources
+
+The `.extendedResources` field is a list of extended resources to advertise.
+See [extended resources](#extended-resources) for more details.
+
+Take this rule as a referential example:
+
+```yaml
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: my-extended-resource-rule
+spec:
+  rules:
+    - name: "my extended resource rule"
+      extendedResources:
+        vendor.io/dynamic: "@kernel.version.major"
+        vendor.io/static: "123"
+      matchFeatures:
+        - feature: kernel.version
+          matchExpressions:
+            major: {op: Exists}
+```
+
+The extended resource `vendor.io/dynamic` is defined in the form `@feature.attribute`.
+The value of the extended resource will be the value of the attribute `major`
+of the feature `kernel.version`.
+
+The `@<feature-name>.<element-name>` format can be used to inject values of
+detected features to the extended resource. See
+[available features](#available-features) for possible values to use. Note that
+the value must be eligible as a
+Kubernetes resource quantity.
+
+This will yield into the following node status:
+
+```yaml
+  allocatable:
+    ...
+    vendor.io/dynamic: "5"
+    vendor.io/static: "123"
+    ...
+  capacity:
+    ...
+    vendor.io/dynamic: "5"
+    vendor.io/static: "123"
+    ...
+```
+
+There are some limitations to the namespace part (i.e. prefix)/ of the Extended
+Resources names:
+
+- `kubernetes.io/` and its sub-namespaces (like `sub.ns.kubernetes.io/`) cannot
+  generally be used
+- the only exception is `feature.node.kubernetes.io/` and its sub-namespaces
+  (like `sub.ns.feature.node.kubernetes.io`)
+- unprefixed names will get prefixed with `feature.node.kubernetes.io/`
+  automatically (e.g. `foo` becomes `feature.node.kubernetes.io/foo`)
 
 #### Vars template
 
--- a/examples/nodefeaturerule.yaml
+++ b/examples/nodefeaturerule.yaml
@@ -14,3 +14,18 @@ spec:
         - feature: kernel.config
           matchExpressions:
             X86: {op: In, value: ["y"]}
+---
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: my-sample-extened-resource
+spec:
+  rules:
+    - name: "my sample rule"
+      extendedResources:
+        vendor.io/dynamic: "@kernel.version.major"
+        vendor.io/static: "123"
+      matchFeatures:
+        - feature: kernel.version
+          matchExpressions:
+            major: {op: Exists}
--- a/pkg/apis/nfd/v1alpha1/annotations_labels.go
+++ b/pkg/apis/nfd/v1alpha1/annotations_labels.go
@@ -38,6 +38,12 @@ const (
 	// AnnotationNs namespace for all NFD-related annotations.
 	AnnotationNs = "nfd.node.kubernetes.io"
 
+	// ExtendedResourceNs is the namespace for extended resources.
+	ExtendedResourceNs = "feature.node.kubernetes.io"
+
+	// ExtendedResourceSubNsSuffix is the suffix for allowed extended resources sub-namespaces.
+	ExtendedResourceSubNsSuffix = "." + ExtendedResourceNs
+
 	// ExtendedResourceAnnotation is the annotation that holds all extended resources managed by NFD.
 	ExtendedResourceAnnotation = AnnotationNs + "/extended-resources"
 
--- a/pkg/apis/nfd/v1alpha1/rule.go
+++ b/pkg/apis/nfd/v1alpha1/rule.go
@@ -30,13 +30,15 @@ import (
 // RuleOutput contains the output out rule execution.
 // +k8s:deepcopy-gen=false
 type RuleOutput struct {
-	Labels map[string]string
-	Vars   map[string]string
-	Taints []corev1.Taint
+	ExtendedResources map[string]string
+	Labels            map[string]string
+	Vars              map[string]string
+	Taints            []corev1.Taint
 }
 
 // Execute the rule against a set of input features.
 func (r *Rule) Execute(features *Features) (RuleOutput, error) {
+	extendedResources := make(map[string]string)
 	labels := make(map[string]string)
 	vars := make(map[string]string)
 
@@ -88,6 +90,10 @@ func (r *Rule) Execute(features *Features) (RuleOutput, error) {
 		}
 	}
 
+	for k, v := range r.ExtendedResources {
+		extendedResources[k] = v
+	}
+
 	for k, v := range r.Labels {
 		labels[k] = v
 	}
@@ -95,7 +101,7 @@ func (r *Rule) Execute(features *Features) (RuleOutput, error) {
 		vars[k] = v
 	}
 
-	ret := RuleOutput{Labels: labels, Vars: vars, Taints: r.Taints}
+	ret := RuleOutput{ExtendedResources: extendedResources, Labels: labels, Vars: vars, Taints: r.Taints}
 	utils.KlogDump(2, fmt.Sprintf("rule %q matched with: ", r.Name), "  ", ret)
 	return ret, nil
 }
--- a/pkg/apis/nfd/v1alpha1/types.go
+++ b/pkg/apis/nfd/v1alpha1/types.go
@@ -163,6 +163,10 @@ type Rule struct {
 	// +optional
 	Taints []corev1.Taint `json:"taints,omitempty"`
 
+	// ExtendedResources to create if the rule matches.
+	// +optional
+	ExtendedResources map[string]string `json:"extendedResources"`
+
 	// MatchFeatures specifies a set of matcher terms all of which must match.
 	// +optional
 	MatchFeatures FeatureMatcher `json:"matchFeatures"`
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -37,8 +37,8 @@ import (
 	"google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/peer"
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/labels"
-	label "k8s.io/apimachinery/pkg/labels"
+	k8sQuantity "k8s.io/apimachinery/pkg/api/resource"
+	k8sLabels "k8s.io/apimachinery/pkg/labels"
 	"k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/klog/v2"
@@ -447,7 +447,6 @@ func (m *nfdMaster) updateMasterNode() error {
 // arriving through the gRPC API.
 func (m *nfdMaster) filterFeatureLabels(labels Labels) (Labels, ExtendedResources) {
 	outLabels := Labels{}
-
 	for label, value := range labels {
 		// Add possibly missing default ns
 		label := addNs(label, nfdv1alpha1.FeatureLabelNs)
@@ -541,6 +540,20 @@ func isNamespaceDenied(labelNs string, wildcardDeniedNs map[string]struct{}, nor
 	return false
 }
 
+func isNamespaceAllowed(labelNs string, wildcardAllowedNs map[string]struct{}, normalAllowedNs map[string]struct{}) bool {
+	for allowedNs := range normalAllowedNs {
+		if labelNs == allowedNs {
+			return true
+		}
+	}
+	for allowedNs := range wildcardAllowedNs {
+		if strings.HasSuffix(labelNs, allowedNs) {
+			return true
+		}
+	}
+	return false
+}
+
 // SetLabels implements LabelerServer
 func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.SetLabelsReply, error) {
 	err := authorizeClient(c, m.args.VerifyNodeName, r.NodeName)
@@ -596,7 +609,7 @@ func (m *nfdMaster) nfdAPIUpdateAllNodes() error {
 }
 
 func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
-	sel := labels.SelectorFromSet(labels.Set{nfdv1alpha1.NodeFeatureObjNodeNameLabel: nodeName})
+	sel := k8sLabels.SelectorFromSet(k8sLabels.Set{nfdv1alpha1.NodeFeatureObjNodeNameLabel: nodeName})
 	objs, err := m.nfdController.featureLister.List(sel)
 	if err != nil {
 		return fmt.Errorf("failed to get NodeFeature resources for node %q: %w", nodeName, err)
@@ -662,19 +675,86 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
 	return nil
 }
 
-func (m *nfdMaster) refreshNodeFeatures(cli *kubernetes.Clientset, nodeName string, annotations, labels map[string]string, features *nfdv1alpha1.Features) error {
+// filterExtendedResources filters extended resources and returns a map
+// of valid extended resources.
+func (m *nfdMaster) filterExtendedResources(features *nfdv1alpha1.Features, extendedResources ExtendedResources) ExtendedResources {
+	outExtendedResources := ExtendedResources{}
+	deniedNs := map[string]struct{}{"kubernetes.io": {}}
+	deniedWildCarNs := map[string]struct{}{".kubernetes.io": {}}
+	allowedNs := map[string]struct{}{nfdv1alpha1.ExtendedResourceNs: {}}
+	allowedWildCardNs := map[string]struct{}{nfdv1alpha1.ExtendedResourceSubNsSuffix: {}}
+	for extendedResource, capacity := range extendedResources {
+		if strings.Contains(extendedResource, "/") {
+			// Check if given NS is allowed
+			ns, _ := splitNs(extendedResource)
+			if isNamespaceDenied(ns, deniedWildCarNs, deniedNs) {
+				if !isNamespaceAllowed(ns, allowedWildCardNs, allowedNs) {
+					klog.Errorf("namespace %q is not allowed. Ignoring Extended Resource  %q", ns, extendedResource)
+					continue
+				}
+			}
+		} else {
+			// Add possibly missing default ns
+			extendedResource = path.Join(nfdv1alpha1.ExtendedResourceNs, extendedResource)
+		}
+
+		// Dynamic Value
+		if strings.HasPrefix(capacity, "@") {
+			// capacity is a string in the form of attribute.featureset.elements
+			split := strings.SplitN(capacity[1:], ".", 3)
+			featureName := split[0] + "." + split[1]
+			elementName := split[2]
+			attrFeatureSet, ok := features.Attributes[featureName]
+			if !ok {
+				klog.Errorf("feature %s not found. Ignoring Extended Resource %q", featureName, extendedResource)
+				continue
+			}
+			element, ok := attrFeatureSet.Elements[elementName]
+			if !ok {
+				klog.Errorf("element %s not foundon feature %s. Ignoring Extended Resource %q", elementName, featureName, extendedResource)
+				continue
+			}
+			q, err := k8sQuantity.ParseQuantity(element)
+			if err != nil {
+				klog.Errorf("bad label value %s encountered for extended resource: %s", q.String(), extendedResource, err)
+				continue
+			}
+			outExtendedResources[extendedResource] = q.String()
+			continue
+		}
+		// Static Value (Pre-Defined at the NodeFeatureRule)
+		q, err := k8sQuantity.ParseQuantity(capacity)
+		if err != nil {
+			klog.Errorf("bad label value %s encountered for extended resource: %s", capacity, extendedResource, err)
+			continue
+		}
+		outExtendedResources[extendedResource] = q.String()
+	}
+	return outExtendedResources
+}
+
+func (m *nfdMaster) refreshNodeFeatures(cli *kubernetes.Clientset, nodeName string, annotations Annotations, labels map[string]string, features *nfdv1alpha1.Features) error {
+
 	if labels == nil {
 		labels = make(map[string]string)
 	}
 
-	crLabels, crTaints := m.processNodeFeatureRule(features)
+	crLabels, crExtendedResources, crTaints := m.processNodeFeatureRule(features)
 
 	// Mix in CR-originated labels
 	for k, v := range crLabels {
 		labels[k] = v
 	}
 
+	// Remove labels which are intended to be extended resources via
+	// -resource-labels or their NS is not whitelisted
 	labels, extendedResources := m.filterFeatureLabels(labels)
+
+	// Mix in CR-originated extended resources with -resource-labels
+	for k, v := range crExtendedResources {
+		extendedResources[k] = v
+	}
+	extendedResources = m.filterExtendedResources(features, extendedResources)
 
 	var taints []corev1.Taint
 	if m.config.EnableTaints {
@@ -795,21 +875,22 @@ func authorizeClient(c context.Context, checkNodeName bool, nodeName string) err
 	return nil
 }
 
-func (m *nfdMaster) processNodeFeatureRule(features *nfdv1alpha1.Features) (map[string]string, []corev1.Taint) {
+func (m *nfdMaster) processNodeFeatureRule(features *nfdv1alpha1.Features) (Labels, ExtendedResources, []corev1.Taint) {
 	if m.nfdController == nil {
-		return nil, nil
+		return nil, nil, nil
 	}
 
+	extendedResources := ExtendedResources{}
 	labels := make(map[string]string)
 	var taints []corev1.Taint
-	ruleSpecs, err := m.nfdController.ruleLister.List(label.Everything())
+	ruleSpecs, err := m.nfdController.ruleLister.List(k8sLabels.Everything())
 	sort.Slice(ruleSpecs, func(i, j int) bool {
 		return ruleSpecs[i].Name < ruleSpecs[j].Name
 	})
 
 	if err != nil {
 		klog.Errorf("failed to list NodeFeatureRule resources: %v", err)
-		return nil, nil
+		return nil, nil, nil
 	}
 
 	// Process all rule CRs
@@ -831,6 +912,9 @@ func (m *nfdMaster) processNodeFeatureRule(features *nfdv1alpha1.Features) (map[
 			for k, v := range ruleOut.Labels {
 				labels[k] = v
 			}
+			for k, v := range ruleOut.ExtendedResources {
+				extendedResources[k] = v
+			}
 
 			// Feed back rule output to features map for subsequent rules to match
 			features.InsertAttributeFeatures(nfdv1alpha1.RuleBackrefDomain, nfdv1alpha1.RuleBackrefFeature, ruleOut.Labels)
@@ -838,7 +922,7 @@ func (m *nfdMaster) processNodeFeatureRule(features *nfdv1alpha1.Features) (map[
 		}
 	}
 
-	return labels, taints
+	return labels, extendedResources, taints
 }
 
 // updateNodeObject ensures the Kubernetes node object is up to date,
--- a/test/e2e/data/nodefeaturerule-4.yaml
+++ b/test/e2e/data/nodefeaturerule-4.yaml
@@ -1,0 +1,32 @@
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: e2e-extened-resource-test
+spec:
+  rules:
+    - name: "e2e no ns rule"
+      extendedResources:
+        nons: "123"
+      matchFeatures:
+        - feature: "fake.attribute"
+          matchExpressions:
+            "attr_1": {op: IsTrue}
+            "attr_2": {op: IsFalse}
+
+    - name: "e2e Dynamic rule"
+      extendedResources:
+        vendor.io/dynamic: "@fake.attribute.attr_3"
+      matchFeatures:
+        - feature: "fake.attribute"
+          matchExpressions:
+            "attr_3": {op: Exists}
+
+    - name: "e2e static rule"
+      extendedResources:
+        vendor.io/static: "123"
+      matchFeatures:
+
+    - name: "e2e not allowed rule"
+      extendedResources:
+        bad.kubernetes.io/malo: "999"
+      matchFeatures:
--- a/test/e2e/node_feature_discovery_test.go
+++ b/test/e2e/node_feature_discovery_test.go
@@ -31,6 +31,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	extclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
+	resourcev1 "k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	taintutils "k8s.io/kubernetes/pkg/util/taints"
@@ -805,6 +806,35 @@ core:
 					Expect(waitForNfdNodeTaints(f.ClientSet, expectedTaintsUpdated, nodes)).NotTo(HaveOccurred())
 					Expect(waitForNfdNodeAnnotations(f.ClientSet, expectedAnnotationUpdated)).NotTo(HaveOccurred())
 
+					By("Deleting NodeFeatureRule object")
+					err = nfdClient.NfdV1alpha1().NodeFeatureRules().Delete(context.TODO(), "e2e-test-3", metav1.DeleteOptions{})
+					Expect(err).NotTo(HaveOccurred())
+
+					expectedERAnnotation := map[string]string{
+						"nfd.node.kubernetes.io/extended-resources": "nons,vendor.io/dynamic,vendor.io/static"}
+
+					expectedCapacity := corev1.ResourceList{
+						"feature.node.kubernetes.io/nons": resourcev1.MustParse("123"),
+						"vendor.io/dynamic":               resourcev1.MustParse("10"),
+						"vendor.io/static":                resourcev1.MustParse("123"),
+					}
+
+					By("Creating NodeFeatureRules #4")
+					Expect(testutils.CreateNodeFeatureRulesFromFile(nfdClient, "nodefeaturerule-4.yaml")).NotTo(HaveOccurred())
+
+					By("Verifying node annotations from NodeFeatureRules #4")
+					Expect(waitForNfdNodeAnnotations(f.ClientSet, expectedERAnnotation)).NotTo(HaveOccurred())
+
+					By("Verfiying node status capacity from NodeFeatureRules #4")
+					Expect(waitForCapacity(f.ClientSet, expectedCapacity, nodes)).NotTo(HaveOccurred())
+
+					By("Deleting NodeFeatureRule object")
+					err = nfdClient.NfdV1alpha1().NodeFeatureRules().Delete(context.TODO(), "e2e-extened-resource-test", metav1.DeleteOptions{})
+					Expect(err).NotTo(HaveOccurred())
+
+					By("Verfiying node status capacity from NodeFeatureRules #4")
+					Expect(waitForCapacity(f.ClientSet, nil, nodes)).NotTo(HaveOccurred())
+
 					By("Deleting nfd-worker daemonset")
 					err = f.ClientSet.AppsV1().DaemonSets(f.Namespace.Name).Delete(context.TODO(), workerDS.Name, metav1.DeleteOptions{})
 					Expect(err).NotTo(HaveOccurred())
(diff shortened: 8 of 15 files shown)


docs/usage/customization-guide.md
```diff
@@ -491,6 +491,42 @@ rules to use. In other words, these are variables that are not advertised as
...
+kind: NodeFeatureRule
+metadata:
+  name: my-sample-rule-object
+spec:
+  rules:
+    - name: "my sample rule"
+      extendedResources:
+        - "cpu-security.tdx.total_keys"
+      matchFeatures:
+        - feature: cpu-security.tdx.enalbed
```
PiotrProkop, line 512: s/enalbed/enabled

PiotrProkop, line 512: From the documentation here https://kubernetes-sigs.github.io/node-feature-discovery/v0.12/usage/features#cpu it looks like `cpu-security.tdx.enabled` is a boolean, so how it will be converted to`31`?

ArangoGutierrez: ups my bad

fidencio: @PiotrProkop, the example given may not have been the best, as I'm working on exposing the TDX values as part of https://github.com/kubernetes-sigs/node-feature-discovery/pull/1079

ArangoGutierrez: jejeje yeah, I used a non-merged PR for the example, my bad 😅

fidencio:
Using this PR, [combined with mine](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1079) , with the following rule ...
```
kind: NodeFeatureRule
metadata:
  name: tdx-total-keys
spec:
  rules:
    - name: "TDX total keys rule"
      extendedResources:
        - "cpu-security.tdx.total_keys"
      matchFeatures:
        - feature: cpu.security
          matchExpressions:
            tdx.enabled: {op: Exists}
```

... I was able to get ...

```
kubectl get node 984fee00befb.jf.intel.com -o jsonpath='{.status.capacity}' | jq
{
  "cpu": "128",
  "ephemeral-storage": "383957772Ki",
  "feature.node.kubernetes.io/cpu-security.tdx.total_keys": "31",
  "hugepages-1Gi": "0",
  "hugepages-2Mi": "0",
  "memory": "131192364Ki",
  "pods": "110"
}
```

cmd/nfd-master/main.go
```diff
@@ -55,6 +55,8 @@ func main() {
 		switch f.Name {
 		case "featurerules-controller":
 			klog.Warningf("-featurerules-controller is deprecated, use '-crd-controller' flag instead")
+		case "resource-labels":
+			klog.Warningf("-resource-labels is deprecated, use a NodeFeatureRule CR instead")
```
AhmedGrati, line 59: Shouldn't we indicate in the docs, that we'll deprecate this flag?

ArangoGutierrez: Agree

zvonkok:
/lgtm @fidencio Thanks for the examples here and on the depending PR, this explains the bigger picture. 

@ArangoGutierrez AMD told me that the only countable resources for SNP is the number of ASIDs. We can do the same for SNP what Fabiano has done here for TDX.

mythi:
> @fidencio Thanks for the examples here and on the depending PR, this explains the bigger picture.
> 

@zvonkok one hiccup we noticed yesterday was that these extended resources are not correctly shown in nodes' `Allocatable`. However, the accounting works fine. See https://github.com/kubernetes/kubernetes/issues/115190

pkg/nfd-master/nfd-master.go
```diff
@@ -449,12 +449,27 @@ func (m *nfdMaster) filterFeatureLabels(labels Labels) (Labels, ExtendedResource
...
+			outExtendedResources[extendedResourceName] = value
 			delete(outLabels, extendedResourceName)
 		}
 	}
 
-	return outLabels, extendedResources
+	// via NodeFeatureRules CR
+	for _, extendedResourceName := range extendedResources {
+		// Add possibly missing default ns
+		extendedResourceName = addNs(extendedResourceName, nfdv1alpha1.FeatureLabelNs)
```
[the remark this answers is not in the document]
ArangoGutierrez: This feels more like a TODO than a change to the patch right?

docs/usage/customization-guide.md
```diff
@@ -491,6 +491,42 @@ rules to use. In other words, these are variables that are not advertised as
...
+```yaml
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: tdx-total-keys
+spec:
+  rules:
+    - name: "TDX total keys rule"
+      extendedResources:
+        - "cpu-security.tdx.total_keys"
```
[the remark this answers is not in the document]
ArangoGutierrez: This feels more like a TODO than a change to the patch right?

pkg/apis/nfd/v1alpha1/annotations_labels.go
```diff
@@ -32,6 +32,9 @@ const (
 	// AnnotationNs namespace for all NFD-related annotations.
 	AnnotationNs = "nfd.node.kubernetes.io"
 
+	// ExtendedResourceNs is the namespace for extended resources.
+	ExtendedResourceNs = "nfd.node.kubernetes.io"
```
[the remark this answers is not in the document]
ArangoGutierrez: good catch

docs/usage/customization-guide.md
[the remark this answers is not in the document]
mythi: I have two concerns with that: 1) we are deprecating `-resource-labels` but the new functionality is not capable to do the same what we get with `-resource-labels` + `-extra-label-ns`, 2) I cannot make commitments to any user facing resources if these will change in the future

pkg/apis/nfd/v1alpha1/rule.go
```diff
@@ -30,13 +30,15 @@ import (
 // RuleOutput contains the output out rule execution.
 // +k8s:deepcopy-gen=false
 type RuleOutput struct {
-	Labels map[string]string
-	Vars   map[string]string
-	Taints []corev1.Taint
+	ExtendedResources []string
```
marquiz, line 33: Thanks for the PR @ArangoGutierrez. This is a great enhancement! Extended resources have a name and a value so in that sense this should 

```
	ExtendedResources map[string]string
```

or even

```
	ExtendedResources map[string]resource.Quantity
```

[ with `resource.Quantity` being the type from k8s apimachinery ]

OTOH, static quantity specified in the rule like 

```yaml
      extendedResources:
        - tdx.total_keys: 123
```

is not generally very useful so we need some dynamic solution. We already support templating for labels (and vars) so maybe following and extending that convention could be the solution. That is, being able to specify smth like this:

```yaml
      extendedResources:
        - tdx_keys: "{{ cpu.security.tdx.total_keys }}"
```

The stuff inside `{{ }}` could be fed to a real templating engine (what we do for `labelsTemplate` and `varsTemplate`) or then just do poor-man's stripping and splitting to figure out the feature value that the user wanted to expose. Thoughts @ArangoGutierrez @fidencio @mythi @zvonkok et al?

ArangoGutierrez: Yeah I decided to change 
```go
	ExtendedResources map[string]string
```
to
```go
	ExtendedResources []string
```
to be able to define just as string, and then give the quantity and create the label. so we only take the name, and internally assign the value, that way we can 
```yaml
      extendedResources:
        - "cpu-security.tdx.total_keys"
```
So the `list` get's "enabled" if the rules match. 
That was my train of thought.  I do think `static quantity` is a bad approach, we should look for a solution like the provided in this PR or with templating.

marquiz: With this we really want to decouple extended resources from labels. I think we want the cabability to specify extended resource like `example.com/foo: 4` or `vendor.io/bar: "{{ cpu.security.tdx.total_keys }}"`

Also, we don't want to change the API from release to release (unless absolutely necessary) so we'd need to try to get it right so with that in mind `map[string]string`

mythi: That's third item on top the list started here https://github.com/kubernetes-sigs/node-feature-discovery/pull/1099#discussion_r1153088587 :-)

ArangoGutierrez: Got it, will work on the changes

marquiz: Just thought about the format, maybe we just go with simple @ to indicate value injection, like

```yaml
extendedResources:
  vendor.io/bar: @cpu.security.tdx.total_keys
```

`{{ }}` is a bit clumsy and possibly misleading if we're not doing real templating here. WDYT?

ArangoGutierrez:
@mythi @marquiz @fidencio Ready for review, changes have been made according to all the comments and discussion 
Thanks all for the feedback and comments

pkg/nfd-master/nfd-master.go
```diff
@@ -439,7 +439,7 @@ func (m *nfdMaster) filterFeatureLabels(labels Labels) (Labels, ExtendedResource
 	}
 
 	// Remove labels which are intended to be extended resources
-	extendedResources := ExtendedResources{}
+	extendedResources := make(map[string]string)
```
marquiz, line 442: nit: No need to change this I guess

```suggestion
```

examples/nodefeaturerule.yaml
```diff
@@ -5,12 +5,10 @@ metadata:
 spec:
   rules:
     - name: "my sample rule"
-      labels:
-        "my-sample-feature": "true"
+      extendedResources:
+        vendor.io/dynamic: "@kernel.version.major"
+        vendor.io/static: "123"
```
marquiz, line 10: Please don't remove the example that shows how to create labels. You can append a new rule demonstrating extended resources or alternatively extend the existing rule adding the extendedResources field there. The latter could be better because in principle you don't need to match the feature (in the matchFeatures`) in order to use its value in an extended resource (of course with that you can make sure that it exists)

pkg/nfd-master/nfd-master.go
```diff
@@ -605,19 +605,74 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
 	return nil
 }
 
-func (m *nfdMaster) refreshNodeFeatures(cli *kubernetes.Clientset, nodeName string, annotations, labels map[string]string, features *nfdv1alpha1.Features) error {
+// validateExtendedResources validates extended resources and returns a map
+// of valid extended resources.
+func (m *nfdMaster) validateExtendedResources(features *nfdv1alpha1.Features, resourceLabels map[string]string) map[string]string {
```
marquiz, line 610: nit: the name resourceLabels is very misleading here IMO

```suggestion
func (m *nfdMaster) validateExtendedResources(features *nfdv1alpha1.Features, extendedResources map[string]string) map[string]string {
```

pkg/nfd-master/nfd-master.go
```diff
@@ -738,21 +793,22 @@ func authorizeClient(c context.Context, checkNodeName bool, nodeName string) err
 	return nil
 }
 
-func (m *nfdMaster) processNodeFeatureRule(features *nfdv1alpha1.Features) (map[string]string, []corev1.Taint) {
+func (m *nfdMaster) processNodeFeatureRule(features *nfdv1alpha1.Features) (map[string]string, map[string]string, []corev1.Taint) {
```
marquiz, line 796: Would it be possible to use the ExtendedResources type here? Like

```suggestion
func (m *nfdMaster) processNodeFeatureRule(features *nfdv1alpha1.Features) (map[string]string, ExtendedResources, []corev1.Taint) {
```

...if so then we'd probably want to use the Labels type for labels, too...

pkg/nfd-master/nfd-master.go
```diff
@@ -774,14 +830,17 @@ func (m *nfdMaster) processNodeFeatureRule(features *nfdv1alpha1.Features) (map[
...
+			}
 
 			// Feed back rule output to features map for subsequent rules to match
 			features.InsertAttributeFeatures(nfdv1alpha1.RuleBackrefDomain, nfdv1alpha1.RuleBackrefFeature, ruleOut.Labels)
 			features.InsertAttributeFeatures(nfdv1alpha1.RuleBackrefDomain, nfdv1alpha1.RuleBackrefFeature, ruleOut.Vars)
 		}
 	}
 
-	return labels, taints
+	return extendedResources, labels, taints
```
marquiz, line 843: nit: just a matter of style but it looks somehow strange to return the extended resources first 🧐

```suggestion
	return extendedResources, labels, taints
```

pkg/nfd-master/nfd-master.go
```diff
@@ -605,19 +605,74 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
 
 	// Mix in CR-originated labels
 	for k, v := range crLabels {
 		labels[k] = v
 	}
 
-	labels, extendedResources := m.filterFeatureLabels(labels)
+	// Remove labels which are intended to be extended resources via
+	// -resource-labels or their NS is not whitelisted
+	labels, resourceLabels := m.filterFeatureLabels(labels)
```
marquiz, line 669: maybe we could name the var just `extendedResources`?

```suggestion
	labels, extendedResources := m.filterFeatureLabels(labels)
```

pkg/apis/nfd/v1alpha1/types.go
```diff
@@ -51,6 +51,10 @@ type NodeFeatureSpec struct {
 	// Labels is the set of node labels that are requested to be created.
 	// +optional
 	Labels map[string]string `json:"labels"`
+	// ExtendedResources is the set of extended resources that are requested to
+	// be created.
+	// +optional
+	ExtendedResources map[string]string `json:"extendedResources"`
```
marquiz, line 57: This makes ERs supported in the NodeFeature object, too. Maybe we want it but then we need to document it also everywhere (where NFs) are discussed and have separate e2e tests for it

pkg/nfd-master/nfd-master.go
```diff
@@ -605,19 +605,74 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+	extendedResources := make(map[string]string)
+	for extendedResource, capacity := range resourceLabels {
+		// Add possibly missing default ns
+		extendedResource = addNs(extendedResource, nfdv1alpha1.ExtendedResourceNs)
+		// Check if given NS is allowed
+		ns, _ := splitNs(extendedResource)
+		if isNamespaceDenied(ns, m.deniedNs.wildcard, m.deniedNs.normal) {
+			if _, ok := m.args.ExtraLabelNs[ns]; !ok {
+				klog.Errorf("Namespace %q is not allowed. Ignoring Extended Resource  %q\n", ns, extendedResource)
+				continue
```
marquiz, line 620: I think we could have a static deny list for now. Anyway keep it separate from the labels (the config option is also named `denyLabelNs`).

pkg/nfd-master/nfd-master.go
```diff
@@ -605,19 +605,74 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+			}
+		}
+
+		// Dynamic Value
+		if strings.HasPrefix(capacity, "@") {
+			// Remove the @ prefix
+			capacity = strings.TrimPrefix(capacity, "@")
+			// capacity is a string in the form of attribute.featureset.elements
+			split := strings.SplitN(capacity, ".", 3)
+			if value, ok := features.Attributes[split[0]+"."+split[1]]; ok {
```
marquiz, line 630: I think we want to improve the error messages here a bit. Let's call `split[0]+"."+split[1]` _feature_ and  `split[2]` _element_. We want an error message if _feature_ is not found (we could have another error message of _feature_ is found from Flags or Instances instead of Attributes). We also want an error message if _element_ is not found inside the _feature_. Third error is then the one that the value is not qualified as a quantity,

pkg/nfd-master/nfd-master.go
```diff
@@ -605,19 +605,74 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+			continue
+		}
+		// Static Value (Pre-Defined at the NodeFeatureRule)
+		q, err := k8sQuantity.ParseQuantity(capacity)
+		if err != nil {
+			klog.Errorf("bad label value %s encountered for extended resource: %s", capacity, extendedResource, err.Error())
+			continue
+		}
+		extendedResources[extendedResource] = q.String()
+		continue
```
marquiz, line 649: Maybe we wouldn't need so many `continue`s if we put this inside else
```suggestion
		} else {
			// Static Value (Pre-Defined at the NodeFeatureRule)
			q, err := k8sQuantity.ParseQuantity(capacity)
			if err != nil {
				klog.Errorf("bad label value %s encountered for extended resource: %s", capacity, extendedResource, err.Error())
				continue
		}
		extendedResources[extendedResource] = q.String()
```

marquiz (commented):
Thanks @ArangoGutierrez for the update. Shaping up nicely 👍 I left a few comments

docs/usage/customization-guide.md
```diff
@@ -491,6 +491,45 @@ rules to use. In other words, these are variables that are not advertised as
...
+```yaml
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
+metadata:
+  name: my-sample-rule-object
+spec:
+  rules:
+    - name: "my sample rule"
+      extendedResources:
+        vendor.io/dynamic: "@kernel.version.major"
```
marquiz, line 510: I think we should explain this `@feature.name` format in written form below. Referencing to the table in "Available features" section below.

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,81 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+	deniedWildCarNs := map[string]struct{}{".kubernetes.io": {}}
+	allowedNS := map[string]struct{}{nfdv1alpha1.ExtendedResourceNs: {},
+		nfdv1alpha1.FeatureLabelNs: {}}
+	for extendedResource, capacity := range extendedResources {
+		// Add possibly missing default ns
+		extendedResource = addNs(extendedResource, nfdv1alpha1.ExtendedResourceNs)
+		// Check if given NS is allowed
+		ns, _ := splitNs(extendedResource)
+		if isNamespaceDenied(ns, deniedWildCarNs, deniedNs) {
+			if _, ok := allowedNS[ns]; !ok {
```
marquiz, line 621: I think we should also allow sub-ns of nfdv1alpha1.ExtendedResourceNs, i.e.
`HasPrefix(ns, nfdv1alpha1.FeatureLabelSubNsSuffix`)

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,81 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+
+		// Dynamic Value
+		if strings.HasPrefix(capacity, "@") {
+			// Remove the @ prefix
+			capacity = strings.TrimPrefix(capacity, "@")
+			// capacity is a string in the form of attribute.featureset.elements
+			split := strings.SplitN(capacity, ".", 3)
+			if attrFeatureSet, ok := features.Attributes[split[0]+"."+split[1]]; ok {
+				if element, ok := attrFeatureSet.Elements[split[2]]; ok {
+					q, err := k8sQuantity.ParseQuantity(element)
```
marquiz, line 635: nit: I think this is a matter of style but somehow this deep nested if structure with error cases very far below is a bit hard to follow. I'd suggest to handle the error case directly after the operation, i.e. smth like

```suggestion
			featureName := split[0]+"."+split[1]
			elementName := split[2]
			attrFeatureSet, ok := features.Attributes[featureName]
			if !ok {
				klog.Errorf("feature %s not found. Ignoring Extended Resource %q", featureName, extendedResource)
				continue
			}
			element, ok := attrFeatureSet.Elements[elementName]
			if !ok {
				klog.Errorf("element %s not foundon feature %s. Ignoring Extended Resource %q", elementName, split[0]+"."+split[1], extendedResource)
				continue
			}
			q, err := k8sQuantity.ParseQuantity(element)
```


WDYT?

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,81 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+					if err != nil {
+						klog.Errorf("bad label value %s encountered for extended resource: %s", q.String(), extendedResource, err.Error())
+						continue
+					}
+					outExtendedResources[extendedResource] = q.String()
+					continue
+				}
+				klog.Errorf("element %s not foundon feature %s. Ignoring Extended Resource %q/n", split[2], split[0]+"."+split[1], extendedResource)
```
marquiz, line 643: remove `/n`
```suggestion
				klog.Errorf("element %s not foundon feature %s. Ignoring Extended Resource %q", split[2], split[0]+"."+split[1], extendedResource)
```

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,81 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+			}
+			klog.Errorf("feature %s not found. Ignoring Extended Resource %q/n", split[0]+"."+split[1], extendedResource)
```
marquiz, line 645: remove `/n`
```suggestion
			klog.Errorf("feature %s not found. Ignoring Extended Resource %q", split[0]+"."+split[1], extendedResource)
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -766,6 +766,56 @@ core:
...
+						testpod.SpecWithContainerImage(dockerImage()),
+						testpod.SpecWithConfigMap(cm.Name, "/etc/kubernetes/node-feature-discovery"),
+						testpod.SpecWithTolerations(testTolerations),
+					)
+					workerDS := testds.NFDWorker(podSpecOpts...)
+					workerDS, err = f.ClientSet.AppsV1().DaemonSets(f.Namespace.Name).Create(context.TODO(), workerDS, metav1.CreateOptions{})
+					Expect(err).NotTo(HaveOccurred())
+
+					By("Waiting for worker daemonset pods to be ready")
+					Expect(testpod.WaitForReady(f.ClientSet, f.Namespace.Name, workerDS.Spec.Template.Labels["name"], 2)).NotTo(HaveOccurred())
```
marquiz, line 804: I think we can just append your new test to the existing test case above, to eliminate duplicating all the identical test setup code

```suggestion
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -766,6 +766,56 @@ core:
...
+
+					expectedAnnotation := map[string]string{
+						"nfd.node.kubernetes.io/extended-resources": "nfd.node.kubernetes.io/nons,vendor.io/static"}
+
+					By("Creating NodeFeatureRules #4")
+					Expect(testutils.CreateNodeFeatureRulesFromFile(nfdClient, "nodefeaturerule-4.yaml")).NotTo(HaveOccurred())
```
marquiz, line 810: `nodefeaturerule-4.yaml` is missing from the PR

test/e2e/node_feature_discovery_test.go
```diff
@@ -766,6 +766,56 @@ core:
...
+					)
+					workerDS := testds.NFDWorker(podSpecOpts...)
+					workerDS, err = f.ClientSet.AppsV1().DaemonSets(f.Namespace.Name).Create(context.TODO(), workerDS, metav1.CreateOptions{})
+					Expect(err).NotTo(HaveOccurred())
+
+					By("Waiting for worker daemonset pods to be ready")
+					Expect(testpod.WaitForReady(f.ClientSet, f.Namespace.Name, workerDS.Spec.Template.Labels["name"], 2)).NotTo(HaveOccurred())
+
+					expectedAnnotation := map[string]string{
+						"nfd.node.kubernetes.io/extended-resources": "nfd.node.kubernetes.io/nons,vendor.io/static"}
```
marquiz, line 807: There is also `fake.attribute.attr_3` feature that has integer value `10` so we could use that to test the dynamic feature value `@fake.attribute.attr_3` (which should be set to 10).

marquiz (commented):
Good work @ArangoGutierrez! Almost there, I think, just a few more comments/sugggestions 😇

pkg/nfd-master/nfd-master.go
ArangoGutierrez: After taking https://github.com/kubernetes-sigs/node-feature-discovery/pull/1099#discussion_r1157614130 , this comment is invalid

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,85 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+	deniedWildCarNs := map[string]struct{}{".kubernetes.io": {}}
+	for extendedResource, capacity := range extendedResources {
+		// Add possibly missing default ns
+		extendedResource = addNs(extendedResource, nfdv1alpha1.ExtendedResourceNs)
+		// Check if given NS is allowed
+		ns, _ := splitNs(extendedResource)
+		if isNamespaceDenied(ns, deniedWildCarNs, deniedNs) {
+			if ns != nfdv1alpha1.ExtendedResourceNs && ns != nfdv1alpha1.FeatureLabelNs &&
+				!strings.HasSuffix(ns, nfdv1alpha1.ExtendedResourceSubNsSuffix) && !strings.HasSuffix(ns, nfdv1alpha1.FeatureLabelSubNsSuffix) {
+				klog.Errorf("Namespace %q is not allowed. Ignoring Extended Resource  %q\n", ns, extendedResource)
```
marquiz, line 621: You don't need the linefeed. Also, just remembered that we've tried to have a convention of starting log messages with lower case letter...

```suggestion
				klog.Errorf("namespace %q is not allowed. Ignoring Extended Resource  %q", ns, extendedResource)
```

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,85 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+		if strings.HasPrefix(capacity, "@") {
+			// Remove the @ prefix
+			capacity = strings.TrimPrefix(capacity, "@")
+			// capacity is a string in the form of attribute.featureset.elements
+			split := strings.SplitN(capacity, ".", 3)
+			featureName := split[0] + "." + split[1]
+			elementName := split[2]
+			attrFeatureSet, ok := features.Attributes[featureName]
+			if !ok {
+				klog.Errorf("feature %s not found. Ignoring Extended Resource %q/n", featureName, extendedResource)
```
marquiz, line 636: Invalid linefeed, and we don't need linefeeds anyway

```suggestion
				klog.Errorf("feature %s not found. Ignoring Extended Resource %q", featureName, extendedResource)
```

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,85 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+				continue
+			}
+			element, ok := attrFeatureSet.Elements[elementName]
+			if !ok {
+				klog.Errorf("element %s not foundon feature %s. Ignoring Extended Resource %q/n", elementName, featureName, extendedResource)
```
marquiz, line 641: Invalid linefeed, and we don't need linefeeds anyway


```suggestion
				klog.Errorf("element %s not foundon feature %s. Ignoring Extended Resource %q", elementName, featureName, extendedResource)
```

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,85 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+				continue
+			}
+			q, err := k8sQuantity.ParseQuantity(element)
+			if err != nil {
+				klog.Errorf("bad label value %s encountered for extended resource: %s", q.String(), extendedResource, err.Error())
```
marquiz, line 646:
```suggestion
				klog.Errorf("bad label value %s encountered for extended resource: %v", q.String(), extendedResource, err)
```

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,85 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+				continue
+			}
+			outExtendedResources[extendedResource] = q.String()
+			continue
+		}
+		// Static Value (Pre-Defined at the NodeFeatureRule)
+		q, err := k8sQuantity.ParseQuantity(capacity)
+		if err != nil {
+			klog.Errorf("bad label value %s encountered for extended resource: %s", capacity, extendedResource, err.Error())
```
marquiz, line 655:
```suggestion
			klog.Errorf("bad label value %s encountered for extended resource: %c", capacity, extendedResource, err)
```

test/e2e/data/nodefeaturerule-4.yaml
```diff
@@ -0,0 +1,40 @@
...
+            "attr_2": {op: IsFalse}
+
+    - name: "e2e not allowed rule"
+      extendedResources:
+        bad.kubernetes.io/malo: "123"
+      matchFeatures:
+        - feature: "fake.attribute"
+          matchExpressions:
+            "attr_1": {op: IsTrue}
+            "attr_2": {op: IsFalse}
```
marquiz, line 40: You could change this so that it always matches so we don't accidentally break this negative test rule e.g. by changing the fake features

```suggestion
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -766,6 +766,23 @@ core:
...
+					Expect(err).NotTo(HaveOccurred())
+
+					expectedERAnnotation := map[string]string{
+						"nfd.node.kubernetes.io/extended-resources": "nfd.node.kubernetes.io/nons,vendor.io/dynamic,vendor.io/static"}
+
+					By("Creating NodeFeatureRules #4")
+					Expect(testutils.CreateNodeFeatureRulesFromFile(nfdClient, "nodefeaturerule-4.yaml")).NotTo(HaveOccurred())
+
+					By("Verifying node annotations from NodeFeatureRules #4")
+					Expect(waitForNfdNodeAnnotations(f.ClientSet, expectedERAnnotation)).NotTo(HaveOccurred())
```
marquiz, line 780: In addition to the annotation we really want to verify that the ERs are created, too (and the denied on isn't)

marquiz (commented):
Good progress 👍 Next review round... 😎

pkg/apis/nfd/v1alpha1/annotations_labels.go
```diff
@@ -32,6 +32,12 @@ const (
...
```
marquiz, line 36: I'm a bit undecided should this be `nfd.node.kubernetes.io` or `feature.node.kubernetes.io` 🧐 Any thoughts? @zvonkok @fidencio @ArangoGutierrez

fidencio: Hmm.  Fair question, and I think a better fit would be `feature.node.kubernetes.io`.

ArangoGutierrez: `feature.node.kubernetes.io` it is!

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,85 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+	outExtendedResources := ExtendedResources{}
+	deniedNs := map[string]struct{}{"kubernetes.io": {}}
+	deniedWildCarNs := map[string]struct{}{".kubernetes.io": {}}
+	for extendedResource, capacity := range extendedResources {
+		// Add possibly missing default ns
+		extendedResource = addNs(extendedResource, nfdv1alpha1.ExtendedResourceNs)
+		// Check if given NS is allowed
+		ns, _ := splitNs(extendedResource)
+		if isNamespaceDenied(ns, deniedWildCarNs, deniedNs) {
+			if ns != nfdv1alpha1.ExtendedResourceNs && ns != nfdv1alpha1.FeatureLabelNs &&
```
PiotrProkop, line 619: can you move this check to separate func like `IsAllowedNamespace`? It would improve readability

pkg/nfd-master/nfd-master.go
```diff
@@ -605,20 +604,85 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
...
+				continue
+			}
+		}
+
+		// Dynamic Value
+		if strings.HasPrefix(capacity, "@") {
+			// Remove the @ prefix
+			capacity = strings.TrimPrefix(capacity, "@")
+			// capacity is a string in the form of attribute.featureset.elements
+			split := strings.SplitN(capacity, ".", 3)
```
PiotrProkop, line 631: you know for sure that 1 char is `@` so you can just
```suggestion

			// capacity is a string in the form of attribute.featureset.elements
			split := strings.SplitN(capacity[1:], ".", 3)
```

PiotrProkop (commented):
Just some nits. Overall good work!

PiotrProkop:
/lgtm

@marquiz for final approval

pkg/nfd-master/nfd-master.go
```diff
@@ -519,6 +518,22 @@ func isNamespaceDenied(labelNs string, wildcardDeniedNs map[string]struct{}, nor
 	return false
 }
 
+// if ns != nfdv1alpha1.ExtendedResourceNs && ns != nfdv1alpha1.FeatureLabelNs &&
+// !strings.HasSuffix(ns, nfdv1alpha1.ExtendedResourceSubNsSuffix) && !strings.HasSuffix(ns, nfdv1alpha1.FeatureLabelSubNsSuffix)
```
marquiz, line 522: this comment is rather strange

test/e2e/data/nodefeaturerule-4.yaml
marquiz: How about this change?

test/e2e/node_feature_discovery_test.go
```diff
@@ -899,6 +926,39 @@ denyLabelNs: []
 
 })
 
+// simplyAndStupidRetryLoop is a simple and stupid re-try loop
+func simplyAndStupidRetryLoop(poll func() error) error {
```
marquiz, line 930: nit: I'd suggest to rename this to smth more appropriate 😅

```suggestion
// simplePoll is a simple and stupid re-try loop
func simplePoll(poll func() error) error {
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -899,6 +926,39 @@ denyLabelNs: []
...
+		if err = poll(); err == nil {
+			return nil
+		}
+		time.Sleep(1 * time.Second)
+	}
+	return err
+}
+
+// waitForCapacity waits for the capacity to be updated in the node status
+func waitForCapacity(cli clientset.Interface, expected map[corev1.ResourceName]resourcev1.Quantity) error {
```
marquiz, line 942: With the current test data we also want to verify that no extra ERs (those ones that should be denied) are created. You can check what [checkForNodeLabels()](https://github.com/kubernetes-sigs/node-feature-discovery/blob/60f052f086154fd8f7eb4bd24208bbfb32c40578/test/e2e/node_feature_discovery_test.go#L933) does

Would be smth like:

```suggestion
func waitForCapacity(cli clientset.Interface, expectedNewERs map[corev1.ResourceName]resourcev1.Quantity, oldNodes []corev1.Node) error {
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -809,6 +810,32 @@ core:
...
+
+					By("Verifying node annotations from NodeFeatureRules #4")
+					Expect(waitForNfdNodeAnnotations(f.ClientSet, expectedERAnnotation)).NotTo(HaveOccurred())
+
+					By("Verfiying node status capacity from NodeFeatureRules #4")
+					Expect(waitForCapacity(f.ClientSet, expectedCapacity)).NotTo(HaveOccurred())
+
+					By("Deleting NodeFeatureRule object")
+					err = nfdClient.NfdV1alpha1().NodeFeatureRules().Delete(context.TODO(), "e2e-extened-resource-test", metav1.DeleteOptions{})
+					Expect(err).NotTo(HaveOccurred())
```
marquiz, line 837: If/when we update the `waitForCapacity()` function I'd suggest to add a test that verifies that the ERs were removed with the NFR object:

```golang
Expect(waitForCapacity(f.ClientSet, nil, nodes)).NotTo(HaveOccurred())
```

pkg/nfd-master/nfd-master.go
ArangoGutierrez: ups, yeah I had it up there as a guide when writing the func

marquiz:
@ArangoGutierrez the commit message is virtually empty. Could you add a bit flesh on that part, explaining what is being done (and if possible also why) 🥸
https://www.kubernetes.dev/docs/guide/pull-requests/#use-the-commit-message-body-to-explain-the-_what_-and-_why_-of-the-commit

Makefile
```diff
@@ -192,13 +192,13 @@ e2e-test:
 	    -kubeconfig=$(KUBECONFIG) \
 	    -nfd.e2e-config=$(E2E_TEST_CONFIG) \
 	    -nfd.pull-if-not-present=$(E2E_PULL_IF_NOT_PRESENT) \
-	    -ginkgo.focus="\[kubernetes-sigs\]" \
+	    -ginkgo.focus="\[kubernetes-sigs\] NFD master" \
```
marquiz, line 195: don't change these

docs/usage/customization-guide.md
```diff
@@ -491,6 +491,65 @@ rules to use. In other words, these are variables that are not advertised as
...
+    ...
+  capacity:
+    ...
+    vendor.io/dynamic: "5"
+    vendor.io/static: "123"
+    ...
+```
+
+There are some limitations to the namespace part (i.e. prefix)/ of the Extended
+Resources keys
```
marquiz, line 544:
```suggestion
Resources names:
```

docs/usage/customization-guide.md
```diff
@@ -491,6 +491,65 @@ rules to use. In other words, these are variables that are not advertised as
...
+
+- `kubernetes.io/` and its sub-namespaces (like `sub.ns.kubernetes.io/`) cannot
+  generally be used
+- the only exception is `feature.node.kubernetes.io/` and its sub-namespaces
+  (like `sub.ns.feature.node.kubernetes.io`)
+- unprefixed keys (like `foo`) will get prefixed with
+  `feature.node.kubernetes.io/` automatically.
```
marquiz, line 551:
```suggestion
- unprefixed names will get prefixed with `feature.node.kubernetes.io/`
  automatically (e.g. `foo` becomes `feature.node.kubernetes.io/foo`)
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -899,6 +929,39 @@ denyLabelNs: []
...
+		if err = poll(); err == nil {
+			return nil
+		}
+		time.Sleep(1 * time.Second)
+	}
+	return err
+}
+
+// waitForCapacity waits for the capacity to be updated in the node status
+func waitForCapacity(cli clientset.Interface, expectedNewERs map[corev1.ResourceName]resourcev1.Quantity, oldNodes []corev1.Node) error {
```
marquiz, line 945: You should use the `oldNodes` arg for something, i.e. expectedNewERs were added, and only those, nothing more

ArangoGutierrez: `waitForCapacity` is inspired on `waitForNfdNodeTaints` and `waitForNfdNodeAnnotations` , that's why it doesn't take `oldNode` as an arg

marquiz: Look what waitForNfdNodeTaints does in #1118 😆

marquiz (approved):
Big hand for this work @ArangoGutierrez. I'd be happy to get this in 😎 Really cool new feature and nice e2e coverage, too. +1

ArangoGutierrez:
@PiotrProkop @zvonkok @fidencio  PR is ready for your `lgtm` now :)

marquiz:
ping @fidencio @PiotrProkop @mythi @zvonkok

ArangoGutierrez:
> Hmm, e2e failed /hold

```bash
Ran 8 of 263 Specs in 279.768 seconds
SUCCESS! -- 8 Passed | 0 Failed | 0 Pending | 255 Skipped
--- PASS: TestE2E (279.77s)
PASS
ok      sigs.k8s.io/node-feature-discovery/test/e2e     279.813s
```
I got a successful run before pushing

marquiz:
There's an ancient bug in nfd-master that this PR exposed. Fixed by #1119. I'd suggest to merge that one first either way is fine, we need both of these. Anyway, this PR (1099) still legit
/unhold

ArangoGutierrez:
> There's an ancient bug in nfd-master that this PR exposed. Fixed by #1119. I'd suggest to merge that one first either way is fine, we need both of these. Anyway, this PR (1099) still legit /unhold

And we are back in business here!

PiotrProkop:
I am happy to LGTM this PR but something is wrong with Netlify :(

ArangoGutierrez:
> /retest

Prow comments don't retrigger GitHub actions. 
And I don't think is the PR changes fault for Netify errors .... I think jejeje 🤞

ArangoGutierrez:
@PiotrProkop  Netlify is happy now

PiotrProkop:
/lgtm

Good job! @ArangoGutierrez @fidencio

PR #1101, opened 2023-03-29
nfd-worker: add annotation with worker-conf sha256 hash

tariq1890:
This PR allows for automatic rolling of worker pods whenever there is an update to the `nfd-worker-conf` configmap.

By default helm will not refresh the DaemonSet/Deployment when there is no change observed in the manifest. This gets around that. 

For more details: see [here](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments)

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/worker.yaml
+++ b/deployment/helm/node-feature-discovery/templates/worker.yaml
@@ -7,7 +7,10 @@ metadata:
     {{- include "node-feature-discovery.labels" . | nindent 4 }}
     role: worker
   annotations:
+    checksum/worker-conf: {{ include (print $.Template.BasePath "/nfd-worker-conf.yaml") . | sha256sum }}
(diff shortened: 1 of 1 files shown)


ArangoGutierrez:
My HELM-fu is not so good, @adrianchiris could you take a look? 
/cc @adrianchiris

marquiz:
> calling a rolling update is not necessary since the worker is constantly reading the file.

Yeah, this should not be needed. As @ArangoGutierrez described nfd-worker has an fsnotify watch on the config file. When the file changes the worker immediately reconfigures itself (and does feature rediscovery / re-labeling). It may take a while (up to 1 minute) for K8s to propagate the updated configmap to the actual file inside the container, though, so there is some delay.

tariq1890:
@marquiz @ArangoGutierrez  Thanks for the information. I do agree then, that we don't need this change. I am going to close this PR

Issue #1113, opened 2023-04-04
Deprecate RDT labels


marquiz:
**What would you like to be added**:

Deprecate built-in [RDT labels](https://kubernetes-sigs.github.io/node-feature-discovery/v0.12/usage/features#intel-rdt-flags). The default built-in RDT labels would not be created anymore, but the RDT features would still be available for NodeFeatureRules to consume if/when somebody wants to expose them

For example:
- v0.13: deprecate the labels
- v0.13 + 2: remove the labels

**Why is this needed**:

The RDT labels are not very useful in their current form (they don't e.g indicate if the features are really enabled i.e. kernel supports them or the resctrlfs is mounted). This would also reduce the flood/clutter of labels created by NFD.

marquiz:
> @ArangoGutierrez: You must be a member of the [kubernetes-sigs/node-feature-discovery-maintainers](https://github.com/orgs/kubernetes-sigs/teams/node-feature-discovery-maintainers/members)

Soon we'll get this fixed 😉

marquiz:
IF we follow what I suggested in #1131 we'd want to move this into v0.16: RDT labels were marked as deprecated in v0.13 and we'd keep the for two more releases (v0.14 and v0.15) and finally drop in v0.16

PR #1114, opened 2023-04-04
source/cpu: deprecate cpu-rdt.* labels

marquiz:
Document built-in RDT labels to be deprecated and removed in a future release. The plan is that the default built-in RDT labels would not be created anymore, but the RDT features would still be available for NodeFeatureRules to consume.

The RDT labels are not very useful (they don't e.g indicate if the features are really enabled in kernel or if the resctrlfs is mounted).

Refs #1113


marquiz:
@PiotrProkop any thoughts on this?

PiotrProkop:
> @PiotrProkop any thoughts on this?

For now we are consuming RDT labels via our custom `worker` config, something like this:
```json
   "sources": {
      "custom":[
        {
          "name": "RDTL3CA",
          "labels": {
            "feature.node.kubernetes.io/cpu-rdt.RDTL3CA": "true"
          },
          "matchFeatures": [
           {
             "feature": "cpu.rdt",
             "matchExpressions": {
               "RDTL3CA": {"op": "Exists"}
             }
           }
         ]
        }
      ],
```
but the plan is to migrate to `NodeFeatureRules` so it's
/lgtm for me

ArangoGutierrez (approved):
/lgtm

PR #1115, opened 2023-04-05
test/e2e: fix node cleanup function

marquiz:
The node cleanup function was not removing all NFD-labels. It omitted NFD-originated labels that used a non-default label namespace. This patch fixes the issue by getting all NFD-managed labels from the special annotation (nfd.node.kubernetes.io/feature-labels).

The patch also adds the ability to cleanup extended resources in a similar way. This will be needed by future work.

Also changes the order of cleaning up CRs and the node. It is the right order as cleaning up the CRs may still update the node.


marquiz:
/assign @ArangoGutierrez 
ping @AhmedGrati

ArangoGutierrez (approved):
/lgtm

PR #1118, opened 2023-04-05
nfd-master: disallow unprefixed and kubernetes taints

marquiz:
Disallow taints having a key with "kubernetes.io/" or "*.kubernetes.io/"
prefix. This is a precaution to protect the user from messing up with
the "official" well-known taints from Kubernetes itself. The only
exception is that the "nfd.node.kubernetes.io/" prefix is allowed.

However, there is one allowed NFD-specific namespace (and its
sub-namespaces) i.e. "feature.node.kubernetes.io" under the
kubernetes.io domain that can be used for NFD-managed taints.

Also disallow unprefixed taint keys. We don't add a default prefix to
unprefixed taints (like we do for labels) from NodeFeatureRules. This is
to prevent unpleasant surprises to users that need to manage matching
tolerations for their workloads.

### DIFF
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -159,6 +159,7 @@ re-labeling delay up to the sleep-interval of nfd-worker (1 minute by default).
 
 See [Label rule format](#label-rule-format) for detailed description of
 available fields and how to write labeling rules.
+
 ### NodeFeatureRule tainting feature
 
 This feature is experimental.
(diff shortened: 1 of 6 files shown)


marquiz:
Depends on #1117 
/hold

docs/usage/customization-guide.md
```diff
@@ -209,6 +210,15 @@ spec:
 In this example, if the `my sample taint rule` rule is matched, `feature.node.kubernetes.io/pci-0300_1d0f.present=true:NoExecute`
 and `feature.node.kubernetes.io/cpu-cpuid.ADX:NoExecute` taints are set on the node.
 
+There are some limitations to the namespace part (i.e. prefix)/ of the taint
+key:
+
+- `kubernetes.io/` and its sub-namespaces (like `sub.ns.kubernetes.io/`) cannot
+  generally be used
+- tbe only exception is `feature.node.kubernetes.io/` and its sub-namespaces
```
ArangoGutierrez, line 218:
```suggestion
- the only exception is `feature.node.kubernetes.io/` and its sub-namespaces
```

ArangoGutierrez (changes requested):

ArangoGutierrez (changes requested):

ArangoGutierrez:
> Depends on #1117 /hold

/unhold

ArangoGutierrez (approved):
/lgtm

PR #1093, opened 2023-03-19
Discover node features as annotations

bebc:
resolve issue #863

PR #1120, opened 2023-04-09
Discover node features as annotations

### LINKED ISSUE #863: Discover node features as annotations

bebc:
resolve #863 

#1093 has too match conflict when extender resource featrue merge to master.So I have to open another pr.
And this pr I think is more clear.Thanks to review this again.

### DIFF
--- a/deployment/base/nfd-crds/nfd-api-crds.yaml
+++ b/deployment/base/nfd-crds/nfd-api-crds.yaml
@@ -155,6 +155,11 @@ spec:
                   description: Rule defines a rule for node customization such as
                     labeling.
                   properties:
+                    annotations:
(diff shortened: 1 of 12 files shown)


marquiz:
Please you don't need to open a new pull request 😱 We lose all the review history with this. You can just do git force push (`git push -f `) to overwrite the old branch instead of creating a new one. Can you restore the branchin #1093 and re-open the PR?

/close

PR #1122, opened 2023-04-11
docs: add missing mentions of extended resources and taints

marquiz:
A small update to fix some missing mentions of extended resources and taints as assets managed by NFD.


ArangoGutierrez (approved):
/lgtm

PR #1123, opened 2023-04-12
nfd-master: increase controller resync period to 1 hour

marquiz:
Increase the NFD API controller resync period from 5 minutes to 1 hour.
The resync causes nfd-master to replay all NodeFeature and
NodeFeatureRule objects, being effectively a "big hammer reset all"
button. This should only be needed as an "insurance" to fix labels et al
in case they have been manually tampered (outside NFD) and against
certain bugs in nfd itself. NFD is not supposed to manage anything
fast-changing so 1 hour should be enough.

This change only affects behavior when the NodeFeature API has been
enabled (with -enable-nodefeature-api).

### DIFF
--- a/pkg/nfd-master/nfd-api-controller.go
+++ b/pkg/nfd-master/nfd-api-controller.go
@@ -50,7 +50,7 @@ func newNfdController(config *restclient.Config, disableNodeFeature bool) (*nfdC
 
 	nfdClient := nfdclientset.NewForConfigOrDie(config)
 
-	informerFactory := nfdinformers.NewSharedInformerFactory(nfdClient, 5*time.Minute)
+	informerFactory := nfdinformers.NewSharedInformerFactory(nfdClient, 1*time.Hour)
 
 	// Add informer for NodeFeature objects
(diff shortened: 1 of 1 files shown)


marquiz:
TBH I have little clue what would be the "right value" for the resync period, it could be even hours I think 😇 Any thoughts @ArangoGutierrez @zvonkok @PiotrProkop et al?

PiotrProkop:
> TBH I have little clue what would be the "right value" for the resync period, it could be even hours I think 😇 Any thoughts @ArangoGutierrez @zvonkok @PiotrProkop et al?

I would go ahead and change it to `1 * time.Hour`.

marquiz:
> I would go ahead and change it to `1 * time.Hour`.

Sold 😅 Increased to one hour now

/retitle nfd-master: increase controller resync period to 1 hour

marquiz:
> We should add a `flag/NFDmaster config` option to edit this by the user in the future

Agree, for v0.14.0

marquiz:
Added #1124

PR #1130, opened 2023-04-13
source/cpu: don't create cpu-security.tdx.total_keys label

marquiz:
Just have that as a feature for NodeFeatureRules to consume.

### DIFF
--- a/source/cpu/cpu.go
+++ b/source/cpu/cpu.go
@@ -20,6 +20,7 @@ import (
 	"os"
 	"strconv"
 
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
 
 	"github.com/klauspost/cpuid/v2"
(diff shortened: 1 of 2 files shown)


marquiz:
My original thinking with #1079 was to NOT have tdx.total_keys as a default built-in label. Just as a feature for NodeFeatureRule(s) to consume and expose. But I lost that idea during the review process. This PR kinda implements what I had in mind. 

@ping @fidencio @mythi @zvonkok WDYT?

fidencio (approved):
lgtm, thanks @marquiz!

source/cpu/cpu.go
```diff
@@ -173,8 +173,13 @@ func (s *cpuSource) GetLabels() (source.FeatureLabels, error) {
 	}
 
 	// Security
+	skipLabel := map[string]struct{}{
```
PiotrProkop, line 176: you can use `k8s.io/apimachinery/pkg/util/sets` and then just:
```
sets.NewString("tdx.total_keys")
```
and check if given feature is in skipLabel by:
```
if !skipLabel.Has(k){
...
}
```

marquiz: @PiotrProkop nice suggestion, I now changed the code like that

PiotrProkop:
lgtm, will wait for others to comment before giving proper one!

source/cpu/cpu.go
```diff
@@ -173,8 +174,11 @@ func (s *cpuSource) GetLabels() (source.FeatureLabels, error) {
 	}
 
 	// Security
+	skipLabel := sets.NewString("tdx.total_keys")
```
mythi, line 177: Maybe add a comment?
```suggestion
        // A set of features that will not have labels created but are only made available for
        // NodeFeatureRules (e.g. to be published via extended resources instead)
	skipLabel := sets.NewString("tdx.total_keys")
```

marquiz: 👍 Added a commet

mythi (approved):
/lgtm

PR #1133, opened 2023-04-14
feat: parallelize nodes update

### LINKED ISSUE #1096: Enhancement: parallelize nodes update
Currently, we are updating nodes in a sequential way.
https://github.com/kubernetes-sigs/node-feature-discovery/blob/821e042dbfde7fe509a26f2215385d7302d9fb51/pkg/nfd-master/nfd-master.go#L295-L306
I think that we can optimize this step by parallelizing this process.

**Why is this needed**:
Optimization of updating nodes process, in terms of execution time.

AhmedGrati:
Resolves #1096.

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -70,6 +70,8 @@ func main() {
 			args.Overrides.NoPublish = overrides.NoPublish
 		case "resync-period":
 			args.Overrides.ResyncPeriod = overrides.ResyncPeriod
+		case "nfd-api-parallelism":
+			args.Overrides.NfdApiParallelism = overrides.NfdApiParallelism
 		}
 	})
 
@@ -156,5 +158,8 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 	flagset.Var(overrides.ResyncPeriod, "resync-period",
 		"Specify the NFD API controller resync period."+
 			"It has an effect when the NodeFeature API has been enabled (with -enable-nodefeature-api).")
+	overrides.NfdApiParallelism = flagset.Int("nfd-api-parallelism", 10, "Defines the maximum number of goroutines responsible of updating nodes. "+
+		"Can be used for the throttling mechanism. It has effect only when -enable-nodefeature-api has been set.")
+
 	return args, overrides
 }
--- a/deployment/helm/node-feature-discovery/values.yaml
+++ b/deployment/helm/node-feature-discovery/values.yaml
@@ -27,6 +27,7 @@ master:
     #   renewDeadline: 10s
     #   # this value has to be greater than 0
     #   retryPeriod: 2s
+    # nfdApiParallelism: 10
   ### <NFD-MASTER-CONF-END-DO-NOT-REMOVE>
   # The TCP port that nfd-master listens for incoming requests. Default: 8080
   port: 8080
@@ -39,6 +40,7 @@ master:
   enableTaints: false
   crdController: null
   featureRulesController: null
+  nfdApiParallelism: null
   deploymentAnnotations: {}
   replicaCount: 1
 
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -69,14 +69,15 @@ type Annotations map[string]string
 
 // NFDConfig contains the configuration settings of NfdMaster.
 type NFDConfig struct {
-	DenyLabelNs    utils.StringSetVal
-	ExtraLabelNs   utils.StringSetVal
-	LabelWhiteList utils.RegexpVal
-	NoPublish      bool
-	ResourceLabels utils.StringSetVal
-	EnableTaints   bool
-	ResyncPeriod   utils.DurationVal
-	LeaderElection LeaderElectionConfig
+	DenyLabelNs       utils.StringSetVal
+	ExtraLabelNs      utils.StringSetVal
+	LabelWhiteList    utils.RegexpVal
+	NoPublish         bool
+	ResourceLabels    utils.StringSetVal
+	EnableTaints      bool
+	ResyncPeriod      utils.DurationVal
+	LeaderElection    LeaderElectionConfig
+	NfdApiParallelism int
 }
 
 // LeaderElectionConfig contains the configuration for leader election
@@ -88,13 +89,14 @@ type LeaderElectionConfig struct {
 
 // ConfigOverrideArgs are args that override config file options
 type ConfigOverrideArgs struct {
-	DenyLabelNs    *utils.StringSetVal
-	ExtraLabelNs   *utils.StringSetVal
-	LabelWhiteList *utils.RegexpVal
-	ResourceLabels *utils.StringSetVal
-	EnableTaints   *bool
-	NoPublish      *bool
-	ResyncPeriod   *utils.DurationVal
+	DenyLabelNs       *utils.StringSetVal
+	ExtraLabelNs      *utils.StringSetVal
+	LabelWhiteList    *utils.RegexpVal
+	ResourceLabels    *utils.StringSetVal
+	EnableTaints      *bool
+	NoPublish         *bool
+	ResyncPeriod      *utils.DurationVal
+	NfdApiParallelism *int
 }
 
 // Args holds command line arguments
@@ -130,15 +132,16 @@ type NfdMaster interface {
 type nfdMaster struct {
 	*nfdController
 
-	args           Args
-	namespace      string
-	nodeName       string
-	configFilePath string
-	server         *grpc.Server
-	stop           chan struct{}
-	ready          chan bool
-	apihelper      apihelper.APIHelpers
-	kubeconfig     *restclient.Config
+	args            Args
+	namespace       string
+	nodeName        string
+	configFilePath  string
+	server          *grpc.Server
+	stop            chan struct{}
+	ready           chan bool
+	apihelper       apihelper.APIHelpers
+	kubeconfig      *restclient.Config
+	nodeUpdaterPool *nodeUpdaterPool
 	deniedNs
 	config *NFDConfig
 }
@@ -176,18 +179,22 @@ func NewNfdMaster(args *Args) (NfdMaster, error) {
 	if args.ConfigFile != "" {
 		nfd.configFilePath = filepath.Clean(args.ConfigFile)
 	}
+
+	nfd.nodeUpdaterPool = newNodeUpdaterPool(nfd)
+
 	return nfd, nil
 }
 
 func newDefaultConfig() *NFDConfig {
 	return &NFDConfig{
-		LabelWhiteList: utils.RegexpVal{Regexp: *regexp.MustCompile("")},
-		DenyLabelNs:    utils.StringSetVal{},
-		ExtraLabelNs:   utils.StringSetVal{},
-		NoPublish:      false,
-		ResourceLabels: utils.StringSetVal{},
-		EnableTaints:   false,
-		ResyncPeriod:   utils.DurationVal{Duration: time.Duration(1) * time.Hour},
+		LabelWhiteList:    utils.RegexpVal{Regexp: *regexp.MustCompile("")},
+		DenyLabelNs:       utils.StringSetVal{},
+		ExtraLabelNs:      utils.StringSetVal{},
+		NoPublish:         false,
+		NfdApiParallelism: 10,
+		ResourceLabels:    utils.StringSetVal{},
+		EnableTaints:      false,
+		ResyncPeriod:      utils.DurationVal{Duration: time.Duration(1) * time.Hour},
 		LeaderElection: LeaderElectionConfig{
 			LeaseDuration: utils.DurationVal{Duration: time.Duration(15) * time.Second},
 			RetryPeriod:   utils.DurationVal{Duration: time.Duration(2) * time.Second},
@@ -219,6 +226,8 @@ func (m *nfdMaster) Run() error {
 			return err
 		}
 	}
+
+	m.nodeUpdaterPool.start(m.config.NfdApiParallelism)
 
 	// Create watcher for config file
 	configWatch, err := utils.CreateFsWatcher(time.Second, m.configFilePath)
@@ -276,6 +285,10 @@ func (m *nfdMaster) Run() error {
 			if m.nfdController != nil && m.args.EnableNodeFeatureApi {
 				m.nfdController.updateAllNodesChan <- struct{}{}
 			}
+			// Restart the node updater pool
+			m.nodeUpdaterPool.stop()
+			m.nodeUpdaterPool.start(m.config.NfdApiParallelism)
+
 		case <-m.stop:
 			klog.InfoS("shutting down nfd-master")
 			return nil
@@ -359,10 +372,7 @@ func (m *nfdMaster) nfdAPIUpdateHandler() {
 		case nodeName := <-m.nfdController.updateOneNodeChan:
 			updateNodes[nodeName] = struct{}{}
 		case <-rateLimit:
-			// Check what we need to do
-			// TODO: we might want to update multiple nodes in parallel
 			errUpdateAll := false
-			errNodes := make(map[string]struct{})
 			if updateAll {
 				if err := m.nfdAPIUpdateAllNodes(); err != nil {
 					klog.ErrorS(err, "failed to update nodes")
@@ -370,16 +380,13 @@ func (m *nfdMaster) nfdAPIUpdateHandler() {
 				}
 			} else {
 				for nodeName := range updateNodes {
-					if err := m.nfdAPIUpdateOneNode(nodeName); err != nil {
-						klog.ErrorS(err, "failed to update node", "nodeName", nodeName)
-						errNodes[nodeName] = struct{}{}
-					}
+					m.nodeUpdaterPool.queue.Add(nodeName)
 				}
 			}
 
-			// Reset "work queue" and timer, will cause re-try if errors happened
+			// Reset "work queue" and timer
 			updateAll = errUpdateAll
-			updateNodes = errNodes
+			updateNodes = map[string]struct{}{}
 			rateLimit = time.After(time.Second)
 		}
 	}
@@ -392,6 +399,8 @@ func (m *nfdMaster) Stop() {
 	if m.nfdController != nil {
 		m.nfdController.stop()
 	}
+
+	m.nodeUpdaterPool.stop()
 
 	close(m.stop)
 }
@@ -645,9 +654,7 @@ func (m *nfdMaster) nfdAPIUpdateAllNodes() error {
 	}
 
 	for _, node := range nodes.Items {
-		if err := m.nfdAPIUpdateOneNode(node.Name); err != nil {
-			return err
-		}
+		m.nodeUpdaterPool.queue.Add(node.Name)
 	}
 
 	return nil
@@ -1160,6 +1167,13 @@ func (m *nfdMaster) configure(filepath string, overrides string) error {
 	}
 	if m.args.Overrides.ResyncPeriod != nil {
 		c.ResyncPeriod = *m.args.Overrides.ResyncPeriod
+	}
+	if m.args.Overrides.NfdApiParallelism != nil {
+		c.NfdApiParallelism = *m.args.Overrides.NfdApiParallelism
+	}
+
+	if c.NfdApiParallelism <= 0 {
+		return fmt.Errorf("the maximum number of concurrent labelers should be a non-zero positive number")
 	}
 
 	m.config = c
--- a/pkg/nfd-master/node-updater-pool.go
+++ b/pkg/nfd-master/node-updater-pool.go
@@ -1,0 +1,98 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nfdmaster
+
+import (
+	"sync"
+
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog/v2"
+)
+
+type nodeUpdaterPool struct {
+	queue workqueue.RateLimitingInterface
+	sync.Mutex
+
+	wg        sync.WaitGroup
+	nfdMaster *nfdMaster
+}
+
+func newNodeUpdaterPool(nfdMaster *nfdMaster) *nodeUpdaterPool {
+	return &nodeUpdaterPool{
+		nfdMaster: nfdMaster,
+		wg:        sync.WaitGroup{},
+	}
+}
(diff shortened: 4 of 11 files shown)


marquiz (commented):
Thanks @AhmedGrati for the PR.

I think we need to throttle the max parallelism (i.e. how many concurrent go routines are generated) somehow. Say we have a cluster with 2000 nodes we don't probably want the fire 2000 goroutines at the same time

/milestone v0.14.0

PiotrProkop:
Can we encorporate workqueue here? Have a flag than configures the number of consumers(which will update specific node) and we will just `workqueue.Add` nodes when we want to update them. 

We can use https://pkg.go.dev/k8s.io/client-go/util/workqueue for example.

marquiz:
> Can we encorporate workqueue here?

Yeah, for example something like this

AhmedGrati:
Hey @PiotrProkop, I've just come up with this design to include workqueue, and to have a goroutines pool responsible for labeling nodes.
NFD drawio
But at the same time @marquiz, suggested that we can use [SizedWaitGroup.](https://github.com/remeh/sizedwaitgroup)
I wanted to gather your input here, on what should we do next.
1. Keep it simple as is now, and add the SizedWaitGroup.
2. Implement a more sophisticated solution and add the `workqueue` package.

AhmedGrati:
ping @marquiz @PiotrProkop

PiotrProkop:
I think decoupling labeling into separate `workqueue` is better approach for the future. It would move labeling responsibility into separate component and we would have to just `Add` labeling request from the main loop.

On the other hand I know that implementing it is non trivial and I think `SizedWaitGroup` will also work but we would just have to make sure that we move all labeling into single `SizedWaitGroup`  or single `ErrGroup`.

@marquiz WDYT?

marquiz:
> I think decoupling labeling into separate `workqueue` is better approach for the future. It would move labeling responsibility into separate component and we would have to just `Add` labeling request from the main loop.
> 
> On the other hand I know that implementing it is non trivial and I think `SizedWaitGroup` will also work but we would just have to make sure that we move all labeling into single `SizedWaitGroup` or single `ErrGroup`.

I agree with @PiotrProkop here. Using workqueues would be more future proof and architecturally cleaner for the long run.

However, as @PiotrProkop said it's not trivial and I think lot more work (including tests) than the simple-and-stupid SizedWaitGroup. Using SizedWaitGroup should be safe and simple (__currently__) as there is only one place (and goroutine) from where we do labeling.

Thus, I'm fine with either approach. You can choose 😉

AhmedGrati:
Thanks for your input @marquiz @PiotrProkop, I'll implement it using `workqueue`.

AhmedGrati:
PTAL @marquiz @PiotrProkop

PiotrProkop:
> PTAL @marquiz @PiotrProkop

Instead of spinning workers on demand up to the limmit, you can precreate desired amount of workers and make them wait for `node` to be processed. You can take a look at this example: https://github.com/kubernetes/client-go/blob/master/examples/workqueue/main.go

AhmedGrati:
@PiotrProkop I updated it with changes inspired from the example the you provided.

pkg/nfd-master/nfd-master.go
```diff
@@ -162,6 +169,9 @@ func NewNfdMaster(args *Args) (NfdMaster, error) {
 	if args.ConfigFile != "" {
 		nfd.configFilePath = filepath.Clean(args.ConfigFile)
 	}
+	// create the workqueue
+	nfd.queue = workqueue.NewDelayingQueue()
```
PiotrProkop, line 173: why use `delaying` queue when you are not using `AddAfter`? It would be better to use standatd queue:
`workqueue.New()` or rate limited one and make it only retry key/node some specified amount of times as in example I have pasted earlier(to prevent infinite processing of one failing key).

pkg/nfd-master/nfd-master.go
```diff
@@ -262,6 +273,40 @@ func (m *nfdMaster) Run() error {
 	}
 }
 
+func (m *nfdMaster) handleLabelerError(err error, nodeName string) {
```
marquiz, line 276: nit: looks a bit strange that this is a separate method 🧐 How about just doing this error checking inline in `processNodeLabelRequest`

pkg/nfd-master/nfd-master.go
marquiz: Good point @PiotrProkop 👍 We don't want to infinitely retry, e.g. nodes that are deleted from the cluster

AhmedGrati: Yes makes sense thanks, do you suggest a max number of retries? @PiotrProkop @marquiz

pkg/nfd-master/nfd-master.go
```diff
@@ -57,6 +59,8 @@ import (
 	"sigs.k8s.io/node-feature-discovery/pkg/version"
 )
 
+const DefaulMaxNumberOfConcurrentLabelers = 200
```
PiotrProkop, line 62: I think 200 is kinda big number for default concurrent labelers, maybe we can set it to `10` ?

cmd/nfd-master/main.go
```diff
@@ -154,5 +156,8 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 	flagset.Var(overrides.ResyncPeriod, "resync-period",
 		"Specify the NFD API controller resync period."+
 			"It has an effect when the NodeFeature API has been enabled (with -enable-nodefeature-api).")
+	overrides.MaxNumberOfConcurrentLabelers = flagset.Int("max-number-of-concurrent-labelers", 200, "Defines the maximum number of goroutines responsible of labeling nodes. "+
```
PiotrProkop, line 159: This should be `flagset.Uint` instead of `Int` cause negative values are incorrect.

AhmedGrati: Yes 100% true, but if the user inputs for example `-5` it would result in having `18446744073709551611 ` for the `MaxNumberOfConcurrentLabelers`. I would suggest working with `math.Abs` to handle negative values. wdyt?

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -39,6 +40,7 @@ master:
   enableTaints: false
   crdController: null
   featureRulesController: null
+  maxNumberOfConcurrentLabelers: 200
```
PiotrProkop, line 43: s/200/10

AhmedGrati:
@marquiz @PiotrProkop PTAL when you have time.

pkg/nfd-master/nfd-master.go
```diff
@@ -284,6 +295,41 @@ func (m *nfdMaster) Run() error {
...
 
+func (m *nfdMaster) processNodeLabelRequest() bool {
+	nodeName, quit := m.queue.Get()
+	if quit {
+		return false
+	}
+	defer m.queue.Done(nodeName)
+	err := m.nfdAPIUpdateOneNode(nodeName.(string))
+
+	if err != nil {
```
PiotrProkop, line 306: you can inline err definition:
```suggestion
	if err := m.nfdAPIUpdateOneNode(nodeName.(string)); err != nil {
```

PiotrProkop (commented):
small nit otherwise LGTM, will wait for proper approval one @marquiz will review this PR.

pkg/nfd-master/nfd-master.go
```diff
@@ -284,6 +295,39 @@ func (m *nfdMaster) Run() error {
...
+			m.queue.AddRateLimited(nodeName)
+		} else {
+			klog.Errorf("error labeling node %v: %v", nodeName, err)
+		}
+	}
+	m.queue.Forget(nodeName)
+	return true
+}
+
+func (m *nfdMaster) runLabeler() {
```
marquiz, line 316: I'd suggest to rename this to not being about labeling but about updating nodes `runNodeUpdater` or smth

pkg/nfd-master/nfd-master.go
```diff
@@ -284,6 +295,39 @@ func (m *nfdMaster) Run() error {
...
+	for m.processNodeLabelRequest() {
+	}
+}
+
+func (m *nfdMaster) runNodeLabelersPool() {
```
marquiz, line 321: Also, I'd suggest to rename this, e.g. to `runNodeUpdaterPool`

pkg/nfd-master/nfd-master.go
```diff
@@ -284,6 +295,39 @@ func (m *nfdMaster) Run() error {
...
+	klog.Info("Starting the NFD master labelers pool")
+	defer m.queue.ShutDown()
+	for i := 0; i < m.config.MaxNumberOfConcurrentLabelers; i++ {
+		go wait.Until(m.runLabeler, time.Second, m.stop)
```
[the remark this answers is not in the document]
marquiz: Actually, taking a closer look, why do we use `wait.Until()` in the first place as we don't utilize the periodic calling of the function, for example?

Wouldn't it be simpler to do just

```golang
		go runLabeler(stopChan)
```

And then handle the stop channedl in runLabeler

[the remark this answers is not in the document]
AhmedGrati: Makes sense, good point 👌

AhmedGrati:
@marquiz Thank you for the review. It's updated now.

AhmedGrati:
@marquiz any other comments on this?

pkg/nfd-master/nfd-master.go
```diff
@@ -277,13 +287,67 @@ func (m *nfdMaster) Run() error {
...
+
+func (m *nfdMaster) runNodeUpdater() {
+	for {
+		select {
+		case <-m.nodeUpdaterPoolStop:
+			return
+		default:
+			m.processNodeLabelRequest()
+		}
+	}
```
marquiz, line 330: Now that I'm looking at this I think it doesn't work at all: processNodeLabelRequest blocks and in practice the stop signal is probably never received (in particular as m.nodeUpdaterPoolStop is re-initialized before we get to the select again) -> growing number of leaking goroutines on every re-configuration event -> bad

I think the right thing to do is to shutdown the queue in runNodeUpdaterPool so that all blocking get operations get cancelled.
```suggestion
	for m.processNodeLabelRequest() {
	}
```

pkg/nfd-master/nfd-master.go
```diff
@@ -277,13 +287,67 @@ func (m *nfdMaster) Run() error {
...
+			go m.runNodeUpdaterPool()
+
 		case <-m.stop:
 			klog.Infof("shutting down nfd-master")
 			return nil
 		}
 	}
 }
 
+func (m *nfdMaster) processNodeLabelRequest() bool {
```
marquiz, line 301: Safer to use local variable as we operate in different  goroutines and the queue is our "sync channel"

```suggestion
func (m *nfdMaster) processNodeLabelRequest(queue workqueue.RateLimitingInterface) bool {
```

pkg/nfd-master/nfd-master.go
```diff
@@ -277,13 +287,67 @@ func (m *nfdMaster) Run() error {
...
+	for i := 0; i < m.config.NfdApiParallelism; i++ {
+		go m.runNodeUpdater()
+	}
+	<-m.nodeUpdaterPoolStop
+	klog.Info("Stopping the NFD master node updater pool")
+}
+
+func (m *nfdMaster) stopNodeUpdaterPool() {
+	// notifies all node updaters goroutines to stop and shut downs the node update pool
+	close(m.nodeUpdaterPoolStop)
```
marquiz, line 348: Hmm, I think ideally we should wait for the updater goroutines to finish (waitgroup). To be on the safe side, even if an unlikely race. We might e.g. have in-flight retries that get intermixed with node updates from the new pool after re-configuration.

One idea: `runNodeUpdaterPool()` could be changed to non-blocking `startNodeUpdaterPool` that just creates the worker goroutines and adds them to a waitgroup and returns. Then `stopNodeUpdaterPool` would signal these worker goroutines to stop and wait for them to finish (with the waitgroup). 

Also, I don't think we actually need the stop channel for anything. 

```suggestion
	m.queue.ShutDown()
```

marquiz (commented):
Some more comments. I think there's still some work to do on the goroutine management front

marquiz (commented):
Thanks for the update. A few more comments. @PiotrProkop could you also take a look.

Also @AhmedGrati you added some benchmarks. Have you run them? Do you have any data? Results could be added into the commit message as informational

AhmedGrati:
Thanks for the review @marquiz. 
For the benchmarking, in the old implementation, it updates `1000` nodes (ofc we are mocking the communication with k8s API) in a duration that goes between `500 micro-seconds` and `650 micro-seconds`. In the current implementation, the duration is between `300 micro-seconds` and `450 micro-seconds`. So I don't know whether It's worth mentioning in the description of the PR or not.
However, I still believe that macro-benchmarking would produce more realistic results regarding this improvement. WDYT?

marquiz (commented):
Nice work @AhmedGrati. I think we could merge this already.

Although now that I took another comprehensive look at this I think it would make sense to separate the node updater pool into a separate type. I think it would make it cleaner, architecturally, also clearly separating the data.

```golang
type nodeUpdaterPool struct {
	queue                workqueue.RateLimitingInterface
	wg                   sync.WaitGroup
        nfdMaster            nfdMaster
}

func (u *nodeUpdaterPool) runNodeUpdater()

func (u *nodeUpdaterPool) start(parallelism int)

func (u *nodeUpdaterPool) stop()
```

/assign @PiotrProkop

marquiz (commented):
Nice work @AhmedGrati. I think we could merge this already.

Although now that I took another comprehensive look at this I think it would make sense to separate the node updater pool into a separate type. I think it would make it cleaner, architecturally, also clearly separating the data.

```golang
type nodeUpdaterPool struct {
	queue                workqueue.RateLimitingInterface
	wg                   sync.WaitGroup
        nfdMaster            nfdMaster
}

func (u *nodeUpdaterPool) runNodeUpdater()

func (u *nodeUpdaterPool) start(parallelism int)

func (u *nodeUpdaterPool) stop()
```

In `start()` you could also guard against calling it twice (without calling stop() in between)

/assign @PiotrProkop

AhmedGrati:
Nice suggestion, I will work on that.
/hold

pkg/nfd-master/node-updater-pool.go
```diff
@@ -0,0 +1,74 @@
...
+type nodeUpdaterPool struct {
+	queue     workqueue.RateLimitingInterface
+	wg        sync.WaitGroup
+	nfdMaster nfdMaster
+}
+
+var lock = &sync.Mutex{}
+var instance *nodeUpdaterPool
+
+func getNodeUpdaterPool(nfdMaster nfdMaster) *nodeUpdaterPool {
```
marquiz, line 19: This singleton approach looks a bit weird as you give nfdMaster as an argument but that argument is totally omitted in subsequent calls.

How about changing this to

```suggestion
func newNodeUpdaterPool(nfdMaster nfdMaster) *nodeUpdaterPool
```

and then calling this in NewNfdMaster and storing the instance in the nfdMaster struct?

PiotrProkop: I agree, It would be much more readable, since the nodeUpdaterPool is only used from within `nfdMaster`.

pkg/nfd-master/node-updater-pool.go
```diff
@@ -0,0 +1,74 @@
...
+
+func (u *nodeUpdaterPool) processNodeLabelRequest(queue workqueue.RateLimitingInterface) bool {
+	nodeName, quit := queue.Get()
+	if quit {
+		return false
+	}
+
+	defer queue.Done(nodeName)
+
+	if err := u.nfdMaster.nfdAPIUpdateOneNode(nodeName.(string)); err != nil {
```
PiotrProkop, line 41: and if this is the only place where `nfdAPIUpdateOneNode` is called you can just remove this func from `nfdMaster` and make it `nodeUpdaterPoll` func, this way you don't need to pass `nfdMaster` to `nodeUpdaterPool` but prolly just Kubernetes client.

pkg/nfd-master/node-updater-pool.go
AhmedGrati: Wouldn't that result in a circular dependency?

AhmedGrati: `nfdMaster contains nodeUpdaterPool` and `nodeUpdaterPool contains nfdMaster`.

marquiz: Think about it like doubly-linked list

pkg/nfd-master/node-updater-pool.go
```diff
@@ -0,0 +1,86 @@
...
+	"k8s.io/klog/v2"
+)
+
+type nodeUpdaterPool struct {
+	queue     workqueue.RateLimitingInterface
+	wg        sync.WaitGroup
+	nfdMaster *nfdMaster
+	started   bool
+	mutex     sync.Mutex
+}
```
marquiz, line 16: One suggestion to potentially beautify a bit

```suggestion
type nodeUpdaterPool struct {
	sync.Mutex
	workqueue.RateLimitingInterface
	
	wg        sync.WaitGroup
	nfdMaster *nfdMaster
	started   bool
}
```

With this you could do simpler `u.Lock()` instead of `u.mutex.Lock()` and e.g. `pool.Add(nodeName)` instead of `pool.queue.NodeName()`. WDYT?

pkg/nfd-master/node-updater-pool.go
```diff
@@ -0,0 +1,86 @@
...
+	for u.processNodeLabelRequest(queue) {
+	}
+	u.wg.Done()
+}
+
+func (u *nodeUpdaterPool) start(parallelism int) {
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
+
+	if u.started {
```
marquiz, line 57: We could do away with the started bool with smth like

```suggestion
	if u.queue != nil && !u.queue.ShuttingDown() {
```

pkg/nfd-master/node-updater-pool.go
```diff
@@ -0,0 +1,86 @@
...
+		klog.InfoS("The NFD master node updater pool is already running.")
+		return
+	}
+
+	klog.InfoS("Starting the NFD master node updater pool", "number of parallel updaters", parallelism)
```
marquiz, line 62: Use simple "fieldName" type keys for the log data. Also start with lower-case letter

```suggestion
	klog.InfoS("starting the NFD master node updater pool", "parallelism", parallelism)
```

pkg/nfd-master/node-updater-pool.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+}
+
+func (u *nodeUpdaterPool) stop() {
+
+	u.mutex.Lock()
+	defer u.mutex.Unlock()
+
+	if !u.started {
+		klog.InfoS("The NFD master node updater pool is not running.")
```
marquiz, line 78: nit: we've tried to maintain a convention to start log messages with lower-case

```suggestion
		klog.InfoS("the NFD master node updater pool is not running.")
```

ditto other messages, too...

pkg/nfd-master/node-updater-pool.go
AhmedGrati:
```go
	u.queue.ShutDown()
	u.wg.Wait()
```
Would be transformed to:
```go
	u.ShutDown()
	u.wg.Wait()
```
Isn't it misleading to have `u.Shutdown()` and then `u.wg.Wait()`?

AhmedGrati: It gives the intention that we're shutting down the pool and then waiting for the waitgroup, where in reality we're shutting down only the queue.

marquiz: Yeah, maybe it's cleaner to have queue as a separate field

pkg/nfd-master/node-updater-pool.go
```diff
@@ -0,0 +1,83 @@
...
+	u.queue = workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
+
+	for i := 0; i < parallelism; i++ {
+		u.wg.Add(1)
+		go u.runNodeUpdater(u.queue)
+	}
+}
+
+func (u *nodeUpdaterPool) stop() {
+
```
marquiz, line 71:
```suggestion
```

pkg/nfd-master/node-updater-pool.go
```diff
@@ -0,0 +1,83 @@
...
+	u.Lock()
+	defer u.Unlock()
+
+	if u.queue != nil && u.queue.ShuttingDown() {
```
marquiz, line 75:
```suggestion
	if u.queue == nil || u.queue.ShuttingDown() {
```

AhmedGrati:
@marquiz any idea why is this failing?

pkg/nfd-master/nfd-master.go
```diff
@@ -1161,6 +1169,18 @@ func (m *nfdMaster) configure(filepath string, overrides string) error {
...
+	} else {
+		// the default value of the NfdApiParallelism should not be 0
+		if c.NfdApiParallelism == 0 {
+			c.NfdApiParallelism = DefaulMaxNumberOfConcurrentLabelers
+		}
+	}
+
+	if c.NfdApiParallelism < 0 {
+		return fmt.Errorf("the maximum number of concurrent labelers should be a positive number")
+	}
```
marquiz, line 1183: I think we could simplify and remove the (wrongly named) const alltogether by just adding the default (10) to `newDefaultConfig()` and then banning zero.

```suggestion
	}
	if c.NfdApiParallelism <= 0 {
		return fmt.Errorf("the maximum number of concurrent labelers should be a non-zero positive number")
	}
```

WDYT?

AhmedGrati: Yes that's cleaner.

pkg/nfd-master/nfd-master.go
```diff
@@ -58,6 +58,8 @@ import (
 	"sigs.k8s.io/node-feature-discovery/pkg/version"
 )
 
+const DefaulMaxNumberOfConcurrentLabelers = 10
+
```
marquiz, line 62:
```suggestion
```

marquiz:
@AhmedGrati also one with: could you clean up the commit message 😇

pkg/nfd-master/node-updater-pool.go
marquiz, line 1: Just realized that this file is missing the license/copyright header. Could you add that one?

AhmedGrati: Done

pkg/nfd-master/node-updater-pool.go
```diff
@@ -0,0 +1,98 @@
+/*
+Copyright 2019-2021 The Kubernetes Authors.
```
marquiz, line 2: nit 🙈 put 2023 here
```suggestion
Copyright 2023 The Kubernetes Authors.
```

marquiz (approved):
Good job and thanks for the persistence on this @AhmedGrati. I don't think I have anything to add for now -> approved

@PiotrProkop could you take look, with fresh eyes?

marquiz (approved):
Good job and thanks for the persistence on this @AhmedGrati. I don't think I have anything to add for now -> approved

@PiotrProkop could you take look, with fresh eyes?

pkg/nfd-master/node-updater-pool.go
PiotrProkop: Ok, we should create issue to track this work.

PiotrProkop (commented):
/lgtm

PR #1139, opened 2023-04-15
feat: add master resync period configurability

### LINKED ISSUE #1124: nfd-master: config option for NFD API resync period

Add a config option for setting the NFD API controller resync period. The resync causes nfd-master to replay all NodeFeature and
NodeFeatureRule objects, being effectively a "big hammer" to reset everything. Thus, this also causes some load on the cluster, nfd-master replaying all NFD API objects and querying all nodes from the API server. Making this configurable would enable some users to better adapt nfd to their needs, e.g. by making it possible to fully disable resync.

> NOTE: only has an effect when the NodeFeature API has been enabled (with -enable-nodefeature-api).

**Why is this needed**:

User configurability. Makes it possible for users to tune NFD to their usage scenarios.

AhmedGrati:
Resolves #1124.

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	"k8s.io/klog/v2"
 
@@ -67,6 +68,8 @@ func main() {
 			args.Overrides.EnableTaints = overrides.EnableTaints
 		case "no-publish":
 			args.Overrides.NoPublish = overrides.NoPublish
+		case "-resync-period":
+			args.Overrides.ResyncPeriod = overrides.ResyncPeriod
 		}
 	})
 
@@ -131,6 +134,7 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 		DenyLabelNs:    &utils.StringSetVal{},
 		ExtraLabelNs:   &utils.StringSetVal{},
 		ResourceLabels: &utils.StringSetVal{},
+		ResyncPeriod:   &utils.DurationVal{Duration: time.Duration(1) * time.Hour},
 	}
 	flagset.Var(overrides.ExtraLabelNs, "extra-label-ns",
 		"Comma separated list of allowed extra label namespaces")
@@ -145,6 +149,8 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 		"Comma separated list of denied label namespaces")
 	flagset.Var(overrides.ResourceLabels, "resource-labels",
 		"Comma separated list of labels to be exposed as extended resources. DEPRECATED: use NodeFeatureRule objects instead")
-
+	flagset.Var(overrides.ResyncPeriod, "resync-period",
+		"Specify the NFD API controller resync period."+
+			"It has an effect when the NodeFeature API has been enabled (with -enable-nodefeature-api).")
 	return args, overrides
 }
--- a/docs/reference/master-configuration-reference.md
+++ b/docs/reference/master-configuration-reference.md
@@ -113,3 +113,20 @@ Example:
 ```yaml
 labelWhiteList: "foo"
 ```
+
+### resyncPeriod
+
+The `resyncPeriod` option specifies the NFD API controller resync period.
+The resync means nfd-master replaying all NodeFeature and NodeFeatureRule objects,
+thus effectively re-syncing all nodes in the cluster (i.e. ensuring labels, annotations,
+extended resources and taints are in place).
+Only has effect when the [NodeFeature](../usage/custom-resources.md#nodefeature)
+CRD API has been enabled with [`-enable-nodefeature-api`](master-commandline-reference.md#-enable-nodefeature-api).
+
+Default: 1 hour.
+
+Example:
+
+```yaml
+resyncPeriod=2h
+```
--- a/pkg/nfd-master/nfd-api-controller.go
+++ b/pkg/nfd-master/nfd-api-controller.go
@@ -41,7 +41,12 @@ type nfdController struct {
 	updateOneNodeChan  chan string
 }
 
-func newNfdController(config *restclient.Config, disableNodeFeature bool) (*nfdController, error) {
+type nfdApiControllerOptions struct {
+	disableNodeFeature bool
+	resyncPeriod       time.Duration
+}
+
+func newNfdController(config *restclient.Config, nfdApiControllerOptions nfdApiControllerOptions) (*nfdController, error) {
 	c := &nfdController{
 		stopChan:           make(chan struct{}, 1),
 		updateAllNodesChan: make(chan struct{}, 1),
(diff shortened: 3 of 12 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -96,6 +96,7 @@ type Args struct {
 	Prune                bool
 	VerifyNodeName       bool
 	Options              string
+	ResyncPeriod         int
```
PiotrProkop, line 99: this should be `time.Duration` instead of `int`

cmd/nfd-master/main.go
```diff
@@ -125,6 +125,9 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 	flagset.StringVar(&args.Options, "options", "",
 		"Specify config options from command line. Config options are specified "+
 			"in the same format as in the config file (i.e. json or yaml). These options")
+	flagset.DurationVar(&args.ResyncPeriod, "resync-period", 60,
```
PiotrProkop, line 128: the default value should be changed to `time.Duration(1)*time.Hour` and description should be changed that it accepts duration string

AhmedGrati: Yeah I know, I actually questioned that in the issue section, but since I didn't get any response, I used `Moment` as the resync period.

cmd/nfd-master/main.go
```diff
@@ -125,6 +125,9 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 	flagset.StringVar(&args.Options, "options", "",
 		"Specify config options from command line. Config options are specified "+
 			"in the same format as in the config file (i.e. json or yaml). These options")
+	flagset.DurationVar(&args.ResyncPeriod, "resync-period", 1,
```
PiotrProkop, line 128: if you specify `1` as default value it'll means that `ResyncPeriod` is set to 1 nanosecond, which is not what we want.  If you want to specify an hour you should pass `time.Hour`.

AhmedGrati: Yeah I know, it's actually WIP, I'll ping you once I finish.

AhmedGrati:
yo guys @marquiz @PiotrProkop, I just found out that the `sigs.k8s.io/yaml` package cannot unmarshal `time.Duration`. You can test this example here:
```go
package main

import (
	"fmt"
	"time"

	"sigs.k8s.io/yaml"
)

type Config struct {
	Timeout time.Duration `yaml:"timeout"`
}

func main() {
	var cfg Config

	b := []byte(`timeout: 60s`)
	yaml.Unmarshal(b, &cfg)
	fmt.Printf("timeout: %dm", int(cfg.Timeout.Minutes()))
}
```
Please correct me if I'm missing something.
As an alternative, we can go with using integers to indicate the number of hours, or maybe changing the package that we are using to `gopkg.in/yaml.v3`, which actually unmarshal the `time.Duration` type.
Curious to hear your thoughts.

marquiz:
> just found out that the `sigs.k8s.io/yaml` package cannot unmarshal `time.Duration`

Ah, yes. In NFD we have `duration` type in `pkg/nfd-worker` that does unmarhalling. So I'd suggest we just move that to `pkg/utils/flags.go` and use it from there, for now



> As an alternative, we can go with using integers to indicate the number of hours, or maybe changing the package that we are using to `gopkg.in/yaml.v3`, which actually unmarshal the `time.Duration` type.

Hmm, interesting. We could do this, but do it carefully in a separate PR. I.e. understand the differences between the two libraries and what might break

pkg/nfd-worker/nfd-worker.go
```diff
@@ -169,7 +165,7 @@ func newDefaultConfig() *NFDConfig {
 	return &NFDConfig{
 		Core: coreConfig{
 			LabelWhiteList: utils.RegexpVal{Regexp: *regexp.MustCompile("")},
-			SleepInterval:  duration{60 * time.Second},
+			SleepInterval:  utils.DurationVal{Duration: 60 * time.Second},
```
marquiz, line 168: You could leave out the field name (as a sidenote the Regexp above is also unnecessary)

```suggestion
			SleepInterval:  utils.DurationVal{60 * time.Second},
```

pkg/utils/flags.go
```diff
@@ -173,3 +174,46 @@ func NewKlogFlagVal(f *flag.Flag) *KlogFlagVal {
...
+	return nil
+}
+
+// time.Duration implements the flag.Value interface
+func (d *DurationVal) DurationVal() time.Duration {
+	if d == nil {
+		return time.Hour * time.Duration(1)
+	}
+	return d.Duration
+}
```
marquiz, line 209: What's the purpose of this?

```suggestion
```

docs/reference/master-configuration-reference.md
```diff
@@ -113,3 +113,16 @@ Example:
 ```yaml
 labelWhiteList: "foo"
 ```
+
+### resyncPeriod
+
+The `resyncPeriod` option specifies the NFD API controller resync period
+and it's expressed in hours.
```
marquiz, line 120: Not necessarily hours :) Also, I'd suggest to give a bit more detailed description what this actually does

```suggestion
The `resyncPeriod` option specifies the NFD API controller resync period. The resync means nfd-master replaying all NodeFeature and NodeFeatureRule objects, thus effectively re-syncing all nodes in the cluster (i.e. ensuring labels, annotations, extended resources and taints are in place). Only has effect when the [NodeFeature](../usage/custom-resources.md#nodefeature) CRD API has been enabled with [`-enable-nodefeature-api`](master-commandline-reference.md#-enable-nodefeature-api).
```

marquiz (commented):
Thanks @AhmedGrati for the update! A few small comments below

pkg/nfd-worker/nfd-worker.go
AhmedGrati: It made my linter angry xd saying that `DurationVal struct literal uses unkeyed fields`

marquiz: K, I was wrong then. Let's leave it as it is. Might be the reason that the regexp is also like that 😅

pkg/utils/flags.go
```diff
@@ -173,3 +174,38 @@ func NewKlogFlagVal(f *flag.Flag) *KlogFlagVal {
 type boolFlag interface {
 	IsBoolFlag() bool
 }
+
+type DurationVal struct {
```
marquiz, line 178: nit: could you add a "docstring" comment for this new exported type, e.g. like

```suggestion
// DurationVal is a wrapper for handling time.Duration as a command line flag
type DurationVal struct {
```

pkg/nfd-master/nfd-api-controller.go
```diff
@@ -41,19 +41,23 @@ type nfdController struct {
 	updateOneNodeChan  chan string
 }
 
-func newNfdController(config *restclient.Config, disableNodeFeature bool) (*nfdController, error) {
+type NfdApiControllerOptions struct {
```
marquiz, line 44: nit: we don't need to export this (especially because `newNfdController()` is not exported:

```suggestion
type nfdApiControllerOptions struct {
```

marquiz (commented):
Two tiny nits but after that I think I'm good 😊

PiotrProkop:
Looks good to me, If you can apply @marquiz suggestions I'll LGTM 😄 Good job!

marquiz (approved):
Nice addition to NFD @AhmedGrati, big thanks for working on this.  I don't have anything to add 😎

PR #1132, opened 2023-04-14
[release-0.6] docs: mark release as unsupported

marquiz:
Refs #1131

PR #1141, opened 2023-04-17
[release-0.7] docs: mark release v0.7 as unsupported

marquiz:
Copied what #1132 did. Outcome will be similar (https://kubernetes-sigs.github.io/node-feature-discovery/v0.6)

### DIFF
--- a/docs/_includes/class/content-wrap.liquid
+++ b/docs/_includes/class/content-wrap.liquid
@@ -1,0 +1,26 @@
+<div class="content-wrap">
+    <div class="header d-flex flex-justify-between p-2 hide-lg hide-xl" aria-label="top navigation">
+        <button id="toggle" class="btn-octicon p-2 m-0 text-white" type="button">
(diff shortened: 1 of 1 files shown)


marquiz:
/retitle [release-0.7] docs: mark release v0.7 as unsupported

source/cpu/power_stub.go
PiotrProkop, line 1: is this change intentional?

source/cpu/power_stub.go
```diff
@@ -1,3 +1,4 @@
+//go:build !amd64
```
ArangoGutierrez, line 1: what does this has to do with an html change?

ArangoGutierrez (changes requested):

source/cpu/power_stub.go
marquiz: Nothing 🤨 Thanks for noting, fixed. I don't understand what gofmt magic I managed to do get this...

Issue #1131, opened 2023-04-14
Spell out release support policy


marquiz:
**What would you like to be added**:

Decide on and write out a release support policy. E.g. that we always support (provide bug fixes and updates) for two releases, N and N-1. In practice, we have currently only supported one release, ceasing patch releases of N when N+1 is out.

**Why is this needed**:

Make it clear for both users and the NFD team what to expect.

marquiz:
Related to this #1132 adds a red banner in v0.6 documentation that this release is not supported anymore. If that looks ok I think we could (and should) safely copy that banner to other older versions at least up to v0.9

fmuyassarov:
> think we could (and should) safely copy that banner to other older versions at least up to v0.9

+1

fmuyassarov:
This would be good addition.

ArangoGutierrez:
Agree, a support matrix, displaying Go build version, Kubernetes version(go.mod) and git commit, and a big disclaimer of the N-1 support 

/lgtm

marquiz:
Related #1141, #1142, #1143 and #1144 add "old non-supported version" banner to v0.7 to v0.10

marquiz:
> Agree, a support matrix, displaying Go build version, Kubernetes version(go.mod) and git commit, and a big disclaimer of the N-1 support

Any thoughts on "N-1"? Makes sense, is enough?

PiotrProkop:
> > Agree, a support matrix, displaying Go build version, Kubernetes version(go.mod) and git commit, and a big disclaimer of the N-1 support
> 
> Any thoughts on "N-1"? Makes sense, is enough?

Is there any official guideline regarding release schedule? I can see that new major version of NFD is released roughly twice a year. So "N-1" means 1 year of support, which seems enough for me 😄

ArangoGutierrez:
I mean.... unless there's a BIG change on the kube dependency(kube api mostly) , we could in theory do N-2, buuuuuut...... I am happy with N-1

marquiz:
> Is there any official guideline regarding release schedule?

Not really. The team has been so small and stretched and contributions so sporadic that I feel a set-in-stone release schedule hasn't really played out. We could decide/say something about this, too. 

> I can see that new major version of NFD is released roughly twice a year. So "N-1" means 1 year of support, which seems enough for me 😄

I hope we could do more than two releases a year, to get new features to users quicker. 3 releases at least would be nice IMO

[...]

PR #1150, opened 2023-04-17
[release-0.11] docs: mark release v0.11 as unsupported

marquiz:


marquiz:
Following the discussion in #1131 it looks like we're find marking v0.11 as unsupported after v0.13 is out. It's over a year old
/hold

PR #1155, opened 2023-04-18
deps: Update kubernetes to v1.27.1

marquiz:
In addition to just updating deps in go.mod (and go.sum) we needed two additional patches:

- Update mocked implementation of k8s.io/kubelet/pkg/apis/podresources/v1.PodResourcesListerClient. The mocked implementation is moved to a separate "mocks" subpackage as it's for an external interface.

  This patch also adds code for auto-generation for the mocked interface.
- test/e2e: adapt tests to updates in k8s e2e-framework


ArangoGutierrez (approved):
/lgtm

PR #1160, opened 2023-04-19
nfd-master: fix -prune

marquiz:
- configure before prune
  Otherwise prune will crash because of uninitialized configuration.
- deployment/kustomize: add master config to prune overlay
  Otherwise pods error out with failed mount of nfd-master-conf ConfigMap.

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -185,6 +185,11 @@ func (m *nfdMaster) Run() error {
 	klog.Infof("NodeName: %q", m.nodeName)
 	klog.Infof("Kubernetes namespace: %q", m.namespace)
 
+	// Read initial configuration
+	if err := m.configure(m.configFilePath, m.args.Options); err != nil {
+		return err
+	}
+
 	if m.args.Prune {
 		return m.prune()
 	}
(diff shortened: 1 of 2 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -185,6 +185,15 @@ func (m *nfdMaster) Run() error {
 	klog.Infof("NodeName: %q", m.nodeName)
 	klog.Infof("Kubernetes namespace: %q", m.namespace)
 
+	// Create watcher for config file and read initial configuration
+	configWatch, err := utils.CreateFsWatcher(time.Second, m.configFilePath)
```
PiotrProkop, line 189: do we need to setup `configWatch` for prune to work?

ArangoGutierrez (approved):
/lgtm
there are comments
/hold

pkg/nfd-master/nfd-master.go
marquiz: Good point @PiotrProkop, we don't need it as `-prune` is sub-command like option which just cleans up the nodes and exits right away. Wouldn't do much harm to set it but sort of "beauty defect" -> fixed it

marquiz:
ping @ArangoGutierrez

PR #1166, opened 2023-04-21
nfd-master: don't create emtpy annotations

marquiz:
Make the nfd.node.kubernetes.io/feature-labels and nfd.node.kubernetes.io/extended-resources annotations behave similary to the taints annotation: only create the annotations if some labels or extended resources are created.


ArangoGutierrez (approved):
/lgtm

PR #1167, opened 2023-04-21
nfd-topology-updater: fix wrong kubelet_internal_checkpoint path and compare basename to full path

PiotrProkop:
This PR fixes 2 problems with kubeletnotifier:

- `kubelet_internal_checkpoint` file is in `/var/lib/kubelet/device-plugins` (as defined [here](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kubelet/pkg/apis/deviceplugin/v1beta1/constants.go#L30) and fsWatcher doesn't watch dirs recursively
- `e.Name` returned from fsWatcher events is a full path not a basename 

We really need tests for this feature 😄

### DIFF
--- a/pkg/nfd-topology-updater/kubeletnotifier/kubeletnotifier.go
+++ b/pkg/nfd-topology-updater/kubeletnotifier/kubeletnotifier.go
@@ -18,6 +18,7 @@ package kubeletnotifier
 
 import (
 	"fmt"
+	"path"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -31,6 +32,8 @@ type EventType string
 const (
 	IntervalBased EventType = "intervalBased"
 	FSUpdate      EventType = "fsUpdate"
+
+	devicePluginsDirName = "device-plugins"
 )
 
 var stateFiles = sets.NewString(
@@ -51,7 +54,8 @@ type Info struct {
 }
 
 func New(sleepInterval time.Duration, dest chan<- Info, kubeletStateDir string) (*Notifier, error) {
-	ch, err := createFSWatcherEvent([]string{kubeletStateDir})
+	devicePluginsDir := path.Join(kubeletStateDir, devicePluginsDirName)
+	ch, err := createFSWatcherEvent([]string{kubeletStateDir, devicePluginsDir})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


marquiz (approved):
Thanks @PiotrProkop. This looks legit to me 👍

> We really need tests for this feature

Yes. I've also been bitten by missing tests today elsewhere in the codebase...

marquiz:
@PiotrProkop one suggestion still: could you add the detailed description you have in the PR into the commit message as well. With that you could shorten the commit message title a bit like `nfd-topology-updater: fix kubelet state file notifier` or smth

pkg/nfd-topology-updater/kubeletnotifier/kubeletnotifier.go
```diff
@@ -31,6 +32,8 @@ type EventType string
 const (
 	IntervalBased EventType = "intervalBased"
 	FSUpdate      EventType = "fsUpdate"
+
+	devicePluginsDirName = "device-plugins"
```
Tal-or, line 36: Maybe it is better to change Kubelet-state-dir to a [slice instead of string](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/cmd/nfd-topology-updater/main.go#L149) and specifies the device-plugins as part of the configuration. 
Then in the configuration we'll specify
 `--kubelet-state-dir=/var/lib/kubelet --kubelet-state-dir=/var/lib/kubelet/device-plugins`

Tal-or:
> We really need tests for this feature

We have e2e for this feature here: https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/test/e2e/topology_updater_test.go#L275

pkg/nfd-topology-updater/kubeletnotifier/kubeletnotifier.go
PiotrProkop: I like the idea of allowing passing more than one dir here but `kubelet_checkpoint_internal` is hardcoded as a default file to watch and we are failing in doing it since we setup watchers for wrong directory. We should either fix what we have right now or allow flexibility in specifying `files` to watch not dirs :)

Tal-or: Switching to files sounds good to me but maybe it deserves a separate PR and the current implementation and fix are good enough for this pr

ArangoGutierrez:
> > @PiotrProkop one suggestion still: could you add the detailed description you have in the PR into the commit message as well. With that you could shorten the commit message title a bit like `nfd-topology-updater: fix kubelet state file notifier` or smth
> 
> commit message changed

Requested: 
`nfd-topology-updater: fix kubelet state file notifier`
Has
`nfd-topology-updater: nfd-topology-updater: fix kubelet state file notifier`

ArangoGutierrez (approved):
Git commit now looks as requested 
/lgtm

PiotrProkop:
> > > @PiotrProkop one suggestion still: could you add the detailed description you have in the PR into the commit message as well. With that you could shorten the commit message title a bit like `nfd-topology-updater: fix kubelet state file notifier` or smth
> > 
> > 
> > commit message changed
> 
> Requested: `nfd-topology-updater: fix kubelet state file notifier` Has `nfd-topology-updater: nfd-topology-updater: fix kubelet state file notifier`

typo, thanks for spotting

PR #1198, opened 2023-04-28
nfd-master: refactor api-controller object handling

marquiz:
Split out resolving of node name (of the node to be updated) into a separate function. Makes it possible to add unit tests. Also. do unconditional type casting in the handler functions – that shouldn't fail unless there is a really serious internal inconsistency in the codebase so it should be ok to panic.


fmuyassarov (commented):
/lgtm

PR #1200, opened 2023-04-28
codecov: drop required minimum coverage ratio of a commit to 0%

fmuyassarov:
not to block currently open PRs from getting merged and deal with target setting separately.

PR #1201, opened 2023-04-28
codecov: drop required minimum coverage ratio at patch level

fmuyassarov:
previously in https://github.com/kubernetes-sigs/node-feature-discovery/pull/1200 we added similar to project and this is doing the same but for the patch level.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,3 +7,6 @@ coverage:
     project:
       default:
         target: 0%
+    patch:
+      default:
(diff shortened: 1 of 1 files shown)


marquiz (approved):
Humpty-dumpty-lallallaa, hope everything's unbroken...
/lgtm

fmuyassarov:
thanks @marquiz for quick reviews :)

PR #1202, opened 2023-04-28
nfd-master: refactor filtering of labels, taints and ERs

marquiz:
Simplify code a bit and get more consistent error messages (in addition to
fixing some of those).

Also preparation for dynamic labels values (that `@` notation currently supported for extended resources).

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -519,22 +519,26 @@ func filterTaints(taints []corev1.Taint) []corev1.Taint {
 	outTaints := []corev1.Taint{}
 
 	for _, taint := range taints {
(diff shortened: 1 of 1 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -510,6 +495,26 @@ func (m *nfdMaster) filterFeatureLabels(labels Labels) (Labels, ExtendedResource
 	return outLabels, extendedResources
 }
 
+func (m *nfdMaster) filterFeatureLabel(name string) error {
```
fmuyassarov, line 498: I would prefer to name the argument `label` instead of `name`, but that's not a big deal.

fmuyassarov (commented):
/lgtm

PR #1204, opened 2023-05-02
helm: fix mount for nfd-master config

marquiz:
Volume/mount setup for the ConfigMap was erroneously inside conditionals so it was not mounted unless TLS was enabled.


fmuyassarov (commented):
/lgtm
Good catch.

fmuyassarov:
@marquiz cherry-pick failed to create a patch due to conflicts. That should be done manually I guess.

PR #1207, opened 2023-05-04
deployment/helm: fix default for kubeletStateDir parameter

marquiz:
This parameter is a path in the host system, not a mount path inside the container.


PiotrProkop (approved):
Good catch!

PR #1218, opened 2023-05-05
topology-updater: use node IP in the default configz URI

marquiz:
Use a separate NODE_ADDRESS environment variable in the default value of -kubelet-config-uri (instead of NODE_NAME that was previously used). Also change the kustomize and Helm deployments to set this variable to node IP address. This should make the default deployment more robust, making it work in scenarios where node name does not resolve to the node ip, e.g. nodename != hostname.

### DIFF
--- a/cmd/nfd-topology-updater/main.go
+++ b/cmd/nfd-topology-updater/main.go
@@ -112,12 +112,13 @@ func parseArgs(flags *flag.FlagSet, osArgs ...string) (*topology.Args, *resource
 	}
 
 	if len(resourcemonitorArgs.KubeletConfigURI) == 0 {
-		if len(utils.NodeName()) == 0 {
-			fmt.Fprintf(flags.Output(), "unable to determine the default kubelet config endpoint 'https://${NODE_NAME}:%d/configz' due to empty NODE_NAME environment, "+
-				"please either define the NODE_NAME environment variable or specify endpoint with the -kubelet-config-uri flag\n", kubeletSecurePort)
+		nodeAddress := os.Getenv("NODE_ADDRESS")
+		if len(nodeAddress) == 0 {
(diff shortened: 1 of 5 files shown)


marquiz:
/assign @PiotrProkop 
ping @Tal-or

cmd/nfd-topology-updater/main.go
```diff
@@ -112,12 +112,13 @@ func parseArgs(flags *flag.FlagSet, osArgs ...string) (*topology.Args, *resource
 	}
 
 	if len(resourcemonitorArgs.KubeletConfigURI) == 0 {
-		if len(utils.NodeName()) == 0 {
-			fmt.Fprintf(flags.Output(), "unable to determine the default kubelet config endpoint 'https://${NODE_NAME}:%d/configz' due to empty NODE_NAME environment, "+
-				"please either define the NODE_NAME environment variable or specify endpoint with the -kubelet-config-uri flag\n", kubeletSecurePort)
+		nodeIP := os.Getenv("NODE_IP")
```
PiotrProkop, line 115: you can use `LookupEnv` here https://pkg.go.dev/os#LookupEnv
```go

PiotrProkop (commented):
small nit otherwise LGTM

cmd/nfd-topology-updater/main.go
PiotrProkop: do we care about checking if `NODE_IP` is a valid ipv4 or ipv6 address?

marquiz: I'm not sure. Now that we think about it we could name it NODE_ADDRESS or something instead so user could put whatever, IP, nodename, hostname... WDYT?

PiotrProkop: sounds good :)

marquiz: If the env var is not defined `os.GetEnv` will return an empty string -> the error case is handled below by `if len(nodeIP) == 0`

marquiz:
PR updated. Env var now named NODE_ADDRESS to make it more generic

Issue #1188, opened 2023-04-25
Implement leader-election for nfd-master


PiotrProkop:
**What would you like to be added**:

Implement leader-election for nfd-master, when `-enable-nodefeature-api` is set to `true`.

**Why is this needed**:

Nfd-master is going through every node and checks if labels/extenden resources needs to be updated. When we deploy more than one instance of nfd-master, we will have multiple containers trying to update same resources.

PR #1219, opened 2023-05-05
Add leader election for nfd-master

PiotrProkop:
This should fix https://github.com/kubernetes-sigs/node-feature-discovery/issues/1188

Right now it is still in progress.

Missing:
- [x] Updating deployment(adding rbac)
- [x] Updating README.md
- [x] Adding tests

### DIFF
--- a/deployment/base/rbac/master-clusterrole.yaml
+++ b/deployment/base/rbac/master-clusterrole.yaml
@@ -22,3 +22,18 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - create
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  resourceNames:
+  - "nfd-master.nfd.kubernetes.io"
+  verbs:
+  - get
+  - update
--- a/deployment/components/master-config/nfd-master.conf.example
+++ b/deployment/components/master-config/nfd-master.conf.example
@@ -5,3 +5,9 @@
 # enableTaints: false
 # labelWhiteList: "foo"
 # resyncPeriod: "2h"
+# leaderElection:
+#   leaseDuration: 15s
+#   # this value has to be lower than leaseDuration and greater than retryPeriod*1.2
+#   renewDeadline: 10s
+#   # this value has to be greater than 0
+#   retryPeriod: 2s
--- a/deployment/helm/node-feature-discovery/templates/clusterrole.yaml
+++ b/deployment/helm/node-feature-discovery/templates/clusterrole.yaml
@@ -25,6 +25,21 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - create
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  resourceNames:
+  - "nfd-master.nfd.kubernetes.io"
+  verbs:
+  - get
+  - update
 {{- end }}
 
 ---
--- a/deployment/helm/node-feature-discovery/templates/master.yaml
+++ b/deployment/helm/node-feature-discovery/templates/master.yaml
@@ -82,6 +82,9 @@ spec:
             - "-port={{ .Values.master.port | default "8080" }}"
             {{- if .Values.enableNodeFeatureApi }}
             - "-enable-nodefeature-api"
+            {{- if  gt (int .Values.master.replicaCount) 1 }}
+            - "-enable-leader-election"
+            {{- end }}
             {{- end }}
             {{- if .Values.master.extraLabelNs | empty | not }}
             - "-extra-label-ns={{- join "," .Values.master.extraLabelNs }}"
--- a/docs/reference/master-configuration-reference.md
+++ b/docs/reference/master-configuration-reference.md
@@ -114,7 +114,7 @@ Example:
 labelWhiteList: "foo"
 ```
 
-### resyncPeriod
+## resyncPeriod
 
 The `resyncPeriod` option specifies the NFD API controller resync period.
 The resync means nfd-master replaying all NodeFeature and NodeFeatureRule objects,
@@ -128,5 +128,64 @@ Default: 1 hour.
 Example:
 
 ```yaml
-resyncPeriod=2h
+resyncPeriod: 2h
+```
+
+## leaderElection
+
+The `leaderElection` section exposes configuration to tweak leader election.
+
+### leaderElection.leaseDuration
+
+`leaderElection.leaseDuration` is the duration that non-leader candidates will
+wait to force acquire leadership. This is measured against time of
+last observed ack.
+
+A client needs to wait a full LeaseDuration without observing a change to
+the record before it can attempt to take over. When all clients are
+shutdown and a new set of clients are started with different names against
+the same leader record, they must wait the full LeaseDuration before
+attempting to acquire the lease. Thus LeaseDuration should be as short as
+possible (within your tolerance for clock skew rate) to avoid a possible
+long waits in the scenario.
+
+Default: 15 seconds.
+
+Example:
+
+```yaml
+leaderElection:
+  leaseDurtation: 15s
+```
+
+### leaderElection.renewDeadline
+
+`leaderElection.renewDeadline` is the duration that the acting master will retry
+refreshing leadership before giving up.
+
+This value has to be lower than leaseDuration and greater than retryPeriod*1.2.
+
+Default: 10 seconds.
+
+Example:
+
+```yaml
+leaderElection:
+  renewDeadline: 10s
+```
+
+### leaderElection.retryPeriod
+
+`leaderElection.retryPeriod` is the duration the LeaderElector clients should wait
+between tries of actions.
+
+It has to be greater than 0.
+
+Default: 2 seconds.
+
+Example:
+
+```yaml
+leaderElection:
+  retryPeriod: 2s
 ```
--- a/docs/usage/nfd-master.md
+++ b/docs/usage/nfd-master.md
@@ -29,6 +29,11 @@ are created on the node (note the allowed
 > its [`-enable-nodefeature-api`](../reference/worker-commandline-reference.md#-enable-nodefeature-api)
 > flag.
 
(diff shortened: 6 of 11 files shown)


PiotrProkop:
@marquiz @ArangoGutierrez PTAL

pkg/nfd-master/nfd-master.go
```diff
@@ -1230,3 +1239,47 @@ func (m *nfdMaster) startNfdApiController() error {
...
+			// add uuid to prevent situation where 2 nfd-master nodes run on same node
+			Identity: m.nodeName + "_" + uuid.NewString(),
+		},
+	}
+	config := leaderelection.LeaderElectionConfig{
+		Lock: lock,
+		// make it configurable?
+		LeaseDuration: 15 * time.Second,
+		RetryPeriod:   2 * time.Second,
+		RenewDeadline: 10 * time.Second,
```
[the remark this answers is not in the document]
PiotrProkop: The values for those fields right now are set to what docs suggests. I think the best way would be to expose this configuration to end user similliar as `scheduler-plugins` does. I'll make this as part of this PR, I guess those values should be exposed via config file and defaults to values above?

PiotrProkop:
> Thanks @PiotrProkop, looks already quite good.
> 
> We need to update the documentation, adjust at least `docs/usage/nfd-master.md` and supplement `docs/reference/master-commandline-reference.md` with the new flag

PR updated.

docs/usage/nfd-master.md
```diff
@@ -29,6 +29,11 @@ are created on the node (note the allowed
 > its [`-enable-nodefeature-api`](../reference/worker-commandline-reference.md#-enable-nodefeature-api)
 > flag.
 
+When `-enable-nodefeature-api` option is enabled and NFD-Master is intended to run
+with more than one replica, it is advised to use `-enable-leader-election` flag.
+This flag turns on leader election for NFD-Master and let only one replica
+to update `NodeFeature` objects.
```
marquiz, line 35: nfd-master only updates Nodes, it just "listens" to NodeFeature and NodeFeatureRule objects (as we don't have e.g. status field in those, yet).

```suggestion
to act on changes in NodeFeature and NodeFeatureRule objects.
```

deployment/base/rbac/master-clusterrole.yaml
```diff
@@ -22,3 +22,18 @@ rules:
...
+  resources:
+  - leases
+  verbs:
+  - create
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  resourceNames:
+  - "nfd-master.nodefeaturediscoveries.nfd.kubernetes.io"
```
marquiz, line 36: nit: not that the name would really matter but there seems to be some unnecessary repeating/tautology here (nodefeaturediscoveries.nfd), how about making it a tad shorter like `nfd-master.nfd.kubernetes.io`, WDYT?

TBH, I have no idea if there are some best practices on naming these...

deployment/components/master-config/nfd-master.conf.example
```diff
@@ -5,3 +5,7 @@
 # enableTaints: false
 # labelWhiteList: "foo"
 # resyncPeriod: "2h"
+# leaderElection:
+#   leaseDuration: 15s
+#   renewDeadline: 10s
+#   retryPeriod: 2s
```
marquiz, line 11: Thanks for adding these. My only small concern is overwhelming the user with detailed configuration options. Is there any advice we could give or point to (in the documentation) for users to be able to pick meaningful values. Like "renewDeadline should be shorter than leaseDuration", "retryPeriod should be shorter than renewDeadline"... Also, would it make sense to have proportional/dynamic defaults, i.e. renewDeadline would be 2/3*leaseDuration and retryPeriod renewDuration/5 if nothing is specified in those fields?

Any thoughts, WDYT?

marquiz (commented):
Thanks for the update @PiotrProkop, a few small comments but basicsally all the pieces seem to be there.

One wish for the Helm chart: could we enable `-enable-leader-election` on nfd-master if `master.replicaCount` parameter is greater than 1?

deployment/base/rbac/master-clusterrole.yaml
PiotrProkop: I have done it in a manner how node-feature-operator defines lock id:
https://github.com/kubernetes-sigs/node-feature-discovery-operator/blob/master/main.go#L99
I am open to changing it :)

deployment/components/master-config/nfd-master.conf.example
PiotrProkop: Right now those values are being defaulted to what is recommended in documentation for leader election.
I can add some comments here what are the rules of specifying them in example conf.

docs/usage/nfd-master.md
PiotrProkop: :+1:

deployment/base/rbac/master-clusterrole.yaml
marquiz: Ach, of course. `nodefeaturediscoveries` is the name of the CRD. In this scheme `nfd-master.nfd.kubernetes.io` would be more logical, IMO. Any thoughts @ArangoGutierrez?

deployment/components/master-config/nfd-master.conf.example
marquiz: > I can add some comments here what are the rules of specifying them in example conf

That would be great 🙏 If you do, please add the same comments to the config-reference in the docs, too

deployment/base/rbac/master-clusterrole.yaml
PiotrProkop: changed to `nfd-master.nfd.kubernetes.io`

deployment/helm/node-feature-discovery/templates/clusterrole.yaml
```diff
@@ -25,6 +25,21 @@ rules:
...
+  resources:
+  - leases
+  verbs:
+  - create
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  resourceNames:
+  - "nfd-master.nodefeaturediscoveries.nfd.kubernetes.io"
```
marquiz, line 39: This needs to be renamed, too 
```suggestion
  - "nfd-master.nfd.kubernetes.io"
```

PiotrProkop: fixed!

Dockerfile
```diff
@@ -17,7 +17,7 @@ COPY go.mod go.sum /go/node-feature-discovery/
 
 WORKDIR /go/node-feature-discovery
 
-RUN go mod download
+# RUN go mod download
```
marquiz, line 20: Why this?

PiotrProkop: It sneaked into PR. I'll remove it right now.

docs/reference/master-configuration-reference.md
```diff
@@ -128,5 +128,64 @@ Default: 1 hour.
...
 ```yaml
-resyncPeriod=2h
+resyncPeriod: 2h
+```
+
+### leaderElection
+
+The `leaderElection` option exposes configuration to tweak leader election.
+
+#### leaseDuration
```
marquiz, line 138: Hmm, I'd suggest to rename this.
```suggestion
#### leaderElection.leaseDuration
```

This would follow the style used in [worker config reference](https://kubernetes-sigs.github.io/node-feature-discovery/v0.13/reference/worker-configuration-reference.html)

...and ditto for the other new options

docs/reference/master-configuration-reference.md
```diff
@@ -128,5 +128,64 @@ Default: 1 hour.
 Example:
 
 ```yaml
-resyncPeriod=2h
+resyncPeriod: 2h
+```
+
+### leaderElection
+
+The `leaderElection` option exposes configuration to tweak leader election.
```
marquiz, line 136: I'd suggest to reword this, e.g.

```suggestion
The `leaderElection` section exposes configuration to tweak leader election.
```

or 

```suggestion
The `leaderElection` options expose configuration to tweak leader election.
```

marquiz (commented):
Thanks for the update @PiotrProkop. I came up with some minor new feedback.

Also, what about the Helm chart improvement I suggested earlier? I.e. setting `-enable-leader-election` on nfd-master in case `master.replicaCount` is more than one

docs/reference/master-configuration-reference.md
PiotrProkop: fixed

docs/reference/master-configuration-reference.md
PiotrProkop: Changed.

deployment/helm/node-feature-discovery/templates/master.yaml
```diff
@@ -82,6 +82,9 @@ spec:
             - "-port={{ .Values.master.port | default "8080" }}"
             {{- if .Values.enableNodeFeatureApi }}
             - "-enable-nodefeature-api"
+            {{- if  gt (float64 .Values.master.replicaCount) 1.0 }}
```
marquiz, line 85: Replicacount is an integer so I think we could simplify here a bit
```suggestion
            {{- if  gt Values.master.replicaCount 1 }}
```

PiotrProkop: I tried it but it seems that Helm is reading yaml numbers as floats:
```sh
$ helm template nfd ./node-feature-discovery --namespace node-feature-discovery --set enableNodeFeatureApi=true
Error: template: node-feature-discovery/templates/master.yaml:85:20: executing "node-feature-discovery/templates/master.yaml" at <gt .Values.master.replicaCount 1>: error calling gt: incompatible types for comparison

Use --debug flag to render out invalid YAML
```

interesting as this works fine:
```sh
$ helm template nfd ./node-feature-discovery --namespace node-feature-discovery --set enableNodeFeatureApi=true,master.replicaCount=1
```

marquiz: Hmm, interesting. I needed to try it out myself and yes, you're right. An evolved suggestion:

```suggestion
            {{- if  gt (int .Values.master.replicaCount) 1 }}
```

marquiz (approved):
Very nice job @PiotrProkop! I don't have anything to add 😊
/approve

PR #1221, opened 2023-05-08
e2e: delete CRs only if found

fmuyassarov:
Delete `NodeFeatureRule` and `NodeFeature` CRs only if found.

### DIFF
--- a/test/e2e/node_feature_discovery_test.go
+++ b/test/e2e/node_feature_discovery_test.go
@@ -150,19 +150,23 @@ func cleanupCRs(ctx context.Context, cli *nfdclient.Clientset, namespace string)
 	nfrs, err := cli.NfdV1alpha1().NodeFeatureRules().List(ctx, metav1.ListOptions{})
 	Expect(err).NotTo(HaveOccurred())
 
-	By("Deleting NodeFeatureRule objects from the cluster")
-	for _, nfr := range nfrs.Items {
-		err = cli.NfdV1alpha1().NodeFeatureRules().Delete(ctx, nfr.Name, metav1.DeleteOptions{})
-		Expect(err).NotTo(HaveOccurred())
+	if len(nfrs.Items) != 0 {
+		By("Deleting NodeFeatureRule objects from the cluster")
+		for _, nfr := range nfrs.Items {
+			err = cli.NfdV1alpha1().NodeFeatureRules().Delete(ctx, nfr.Name, metav1.DeleteOptions{})
+			Expect(err).NotTo(HaveOccurred())
+		}
 	}
 
 	nfs, err := cli.NfdV1alpha1().NodeFeatures(namespace).List(ctx, metav1.ListOptions{})
 	Expect(err).NotTo(HaveOccurred())
 
-	By("Deleting NodeFeature objects from namespace " + namespace)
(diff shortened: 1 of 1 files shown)


marquiz (approved):
Thx, logs look saner this way

fmuyassarov:
> Thx, logs look saner this way

yes and I noticed that GitHub action based e2e would fail if we try to delete object that doesn't exist. I think that because GitHub action is very sensitive to return codes.

marquiz:
> yes and I noticed that GitHub action based e2e would fail if we try to delete object that doesn't exist. I think GitHub action is very sensitive to return codes.

Hmm, that shouldn't happen 🤔 Even with the previous code we won't try to delete if there are no CRs. This PR doesn't seem to fix that problem (if it happens), just makes the log messages saner

ArangoGutierrez (approved):
/lgtm

fmuyassarov:
> > yes and I noticed that GitHub action based e2e would fail if we try to delete object that doesn't exist. I think GitHub action is very sensitive to return codes.
> 
> Hmm, that shouldn't happen thinking Even with the previous code we won't try to delete if there are no CRs. This PR doesn't seem to fix that problem (if it happens), just makes the log messages saner

@marquiz `cleanupCRs` is called from within [`JustBeforeEach`](https://github.com/kubernetes-sigs/node-feature-discovery/blob/50caa92232e0a2acfe900498b1bdfc9354c9db18/test/e2e/node_feature_discovery_test.go#L235). Due to that, cleanupCRs will fail for the first test when there is no CR instances created yet.

One example is https://github.com/fmuyassarov/node-feature-discovery/actions/runs/4912986710/jobs/8773141452 and you can see from the output that it can't delete the `nodefeaturerule`

```
  May  8 08:44:28.501: INFO: Deleting pod-security.kubernetes.io/enforce label from the test namespace
  [FAILED] in [JustBeforeEach] - /home/prow/actions-runner/e2e/node-feature-discovery/node-feature-discovery/test/e2e/node_feature_discovery_test.go:151 @ 05/08/23 08:44:28.563
  STEP: Deleting NodeFeatureRule objects from the cluster @ 05/08/23 08:44:28.869
  STEP: Deleting NodeFeature objects from namespace node-feature-discovery-6117 @ 05/08/23 08:44:29.039
  STEP: Destroying namespace "node-feature-discovery-6117" for this suite. @ 05/08/23 08:44:29.087
  << Timeline

  [FAILED] Unexpected error:
      <*errors.StatusError | 0xc000c2ac80>: 
      the server could not find the requested resource (get nodefeaturerules.nfd.k8s-sigs.io)
      {
          ErrStatus: 
              code: 404
              details:
                causes:
                - message: 404 page not found
                  reason: UnexpectedServerResponse
                group: nfd.k8s-sigs.io
                kind: nodefeaturerules
              message: the server could not find the requested resource (get nodefeaturerules.nfd.k8s-sigs.io)
              metadata: {}
              reason: NotFound
              status: Failure,
      }
  occurred
  In [JustBeforeEach] at: /home/prow/actions-runner/e2e/node-feature-discovery/node-feature-discovery/test/e2e/node_feature_discovery_test.go:151 @ 05/08/23 08:44:28.563
------------------------------
```

PR #1223, opened 2023-05-22
Migrate to structured logging

marquiz:
An attempt to make logs more consistent and easier to consume by logging tools.

Some background for klog structured logging e.g. in this blog post:
https://kubernetes.io/blog/2020/09/04/kubernetes-1-19-introducing-structured-logs/

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -608,7 +608,7 @@ func (m *nfdMaster) SetLabels(c context.Context, r *pb.SetLabelsRequest) (*pb.Se
 
 	switch {
 	case klog.V(4).Enabled():
-		utils.KlogDump(3, "gRPC SetLabels request received", "  ", r)
+		klog.InfoS("gRPC SetLabels request received", "setLabelsRequest", utils.DelayedDumper(r))
 	case klog.V(1).Enabled():
 		klog.InfoS("gRPC SetLabels request received", "nodeName", r.NodeName, "nfdVersion", r.NfdVersion, "labels", r.Labels)
 	default:
@@ -697,7 +697,7 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(nodeName string) error {
 			o.Spec.MergeInto(features)
 		}
 
-		utils.KlogDump(4, "Composite NodeFeatureSpec after merge:", "  ", features)
+		klog.V(4).InfoS("merged nodeFeatureSpecs", "newNodeFeatureSpec", utils.DelayedDumper(features))
 
 		if objs[0].Namespace == m.namespace && objs[0].Name == nodeName {
 			// This is the one created by nfd-worker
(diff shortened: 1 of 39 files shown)


fmuyassarov:
/assign
will take a look today

pkg/nfd-master/nfd-master.go
```diff
@@ -198,12 +198,10 @@ func newDefaultConfig() *NFDConfig {
 // Run NfdMaster server. The method returns in case of fatal errors or if Stop()
 // is called.
 func (m *nfdMaster) Run() error {
-	klog.Infof("Node Feature Discovery Master %s", version.Get())
+	klog.InfoS("Node Feature Discovery Master", "version", version.Get(), "nodeName", m.nodeName, "namespace", m.namespace)
 	if m.args.Instance != "" {
-		klog.Infof("Master instance: %q", m.args.Instance)
+		klog.InfoS("Master instance", "instance", m.args.Instance)
```
fmuyassarov, line 203: do we need `"instance" `addition  here?

pkg/nfd-master/nfd-master.go
```diff
@@ -891,22 +889,18 @@ func authorizeClient(c context.Context, checkNodeName bool, nodeName string) err
...
 			return fmt.Errorf("incorrect client credentials")
 		}
 		if len(tlsAuth.State.VerifiedChains) == 0 || len(tlsAuth.State.VerifiedChains[0]) == 0 {
-			klog.Errorf("gRPC request error: client certificate verification for '%v' failed", client.Addr)
 			return fmt.Errorf("client certificate verification failed")
 		}
 
 		err := verifyNodeName(tlsAuth.State.VerifiedChains[0][0], nodeName)
 		if err != nil {
-			klog.Errorf("gRPC request error: authorization for %v failed: %v", client.Addr, err)
```
fmuyassarov, line 909: are we missing replacement `klog.ErrorS()` here ?

fmuyassarov (commented):
Looks good. Good work, as the next step perhaps we can move to even contextual logging ?

pkg/nfd-master/nfd-master.go
marquiz: Sorry, I didn't quite understand the comment 🤔 Do you mean removing the arg like

```golang
klog.InfoS("Master instance" m.args.Instance)
```

That's not possible because args must be in key-value pairs

pkg/nfd-master/nfd-master.go
marquiz: Hmm, thanks for pointing this out. I wonder what I was thinking with dropping all of these 🤨 Probably cutting corners in the name of duplication (because we get the error on the client side, too). But I think we should not change this -> let me fix this

pkg/nfd-master/nfd-master.go
fmuyassarov: No, I I was thinking  `klog.InfoS("Master", "instance", m.args.Instance)` but this will result in "Master" instance="value" which looks a bit weird. I think then what you have is better.

ArangoGutierrez:
A side from @fmuyassarov comments, my only question/concern is moving some warning to infoS, should we prefix those with a `W` or something?

pkg/nfd-master/nfd-master.go
marquiz: This is now resolved by logging the returned error in `SetLabels()`

marquiz:
> A side from @fmuyassarov comments, my only question/concern is moving some warning to infoS, should we prefix those with a `W` or something?

I don't think so. You can read some of the rationale for the design here: https://github.com/go-logr/logr (and even more radically plain simple [here](http://dave.cheney.net/2015/11/05/lets-talk-about-logging).

What we could think about are the log levels, though. E.g. reserve V(0) for the absolute minimum and warning-type of messages. Then possibly V(1) would be the default etc. But I'm not sure if that (too) should be incorporated in this PR.

ArangoGutierrez (approved):
/lgtm
/milestone v0.14.0

fmuyassarov (commented):
/lgtm

marquiz:
Rebased

marquiz:
ping @ArangoGutierrez

PR #1227, opened 2023-05-24
nfd-master: use close for stop channel

marquiz:
Simpler and more reliable (in case of multiple consumers) to just close the channel.


PiotrProkop (approved):

PiotrProkop:
@marquiz sorry, it took so long. But I have missed notification somehow.
/lgtm

PR #1228, opened 2023-05-26
nfd-master: add validation of label names and values

marquiz:
Validate labels before trying to update the node. Makes us fail early nad prevent useless retries in case invalid labels are tried.


deployment/base/topologyupdater-daemonset/topologyupdater-daemonset.yaml
PiotrProkop, line 1: is this change required for this patch?

marquiz: No, definitely not! How on earth did this change and up in the PR 🤔 Fixed. Thanks for spotting this @PiotrProkop, looks like a bit hasty submission on my part...

PiotrProkop:
I think we can experience -> https://github.com/kubernetes/test-infra/issues/29622

PR #1233, opened 2023-05-31
[release-0.13] nfd-master: add validation of label names and values

marquiz:
Validate labels before trying to update the node. Makes us fail early nad prevent useless retries in case invalid labels are tried.

(backported from commit 2a3c7e4c9357b5426c1203ba7684e6f966a0d65f)


ArangoGutierrez (approved):
/lgtm

PR #1240, opened 2023-06-05
cpu: Discover TDX guests based on cpuid information

fidencio:
NFD already has the capability to discover whether baremetal / host
machines support Intel TDX.  Now, the next step is to add support for
discovering whether a node is TDX protected (as in, a virtual machine
started using Intel TDX).

In order to do so, we've decided to go for a new `cpu-security.tdx`
property, called `protected` (`cpu-security.tdx.protected`).

NOTE: We have to update the cpuid to its v2.2.5 release, released on
June 2nd, 2023, as it brings in information about TDX guests.

### DIFF
--- a/docs/usage/features.md
+++ b/docs/usage/features.md
@@ -58,6 +58,7 @@ option of nfd-worker.
 | **`cpu-security.sgx.enabled`**    | true   | Set to 'true' if Intel SGX is enabled in BIOS (based on a non-zero sum value of SGX EPC section sizes).
 | **`cpu-security.se.enabled`**     | true   | Set to 'true' if IBM Secure Execution for Linux (IBM Z & LinuxONE) is available and enabled (requires `/sys/firmware/uv/prot_virt_host` facility)
 | **`cpu-security.tdx.enabled`**    | true   | Set to 'true' if Intel TDX is available on the host and has been enabled (requires `/sys/module/kvm_intel/parameters/tdx`).
+| **`cpu-security.tdx.protected`**  | true   | Set to 'true' if Intel TDX was used to start the guest node, based on the existence of the "TDX_GUEST" information as part of cpuid features.
 | **`cpu-security.sev.enabled`**    | true   | Set to 'true' if ADM SEV is available on the host and has been enabled (requires `/sys/module/kvm_amd/parameters/sev`).
 | **`cpu-security.sev.es.enabled`** | true   | Set to 'true' if ADM SEV-ES is available on the host and has been enabled (requires `/sys/module/kvm_amd/parameters/sev_es`).
 | **`cpu-security.sev.snp.enabled`**| true   | Set to 'true' if ADM SEV-SNP is available on the host and has been enabled (requires `/sys/module/kvm_amd/parameters/sev_snp`).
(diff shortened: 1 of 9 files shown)


source/cpu/security_amd64.go
```diff
@@ -102,6 +106,17 @@ func tdxEnabled() bool {
 	return false
 }
 
+func tdxProtected() bool {
+	features := cpuid.CPU.FeatureSet()
+	for _, feature := range features {
+		if feature == "TDX_GUEST" {
+			return true
+		}
+	}
```
marquiz, line 115: I think you could use `cpuid.CPU.Supports` instead?

Also, afaiu with the new cpuid we will also get `cpu-cpuid.TDX_GUEST` label, right? If we want this separate tdx label, I'd suggest to add that cpuid.TDX_GUEST label to the default deny list in `newDefaultConfig()` to avoid redundant duplicate information (i.e. excess labels). In this case we also need to update the documentation, and preferably the example config, regarding the default deny list.

fidencio: @marquiz, thanks for both suggestions.

About `cpuid.CPU.Supporst` (or `cpuid.CPU.Has`, I'll do the change right now.

About exposing the `cpuid.TDX_GUEST` label, I'd like to confirm one thing with you.  The main reason I do believe we should have it as `tdx.protected` instead of exposing `cpuid.TDX_GUEST` is for the sake of consistency.  You're 100% correct on  not exposing the `cpuid.TDX_GUEST` in this case, and I'd like to hear whether you agree with my consistency argument in order to expose `tdx.protected`.

marquiz: I think that consistency argument makes sense and I'm good with `tdx.protected`

marquiz: and yes, `cpuid.CPU.Has()` is better...

fidencio:
I've covered all the comments from @marquiz.
Please, let me know if I missed something, or if something else has to be changed.

marquiz (commented):
Thanks for the update @fidencio.

You need to run `make template` to update the default config of Helm chart, too.

Also, we list the filtered out cpuid labels in `docs/usage/features.md` so add TDX_GUEST there as well

marquiz (commented):
Thanks for the update @fidencio.

You need to run `make template` to update the default config of Helm chart, too.

Also, we list the filtered out cpuid labels in `docs/usage/features.md` so add TDX_GUEST there as well

fidencio:
`make templates`  ran, and `docs/usage/feature.md` updated.  Thanks @marquiz!

docs/usage/features.md
```diff
@@ -124,7 +125,7 @@ configuration options for details.
 
 By default, the following CPUID flags have been blacklisted: BMI1, BMI2, CLMUL,
 CMOV, CX16, ERMS, F16C, HTT, LZCNT, MMX, MMXEXT, NX, POPCNT, RDRAND, RDSEED,
-RDTSCP, SGX, SSE, SSE2, SSE3, SSE4, SSE42 and SSSE3. See
+RDTSCP, SGX, SSE, SSE2, SSE3, SSE4, SSE42, SSSE3 and TDX_GUEST. See
```
marquiz, line 128: Thx for adding SSSE3 here 👍

marquiz (approved):
Thanks @fidencio. looks good to me.

/assign @PiotrProkop

PR #1245, opened 2023-06-07
Fix missing apostrophe for jq

sj14:
There seems to be missing apostrophes in the `jq` calls in the documentation.

jq version:

``` console
> jq --version
jq-1.6
```

current behaviour:

```console
> kubectl get no -o json | jq .items[].metadata.labels
zsh: no matches found: .items[].metadata.labels
```

changed behaviour:

```console
> kubectl get no -o json | jq '.items[].metadata.labels'
{
...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ $ kubectl -n node-feature-discovery get all
   pod/nfd-worker-mjg9f              1/1     Running   0          17s
 ...
 
-$ kubectl get no -o json | jq .items[].metadata.labels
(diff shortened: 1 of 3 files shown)


ArangoGutierrez (approved):
Thanks @sj14 for this patch! 
/ok-to-test
/lgtm

marquiz:
zsh is strange :)

sj14:
Ah, the reason was bash vs zsh. I thought it was due to a change in a jq version :)

PR #1253, opened 2023-07-02
fix typo in helm chart

adrianchiris:
else statement of crd-controller should
also refer to crd-controller flag.


ArangoGutierrez (approved):
as seen in https://kubernetes-sigs.github.io/node-feature-discovery/v0.13/reference/master-commandline-reference.html#-crd-controller
/lgtm

PR #1255, opened 2023-07-04
Stop ticker in time to avoid memory leak

hangscer8:
Clean `ticker` in time.
Because it will cause memory leak if we do not stop `ticker` when the function has completed.

### DIFF
--- a/pkg/nfd-topology-gc/nfd-nrt-gc.go
+++ b/pkg/nfd-topology-gc/nfd-nrt-gc.go
@@ -146,6 +146,7 @@ func (n *topologyGC) runGC() {
 // periodicGC runs garbage collector at every gcPeriod to make sure we haven't missed any node
 func (n *topologyGC) periodicGC(gcPeriod time.Duration) {
 	gcTrigger := time.NewTicker(gcPeriod)
+	defer gcTrigger.Stop()
 	for {
 		select {
 		case <-gcTrigger.C:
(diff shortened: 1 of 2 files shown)


ArangoGutierrez:
/cc @ffromani 
Thanks for this PR @hangscer8 !!!, could you provide more details on why this is needed, and expand the Commit description
`clean ticker` is too short for the git history, something a bit longer will be useful when cutting a release or looking back into CHANGELOG

ffromani:
IIUC this is to prevent goroutine leaks, which can be a common mistake when using `time.Ticker`.
Waiting for the author to confirm the intent

hangscer8:
> IIUC this is to prevent goroutine leaks, which can be a common mistake when using `time.Ticker`. Waiting for the author to confirm the intent

Hi , @ffromani , I have changed the description of this pr.

ffromani:
thanks @hangscer8 ! could you please convey the same information in the commit message? this means editing your local commit message and (force-) pushing again the PR.
While fixing the description of the PR helps the review, we would like to have this information on record in the git history.

Once that is done, LGTM

hangscer8:
> thanks @hangscer8 ! could you please convey the same information in the commit message? this means editing your local commit message and (force-) pushing again the PR. While fixing the description of the PR helps the review, we would like to have this information on record in the git history.
> 
> Once that is done, LGTM

I have changed the commit message.

ffromani:
thanks again @hangscer8 . Since unfortunately the `verify` lane is failing, if you need to resubmit, could you please add the line
```
Because it will cause memory leak if we do not stop ticker when the function has completed.
```
to the commit message? thanks

ArangoGutierrez:
/re-title Stop ticker in time to avoid memory leak

ArangoGutierrez:
/retitle Stop ticker in time to avoid memory leak

ArangoGutierrez (approved):
Thanks for fixing this

ffromani (commented):
/lgtm

Issue #1254, opened 2023-07-03
Can't update topologyManagerPolicy and topologyManagerScope attributes of NodeResourceTopology


PiotrProkop:
**What happened**:
NFD is only detecting policy and scope of Topology Manager when NRT object doesn't exist. This means that `topologyManagerScope` and `topologyManagerPolicy` attributes won't change even if `kubelet` config was changed to use other TopologyManager policy and scope.

**What you expected to happen**:

`topologyManagerScope` and `topologyManagerPolicy` attributes of NRT objects will be updated when kubelet config was changed.

**How to reproduce it (as minimally and precisely as possible)**:

- Change TopologyManager policy
- Restart kubelet
- NRT object is still advertising old policy

ffromani:
it seems a bug indeed. We should update the attributes alongside the zones.

PR #1256, opened 2023-07-04
Fix Topology Manager policy and scope not being updated after NRT creation

PiotrProkop:
fixes https://github.com/kubernetes-sigs/node-feature-discovery/issues/1254

### DIFF
--- a/pkg/nfd-topology-updater/nfd-topology-updater.go
+++ b/pkg/nfd-topology-updater/nfd-topology-updater.go
@@ -18,6 +18,7 @@ package nfdtopologyupdater
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 
@@ -26,6 +27,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
+	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
 
 	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
 	"sigs.k8s.io/node-feature-discovery/pkg/apihelper"
@@ -34,6 +36,7 @@ import (
 	"sigs.k8s.io/node-feature-discovery/pkg/resourcemonitor"
 	"sigs.k8s.io/node-feature-discovery/pkg/topologypolicy"
 	"sigs.k8s.io/node-feature-discovery/pkg/utils"
+	"sigs.k8s.io/node-feature-discovery/pkg/utils/kubeconf"
 	"sigs.k8s.io/node-feature-discovery/pkg/version"
 	"sigs.k8s.io/yaml"
 )
@@ -66,24 +69,8 @@ type NfdTopologyUpdater interface {
 	Stop()
 }
 
-type staticNodeInfo struct {
-	nodeName string
-	tmPolicy string
-	tmScope  string
-}
-
-func newStaticNodeInfo(policy, scope string) staticNodeInfo {
-	nodeName := utils.NodeName()
-	klog.InfoS("detected kubelet Topology Manager configuration", "policy", policy, "scope", scope, "nodeName", nodeName)
-	return staticNodeInfo{
-		nodeName: nodeName,
-		tmPolicy: policy,
-		tmScope:  scope,
-	}
-}
-
 type nfdTopologyUpdater struct {
-	nodeInfo            staticNodeInfo
+	nodeName            string
 	args                Args
 	apihelper           apihelper.APIHelpers
 	resourcemonitorArgs resourcemonitor.Args
@@ -91,10 +78,11 @@ type nfdTopologyUpdater struct {
 	eventSource         <-chan kubeletnotifier.Info
 	configFilePath      string
 	config              *NFDConfig
+	kubeletConfigFunc   func() (*kubeletconfigv1beta1.KubeletConfiguration, error)
 }
 
 // NewTopologyUpdater creates a new NfdTopologyUpdater instance.
-func NewTopologyUpdater(args Args, resourcemonitorArgs resourcemonitor.Args, policy, scope string) (NfdTopologyUpdater, error) {
+func NewTopologyUpdater(args Args, resourcemonitorArgs resourcemonitor.Args) (NfdTopologyUpdater, error) {
 	eventSource := make(chan kubeletnotifier.Info)
 	if args.KubeletStateDir != "" {
 		ntf, err := kubeletnotifier.New(resourcemonitorArgs.SleepInterval, eventSource, args.KubeletStateDir)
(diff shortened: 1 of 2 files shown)


pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -290,3 +305,38 @@ func updateAttributes(lhs *v1alpha2.AttributeList, rhs v1alpha2.AttributeList) {
 		updateAttribute(lhs, attr)
 	}
 }
+
+func getKubeletConfigFunc(uri, apiAuthTokenFile string) (func() (*kubeletconfigv1beta1.KubeletConfiguration, error), error) {
```
ffromani, line 309: Ideally, I'd prefer having a goroutine waiting for changes in the kubeletconfig and re-reading it if needed and only if needed, avoding constant and costly polling. IIRC we use already `fsnotify` in NFD so this would mean no new deps. Have you explored this option? how hard would it be to implement?

PiotrProkop: I feel like if we want to support both endpoints and not to overcomplicate the code. We can spin up another goroutine that will check `kubelet` config every 5mins(or make it configurable) to reduce costly poling and `patch` only `TopologyPolices` or attributes field?

ffromani: Fair point. I'm hesitant to add kubelet config read in this path, this is why I mentioned a different goroutine.
Let me review the code again and think a bit more.

PiotrProkop: I feel like spinning up another goroutine which will be updating NRT object can lead to conflicts when updating same resource it would be best to have one func that's responsible for managing NRT. We can also consider such implementation:
```
-       policy, scope, err := w.detectTopologyPolicyAndScope()
-       if err != nil {
-               return fmt.Errorf("failed to detect TopologyManager's policy and scope: %w", err)
-       }
-
-       tmAttributes := createTopologyAttributes(policy, scope)
-       deprecatedTopologyPolicies := []string{string(topologypolicy.DetectTopologyPolicy(policy, scope))}
-
        nrt, err := cli.TopologyV1alpha2().NodeResourceTopologies().Get(context.TODO(), w.nodeName, metav1.GetOptions{})
        if errors.IsNotFound(err) {
+               policy, scope, err := w.detectTopologyPolicyAndScope()
+               if err != nil {
+                       return fmt.Errorf("failed to detect TopologyManager's policy and scope: %w", err)
+               }
+
+               tmAttributes := createTopologyAttributes(policy, scope)
+               deprecatedTopologyPolicies := []string{string(topologypolicy.DetectTopologyPolicy(policy, scope))}
+
                nrtNew := v1alpha2.NodeResourceTopology{
                                Name: w.nodeName,
@@ -237,15 +239,28 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
        nrtMutated := nrt.DeepCopy()
        nrtMutated.Zones = zoneInfo

-       attributes := append(scanResponse.Attributes, tmAttributes...)
-       updateAttributes(&nrtMutated.Attributes, attributes)
+       if w.lastUpdateTimestamp.IsZero() || time.Since(w.lastUpdateTimestamp).Minutes() > float64(5) {
+               policy, scope, err := w.detectTopologyPolicyAndScope()
+               if err != nil {
+                       return fmt.Errorf("failed to detect TopologyManager's policy and scope: %w", err)
+               }

-       nrtMutated.TopologyPolicies = deprecatedTopologyPolicies
+               tmAttributes := createTopologyAttributes(policy, scope)
+               deprecatedTopologyPolicies := []string{string(topologypolicy.DetectTopologyPolicy(policy, scope))}
+
+               attributes := append(scanResponse.Attributes, tmAttributes...)
+               updateAttributes(&nrtMutated.Attributes, attributes)
+
+               nrtMutated.TopologyPolicies = deprecatedTopologyPolicies
+       }

        nrtUpdated, err := cli.TopologyV1alpha2().NodeResourceTopologies().Update(context.TODO(), nrtMutated, metav1.UpdateOptions{})
        if err != nil {
                return fmt.Errorf("failed to update NodeResourceTopology: %w", err)
        }
+
+       w.lastUpdateTimestamp = time.Now()
+
```

or move reading `scope` and `policy` into seperate goroutine and keep them in shared state somewhere?

PiotrProkop: @ffromani to reduce number of times we need to read Kubelet config file and minimize complexity. I have modified the code only to reread Kubelet config when `Interval based` event is triggered, which default to 1 min(somewhat low). Should we introduce another ticker? 
I feel like we should implement a better solution to this error but I am unsure if we need to do it in this PR.

ffromani: thanks @PiotrProkop my main issue was finding time to review, I'll have another pass ASAP. I tend to agree without a some redesign (which is better served on a separate PR) we can't do much better than the current approach.

marquiz (commented):
Thanks @PiotrProkop for the enhancement. I have just a few notes after a quick read-through, relying mostly on @ffromani here :)

One ask: you have nice description in the related github issue, could you take some of that to the commit message?

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -195,27 +204,34 @@ func (w *nfdTopologyUpdater) Stop() {
 	}
 }
 
-func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneList, scanResponse resourcemonitor.ScanResponse) error {
+func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneList, scanResponse resourcemonitor.ScanResponse, shouldReReadKubeletConfig bool) error {
```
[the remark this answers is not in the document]
PiotrProkop: 👍

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -225,12 +241,28 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
+
+	attributes := scanResponse.Attributes
+
+	if shouldReReadKubeletConfig {
+		policy, scope, err := w.detectTopologyPolicyAndScope()
+		if err != nil {
+			return fmt.Errorf("failed to detect TopologyManager's policy and scope: %w", err)
+		}
+		tmAttributes := createTopologyAttributes(policy, scope)
+		deprecatedTopologyPolicies := []string{string(topologypolicy.DetectTopologyPolicy(policy, scope))}
```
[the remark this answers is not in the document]
PiotrProkop: I'll try to think of something.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -169,8 +173,13 @@ func (w *nfdTopologyUpdater) Run() error {
 			}
 			zones = resAggr.Aggregate(scanResponse.PodResources)
 			klog.V(1).InfoS("aggregated resources identified", "resourceZones", utils.DelayedDumper(zones))
+			shouldReReadKubeletConfig := false
+			if info.Event == kubeletnotifier.IntervalBased {
+				shouldReReadKubeletConfig = true
```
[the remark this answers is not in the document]
PiotrProkop: Checking now.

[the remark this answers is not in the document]
PiotrProkop: If I understand correctly current behavior there is no first run during startup. The first update happens either after `sleepInterval` or when nfd detects changes to checkpoint files.

pkg/nfd-topology-updater/nfd-topology-updater.go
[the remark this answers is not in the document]
marquiz: I think this is cleaner now 👍

marquiz:
Thx for the updates @PiotrProkop. I don't think I have much to add. I'll leave this to @ffromani now and will approve when he's happy

marquiz:
ping @ffromani

ffromani:
> ping @ffromani

Hi! sorry for the delay I'll review by the end of this week (July 28)

marquiz:
> Hi! sorry for the delay I'll review by the end of this week (July 28)

Thanks @ffromani, no panic with this (yet), just wanted to remind so that this wouldn't be the last one pending

marquiz (approved):
Thanks @ffromani. I agree that going forward some re-architecting would be beneficial, e.g. detecting kubelet restarts or smth

PiotrProkop:
```
 /home/prow/go/src/github.com/kubernetes-sigs/node-feature-discovery/pkg/utils/memory_resources.go:54:17: n.Name undefined (type fs.DirEntry has no field or method Name) 
 ```
 should I rebase my branch to fix it? @marquiz

marquiz:
> should I rebase my branch to fix it? @marquiz

Hmm, no, the logcheck seems to be flaky 🧐
/retest

PR #1259, opened 2023-07-17
[release-0.13] nfd-master: fix node updates on config change

marquiz:
Don't try to update all nodes on config change when the NodeFeature API is disabled. In this case we rely on gRPC and only act on gRPC requests from the worker.


fmuyassarov (commented):
/lgtm

PR #1265, opened 2023-07-19
cpu: Take cgroupsv1 into account when reading misc.capacity

fidencio:
We've been only considering cgroupsv2 when trying to read misc.capacity. However, there are still a bunch of systems out there relying on cgroupsv1.

### DIFF
--- a/source/cpu/security_amd64.go
+++ b/source/cpu/security_amd64.go
@@ -122,26 +122,10 @@ func sevParameterEnabled(parameter string) bool {
 	return false
 }
 
-func getCgroupMiscCapacity(resource string) int64 {
+func retrieveCgroupMiscCapacityValue(miscCgroupPath *os.File, resource string) int64 {
 	var totalResources int64 = -1
-	var err error = nil
-	var f *os.File = nil
 
-	miscCgroupsPaths := []string{"fs/cgroup/misc.capacity", "fs/cgroup/misc/misc.capacity"}
-	for _, miscCgroupsPath := range miscCgroupsPaths {
-		miscCgroups := hostpath.SysfsDir.Path(miscCgroupsPath)
-		f, err = os.Open(miscCgroups)
-		if err == nil {
-			defer f.Close()
-			break
-		}
-	}
-
-	if err != nil {
-		return totalResources
-	}
-
-	r := bufio.NewReader(f)
+	r := bufio.NewReader(miscCgroupPath)
 	for {
 		line, _, err := r.ReadLine()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


fidencio:
/cc @mythi, @marquiz

@arronwy is testing the patch right now.

source/cpu/security_amd64.go
```diff
@@ -124,13 +124,22 @@ func sevParameterEnabled(parameter string) bool {
...
+	var err error = nil
+	var f *os.File = nil
+
+	miscCgroupsPaths := []string{"fs/cgroup/misc.capacity", "fs/cgroup/misc/misc.capacity"}
+	for _, miscCgroupsPath := range miscCgroupsPaths {
+		miscCgroups := hostpath.SysfsDir.Path(miscCgroupsPath)
+		f, err = os.Open(miscCgroups)
+		if err != nil {
+			continue	
+		}
```
marquiz, line 136: Should we break out in case err==nil?

fidencio: We actually should. :-)

fidencio: Updated, thanks @marquiz!

fidencio:
`quay.io/fidencio/nfd:v0.14.0-devel-137-g8ed5a234` is an image that can be used for testing it out.

ArangoGutierrez:
> If there's a way to get this to the stable releases, this would be awesome.

I am ok with cherry-picking this into v0.13.x , the argument makes sense, CG V1 is still around

fidencio:
> CG V1 is still around

We like it or not. :-)

marquiz (commented):
Thanks @fidencio.

One suggestion: I think it would make it cleaner if you refactored the (`r := bufio.NewReader(f)...` stuff) into a separate function. WDYT?

Also, should we cherry-pick this for the release-0.13 branch?

ArangoGutierrez (commented):
/lgtm
Leaving cheery pick and final approval to @marquiz

fidencio:
> One suggestion: I think it would make it cleaner if you refactored the (`r := bufio.NewReader(f)...` stuff) into a separate function. WDYT?

No problem from my side, I was trying to keep the change as small as possible in order to get it easily backported.
Let me do the refactoring and I will update the PR in a few minutes.

fidencio:
> Also, should we cherry-pick this for the release-0.13 branch?

If possible, please, let's do it.

fidencio:
Updated, and here's the new image @arronwy can use to test it:
`quay.io/fidencio/nfd:v0.14.0-devel-138-g1f6e31cd`

source/cpu/security_amd64.go
```diff
@@ -158,3 +151,20 @@ func getCgroupMiscCapacity(resource string) int64 {
 
 	return totalResources
 }
+
+func getCgroupMiscCapacity(resource string) int64 {
+	var totalResources int64 = -1
```
marquiz, line 156: You could just drop this and return -1 in the end

```suggestion
```

fidencio: Updated.

marquiz:
@fidencio @mythi this was backported and released as part of [v0.13.3](https://github.com/kubernetes-sigs/node-feature-discovery/releases/tag/v0.13.3)

Issue #1257, opened 2023-07-14
Document supported k8s versions


Agalin:
**What would you like to be added**:

Document currently compatible versions of k8s in readme or docs and last verison of NFD working on a particular k8s version.

**Why is this needed**:

I'm currently working on deploying NFD in our environment. We have multiple clusters using k8s 1.23 and 1.27. To my surprise, Helm deployment working just fine on 1.27 cluster failed with master pod crashlooping on 1.23 clusters. It seems that recent kubernetes API library version bump is the problem - master works fine on 0.13.1.

Ensuring compatibility with old (I'm aware of 1.23 not being supported any more) k8s versions makes no sense but compatibility matrix would be a welcome addition and would save me a few hours of debugging. 😄

marquiz:
Hi @Agalin. NFD really should practically work on any kubernetes version that is still being used as we're only using stable apis. Bumping the patch version of the k8s deps really shouldn't break anything

Could you share more details about your deployment and the failure?

Agalin:
It seems to be some kind of an incompatibility between the latest version of the Kubernetes API library and K8s 1.23 (1.23.10 to be precise).

Master deployed from 0.13.2 crashloops with the following error in logs:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1724548]

goroutine 14 [running]:
sigs.k8s.io/node-feature-discovery/pkg/nfd-master.(*nfdMaster).nfdAPIUpdateOneNode(0xc000482a00, {0xc000158f60, 0xc})
        /go/node-feature-discovery/pkg/nfd-master/nfd-master.go:637 +0x108
sigs.k8s.io/node-feature-discovery/pkg/nfd-master.(*nfdMaster).nfdAPIUpdateAllNodes(0xc000482a00)
        /go/node-feature-discovery/pkg/nfd-master/nfd-master.go:627 +0x170
sigs.k8s.io/node-feature-discovery/pkg/nfd-master.(*nfdMaster).nfdAPIUpdateHandler(0xc000482a00)
        /go/node-feature-discovery/pkg/nfd-master/nfd-master.go:338 +0x1cc
created by sigs.k8s.io/node-feature-discovery/pkg/nfd-master.(*nfdMaster).Run
        /go/node-feature-discovery/pkg/nfd-master/nfd-master.go:228 +0x6f8
```

[...]

PR #1277, opened 2023-07-21
docs: describe supported Kubernetes versions

marquiz:
Fixes: #1257

### DIFF
--- a/docs/deployment/index.md
+++ b/docs/deployment/index.md
@@ -6,6 +6,9 @@ sort: 2
 
 # Deployment
 
+Node Feature Discovery can be deployed on any recent version of Kubernetes
+(v1.21+).
+
 See [Image variants](image-variants.md) for description of the different NFD
 container images available.
 
--- a/docs/get-started/introduction.md
+++ b/docs/get-started/introduction.md
@@ -18,7 +18,8 @@ sort: 1
 This software enables node feature discovery for Kubernetes. It detects
 hardware features available on each node in a Kubernetes cluster, and
 advertises those features using node labels and optionally node extended
-resources and node taints.
+resources and node taints. Node Feature Discovery is compatible with any recent
+version of Kubernetes (v1.21+).
 
 NFD consists of four software components:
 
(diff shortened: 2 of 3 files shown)


ArangoGutierrez:
Question for @ffromani @swatisehgal  do we have all we need for `Topology-updater` on Kubernetes 1.26?

marquiz:
Changed from K8s v1.16 to v1.21 as that's the minimum version required by nfd-topology-updater. v1.20 is already so old (EOL'd nearly a year and a half ago) that didn't feel it necessary to describe minimum k8s version of different components

/unhold

swatisehgal:
> > Question for @ffromani @swatisehgal do we have all we need for `Topology-updater` on Kubernetes 1.16?
> 
> Ha, good point @ArangoGutierrez. I totally forgot about topology-updater see_no_evil It doesn't work with k8s v1.16. When did the podresources improvement get beta? monocle_face I need to edit this pr /hold

Apologies for the delay in response, some PodResource API [improvements](https://github.com/kubernetes/kubernetes/pull/93243) were made in 1.20 and the GetallocatableResources endpoint was [introduced](https://github.com/kubernetes/kubernetes/pull/95734) as an alpha feature in 1.21 and [graduated to Beta](https://github.com/kubernetes/enhancements/pull/2932) in 1.23 so NFD-topology updater has everything it needs 1.21 onwards. In 1.21 and 1.22, feature gate needs to be enabled but is not needed from 1.23 onwards.

marquiz:
@swatisehgal thanks. Just checked the docs and the [nfd-topology-updater](https://kubernetes-sigs.github.io/node-feature-discovery/v0.13/usage/nfd-topology-updater.html) page actually has detailed deployment notes (pre-1.23 must have the feature gate enabled). So I think v1.21+ on the intro page is good enough. I now added a bit more details on the usage/nfd-topology-update page

docs/deployment/index.md
```diff
@@ -6,6 +6,9 @@ sort: 2
 
 # Deployment
 
+Node Feature Discovery can be deployed on any recent version of Kubernetes
+(v1.21+).
```
ArangoGutierrez, line 10: lgtm now

PR #1278, opened 2023-07-21
docs: fix toc of topology-updater and topology-gc reference

marquiz:
Exclude the main title from to (with the empty line the "no_toc" directive took no effect).


ArangoGutierrez (approved):
/lgtm

PR #1280, opened 2023-07-21
docs: remove useless TOCs

marquiz:
Drop table of contents from short pages where it is only cluttering the page.


ArangoGutierrez (approved):
/lgtm

PR #1282, opened 2023-07-21
Add optional labels to the podmonitor

jcpunk:
Prometheus is able to select monitors based on their labels.  This lets you easily set a label for the newly created monitor.


ArangoGutierrez (approved):
Awesome contribution, 1 request, could you add some documentation as well, thanks

marquiz (commented):
Thanks @ArangoGutierrez, an essential improvement indeed.

Could you also document this parameter in the table and possibly mention it in the metrics#helm section

jcpunk:
I took a stab at documentation for this parameter.

marquiz (approved):
Thanks @jcpunk for the update (and apilogies for calling you @ArangoGutierrez above 🙈)
/lgtm

PR #1284, opened 2023-07-27
generate: bump tools to their latest versions

marquiz:
Bump tools versions and re-auto-generate files.


ArangoGutierrez (approved):
Latest and greatest!!!
/lgtm

PR #1285, opened 2023-07-29
Feat: add expiry date for feature files

### LINKED ISSUE #857: Life-cycle management for external features

Implementation and documentation for life-cycle management of external features (split from #855).

**Why is this needed**:

Issues / missing:

* Documentation: what party is supposed to clean out the obsolete features and how?
* Functionality: what happens if e.g. Pod adds differently named hook binary or feature file on each run, or some test Pods run year ago were obsolete the next day, but their features still persist in /etc/? Does NFD just continue running them, slowing down with time, and bloating k8s nodes with obsolete labels?
* Documentation: what happens if multiple hooks or feature files update same labels, does NFD complain loudly, or just take the last value it gets?

**Proposed solution**:

* Features files should include best-before timestamp, after which NFD would ignore, or even remove them.  Timestamp could be either offset to feature file date, or absolute date
* Admin could put features directory to a tmpfs to make sure obsolete stuff with inappropriate timestamps get cleaned out eventually (on next reboot)
* NFD could trivially reject + log (maybe even remove) "obviously too large" feature files, to avoid features dir accidentally growing too large
* Duplicate handling:
  * Each label has origin ID (feature file name?)
  * Message is logged for each additional value for the same label
  * If new label value has a same origin value, last value is used, and message is just a warning.  Otherwise it's an error having more severe consequences

=> If pod changes the name of a feature file it installs, without changing all the label names, it must remove old feature file before installing the new one

Note: Above ignores life-cycle management for hooks, as they're going away (#856).

AhmedGrati:
Resolves #857.

This PR adds the expiration feature of feature files. The implemented solution suggests that the expiration date should follow this format:
```plaintext
# +expiry-time=2023-07-28T11:22:33Z
```

### DIFF
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -316,6 +316,14 @@ Label namespace may be specified with `<namespace>/<name>[=<value>]`.
 
 Comment lines (starting with `#`) are ignored.
 
+You can include the following block if you want to define the expiration date 
+of features that are described in the feature files:
+
+```plaintext
+# +expiry-time: 2023-07-29T11:22:33Z
+```
+**Note: The time format that we are supporting is RFC3339.**
+
 ### Mounts
 
 The standard NFD deployments contain `hostPath` mounts for
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -316,13 +316,37 @@ Label namespace may be specified with `<namespace>/<name>[=<value>]`.
 
 Comment lines (starting with `#`) are ignored.
 
-You can include the following block if you want to define the expiration date 
-of features that are described in the feature files:
+Adding following line anywhere to feature file defines date when
+its content expires / is ignored:
 
 ```plaintext
-# +expiry-time: 2023-07-29T11:22:33Z
+# +expiry-time=2023-07-29T11:22:33Z
 ```
-**Note: The time format that we are supporting is RFC3339.**
+
+Also, the expiry-time value would stay the same during the processing of the
+feature file until another expiry-time directive is encountered.
+Considering the following file:
+
+```plaintext
+# +expiry-time=2012-07-28T11:22:33Z
+featureKey=featureValue
+
+# +expiry-time=2080-07-28T11:22:33Z
+featureKey2=featureValue2
+
+# +expiry-time=2070-07-28T11:22:33Z
+featureKey3=featureValue3
+
+# +expiry-time=2002-07-28T11:22:33Z
+featureKey4=featureValue4
+```
+
+After processing the above file, only `featureKey2` and `featureKey3` would be
+included in the list of accepted features.
+
+> **NOTE:** The time format that we are supporting is RFC3339. Also, the `expiry-time`
+> tag is only evaluated in each re-discovery period, and the expiration of
+> node labels is not tracked.
 
 ### Mounts
 
--- a/source/local/local.go
+++ b/source/local/local.go
@@ -38,9 +38,12 @@ const Name = "local"
 // LabelFeature of this feature source
 const LabelFeature = "label"
 
-// ExpiryDateKey is the key of this feature source indicating
+// ExpiryTimeKey is the key of this feature source indicating
 // when features should be removed.
-const ExpiryDateKey = "# +expiry-time"
+const ExpiryTimeKey = "expiry-time"
+
+// DirectivePrefix defines the prefix of directives that should be parsed
+const DirectivePrefix = "# +"
 
 // Config
 var (
@@ -56,6 +59,11 @@ type localSource struct {
 
 type Config struct {
 	HooksEnabled bool `json:"hooksEnabled,omitempty"`
+}
+
+// parsingOpts contains options used for directives parsing
+type parsingOpts struct {
+	ExpiryTime time.Time
 }
 
 // Singleton source instance
@@ -149,14 +157,56 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
 	return s.features
 }
 
-func parseFeatures(lines [][]byte) map[string]string {
+func parseDirectives(line string, opts *parsingOpts) error {
+	if !strings.HasPrefix(line, DirectivePrefix) {
+		return nil
+	}
+
+	directive := line[len(DirectivePrefix):]
+	split := strings.SplitN(directive, "=", 2)
+	key := split[0]
+
+	if len(split) == 1 {
+		return fmt.Errorf("invalid directive format in %q, should be '# +key=value'", line)
+	}
+	value := split[1]
+
+	switch key {
+	case ExpiryTimeKey:
+		expiryDate, err := time.Parse(time.RFC3339, strings.TrimSpace(value))
+		if err != nil {
+			return fmt.Errorf("failed to parse expiry-date directive: %w", err)
+		}
+		opts.ExpiryTime = expiryDate
+	default:
+		return fmt.Errorf("unknown feature file directive %q", key)
+	}
+
+	return nil
+}
+
+func parseFeatures(lines [][]byte, fileName string) map[string]string {
 	features := make(map[string]string)
+	now := time.Now()
+	parsingOpts := &parsingOpts{
+		ExpiryTime: now,
+	}
 
 	for _, l := range lines {
 		line := strings.TrimSpace(string(l))
 		if len(line) > 0 {
-			// Skip comment lines
 			if strings.HasPrefix(line, "#") {
+				// Parse directives
+				err := parseDirectives(line, parsingOpts)
+				if err != nil {
+					klog.ErrorS(err, "error while parsing directives", "fileName", fileName)
+				}
+
+				continue
+			}
+
+			// handle expiration
+			if parsingOpts.ExpiryTime.Before(now) {
 				continue
 			}
 
@@ -202,7 +252,7 @@ func getFeaturesFromHooks() (map[string]string, error) {
 		}
 
 		// Append features
-		fileFeatures := parseFeatures(lines)
+		fileFeatures := parseFeatures(lines, fileName)
 		klog.V(4).InfoS("hook executed", "fileName", fileName, "features", utils.DelayedDumper(fileFeatures))
 		for k, v := range fileFeatures {
 			if old, ok := features[k]; ok {
@@ -278,18 +328,7 @@ func getFeaturesFromFiles() (map[string]string, error) {
 		}
 
 		// Append features
-		fileFeatures := parseFeatures(lines)
-
-		// Check expiration of file features
-		expiryDate, err := getExpirationDate(lines)
-		if err != nil {
-			klog.ErrorS(err, "failed to parse feature file expiry date", "fileName", fileName)
-			continue
-		}
-
-		if expiryDate.Before(time.Now()) {
-			continue
-		}
+		fileFeatures := parseFeatures(lines, fileName)
 
 		klog.V(4).InfoS("feature file read", "fileName", fileName, "features", utils.DelayedDumper(fileFeatures))
 		for k, v := range fileFeatures {
@@ -301,27 +340,6 @@ func getFeaturesFromFiles() (map[string]string, error) {
 	}
 
 	return features, nil
-}
-
-// Return the expiration date of a feature file
-func getExpirationDate(lines [][]byte) (time.Time, error) {
-	for _, line := range lines {
-		if len(line) > 0 {
-			lineSplit := strings.SplitN(string(line), ":", 2)
-
-			key := lineSplit[0]
-
-			if key == ExpiryDateKey {
-				expiryDate, err := time.Parse(time.RFC3339, lineSplit[1])
-				if err != nil {
-					return time.Now(), err
-				}
-				return expiryDate, nil
-			}
-		}
-	}
-
-	return time.Now(), nil
 }
 
 // Read one file
--- a/source/local/local.go
+++ b/source/local/local.go
@@ -276,22 +276,19 @@ func getFeaturesFromFiles() (map[string]string, error) {
 			klog.ErrorS(err, "failed to read file", "fileName", fileName)
 			continue
 		}
-		
+
 		// Append features
 		fileFeatures := parseFeatures(lines)
 
 		// Check expiration of file features
-		if expiryDate, ok := fileFeatures[ExpiryDateKey]; ok {
-			expiryDate, err := time.Parse(time.RFC3339, expiryDate)
-			if err != nil {
-				klog.ErrorS(err, "failed to parse feature file expiry date", "fileName", fileName)
-				continue
-			}
+		expiryDate, err := getExpirationDate(lines)
+		if err != nil {
+			klog.ErrorS(err, "failed to parse feature file expiry date", "fileName", fileName)
+			continue
+		}
 
-			// Features should not be included if they're expired
-			if expiryDate.Before(time.Now()) {
-				continue
-			}
+		if expiryDate.Before(time.Now()) {
+			continue
 		}
 
 		klog.V(4).InfoS("feature file read", "fileName", fileName, "features", utils.DelayedDumper(fileFeatures))
@@ -304,6 +301,27 @@ func getFeaturesFromFiles() (map[string]string, error) {
 	}
 
 	return features, nil
+}
+
+// Return the expiration date of a feature file
+func getExpirationDate(lines [][]byte) (time.Time, error) {
+	for _, line := range lines {
+		if len(line) > 0 {
+			lineSplit := strings.SplitN(string(line), ":", 2)
+
+			key := lineSplit[0]
+
+			if key == ExpiryDateKey {
+				expiryDate, err := time.Parse(time.RFC3339, lineSplit[1])
+				if err != nil {
+					return time.Now(), err
+				}
+				return expiryDate, nil
+			}
+		}
+	}
+
+	return time.Now(), nil
 }
 
 // Read one file
--- a/source/local/local.go
+++ b/source/local/local.go
@@ -23,6 +23,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"k8s.io/klog/v2"
 
@@ -36,6 +37,10 @@ const Name = "local"
 
 // LabelFeature of this feature source
 const LabelFeature = "label"
+
+// ExpiryDateKey is the key of this feature source indicating
+// when features should be removed.
+const ExpiryDateKey = "# +expiry-time"
 
 // Config
 var (
@@ -271,9 +276,24 @@ func getFeaturesFromFiles() (map[string]string, error) {
 			klog.ErrorS(err, "failed to read file", "fileName", fileName)
 			continue
 		}
-
+		
 		// Append features
 		fileFeatures := parseFeatures(lines)
+
+		// Check expiration of file features
+		if expiryDate, ok := fileFeatures[ExpiryDateKey]; ok {
+			expiryDate, err := time.Parse(time.RFC3339, expiryDate)
+			if err != nil {
+				klog.ErrorS(err, "failed to parse feature file expiry date", "fileName", fileName)
+				continue
+			}
+
+			// Features should not be included if they're expired
+			if expiryDate.Before(time.Now()) {
+				continue
+			}
+		}
+
 		klog.V(4).InfoS("feature file read", "fileName", fileName, "features", utils.DelayedDumper(fileFeatures))
 		for k, v := range fileFeatures {
 			if old, ok := features[k]; ok {
--- a/source/local/local_test.go
+++ b/source/local/local_test.go
@@ -17,7 +17,11 @@ limitations under the License.
 package local
 
 import (
+	"fmt"
+	"os"
+	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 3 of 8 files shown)


source/local/local_test.go
```diff
@@ -33,3 +35,22 @@ func TestLocalSource(t *testing.T) {
 	assert.Empty(t, l)
 
 }
+
+func TestGetExpirationDate(t *testing.T) {
+
```
ArangoGutierrez, line 40: stale empty line

ArangoGutierrez (changes requested):

source/local/local.go
```diff
@@ -280,6 +298,27 @@ func getFeaturesFromFiles() (map[string]string, error) {
...
+				fmt.Println(err)
+				if err != nil {
+					return time.Now(), err
+				}
+				return expiryDate, nil
+			}
+		}
+	}
+
+	return time.Now(), nil
```
marquiz, line 319: I think this is bad as it basically makes all files that don't have the `+expiry-time` directive expired (as this timestamp is checked like 1 microsecond after `Now()` and it has already passed)

AhmedGrati:
ping @ArangoGutierrez

source/local/local.go
```diff
@@ -268,6 +273,19 @@ func getFeaturesFromFiles() (map[string]string, error) {
 
 		// Append features
 		fileFeatures := parseFeatures(lines)
+
+		// Check expiration of file features
+		expiryDate, err := getExpirationDate(lines)
```
marquiz, line 278: Hmm, I just realized that we don't have support for `# comment` lines in the feature files (I thought we had). I think that should be added (inside `parseFeatures()`, I guess) , otherwise it tries to interpret the `# +expiry-time: ...` line as a feature label, too, and throw some errors in the logs.

You could make parsing `# +<some-tag>: <value>` (I don't know whether they should be called tags or directives or something else) a generic function that updates an opts struct or smth. We might want to add other tags/directives later.

Also, you could do this parsing inside `parseFeatures()` so that the tag/directive could effect only after it's encountered. This would add some flexibility. E.g. in the case of expiry-time there could be some features that don't expire and others that do (even have different expiry times). I don't know if this is needed, and we can have expiry-time as a file-wide global setting, but for some others we might choose differently.

WDYT?

marquiz (commented):
@AhmedGrati thank you for the enhancement. Some comments and speculation below. Looks good overall, but some refinements are needed before merging

source/local/local.go
```diff
@@ -157,7 +162,21 @@ func parseFeatures(lines [][]byte) map[string]string {
 			if len(lineSplit) == 1 {
 				features[key] = "true"
 			} else {
-				features[key] = lineSplit[1]
+				value := lineSplit[1]
```
marquiz, line 165: What I meant in https://github.com/kubernetes-sigs/node-feature-discovery/pull/1285#discussion_r1280770365 is that we should improve this parsing.

First, filter out all `# comments`, i.e. lines starting with a `#`). Plus probably do `stings.TrimSpace()` for each line to remove leading and trailing whitespace.

Second, for each "`#`" line do separate (generic) parsing to catch any `+<a-tag>=<a-value>`. This will make it easy to add new directives/tags in the future. Plus, we can catch unknown tags and log an error (e.g. if there is a typo like `+expiru-time=...`

Makes sense?

marquiz: #1293 and #1294 provide the first work item from above

AhmedGrati: Thanks for the clarification.

docs/usage/customization-guide.md
```diff
@@ -316,6 +316,17 @@ Label namespace may be specified with `<namespace>/<name>[=<value>]`.
...
+You can include the following block if you want to define the expiration date
+of features that are described in the feature files:
+
+```plaintext
+# +expiry-time=2023-07-29T11:22:33Z
+```
+
+**Note: The time format that we are supporting is RFC3339. Also, the `expiry-time`
+tag is only evaluated in each re-discovery period, and the expiration of
+node labels is not tracked.**
```
marquiz, line 328: nit: we just (re-)formatted all notes to a common format

```suggestion
> **NOTE:** The time format that we are supporting is RFC3339. Also, the `expiry-time`
> tag is only evaluated in each re-discovery period, and the expiration of
> node labels is not tracked.
```

source/local/local.go
```diff
@@ -272,8 +297,20 @@ func getFeaturesFromFiles() (map[string]string, error) {
 			continue
 		}
 
+		// Parse comments
+		comments := parseComments(lines, fileName)
```
marquiz, line 301: We' re already parsing line by line here. No need/sense to parse all comment lines per each line. I suggest to simplify this a lot. E.g. above if `HasPrefix(line, "#")` then do parseComment() for that (one) line.

source/local/local.go
```diff
@@ -272,8 +297,20 @@ func getFeaturesFromFiles() (map[string]string, error) {
...
+
+		// handle expiry date
+		if expiryDate, ok := comments[ExpiryDateKey]; ok {
+			if expiryDate.(time.Time).Before(time.Now()) {
```
marquiz, line 305: We need to take `time.Now()` in the beginning of this function so that (for consistency) the same timestamp is used for all lines in the file

source/local/local.go
```diff
@@ -144,17 +149,37 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
...
+func parseComments(lines [][]byte, fileName string) map[string]interface{} {
+	comments := make(map[string]interface{})
+	for _, l := range lines {
+		line := strings.TrimSpace(string(l))
+		if len(line) > 0 && strings.HasPrefix(line, "#") {
+			lineSplit := strings.SplitN(line, "=", 2)
+
+			key := lineSplit[0]
+			value := lineSplit[1]
+			switch key {
```
marquiz, line 161: Make this more generic to check if the comment starts with a `+`, in that case extract the key (and value)

source/local/local.go
```diff
@@ -144,17 +149,37 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
...
+			case ExpiryDateKey:
+				expiryDate, err := time.Parse(time.RFC3339, strings.TrimSpace(value))
+				if err != nil {
+					klog.ErrorS(err, "failed to parse feature file expiry date", "fileName", fileName)
+					continue
+				}
+				comments[ExpiryDateKey] = expiryDate
+			default:
+				klog.ErrorS(fmt.Errorf("unknown comment encountered while parsing features' comments"), "unknown comment", "commentKey", key)
```
marquiz, line 170: We don't want to log all comment lines as errors (like `# Some regular comment` should just be skipped). Only if it's `# +some-key ...` we consider it as a tag/directive and log an error if we don't recognize the `some-key`

source/local/local.go
AhmedGrati: You mean in the `parseFeatures()` function we check for each line if it's a comment we parse it?

marquiz: Yeah, that's what I was thinking. In any case it doesn't make any sense to for every line to re-parse the whole file for comments (plus we lose the trimming of whitespace we did earlier).

source/local/local.go
```diff
@@ -37,6 +38,13 @@ const Name = "local"
 // LabelFeature of this feature source
 const LabelFeature = "label"
 
+// ExpiryDateKey is the key of this feature source indicating
+// when features should be removed.
+const ExpiryDateKey = "# +expiry-time"
+
+// CommentsPrefix defines the prefix of comments that should be parsed
+const CommentsPrefix = "# +"
```
marquiz, line 46: nit: this is not comment prefix (comments start with `#`). Should be e.g. `TagPrefix` or `OptPrefix` or `DirectivePrefix`  or smth.

source/local/local.go
```diff
@@ -37,6 +38,13 @@ const Name = "local"
 // LabelFeature of this feature source
 const LabelFeature = "label"
 
+// ExpiryDateKey is the key of this feature source indicating
+// when features should be removed.
+const ExpiryDateKey = "# +expiry-time"
```
marquiz, line 43: If we want to define a const, align the name

```suggestion
const ExpiryTimeKey = "# +expiry-time"
```

source/local/local.go
```diff
@@ -144,17 +152,48 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
 	return s.features
 }
 
-func parseFeatures(lines [][]byte) map[string]string {
-	features := make(map[string]string)
+func parseComments(line string, fileName string) map[string]interface{} {
+	comments := make(map[string]interface{})
+	if strings.HasPrefix(line, CommentsPrefix) {
+		lineSplit := strings.SplitN(line, "=", 2)
```
marquiz, line 158: I think it would be cleaner to drop the `CommentPrefix` at this point. `ExpiryTimeKey` would become just `expiry-time`, without the `# +` prefix

source/local/local.go
```diff
@@ -144,17 +152,48 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
 	return s.features
 }
 
-func parseFeatures(lines [][]byte) map[string]string {
-	features := make(map[string]string)
+func parseComments(line string, fileName string) map[string]interface{} {
```
marquiz, line 155: Hmm,  I don't like the ambiguous `map[string]interface{}` here. How about doing smth like

```golang
type parsingOpts struct {
    ExpiryTime time.Time
}

func parseComment(line string, opts *ParsingOpts) error {
...
```

I.e. the parsing func would update an opts struct. Also return an error (can be then logged on the upper level) which makes it easier to (unit)test, too.

source/local/local.go
AhmedGrati: How would we differentiate then between comments and features? It can also be ambiguous for users.

marquiz: Comment lines start with a `#`.  This is now already supported in ` master`. We don't need to parse features in `parseComment()`

source/local/local.go
marquiz: No no 😅 What I tried to suggest is to not have the `# +` prefix included in the `ExpiryTimeKey` const. All tags/opts/directives would start with `# +` so we don't need to repeat it

source/local/local.go
AhmedGrati: Got it! I was confused by the previous comment above. 😅

AhmedGrati:
ping @marquiz.

docs/usage/customization-guide.md
```diff
@@ -316,6 +316,17 @@ Label namespace may be specified with `<namespace>/<name>[=<value>]`.
 
 Comment lines (starting with `#`) are ignored.
 
+You can include the following block if you want to define the expiration date
+of features that are described in the feature files:
```
eero-t, line 320: I would rephrase this simply as:
```suggestion
Adding following line anywhere to feature file defines date when its content expires / is ignored:
```

source/local/local_test.go
```diff
@@ -33,3 +36,43 @@ func TestLocalSource(t *testing.T) {
...
+		{
+			name:        "valid comment",
+			comment:     "# +expiry-time=2080-07-28T11:22:33Z",
+			expectedErr: nil,
+		},
+		{
+			name:        "unknown comment",
+			comment:     "# +random-key=random-value",
+			expectedErr: ErrorUnknownDirective,
+		},
```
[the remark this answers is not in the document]
AhmedGrati: For the `# + something` I added it. However, the comment `# Notes:` would not even pass to the `parseDirectives` function, because we are doing a filter before parsing directives, so general comments are excluded. Therefore, I added another file in the testdata folder that has the `# Notes:` comment.

source/local/local.go
```diff
@@ -144,17 +160,54 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
...
+				// Parse directives
+				err := parseDirectives(line, fileName, parsingOpts)
+				if err != nil {
+					klog.ErrorS(err, "error while parsing directives", "fileName", fileName)
+				}
+
+				// handle expiration
+				if parsingOpts.ExpiryTime.Before(now) {
+					return map[string]string{}
+				}
```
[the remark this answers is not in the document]
AhmedGrati: For this feature, we are only supporting one `expiry-time` for the whole file. it's not customizable for each label. Maybe if there's a need for that kind of customization we can adjust. wdyt @marquiz @eero-t?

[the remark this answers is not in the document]
eero-t: Making it more flexible would be just one line change (+ doc update).   Otherwise I think code should return error if there are features specified before directives (+ doc change).

source/local/local_test.go
```diff
@@ -33,3 +36,53 @@ func TestLocalSource(t *testing.T) {
...
+			directive:   "# +random-key=random-value",
+			expectedErr: ErrorUnknownDirective,
+		},
+		{
+			name:        "unknown directive",
+			directive:   "# +random-key=random-value",
+			expectedErr: ErrorUnknownDirective,
+		},
+		{
+			name:        "unkown directive structure",
```
eero-t, line 73:
```suggestion
			name:        "unknown directive format",
```
(Or maybe "invalid" would suit better here than "unknown".)

source/local/local.go
```diff
@@ -144,17 +163,57 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
 	return s.features
 }
 
-func parseFeatures(lines [][]byte) map[string]string {
+func parseDirectives(line string, fileName string, opts *parsingOpts) error {
```
marquiz, line 166: I think it would be just simpler to not care about the filename here. Just return an error and log it in the caller function.

```suggestion
func parseDirectives(line string, opts *parsingOpts) error {
```

source/local/local.go
```diff
@@ -144,17 +163,57 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
...
+		return ErrorWrongDirectiveFormat
+	}
+	value := lineSplit[1]
+
+	switch key {
+	case ExpiryTimeKey:
+		expiryDate, err := time.Parse(time.RFC3339, strings.TrimSpace(value))
+		if err != nil {
+			klog.ErrorS(err, "failed to parse feature file expiry date", "fileName", fileName)
+			return err
```
marquiz, line 183: Why not just return an error?

```suggestion
			return fmt.Errorf(err, "failed to parse expiry-date directive: %w", err)
```

source/local/local.go
```diff
@@ -144,17 +163,57 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
...
+		}
+		opts.ExpiryTime = expiryDate
+	default:
+		klog.ErrorS(ErrorUnknownDirective, "unknown directive", "directiveKey", key)
+		return ErrorUnknownDirective
```
marquiz, line 188: Same here

```suggestion
		return fmt.Errorf("unknown feature file directive %q", key)
```

...or something

source/local/local.go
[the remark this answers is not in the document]
marquiz: Thanks for review @eero-t . Now that I look at this with fresher eyes after a break I think @eero-t's comment/suggestion above makes sense.

I feel that the logical thing to do would be to treat this like batch processing/shell scripts: read and execute line by line. The directive takes effect when it is encountered, think about bash `set +x ... set -x for example. That is, you could have different expiry times for different features as outlined above). If would be directives that are for the whole file, they should be placed at the very top of the file (like go build tags or bash shebang or similar). I don't think we need to have this limitation for the `expiry-time` directive.

WDYT @AhmedGrati @eero-t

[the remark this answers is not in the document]
eero-t: To me that makes sense.

Doc would then need to be changed to something like this:
> `expiry-time` option defines date/time when all features listed after it will expire / are ignored:
> ```
> # +expiry-time=2023-12-31
> yearly_label=value1
> 
> # +expiry-time=2023-08-31
> monthly_label=value2
> ```
> (Meaning that any `expiry-time` having later date/time than previous ones, will have no effect.)

PS. If support for directives that can apply only to whole file is added later on, it's more consistent for the user if their implementation requires that they are set before any of the features are specified.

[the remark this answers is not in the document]
marquiz: > If support for directives that can apply only to whole file is added later on, it's more consistent for the user if their implementation requires that they are set before any of the features are specified.

Yes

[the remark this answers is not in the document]
AhmedGrati: @marquiz how should we know what are the specific features that a directive would be applied on? e.g.: Should we apply the expiry directive on all features until we encounter the same directive key, or we should have a delimiter like `\n`? 

Also, we should consider the case where we can add more directives, how should we handle these batches? e.g.:
```
# +expiry-time=2023-12-31
# +key1=value1
yearly_label=value1

# +expiry-time=2023-08-31
monthly_label=value2
```

would `key1` be applied only on `yearly_label`, or on both `yearly_label` and `monthly_label`.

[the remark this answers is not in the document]
eero-t: It depends at which point in the file NFD encounters expiration.

If it's 2024, none of them is applied, because file processing expires on first line.

If its late 2023, processing expires on the `+expiry-time=2023-08-31` line.

If it's early 2023, end of that file is reached without file processing expiring.

This is all you need for that (besides doc update):
```suggestion
				if parsingOpts.ExpiryTime.Before(now) {
					return feature
				}
```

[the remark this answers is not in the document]
marquiz: To make it stupid and simple, I think those directives could/should be treated like "variables". Expiry-time (or whatever key) is applied to all features until a new expiry time is specified. WDYT?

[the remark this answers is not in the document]
AhmedGrati: @marquiz I got your point. The question is, when we encounter an expired feature, should we continue parsing the rest of the file, or we just return features that we already parsed and not expired?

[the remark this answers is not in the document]
marquiz: I think the logical thing to do would be to always process the whole file

docs/usage/customization-guide.md
```diff
@@ -316,6 +316,38 @@ Label namespace may be specified with `<namespace>/<name>[=<value>]`.
...
+featureKey2=featureValue2
+
+# +expiry-time=2070-07-28T11:22:33Z
+featureKey3=featureValue3
+
+# +expiry-time=2002-07-28T11:22:33Z
+featureKey4=featureValue4
+```
+
+After processing the above file, only `featureKey` and `featureKey3` would be
```
marquiz, line 344: Wouldn't it be 2 and 3?
```suggestion
After processing the above file, only `featureKey2` and `featureKey3` would be
```

source/local/local.go
```diff
@@ -144,14 +160,54 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
...
+	lineSplit := strings.SplitN(line, "=", 2)
+
+	// If we have the following directive: # +expiry-time=xxx
+	// we need to get the key '# +expiry-time' and then we need to extract
+	// the following key 'expiry-time'.
+	key := strings.Split(lineSplit[0], DirectivePrefix)[1]
+	if len(lineSplit) == 1 {
+		return ErrorInvalidDirectiveFormat
+	}
+	value := lineSplit[1]
```
marquiz, line 173: Using `strings.Split()` for dropping the prefix looks odd and might cause unexpected results in the future (if the value of some directive contained `# +`). How about moving the checking of the prefix from `parseFeatures` over to this `parseDirectives` function?

```suggestion
	if !strings.HasPrefix(line, DirectivePrefix) {
		return nil
	}
	directive := line[len(DirectivePrefix):len(line)]
	split := strings.SplitN(directive, "=", 2)
	key := split[0]
```

source/local/local.go
```diff
@@ -144,14 +160,54 @@ func (s *localSource) GetFeatures() *nfdv1alpha1.Features {
...
+	parsingOpts := &parsingOpts{
+		ExpiryTime: now,
+	}
 
 	for _, l := range lines {
 		line := strings.TrimSpace(string(l))
 		if len(line) > 0 {
-			// Skip comment lines
-			if strings.HasPrefix(line, "#") {
+			if strings.HasPrefix(line, DirectivePrefix) {
```
marquiz, line 199: We should still allow "normal" comment lines (like `# Note: foo bar`) without trying to parse these as features

```suggestion
			if strings.HasPrefix(line, "#") {
```

marquiz (commented):
Thanks for the update @AhmedGrati. A few more comments below

AhmedGrati:
ping @marquiz.

marquiz:
Thanks @AhmedGrati, I only had one more small suggestion

@eero-t @ArangoGutierrez do you have anything to add? I think we can squeeze this into v0.14
/milestone v0.14

source/local/testdata/features.d/mutliple_expiration_dates
```diff
@@ -0,0 +1,11 @@
+# +expiry-time=2012-07-28T11:22:33Z
```
eero-t, line 1: filename: mutliple_expiration_dates -> multiple_expiration_dates

eero-t (commented):
Did cursory review.

Besides the file name typo, I wondered whether feature handling has been tested with e.g. 1GB binary file?

I mean, `local.go` does not size checks, it just reads whole thing in, splits it by newlines, before passing it to feature parsing...

AhmedGrati:
@marquiz @eero-t Thanks for the review. For the 1GB file test, I didn't actually do that. However, I think it can be optimized by reading the file in chunks. I think that it can be done in another PR. wdyt?

marquiz:
> @marquiz @eero-t Thanks for the review. For the 1GB file test, I didn't actually do that. However, I think it can be optimized by reading the file in chunks. I think that it can be done in another PR. wdyt?

I think that's an issue that should be addressed in a separate PR. The feature files are considered to be from a trusted source so we haven't been too concerned about this. We could put a limit for the maximum size of the file, e.g. 64kB or even 32kB should definitely be enough for a big number of labels and massive comments

marquiz (approved):
Thanks @AhmedGrati for this enhancement (and @eero-t for the reviews). This now looks good to me

/assign @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

Issue #1286, opened 2023-07-31
nfd-topology-updater: can't run normally when there are empty huagepages in some numa nodes


freelizhun:
**What happened**:
nfd-topology-updater can't running normally when there are empty huagepages in some numa nodes
eg: 
```bash
[root@master1 node-feature-discovery]# kubectl -n node-feature-discovery get pods -o wide
NAME                               READY   STATUS             RESTARTS       AGE     IP             NODE      NOMINATED NODE   READINESS GATES
nfd-topology-updater-mwhx2         1/1     Running            0              7m52s   10.119.1.203   node1     <none>           <none>
nfd-topology-updater-rv6kp         0/1     CrashLoopBackOff   6 (117s ago)   7m52s   10.119.0.75    master1   <none>           <none>

[root@master1 node-feature-discovery]# kubectl -n node-feature-discovery logs nfd-topology-updater-rv6kp 
I0731 08:11:40.262277       1 nfd-topology-updater.go:127] "Node Feature Discovery Topology Updater" version="v0.14.0-devel-161-ge0f10a81-dirty" nodeName="master1"
I0731 08:11:40.262395       1 util_unix.go:103] "Using this endpoint is deprecated, please consider using full URL format" endpoint="/host-var/lib/kubelet/pod-resources/kubelet.sock" URL="unix:///host-var/lib/kubelet/pod-resources/kubelet.sock"
I0731 08:11:40.262455       1 component.go:36] [core][Channel #1] Channel created
I0731 08:11:40.262473       1 component.go:36] [core][Channel #1] original dial target is: "/host-var/lib/kubelet/pod-resources/kubelet.sock"
I0731 08:11:40.262506       1 component.go:36] [core][Channel #1] parsed dial target is: {Scheme: Authority: Endpoint:host-var/lib/kubelet/pod-resources/kubelet.sock URL:{Scheme: Opaque: User: Host: Path:/host-var/lib/kubelet/pod-resources/kubelet.sock RawPath: OmitHost:false ForceQuery:false RawQuery: Fragment: RawFragment:}}
I0731 08:11:40.262519       1 component.go:36] [core][Channel #1] fallback to scheme "passthrough"
[...]

PR #1287, opened 2023-07-31
fix empty hugepages in some numa nodes caused no such file or directory errors

freelizhun:


freelizhun:
fixes #1286

pkg/utils/memory_resources.go
```diff
@@ -67,6 +67,10 @@ func GetNumaMemoryResources() (NumaMemoryResources, error) {
 		info[corev1.ResourceMemory] = nodeTotalMemory
 
 		// Get hugepages
+		_, err = os.Stat(filepath.Join(sysBusNodeBasepath, numaNode, "hugepages"))
```
marquiz, line 70: nit: I suggest to put the path in a variable and use that same value in the `getHugepagesBytes()` below to make it evident and sure that we actually use the same path

/cc @PiotrProkop @ffromani

marquiz: ping @freelizhun

freelizhun: is this ok ? @marquiz
```golang
               // Get hugepages
		huagePagePath := filepath.Join(sysBusNodeBasepath, numaNode, "hugepages")
		_, err = os.Stat(huagePagePath)
		if err != nil && os.IsNotExist(err) {
			continue
		}
		hugepageBytes, err := getHugepagesBytes(huagePagePath)
```

marquiz: Yeah, sure

pkg/utils/memory_resources.go
```diff
@@ -67,7 +67,12 @@ func GetNumaMemoryResources() (NumaMemoryResources, error) {
 		info[corev1.ResourceMemory] = nodeTotalMemory
 
 		// Get hugepages
-		hugepageBytes, err := getHugepagesBytes(filepath.Join(sysBusNodeBasepath, numaNode, "hugepages"))
```
ffromani, line 70: would it work if we just swallow IsNotExist here?
```
hugepageBytes, err := getHugepagesBytes(filepath.Join(sysBusNodeBasepath, numaNode, "hugepages"))
if err != nil && os.IsNotExist(err) {
    continue
}
```

ffromani (commented):
I think this works but I'd like the explore alternative implementions

pkg/utils/memory_resources.go
marquiz: Good point 😅 I think it should work

freelizhun: Ok, i will change it

pkg/utils/memory_resources.go
```diff
@@ -68,8 +68,8 @@ func GetNumaMemoryResources() (NumaMemoryResources, error) {
 
 		// Get hugepages
 		hugepageBytes, err := getHugepagesBytes(filepath.Join(sysBusNodeBasepath, numaNode, "hugepages"))
-		if err != nil {
-			return nil, err
+		if err != nil && os.IsNotExist(err) {
+			continue
```
marquiz, line 72: We still want to return an error is other error cases

freelizhun: this may be the best chocies?😅
```
               // Get hugepages
		huagePagePath := filepath.Join(sysBusNodeBasepath, numaNode, "hugepages")
		_, err = os.Stat(huagePagePath)
		if err != nil && os.IsNotExist(err) {
			continue
		}
		hugepageBytes, err := getHugepagesBytes(huagePagePath)
```

marquiz: How about using `else`?

freelizhun: Thank for your suggestion,@marquiz,is this ok?

marquiz: Looks legit to me 👍

marquiz (approved):
Looks good to me
ping @ffromani @PiotrProkop

ffromani (commented):
/lgtm

thanks for the updates!

ffromani:
/test pull-node-feature-discovery-verify-master

is this a flake? Let's retry 1 (one) time

marquiz:
It's the stupid logcheck that is flaky, maybe it should be disabled :/
/retest

PR #1288, opened 2023-07-31
Improve metrics

marquiz:
Some changes to the initial metrics we added:
- improve nfr processing time metric
  Change the metric from a simple gauge (that basically was a single value
for the whole cluster) into a HistogramVec, aligning with the feature
discovery duration metric in nfd-worker. This improved metric now has
prometheus labels for the NFR name and node name, i.e. it is tracking
per-NFR metric for each node being processed. Also, change the naming to
better comply with prometheus suggested conventions.
- rename the node updates gauge
  Better describe what we're measuring and better comply with prometheus
naming conventions.
- rename prometheus labell
  Use "node" instead of "NodeName". Aligns with prometheus conventions.


ArangoGutierrez (approved):
/lgtm

marquiz:
Some testing to do
/hold

marquiz:
Changed the "node updates" gauge to count actual updates of the node objects on the api server.

Also configured buckets for the NFR processing time and feature discovery time histograms. Only time and practice will tell how useful these 🙄
/unhold

ArangoGutierrez (approved):
/lgtm

PR #1291, opened 2023-08-01
nfd-master: use term node update instead of labeling

marquiz:
Rename symbols and reword log messages to correlate with the
functionality (we may do other updates than just modify labels
nowadays).

This PR also containers a oneliner patch for removing one stale empty line from the codebase.


AhmedGrati (approved):
Good catch thanks!

fmuyassarov (commented):
/lgtm

PR #1295, opened 2023-08-04
nfd-topology-updater: add metrics support

### LINKED ISSUE #1281: Topology-updater metrics

Add metrics support for nfd-topology-updater (and nfd-topology-gc?). From https://github.com/kubernetes-sigs/node-feature-discovery/issues/1162#issuecomment-1578007917 from @Tal-or 👍 

- Number of updates done to NRT objects, each update increase the counter by one.
- Number of PodresourceAPI calls, each call increases the counter by one.
- We can also expose the time the update took for each node

NOTE: tracking as a separate issue from metrics improvements for nfd-master and nfd-worker (#1275)

**Why is this needed**:

Observability for nfd-topology-updater, get it on-par with nfd-master and nfd-worker.

marquiz:
For now, add only one metric, a counter for the errors occurring while scanning pod resources on the node.

Fixes #1281


marquiz:
Not fully tested/verified
/hold

ffromani (commented):
LGTM. Do we want to add a e2e tests which queries the metrics endpoint (could do a simple `curl -L` from localhost) and checks any data is present?

marquiz:
> LGTM.

Seems to work
/unhold

> Do we want to add a e2e tests which queries the metrics endpoint (could do a simple `curl -L` from localhost) and checks any data is present?

I think we should. But probably in a separate PR for all the daemons(?)

ffromani:
> > Do we want to add a e2e tests which queries the metrics endpoint (could do a simple `curl -L` from localhost) and checks any data is present?
> 
> I think we should. But probably in a separate PR for all the daemons(?)

very fair.

ffromani:
LGTM, not adding real tag to enable other reviewers chime in.

PR #1297, opened 2023-08-07
topology-updater: make -version always runnable

marquiz:
Make it possible to run -version in an environment whithout the NODE_ADDRESS environment variable set.


ArangoGutierrez (approved):
/lgtm

PR #1305, opened 2023-08-18
Garbage collection of NodeFeature objects

### LINKED ISSUE #1304: NodeFeature CRD not deleted when node is deleted
**What happened**:
When running with `"enableNodeFeatureApi": true` I've noticed that NodeFeature CRDs seem to accumulate endlessly. Our workload scales up and down frequently and the NodeFeature CRDs for deleted nodes are not removed. After a couple months, this led to almost five thousand NodeFeatures resources in one of our clusters, at which point it seems like NFD runs into intermittent performance/scale issues where new nodes sometimes do not get NodeFeatures created for them.
 
I expect a NodeFeature to be deleted when the node in question is removed.

1. Install the Node Feature Discovery Helm chart with `"enableNodeFeatureApi": true`
2. Scale up the cluster (provider specific: [here are instructions](https://learn.microsoft.com/en-us/azure/aks/scale-cluster?tabs=azure-cli) for how to do this in Azure Kubernetes Service)
3. Confirm that a NodeFeature has been created for the new node.
4. Scale down the cluster
5. Confirm that the NodeFeature for the deleted node still exists (this is the bug)

**Environment**:
- Kubernetes version (use `kubectl version`): 1.26.3
- Cloud provider or hardware configuration: Azure Kubernetes Service
- OS (e.g: `cat /etc/os-release`): Ubuntu 20.04.1 LTS
- Kernel (e.g. `uname -a`): `Linux <vmname here> 5.15.0-1039-azure #46-Ubuntu SMP Mon May 22 15:18:07 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
- Install tools: Helm
- Network plugin and version (if this is a network-related bug): Not networking
- Others:

marquiz:
This PR implements garbage collection for NodeFeature objects. It achieves this by renaming the `nfd-topology-gc` daemon to more generic `nfd-gc` and extends it to clean up stale NodeFeatures.
 
Fixes #1304


marquiz:
This is a "mega-PR" also containing some refactoring. I could try to split out some smaller chunks if that would help the review.

/assign @ArangoGutierrez @PiotrProkop

marquiz:
Rebased, still depends on #1311 but otherwise ready

pkg/nfd-gc/nfd-gc.go
```diff
@@ -126,20 +153,40 @@ func (n *nfdGarbageCollector) garbageCollect() {
...
 	}
 
-	nrts, err := n.topoClient.TopologyV1alpha2().NodeResourceTopologies().List(context.TODO(), metav1.ListOptions{})
-	if err != nil {
-		klog.ErrorS(err, "failed to list NodeResourceTopology objects")
-		return
+	// Handle NodeFeature objects
+	nfs, err := n.nfdClient.NfdV1alpha1().NodeFeatures("").List(context.TODO(), metav1.ListOptions{})
+	if errors.IsNotFound(err) {
+		klog.V(2).InfoS("NodeFeature CRD does not exist")
```
fmuyassarov, line 159: nit: Perhaps should be CR instead of CRD?

fmuyassarov (commented):
Looks good to me but I will leave it for others also to review. Once you think it has gone through enough reviews, please tag me and i will add lgtm label.

pkg/nfd-gc/nfd-gc.go
marquiz: In this case it is CRD indeed. If you cannot list the resource (and apiserver returns notfound) it means the whole api resource type does not exist

marquiz:
Now all dependencies (i.e. PRs) have been merged and this is ready for review

pkg/nfd-gc/nfd-gc.go
```diff
@@ -126,26 +153,46 @@ func (n *topologyGC) garbageCollect() {
...
+	} else {
+		for _, nrt := range nrts.Items {
+			key, err := cache.MetaNamespaceKeyFunc(&nrt)
+			if err != nil {
+				klog.ErrorS(err, "failed to create key", "noderesourcetopology", klog.KObj(&nrt))
+				continue
+			}
+			if !nodeNames.Has(key) {
+				n.deleteNRT(key)
+			}
```
PiotrProkop, line 189:
```suggestion
			if !nodeNames.Has(nrt.Name) {
				n.deleteNRT(key)
			}
```

marquiz: Thanks @PiotrProkop, makes total sense (I was blindly moving code without thinking much). I incorporated this change in the ` nfd-gc: support garbage collection of NodeFeatures` patch

PiotrProkop:
Good job!
/lgtm

fmuyassarov (commented):
/lgtm
Thanks @marquiz 
Feel free to remove the hold.

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -450,7 +450,7 @@ topologyUpdater:
   affinity: {}
   podSetFingerprint: true
 
-topologyGC:
+gc:
   enable: true
```
ArangoGutierrez, line 454: I Think we should tie this to `enableNodeFeatureApi: false` (line 13 of the same file) 
why having it set to `true` if currently default `enableNodeFeatureApi` is false?

ArangoGutierrez (changes requested):

deployment/helm/node-feature-discovery/values.yaml
ArangoGutierrez: in the `nfd-gc.yaml` we could 
```yaml
{{- if and .Values.enableNodeFeatureApi (or .Values.topologyUpdater.enable) -}}
```

something like that, could work?.....

PR #1307, opened 2023-08-18
topology-gc: refactor unit tests

marquiz:
Remove a lot of boilerplate code by defining reusable functions.

Also, test the `Run()` method instead of the functions callees of `Run()` as it is the top level functionality that was tested in practice (we don't have separate unit tests for the callee functions).


marquiz:
Depends on #1306 
/hold

marquiz:
#1306 merged
/unhold

pkg/nfd-topology-gc/nfd-nrt-gc_test.go
```diff
@@ -22,212 +22,130 @@ import (
...
+			topoClient: faketopologyv1alpha2.NewSimpleClientset(createFakeNRTs(nrts...)...),
+			stopChan:   make(chan struct{}, 1),
+			gcPeriod:   10 * time.Minute,
+		},
+		k8sClient: k8sClient,
+	}
+}
 
-		stopChan := make(chan struct{}, 1)
+func createFakeNodes(names ...string) []runtime.Object {
```
PiotrProkop, line 106: it would be more readable to return `[]*corev1.Node`, wdyt?

pkg/nfd-topology-gc/nfd-nrt-gc_test.go
```diff
@@ -22,212 +22,130 @@ import (
...
+}
 
-		factory := informers.NewSharedInformerFactory(k8sClient, 5*time.Minute)
-		gc := &topologyGC{
-			factory:    factory,
-			topoClient: fakeClient,
-			stopChan:   stopChan,
-			gcPeriod:   time.Second,
-		}
+func createFakeNRTs(names ...string) []runtime.Object {
```
PiotrProkop, line 117: ditto

PiotrProkop (commented):
small comment, otherwise looks good

pkg/nfd-topology-gc/nfd-nrt-gc_test.go
marquiz: It would (and I actually wrote it like that initially) but `NewSimpleClientset()` wants `runtime.Object` type so I changed it like this to avoid converting `[]*corev1.Node` to `[]runtime.Object` (plus the same conversion loop for NRTs).

PiotrProkop: Aaa ok, slice to interface is a different type than interface. Still a bit rusty after PTO 😄

marquiz: Yes, that's the problem

PR #1319, opened 2023-08-30
docs: use ruby docker image for building docs

marquiz:
Get away with the jekyll:3.8 image which is already four years old.  Use
the ruby instead. The jekyll image did not bring any value (more
problems, if anything) as we install/update jekyll and all other gems
with byndler nevertheless (image was jekyll:3.8 but we use jekyll
v3.9.3).

Also updates docs build dependencies.

Replaces #1316 and #1317


fmuyassarov (commented):
/lgtm

Issue #1312, opened 2023-08-27
Allocatable CPU not updating


cl-rf:
**What happened**:
Deploying a pod with a cpu request/limit does not change the allocatable value for the noderesourcetopology.

**What you expected to happen**:
I expect the value to decrease based on the request/limit of the pod.

**How to reproduce it (as minimally and precisely as possible)**:
 `kubectl get noderesourcetopology.topology.node.k8s.io -o yaml`
```
apiVersion: v1
items:
- apiVersion: topology.node.k8s.io/v1alpha2
  attributes:
  - name: topologyManagerPolicy
    value: single-numa-node
  - name: topologyManagerScope
    value: container
  kind: NodeResourceTopology
  metadata:
    creationTimestamp: "2023-08-26T19:04:07Z"
    generation: 3242
    name: ip-1.1.1.1-west-1.compute.internal
    resourceVersion: "67906"
    uid: 3f1f8a64-edbf-459a-a51b-12404b1f40d4
  topologyPolicies:
  - SingleNUMANodeContainerLevel
  zones:
  - costs:
    - name: node-0
      value: 10
    - name: node-1
      value: 21
    name: node-0
    resources:
    - allocatable: "32"
      available: "32"
      capacity: "36"
      name: cpu
    - allocatable: "98327445504"
      available: "98327445504"
      capacity: "100474929152"
      name: memory
    type: Node
  - costs:
    - name: node-0
      value: 21
    - name: node-1
      value: 10
    name: node-1
    resources:
    - allocatable: "36"
      available: "36"
      capacity: "36"
      name: cpu
    - allocatable: "99261173760"
      available: "99261173760"
      capacity: "101408657408"
      name: memory
    type: Node
kind: List
metadata:
  resourceVersion: ""`
```

Deploy a test.

```
`apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deployment1
spec:
  selector:
      matchLabels:
        name: test
  template:
    metadata:
      labels:
        name: test
    spec:
     containers:
      - name: test-deployment-1-container-1
        image: quay.io/fromani/numalign
        imagePullPolicy: IfNotPresent
        command: ["/bin/sh", "-c"]
        args: [ "while true; do numalign; sleep 100000; done;" ]
        resources:
          limits:
            cpu: 20
          requests:
            cpu: 20`
```

The available and allocatable does not decrease.

```
`apiVersion: v1
items:
[...]

PR #1321, opened 2023-08-31
docs: nfd-updater: clarify accounting

ffromani:
Clarify that we account, and we can account, only
resources exclusively allocated to Guaranteed QoS pods.

Fixes: #1312

### DIFF
--- a/docs/usage/nfd-topology-updater.md
+++ b/docs/usage/nfd-topology-updater.md
@@ -17,6 +17,15 @@ resources and hence the allocatable resources on a per-zone basis by updating
 It makes sure that new NodeResourceTopology instances are created for each new
 nodes that get added to the cluster.
 
+Because of the design and implementation of Kubernetes, only resources exclusively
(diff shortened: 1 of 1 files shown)


marquiz (commented):
Thanks @ffromani for the update. Even though the potential readers of this are hopefully familiar with the concept of Pod QoS classes, I was thinking should we add some reference to 
https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#guaranteed
and/or
https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy

WDYT?

ffromani:
> Thanks @ffromani for the update. Even though the potential readers of this are hopefully familiar with the concept of Pod QoS classes, I was thinking should we add some reference to https://kubernetes.io/docs/concepts/workloads/pods/pod-qos/#guaranteed and/or https://kubernetes.io/docs/tasks/administer-cluster/cpu-management-policies/#static-policy
> 
> WDYT?

Thanks @marquiz . Totally makes sense. I'll update shortly.

ffromani:
> ping @ffromani

thanks, updated

marquiz (approved):
Looks good
/assign @PiotrProkop

PR #1325, opened 2023-09-01
nfd-updater: events: enable timer-only flow

ffromani:
The nfd-topology-updater has state-directories notification enabled by default. In theory, we can have only timer-based updates, but if the option is given to disable the state-directories event source, then all the update mechanism is mistakenly disabled, including the timer-based updates.

The two updaters mechanism should be decoupled.
So this PR changes this to make sure we can enable just and only the timer-based updates.

WIP: missing tests (probably e2e tests is the "only" tests we need, but still, missing)

### DIFF
--- a/pkg/nfd-topology-updater/kubeletnotifier/kubeletnotifier.go
+++ b/pkg/nfd-topology-updater/kubeletnotifier/kubeletnotifier.go
@@ -54,26 +54,35 @@ type Info struct {
 }
 
 func New(sleepInterval time.Duration, dest chan<- Info, kubeletStateDir string) (*Notifier, error) {
-	devicePluginsDir := path.Join(kubeletStateDir, devicePluginsDirName)
-	ch, err := createFSWatcherEvent([]string{kubeletStateDir, devicePluginsDir})
-	if err != nil {
-		return nil, err
-	}
-	return &Notifier{
+	notif := Notifier{
 		sleepInterval: sleepInterval,
 		dest:          dest,
-		fsEvent:       ch,
-	}, nil
+	}
+
+	if kubeletStateDir != "" {
+		devicePluginsDir := path.Join(kubeletStateDir, devicePluginsDirName)
+		ch, err := createFSWatcherEvent([]string{kubeletStateDir, devicePluginsDir})
+		if err != nil {
+			return nil, err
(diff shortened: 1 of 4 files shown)


ffromani:
/hold

figuring out a proper e2e test

marquiz (commented):
Thanks @ffromani, looks basically good to me. There are some errors in the CI to fix. Looks like we'd like to have this for v0.14(?)

EDIT: forgot to mention that does the documentation need any updates in this regard?

ffromani:
> Thanks @ffromani, looks basically good to me. There are some errors in the CI to fix. Looks like we'd like to have this for v0.14(?)
> 
> EDIT: forgot to mention that does the documentation need any updates in this regard?

thanks for the early feedback! I'm on it to fix CI issues and ensure non-regression. I don't have a ETA for missing tests to check the new (and correct) behavior so I can't say if this will make it in 0.14. I'll be happy to manage backports and not block 0.14.0, whatever you think is best for the project.

marquiz:
> I don't have a ETA for missing tests to check the new (and correct) behavior so I can't say if this will make it in 0.14. I'll be happy to manage backports and not block 0.14.0, whatever you think is best for the project.

I think we'd like to have v0.14 out in a week or two 😅 But yes, I think this could be backported as a bugfix for v0.14.1

ffromani:
/hold cancel

begun working on the e2e test

ffromani:
no longer WIP - I'm happy with the e2e test. Ready for review.

marquiz:
@ffromani Looks like the changes to `pkg/nfd-topology-updater/nfd-topology-updater.go` were reverted in the latest force push. To me this looks like reverting to the old behavior (where updates are disabled if `-kubelet-state-dir` is empty).

ffromani:
> @ffromani Looks like the changes to `pkg/nfd-topology-updater/nfd-topology-updater.go` were reverted in the latest force push. To me this looks like reverting to the old behavior (where updates are disabled if `-kubelet-state-dir` is empty).

Oops! silly me was testing making sure the old code was failing the test (which it does) and forgot the re-revert. Fixing now

PiotrProkop:
/lgtm

Thanks!

PR #1331, opened 2023-09-05
Helm: conditionally add annotations if defined

andrewjamesbrown:
With the current implementation, our deployment and daemonset resources are set with an empty set of annotations; versus skipping them like in [other](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/master/charts/aws-ebs-csi-driver/templates/node.yaml#L24) [helm](https://github.com/kubernetes-sigs/aws-efs-csi-driver/blob/master/charts/aws-efs-csi-driver/templates/controller-deployment.yaml#L26) [charts](https://github.com/kubernetes-sigs/external-dns/blob/master/charts/external-dns/templates/deployment.yaml#L31).

This PR proposes a similar method to the other charts, for consistency with the rest of `kubernetes-sigs`

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/master.yaml
+++ b/deployment/helm/node-feature-discovery/templates/master.yaml
@@ -6,8 +6,10 @@ metadata:
   labels:
     {{- include "node-feature-discovery.labels" . | nindent 4 }}
     role: master
+  {{- with .Values.master.deploymentAnnotations }}
   annotations:
-    {{- toYaml .Values.master.deploymentAnnotations | nindent 4 }}
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
   replicas: {{ .Values.master.replicaCount }}
   selector:
@@ -19,8 +21,10 @@ spec:
       labels:
         {{- include "node-feature-discovery.selectorLabels" . | nindent 8 }}
         role: master
+      {{- with .Values.master.annotations }}
       annotations:
-        {{- toYaml .Values.master.annotations | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
     spec:
     {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
--- a/deployment/helm/node-feature-discovery/templates/nfd-gc.yaml
+++ b/deployment/helm/node-feature-discovery/templates/nfd-gc.yaml
@@ -7,6 +7,10 @@ metadata:
   labels:
     {{- include "node-feature-discovery.labels" . | nindent 4 }}
     role: gc
+  {{- with .Values.gc.deploymentAnnotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
 spec:
   replicas: {{ .Values.gc.replicaCount | default 1 }}
   selector:
--- a/deployment/helm/node-feature-discovery/templates/nfd-gc.yaml
+++ b/deployment/helm/node-feature-discovery/templates/nfd-gc.yaml
@@ -18,8 +18,10 @@ spec:
       labels:
         {{- include "node-feature-discovery.selectorLabels" . | nindent 8 }}
         role: gc
+      {{- with .Values.gc.annotations }}
       annotations:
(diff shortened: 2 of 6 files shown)


deployment/helm/node-feature-discovery/templates/master.yaml
```diff
@@ -19,8 +21,10 @@ spec:
       labels:
         {{- include "node-feature-discovery.selectorLabels" . | nindent 8 }}
         role: master
+      {{- with .Values.master.annotations }}
       annotations:
-        {{- toYaml .Values.master.annotations | nindent 8 }}
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
```
marquiz, line 27: Is there some particular reason to change these? If not, I'd suggest not to change

deployment/helm/node-feature-discovery/templates/master.yaml
```diff
@@ -6,8 +6,10 @@ metadata:
   labels:
     {{- include "node-feature-discovery.labels" . | nindent 4 }}
     role: master
+  {{- with .Values.master.deploymentAnnotations }}
   annotations:
-    {{- toYaml .Values.master.deploymentAnnotations | nindent 4 }}
+    {{- toYaml . | nindent 4 }}
```
marquiz, line 11: I'm not a Helm expert but why not use the simpler oneliner format used elsewhere in this chart?

```suggestion
    {{- toYaml .Values.master.deploymentAnnotations | nindent 8 }}
```

deployment/helm/node-feature-discovery/templates/topologyupdater.yaml
marquiz, line 1: For consistency wrt other templates, please add also `.Values.topologyUpdater.daemonsetAnnotations`

deployment/helm/node-feature-discovery/templates/nfd-gc.yaml
marquiz, line 1: For consistency wrt other templates, please add also `.Values.gc.deploymentAnnotations`

marquiz (commented):
Thanks for the contribution @andrewjamesbrown. A few comments below.

Please also document the new parameters in `docs/deployment/helm.md` and by adding them to the `values.yaml`

deployment/helm/node-feature-discovery/templates/master.yaml
andrewjamesbrown: See my explanation in the [comment above](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1331/files#r1317319726)

deployment/helm/node-feature-discovery/templates/master.yaml
marquiz: @andrewjamesbrown thank you for the explanation. Makes sense. As said my Helm-fu is limited 😇 As a bit of a sidenote,  do you think that some other fields (like `labels`) should be handled similarly.

andrewjamesbrown: As far as I can tell, labels are always set through values/helpers, so they'll always be present in the default case.  Which means there's no reason to conditionally add them using `with`

marquiz: K, thanks for educating me

andrewjamesbrown:
> ping @andrewjamesbrown CLA: [#1331 (comment)](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1331#issuecomment-1707446300)

Yep... in progress on my end 👍

marquiz:
> Yep... in progress on my end 👍

👍 It would be cool to get this into v0.14 which is imminent

andrewjamesbrown:
@marquiz CLA is good to go 👍

marquiz:
Nice.

And @andrewjamesbrown thanks for adding the new `topologyUpdater.daemonsetAnnotations` and `gc.deploymentAnnotations` parameters. My request for them in this PR was unjustified (this PR was originally fixing existing stuff not adding new stuff), blame my hasty review 🙈

marquiz:
ping @ArangoGutierrez

PR #1336, opened 2023-09-06
sources/custom: convert static rules to new format

marquiz:
Also add a log message warning about legacy rule format.


ArangoGutierrez (approved):
/lgtm

PR #1338, opened 2023-09-06
nfd-master: add config file options for logging

### LINKED ISSUE #1206: nfd-master: logging parameters in config file

Make klog parameters part of the (dynamic) configuration file, similarly to what we have for [nfd-worker](https://kubernetes-sigs.github.io/node-feature-discovery/v0.13/reference/worker-configuration-reference.html#coreklog).

This feature would make it possible to change logging parameters without re-starting nfd-master daemon.

Config file example

```yaml
klog:
  v: "4"
```

**Why is this needed**:

Better usability, dynamic run-time configuration of logging.

AhmedGrati:
Resolves #1206.

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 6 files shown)


ArangoGutierrez (approved):
I like it! 
but makes me think, since now master and worker have the same functions on `cmd/nfd-*` should we have this `klog` flag parsing fuctions in a shared location? `pkg/utils/klog`?

ArangoGutierrez:
Do we want this in v0.14?

marquiz:
> Do we want this in v0.14?

I think this is a very low-risk addition that would be nice to have. Same functionality is already present in nfd-worker. I quickly verified this last night ✅

marquiz:
> but makes me think, since now master and worker have the same functions on `cmd/nfd-*` should we have this `klog` flag parsing fuctions in a shared location? `pkg/utils/klog`?

Good point. I think this could be future refactoring/enhancement. We could create an issue about this. WDYT?

PR #1350, opened 2023-09-08
cpu: drop the deprecated sgx and se labels

marquiz:
Drop the deprecated cpu-sgx.enabled and cpu-se.enabled labels and the corresponding "raw" features. These have been replaced by cpu-security.sgx.enabled and cpu-security.se.enabled.

Fixes: #832


marquiz:
/assign @fmuyassarov 

ping @mythi

PR #1353, opened 2023-09-08
feat: ignore hidden feature files

### LINKED ISSUE #1337: Create subdir for atomic feature file updates

NFD to document and create a subdir under its feature files dir, which NFD client apps can use for their atomic feature file creation.

**Why is this needed**:

Atomic feature file generation means that the file needs to be done elsewhere and renamed/moved to feature dir.   Because renaming works atomically only within same fs, best place for generating the feature file, is feature dir subdir (clients already have feature dir mounted as writable).

If NFD does not document & provide such dir, it means that clients need to.  That can results feature dir having large number extra entries, for per-client subdirs.  Which can be a problem, if NFD will at some later point start parsing also feature dir subdir content...

AhmedGrati:
Resolves #1337.


AhmedGrati:
/cc @ArangoGutierrez @marquiz 
ping @eero-t @tkatila.

eero-t:
Please describe in docs also why hidden files should be used.    This should preferably be in the same place as where NFD docs say that feature files should be updated atomically (as this change allows NFD client to write features to a dot file, and then atomically rename it to real feature file name).

docs/usage/customization-guide.md
```diff
@@ -237,7 +237,9 @@ further be used in label rules specified in
 > while NFD is running. In order to avoid race conditions you should write into
 > a temporary file (outside the `source.d` and `features.d` directories), and,
 > atomically create/update the original file by doing a filesystem move
-> operation.
+> operation. You can also use a dot file like `.my.feature` that contains
+> features, and then automatically update the file name to a real
+> feature file name like `my.feature`.
```
eero-t, line 242: Because using another dir can cause issues (if it happens to be on another file system), dot file use is supposed to be **the**  new recommendation, e.g. something like this:

> In order to avoid race conditions you should write into a temporary file , and atomically create/update the original file by doing a file rename operation.  NFD ignores dot files in this directory, so temporary file can be written to the same directory and renamed  (`.my.feature` -> `my.feature`) once file is complete.  Both file names  should (obviously) be unique for the given application.

eero-t (changes requested):

docs/usage/customization-guide.md
AhmedGrati: Thanks @eero-t. Updated!

AhmedGrati:
ping @marquiz

docs/usage/customization-guide.md
```diff
@@ -234,10 +234,12 @@ further be used in label rules specified in
 [`custom`](#custom-feature-source) feature source.
 
 > **NOTE:** Be careful when creating and/or updating hook or feature files
-> while NFD is running. In order to avoid race conditions you should write into
-> a temporary file (outside the `source.d` and `features.d` directories), and,
-> atomically create/update the original file by doing a filesystem move
-> operation.
+> while NFD is running. In order to avoid race conditions you should write
```
marquiz, line 237: We talk about feature files AND hooks here so we should then add the dot-file filter to `getFeaturesFromHooks()`, too

docs/usage/customization-guide.md
```diff
@@ -234,10 +234,12 @@ further be used in label rules specified in
...
+> into a temporary file, and atomically create/update the original file by
+> doing a file rename operation. NFD ignores dot files in this directory,
+> so temporary file can be written to the same directory and renamed
+> (.my.feature -> my.feature) once file is complete. Both file names should
```
marquiz, line 241:
```suggestion
> (`.my.feature` -> `my.feature`) once file is complete. Both file names should
```

docs/usage/customization-guide.md
```diff
@@ -234,10 +234,12 @@ further be used in label rules specified in
 [`custom`](#custom-feature-source) feature source.
 
 > **NOTE:** Be careful when creating and/or updating hook or feature files
-> while NFD is running. In order to avoid race conditions you should write into
-> a temporary file (outside the `source.d` and `features.d` directories), and,
-> atomically create/update the original file by doing a filesystem move
-> operation.
+> while NFD is running. In order to avoid race conditions you should write
+> into a temporary file, and atomically create/update the original file by
+> doing a file rename operation. NFD ignores dot files in this directory,
```
marquiz, line 239: IF we support dot-files for both feature files and hooks:

```suggestion
> doing a file rename operation. NFD ignores dot files,
```

marquiz (commented):
Looking at the docs, we now promise that dot-files are ignored for both feature files and hooks. That's probably a good idea (even if hooks are deprecated), see my comments below

eero-t:
> Looking at the docs, we now promise that dot-files are ignored for both feature files and hooks. That's probably a good idea (even if hooks are deprecated)

In the hook binary case, I would assume them to be moved to hook dir from somewhere else (e.g. installation container), which is atomic (at least as long as it's on the same file system) => IMHO dot file handling is not not really needed for hooks, but either updating doc again, or adding that feature also to hooks should both be fine.

AhmedGrati:
I added the same check to hooks also.

eero-t:
Test for checking that normally named feature file is parsed, but one prefixed with dot is not, would also be good.  This could be part of some other feature file test though, it does not necessarily need its own.

marquiz (approved):
Thanks @AhmedGrati, looks good to me now
/assign @ArangoGutierrez

AhmedGrati:
ping @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #1354, opened 2023-09-11
gh-pages: Add OWNERS file

marquiz:
The gh-pages branch is basically automatically managed by the GitHub workflow. However, on really rare occasions there may beed to add a file manually, for example. Thus add an owners file to enable the usual prow-based workflow for these PRs.


marquiz:
This is bad, I have a better idea
/hold

PR #1358, opened 2023-09-15
github: prevent parallel runs of gh-pages update

marquiz:
Ensure sequential updates to the gh-pages branch. Otherwise, if parallel jobs are triggered, onnly one (the first to complete) will succeed and the others will fail with git push error (because the tip of the gh-pages has moved and their commit doesn't apply anymore).


fmuyassarov (commented):
/lgtm

PR #1359, opened 2023-09-15
github: separate workflow for helm repo index update

marquiz:
No need to (re-)build documentation when a release is published. Great simplification of the Helm repo index update script: do not scan all releases but just get the assets from the release that was published.

This separation should make the maintenance of scripts and workflows easier.


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+
+jobs:
+  update-helm-repo:
+    name: Update gh-pages helm repo index
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install helm
+      run: |
+        curl -sfL https://raw.githubusercontent.com/helm/helm/master/scripts/get-helm-3 | bash -s -- --version v3.12.3
+
```
fmuyassarov, line 17:
```suggestion
    - name: Install Helm
      uses: azure/setup-helm@v3
```
I would suggest to use action image here instead which is doing the same thing.

.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
...
+        fetch-depth: 0
+
+    - name: Update repo index
+      run: |
+        git config user.name "Github Actions"
+        git config user.email "no-reply@github.com"
+        ./scripts/github/update-helm-repo.sh ${{ join(github.event.release.assets.*.browser_download_url, ' ') }}
+
+    - name: Push
+      run: git push -f https://${GITHUB_ACTOR}:${{ secrets.GITHUB_TOKEN }}@github.com/${{ github.repository }} gh-pages
```
fmuyassarov, line 30: no strong reasoning, but i found using `stefanzweifel/git-auto-commit-action@v4` easier and cleaner to push to remote. But as said, I'm not against having plain git commands on the workflow, just heads up.

fmuyassarov (commented):
Looks much better in my opinion:)

.github/workflows/release.yml
marquiz: Thanks, makes sense, fixed

.github/workflows/release.yml
marquiz: I'd like to only rely on "official" actions or at least actions from verified publishers to voting against stefanzweifel. It would look slightly cleaner but doesn't even save any lines of yaml

.github/workflows/release.yml
```diff
@@ -0,0 +1,31 @@
+name: gh-pages
+on:
+  release:
+    types: [published, edited]
```
fmuyassarov, line 4: Do you really expect to be editing the releases where there could be some assets added/removed/updated? It is fine perhaps to edit the notes, but assets? Would not that require, another patch releases?

marquiz: Let's say you publish the release but "forget" to add the assets. Then you "edit" the release, adding the missing assets.

PR #1372, opened 2023-09-19
Use the distroless/base-nossl base image for minimal

marquiz:
Smaller footprint. We don't need libssl for anything.


marquiz:
I like #1373 better
/close

PR #1373, opened 2023-09-19
Build statically linked binaries

marquiz:
Switch to fully statically linked binaries and use scratch as a base image.

Switching to the virtually empty scratch base image means that the default/minimal NFD image only supports running hooks that are truly statically linked (e.g.  normal go binaries that are "almost" statically linked stop working).  The documentation has been already stating this (i.e. that only statically-linked binaries are supported) - i.e. we have had no promise of supporting other than that. Also, hooks are now deprecated and even disabled by default so the possibility of real user impact should be small.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,8 @@ FROM ${BUILDER_IMAGE} as builder
 
 # Build and install the grpc-health-probe binary
 RUN GRPC_HEALTH_PROBE_VERSION=v0.4.19 && \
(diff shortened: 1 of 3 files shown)


marquiz:
@tkatila had some good questions (in a private chat) about the images and some bells started to ring. I started to dig into my memory and git repo and found this one. It could be time for this...

/assign @ArangoGutierrez 
/cc @mythi @PiotrProkop

ArangoGutierrez (approved):
I call this "Peace of mind"
/lgtm

PR #1378, opened 2023-09-27
nfd-master: fix filtering of extended resources

marquiz:
Fix a bug in checking the allowed ".feature.node.kubernetes.io" ns suffix for extended resources. Also update e2e-tests to cover this case.


ArangoGutierrez (approved):
/lgtm

PR #1386, opened 2023-10-02
feat: support raw features

### LINKED ISSUE #1376: Labels converted to resources are also kept also as labels
**What happened**:
Previously, we told NFD to convert a node label into a resource by the app arguments. Now when we moved to using NodeFeatureRules for the label->resource conversion, we see that the label is still visible in node's labels. Previously the label vanished when it got converted into a resource.

Labels that are converted into resources should be removed from labels.

Our GPU plugin exports a set of labels via features.d directory:
```
gpu.intel.com/cards=card0
gpu.intel.com/gpu-numbers=0
gpu.intel.com/memory.max=0
gpu.intel.com/millicores=1000
gpu.intel.com/tiles=1
```
And from these, we convert the following resources:
```
gpu.intel.com/memory.max:       0
gpu.intel.com/millicores:       1k
gpu.intel.com/tiles:            1
```
With this rule:
```
kind: NodeFeatureRule
metadata:
  name: intel-gpu-platform-labeling
spec:
  rules:
  - extendedResources:
      gpu.intel.com/millicores: "@local.label.gpu.intel.com/millicores"
      gpu.intel.com/memory.max: "@local.label.gpu.intel.com/memory.max"
      gpu.intel.com/tiles: "@local.label.gpu.intel.com/tiles"
    matchFeatures:
      - feature: local.label
        matchExpressions:
          gpu.intel.com/millicores: {op: Exists}
          gpu.intel.com/memory.max: {op: Exists}
          gpu.intel.com/tiles: {op: Exists}
    name: intel.gpu.fractionalresources
```
Labels for the node after the resources have been converted:
```
kubectl get node -o json | jq .items[0].metadata.labels
{
...
  "gpu.intel.com/cards": "card0",
  "gpu.intel.com/device-id.0300-9a49.count": "1",
  "gpu.intel.com/device-id.0300-9a49.present": "true",
  "gpu.intel.com/gpu-numbers": "0",
  "gpu.intel.com/memory.max": "0",
  "gpu.intel.com/millicores": "1000",
  "gpu.intel.com/tiles": "1",
...
}
```

**Environment**:
NFD 0.14.0

AhmedGrati:
Resolves #1376.

**Note: I used `no-label` and `no-feature` directives instead of `noLabel` and `noFeature`, to make it consistent with `expiry-time` directive.**

### DIFF
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -355,6 +355,41 @@ included in the list of accepted features.
 > tag is only evaluated in each re-discovery period, and the expiration of
 > node labels is not tracked.
 
+To exclude specific features from the `local.feature` Feature, you can use the
+`# +no-feature` directive. The `# +no-label` directive causes the feature to
+be excluded from the `local.label` Feature and a node label not to be generated.
+
+Considering the following file:
+
+```plaintext
+# +no-feature
+label-only=value
+
+my-feature=value
+
+foo=bar
+# +no-label
+foo=baz
+```
+
+Processing the above file would result in the following Features:
+
+```yaml
+local.features:
+  foo: baz
+  my-feature: value
+local.labels:
+  label-only: value
+  my-feature: value
+```
+
+and the following labels added to the Node:
+
+```plaintext
+feature.node.kubernetes.io/label-only=value
+feature.node.kubernetes.io/my-feature=value
+```
+
 ### Mounts
 
 The standard NFD deployments contain `hostPath` mounts for
@@ -775,7 +810,8 @@ The following features are available for matching:
 |                  |              | **`major`** | int     | First component of the kernel version (e.g. ‘4')
 |                  |              | **`minor`** | int     | Second component of the kernel version (e.g. ‘5')
 |                  |              | **`revision`** | int  | Third component of the kernel version (e.g. ‘6')
-| **`local.label`** | attribute   |           |           | Features feature files and hooks, i.e. labels from the [*local* feature source](#local-feature-source)
+| **`local.label`** | attribute   |           |           | Labels from feature files and hooks, i.e. labels from the [*local* feature source](#local-feature-source)
+| **`local.feature`** | attribute   |           |         | Features from feature files and hooks, i.e. features from the [*local* feature source](#local-feature-source)
 |                  |              | **`<label-name>`** | string | Label `<label-name>` created by the local feature source, value equals the value of the label
 | **`memory.nv`**  | instance     |          |            | NVDIMM devices present in the system
 |                  |              | **`<sysfs-attribute>`** | string | Value of the sysfs device attribute, available attributes: `devtype`, `mode`
--- a/source/local/local.go
+++ b/source/local/local.go
@@ -38,9 +38,22 @@ const Name = "local"
 // LabelFeature of this feature source
 const LabelFeature = "label"
 
-// ExpiryTimeKey is the key of this feature source indicating
-// when features should be removed.
-const ExpiryTimeKey = "expiry-time"
+// RawFeature of this feature source
+const RawFeature = "feature"
+
+const (
+	// ExpiryTimeKey is the key of this feature source indicating
+	// when features should be removed.
+	DirectiveExpiryTime = "expiry-time"
+
+	// NoLabel indicates whether the feature should be included
+	// in exposed labels or not.
+	DirectiveNoLabel = "no-label"
+
+	// NoFeature indicates whether the feature should be included
+	// in exposed raw features or not.
+	DirectiveNoFeature = "no-feature"
+)
 
 // DirectivePrefix defines the prefix of directives that should be parsed
 const DirectivePrefix = "# +"
@@ -66,7 +79,9 @@ type Config struct {
 
 // parsingOpts contains options used for directives parsing
 type parsingOpts struct {
-	ExpiryTime time.Time
+	ExpiryTime  time.Time
+	SkipLabel   bool
+	SkipFeature bool
 }
 
 // Singleton source instance
@@ -121,7 +136,7 @@ func newDefaultConfig() *Config {
 func (s *localSource) Discover() error {
 	s.features = nfdv1alpha1.NewFeatures()
 
-	featuresFromFiles, err := getFeaturesFromFiles()
+	featuresFromFiles, labelsFromFiles, err := getFeaturesFromFiles()
 	if err != nil {
 		klog.ErrorS(err, "failed to read feature files")
 	}
@@ -131,7 +146,7 @@ func (s *localSource) Discover() error {
 		klog.InfoS("starting hooks...")
 		klog.InfoS("NOTE: hooks are deprecated and will be completely removed in a future release.")
 
-		featuresFromHooks, err := getFeaturesFromHooks()
+		featuresFromHooks, labelsFromHooks, err := getFeaturesFromHooks()
 		if err != nil {
 			klog.ErrorS(err, "failed to run hooks")
 		}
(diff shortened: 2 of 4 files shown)


source/local/local.go
```diff
@@ -143,9 +156,18 @@ func (s *localSource) Discover() error {
 			}
 			featuresFromFiles[k] = v
 		}
+
+		// Merge labels from hooks and files
+		for k, v := range labelsFromHooks {
+			if old, ok := labelsFromFiles[k]; ok {
+				klog.InfoS("overriding label value", "labelKey", k, "oldValue", old, "newValue", v)
```
marquiz, line 163: The old/existing log message above should be reworded to `overriding feature value..."

source/local/local.go
```diff
@@ -259,17 +305,24 @@ func getFeaturesFromHooks() (map[string]string, error) {
...
 
 		// Append features
-		fileFeatures := parseFeatures(lines, fileName)
-		klog.V(4).InfoS("hook executed", "fileName", fileName, "features", utils.DelayedDumper(fileFeatures))
+		fileFeatures, fileLabels := parseFeatureFile(lines, fileName)
+		klog.V(4).InfoS("hook executed", "fileName", fileName, "features", utils.DelayedDumper(fileFeatures), "labels", utils.DelayedDumper(fileLabels))
 		for k, v := range fileFeatures {
 			if old, ok := features[k]; ok {
-				klog.InfoS("overriding label value from another hook", "labelKey", k, "oldValue", old, "newValue", v, "fileName", fileName)
+				klog.InfoS("overriding label value from another hook", "featureKey", k, "oldValue", old, "newValue", v, "fileName", fileName)
```
marquiz, line 312:
```suggestion
				klog.InfoS("overriding feature value from another hook", "featureKey", k, "oldValue", old, "newValue", v, "fileName", fileName)
```

source/local/local.go
```diff
@@ -66,7 +77,9 @@ type Config struct {
 
 // parsingOpts contains options used for directives parsing
 type parsingOpts struct {
-	ExpiryTime time.Time
+	ExpiryTime           time.Time
+	IsIncludedInLabels   bool
+	IsIncludedInFeatures bool
```
marquiz, line 82: I'd suggest to inverse the logic of the opts. So that the default value (`false`) produces the default behavior (i.e. no directive specified)

```suggestion
	SkipLabel   bool
	SkipFeature bool
```

(field names don't need to be `SkipLabel` and `SkipFeature`, I just picked them as short alternatives...)

source/local/local.go
```diff
@@ -38,10 +38,21 @@ const Name = "local"
...
 // when features should be removed.
 const ExpiryTimeKey = "expiry-time"
 
+// NoLabel indicates whether the feature should be included
+// in exposed labels or not.
+const NoLabel = "no-label"
+
+// NoFeature indicates whether the feature should be included
+// in exposed raw features or not.
+const NoFeature = "no-feature"
```
marquiz, line 54: suggestion: we could join the directive names under one `const` block to keep them in the same "context". Also, we could use a prefix to denote that they are directives.

```suggestion
const (
    // ExpiryTimeKey is the key of this feature source indicating
    // when features should be removed.
    DirectiveExpiryTimeKey = "expiry-time"
    
    // NoLabel indicates whether the feature should be included
    // in exposed labels or not.
    DirectiveNoLabel = "no-label"
  
    // NoFeature indicates whether the feature should be included
    // in exposed raw features or not.
    DirectiveNoFeature = "no-feature"
)
```

WDYT?

marquiz (commented):
Thanks @AhmedGrati. Some really quick comments below. We also need to update the documentation (at least `docs/usage/customization-guide.md`) to mention the new directives and the new `local.features` feature.

/cc @tkatila @eero-t

source/local/local.go
eero-t: Yes, one from line 154.

source/local/local.go
```diff
@@ -216,31 +245,48 @@ func parseFeatures(lines [][]byte, fileName string) map[string]string {
...
+			updateFeatures(features, &parsingOpts.IsIncludedInFeatures, key, lineSplit)
+			updateFeatures(labels, &parsingOpts.IsIncludedInLabels, key, lineSplit)
 		}
 	}
 
-	return features
+	return features, labels
+}
+
+func updateFeatures(m map[string]string, shouldBeIncluded *bool, key string, lineSplit []string) {
```
eero-t, line 256: With option names changed to `skipLabel` / `skipFeature`, argument name should also be changed:
```suggestion
func updateFeatures(m map[string]string, skip *bool, key string, lineSplit []string) {
```

(And function logic regarding the bool inverted accordingly.)

eero-t (commented):
Update also to test code and docs would be good.

docs/usage/customization-guide.md
```diff
@@ -355,6 +355,34 @@ included in the list of accepted features.
 > tag is only evaluated in each re-discovery period, and the expiration of
 > node labels is not tracked.
 
+To exclude specific features from the `local.feature` Feature, you can use the
+`# +no-feature` directive. The same applies to `local.label` Feature where you can
+exclude some features using the `# +no-label` directive.
```
marquiz, line 360: We need to also mention that no labels are generated with `+noLabel`

```suggestion
`# +no-feature` directive. The `# +no-label` directive causes the feature to be excluded from the
`local.label` Feature and a node label not to be generated.
```

or smth

Note: now that we have already three directives we could add e.g. a table to list them all. But this can be left as a future enhancement.

source/local/local.go
```diff
@@ -216,31 +247,48 @@ func parseFeatures(lines [][]byte, fileName string) map[string]string {
...
 			key := lineSplit[0]
-
-			// Check if it's a boolean value
-			if len(lineSplit) == 1 {
-				features[key] = "true"
-			} else {
-				features[key] = lineSplit[1]
-			}
+			updateFeatures(features, &parsingOpts.SkipFeature, key, lineSplit)
+			updateFeatures(labels, &parsingOpts.SkipLabel, key, lineSplit)
```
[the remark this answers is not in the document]
AhmedGrati: sounds good!

source/local/local.go
eero-t: Please set this as resolved (as I do not have option for that).

docs/usage/customization-guide.md
```diff
@@ -775,7 +803,8 @@ The following features are available for matching:
 |                  |              | **`major`** | int     | First component of the kernel version (e.g. ‘4')
 |                  |              | **`minor`** | int     | Second component of the kernel version (e.g. ‘5')
 |                  |              | **`revision`** | int  | Third component of the kernel version (e.g. ‘6')
-| **`local.label`** | attribute   |           |           | Features feature files and hooks, i.e. labels from the [*local* feature source](#local-feature-source)
+| **`local.label`** | attribute   |           |           | Labels feature files and hooks, i.e. labels from the [*local* feature source](#local-feature-source)
```
marquiz, line 806:
```suggestion
| **`local.label`** | attribute   |           |           | Labels from feature files and hooks, i.e. labels from the [*local* feature source](#local-feature-source)
```

docs/usage/customization-guide.md
```diff
@@ -355,6 +355,34 @@ included in the list of accepted features.
...
+
+```yaml
+local.features:
+  foo: baz
+  my-feature: value
+local.labels:
+  label-only: value
+  my-feature: value
+```
+
```
marquiz, line 385: Suggest an addition to mention the node labels, too (the indentation below is wrong to get the formatting right...)

```suggestion
    and the following labels added to the Node:
    
    ```
    feature.node.kubernetes.io/label-only=value
    feature.node.kubernetes.io/my-feature=value
    ```
```

source/local/local.go
```diff
@@ -217,30 +248,49 @@ func parseFeatures(lines [][]byte, fileName string) map[string]string {
...
 	}
 
-	return features
+	return features, labels
+}
+
+func updateFeatures(m map[string]string, key string, lineSplit []string) {
+	// Check if it's a boolean value
+	if len(lineSplit) == 1 {
+		m[key] = "true"
```
marquiz, line 274: `key` is redundant arg

```suggestion
func updateFeatures(m map[string]string, lineSplit []string) {
	key := lineSplit[0]
	// Check if it's a boolean value
	if len(lineSplit) == 1 {
		m[key] = "true"
```

marquiz (approved):
Big thanks @AhmedGrati for the contribution. I think @tkatila et al will be happy 😊
/lgtm

PR #1387, opened 2023-10-03
Helm - Move remaining gPRC related flags to conditional

ArangoGutierrez:
This patch continues to clean up after gRPC deprecation, by moving remaining flags in Helm charts under conditional of enableNodeFeatureAPI being set to false


ArangoGutierrez:
> Thanks @ArangoGutierrez.
> /lgtm
> 
> A further enhancement: make the creation of the Service (`templates/service.yaml`) conditional, too.

Will do!

PR #1388, opened 2023-10-04
[WIP] Enable Tolerations of automated taints

ArangoGutierrez:
Prevent NFD objects from being evicted when creating Taints


marquiz (commented):
This is a bit scary as nfd-master becomes an operator of the nfd-deployment itself 🤔 This also sort of nails down the expected deployment controller (daemonset vs job) of nfd-worker in the nfd codebase. Need to think carefully about the possible consequences.

One quick comment is that at least the nfd-master must not be give cluster-wide privileges to edit daemonsets.

deployment/components/common/env.yaml
```diff
@@ -5,3 +5,5 @@
       valueFrom:
         fieldRef:
           fieldPath: spec.nodeName
+    - name: NFD_WORKER_DS_NAME
+      value: nfd-worker
```
AhmedGrati, line 9: I think there's a missing EOF here.

AhmedGrati (commented):
Great improvement @ArangoGutierrez. Can we add some unit/e2e tests for this?

ArangoGutierrez:
Given the security implications will be good to get all reviewers check on this PR
@zvonkok 
@ffromani 
@AhmedGrati 
@fmuyassarov 
@PiotrProkop

zvonkok:
What happens if I create an NFR that "creates" `node.kubernetes.io/unschedulable`, will this code auto tolerate  `node.kubernetes.io/unschedulable` and essentially create a backdoor for auto tolerating any taint?

ArangoGutierrez:
> Given the security implications will be good to get all reviewers check on this PR @zvonkok @ffromani @AhmedGrati @fmuyassarov @PiotrProkop

Question number 1:
- Do we make Auto-Tolerate an optional feature? 
User might want to use NFD to drain specific nodes based on feature matching, and hence wanting NFD to leave the node after tainting. 
But NFD looses the ability to remove the taint or update features on the node, it will persist with the NodeFeature labels set before the worker is evicted

ArangoGutierrez:
> > > Given the security implications will be good to get all reviewers check on this PR @zvonkok @ffromani @AhmedGrati @fmuyassarov @PiotrProkop
> > 
> > 
> > Question number 1:
> > ```
> > * Do we make Auto-Tolerate an optional feature?
> >   User might want to use NFD to drain specific nodes based on feature matching, and hence wanting NFD to leave the node after tainting.
> >   But NFD looses the ability to remove the taint or update features on the node, it will persist with the NodeFeature labels set before the worker is evicted
> > ```
> 
> Hmm, can we just allow NFD to put taints that are present in its tolerations? This will move responsibilty on operator to add proper tolerations and NFD will just read objects instead of modify them?

Umm The thing is, the taints feature is to allow taints to be dynamic and change over time, if a new taint is needed, that would then require the System Admin to re-deploy NFD via HELM to add the new toleration. then creating burden on the Admin and removing the convenience of automating via NFR

marquiz:
> Hmm, can we just allow NFD to put taints that are present in its tolerations?
> This will move responsibilty on operator to add proper tolerations and NFD will just read objects instead of modify them?

This in another possible option. It could be also implemented as a allowedList so that at least in Helm deployment the taints allowedList would be set in nfd-master and corresponding tolerations would be set on nfd-worker. This way there would be no need for nfd-master to get/parse the nfd-worker daemonset spec.

marquiz:
> * Do we make Auto-Tolerate an optional feature?

Yes, I definitely think we should. Mark it as experimental and disabled by default

>   User might want to use NFD to drain specific nodes based on feature matching, and hence wanting NFD to leave the node after tainting.

Might be the case, too 🤷

>   But NFD looses the ability to remove the taint or update features on the node, it will persist with the NodeFeature labels set before the worker is evicted

Yes, this persists until the NFD deployment is changed or the rule that created the taint is removed (unless nfd-master gets evicted, too). In any case this scenario might be something that we'd want to document more clearly even in the current setup

ArangoGutierrez:
I think the best route here is simply better documentation, the more I think about it, the less I want this patch. will turn NFD into a scary thing to deploy.

PR #1389, opened 2023-10-04
Helm - service to be only deployed when needed

ArangoGutierrez:


marquiz (approved):
Thanks
/lgtm

PR #1393, opened 2023-10-05
nfd-master: correctly clean up annotations

marquiz:
Delete correct annotations if -instance is specified.


ArangoGutierrez (approved):
/lgtm

PR #1394, opened 2023-10-05
nfd-master: stop creating NFD version annotations

### LINKED ISSUE #1352: Drop the version annotations

Now that we have the [metrics](https://kubernetes-sigs.github.io/node-feature-discovery/v0.14/deployment/metrics.html) (with the `nfd_master_build_id` and `nfd_worker_build_id` metrics) there we could drop the `nfd.node.kubernetes.io/master.version` and `nfd.node.kubernetes.io/worker.version` annotations as redundant.

**Why is this needed**:

Drop redundant information. Keep the node annotations cleaner. Also the annotations are "asymmetric" in the sense that we don't have similar annotations for nfd-topology-updater or nfd-gc.

Stale annotations can also give misleading information. For example: say you run one replica of nfd-master, first it runs on node A and you get `master.version=v0.13.0` on node A. Then NFD gets migrated to node B and you update to NFD v0.14.0 and you will get `master.version=v0.14.0` on node B. Now node A has the annotation even if nfd-master is not running there anymore, and moreover, it also gives misleading information about the current version of NFD.

marquiz:
We now have metrics for getting detailed information about the NFD instances running. There should be no need to pollute the node object with NFD version annotations.

One problem with the annotations also that they were incomplete in the sense that they only covered nfd-master and nfd-worker but not nfd-topology-updater or nfd-gc.

Also, there was a problem with stale annotations, giving misleading information. E.g. there was no way to remove old/stale master.version annotations if nfd-master was scheduled on another node where it was previously running.

Fixes #1352

### DIFF
--- a/docs/get-started/introduction.md
+++ b/docs/get-started/introduction.md
@@ -102,16 +102,15 @@ NFD also annotates nodes it is running on:
 
 | Annotation                                                   | Description
 | ------------------------------------------------------------ | -----------
-| [&lt;instance&gt;.]nfd.node.kubernetes.io/master.version     | Version of the nfd-master instance running on the node. Informative use only.
-| [&lt;instance&gt;.]nfd.node.kubernetes.io/worker.version     | Version of the nfd-worker instance running on the node. Informative use only.
 | [&lt;instance&gt;.]nfd.node.kubernetes.io/feature-labels     | Comma-separated list of node labels managed by NFD. NFD uses this internally so must not be edited by users.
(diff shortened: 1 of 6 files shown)


ArangoGutierrez (approved):
/lgtm

marquiz:
I think I need to update the e2e-tests too 🙄
/hold

marquiz:
E2E fixed
/unhold
ping @ArangoGutierrez

marquiz:
Third time's the charm 🤞

marquiz:
Sync with https://github.com/kubernetes/test-infra/pull/30955
/hold

PR #1397, opened 2023-10-05
source/custom: drop support for the legacy rule format

### LINKED ISSUE #1344: Drop the legacy rule format from source/custom

Drop support for the legacy "matchOn" rule format in the `custom` source of nfd-worker.

**Why is this needed**:

The legacy rule format was deprecated in v0.10.0. Dropping it will simplify code. It also better aligns the node-local custom rules to the NodeFeatureRule API (essentially the same format except that NFR supports extended resources and taints which are missing from the custom source).

marquiz:
Fixes #1344


ArangoGutierrez (approved):
/lgtm

PR #1412, opened 2023-10-10
Fix prepare-release to work on MacOs-arm64

ArangoGutierrez:


ArangoGutierrez:
/retile Fix prepare-release to work on MacOs-arm64

ArangoGutierrez:
/retile Fix prepare-release to work on MacOs-arm64

ArangoGutierrez:
/retitle Fix prepare-release to work on MacOs-arm64

marquiz (changes requested):
This breaks usage on Linux so NACK. I don't know if there's a cross-compatible format for `-i` between the Linux and BSD variants of sed 🧐

ArangoGutierrez:
> This breaks usage on Linux so NACK. I don't know if there's a cross-compatible format for `-i` between the Linux and BSD variants of sed 🧐

/close

marquiz:
I you want to invest in this, you could of course (try to) detect which variant of sed (Linux vs BSD) is installed and parameterize the handling of `-i` depending on that...

PR #1415, opened 2023-10-10
feat: add parameters in helm to disable/enable nfd-master and nfd-worker

### LINKED ISSUE #1395: Helm: add parameters for disabling nfd-master and nfd-worker

Add `master.enable` and `worker.enable` parameters that allow users to skip deployment of nfd-master and nfd-worker.

**Why is this needed**:

Enable use cases like
- only run nfd-master: only custom labels from 3rd party extensions (via NodeFeature CRD API) are needed
- only run nfd-topology-updater

AhmedGrati:
Resolves #1395.

### DIFF
--- a/deployment/helm/node-feature-discovery/values.yaml
+++ b/deployment/helm/node-feature-discovery/values.yaml
@@ -13,6 +13,7 @@ namespaceOverride: ""
 enableNodeFeatureApi: true
 
 master:
+  enable: true
   config: ### <NFD-MASTER-CONF-START-DO-NOT-REMOVE>
     # noPublish: false
     # extraLabelNs: ["added.ns.io","added.kubernets.io"]
@@ -132,6 +133,7 @@ master:
                 values: [""]
 
 worker:
+  enable: true
   config: ### <NFD-WORKER-CONF-START-DO-NOT-REMOVE>
     #core:
     #  labelWhiteList:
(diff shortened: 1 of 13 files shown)


deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -13,6 +13,7 @@ namespaceOverride: ""
 enableNodeFeatureApi: true
 
 master:
+  enabled: true
```
marquiz, line 16: We have `gc.enable`, `topologyUpdater.enable`, `prometheus.enable` so for consistency it'd be good to have this `master.enable`

```suggestion
  enable: true
```

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -132,6 +133,7 @@ master:
                 values: [""]
 
 worker:
+  enabled: true
```
marquiz, line 136:
```suggestion
  enable: true
```

marquiz (changes requested):
Thankx @AhmedGrati for this enhancement. One comment regarding the parameter naming below.

In addition, make creation of the RBAC objects (clusterrole, clusterrolebinding, role, rolebinding and serviceaccount) conditionally dependent on the new parameters.

marquiz (approved):
Good improvement @AhmedGrati 
/assign @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #1421, opened 2023-10-20
test/e2e: fix source/custom nodename test

marquiz:
We dropped the legacy rule format so we need to convert the e2e test rules to the new format, accordingly.


marquiz:
Fixes my omission in #1397 
/assign @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #1425, opened 2023-10-20
nfd-master: fix retry of node updates

marquiz:
This patch addresses issues with slow node status (extended resources) updates. Previously we did just a few retries in quick succession which could result in the node update failing, just because node status was updated slower than our retry window. The patch mitigates the issue by increasing the number of tries to 15. In addition, it creates a ratelimiter with a longer per-item (per-node) base delay.

The patch also fixes the e2e-tests to expose the issue.

### DIFF
--- a/pkg/nfd-master/node-updater-pool.go
+++ b/pkg/nfd-master/node-updater-pool.go
@@ -18,7 +18,9 @@ package nfdmaster
 
 import (
 	"sync"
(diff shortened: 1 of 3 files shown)


ArangoGutierrez (approved):
oh yeah, testing annotations was super weird, thanks for this
/lgtm

PR #1426, opened 2023-10-20
test/e2e: stricter validation of node annotations

marquiz:
Now that the hard-to-predict version annotations are gone we can do
strict validation of nfd-generated node annotations.

The PR also contains another patch that drops the now unused ignoreUnexpected arg from custom gomega matcher.


ArangoGutierrez (approved):
Time to rebase annotations PR! 
/lgtm

Issue #1186, opened 2023-04-25
Spiffe support


marquiz:
**What would you like to be added**:

Support [SPIFFE](https://github.com/spiffe/spiffe) for verifying nfd-worker/labeler IDs. For example, utilize spiffe IDs for verifying the identity of creator of NodeFeature objects, so that an actor from node A will not be able to modify properties of node B. This would also add one extra layer of protection (in addition to RBAC) on who is authorized to modify nodes with NodeFeature objects.

To simplify,  gRPC could be possibly left out-of-scope, if needed, as it's being phased out.

Inspired by comment from @AhmedGrati on the node-feature-discovery slack channel.

**Why is this needed**:

Improved security

ArangoGutierrez:
Only comment, is make sure is optional and easy to enable/disable from HELM

AhmedGrati:
@ArangoGutierrez yes that's for sure!

PR #1444, opened 2023-10-28
docs: add KEP of Spiffe integration

AhmedGrati:
Add KEP of #1186.


enhancements/1186-spiffe-integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+<img src="./nfd-spiffe.png" style="display: block; margin-left: auto; margin-right: auto;">
+As shown in the figure above, the proposed solution involves:
+
+1. Installation of Spiffe which is composed of Spire-agents as daemonset, and Spire-server as statefulset. Spiffe installation should be available in Helm as well as Kustomize.
+
+2. The cluster administration (who intends to use NFD), should initialize the Spire-server with the specific Spiffe Verifiable Identity Document ID a.k.a SVID. In fact, we need to have only one SVID shared between all of the nfd-worker workloads. Initialization commands should be provided in docs.
+
+3. After features discovery, nfd-workers need to get their private key from Spiffe and sign the NFD CRD spec with their private keys. The signature should be put in the annotations section of the CRD. e.g.: `annotations.signature`
+
+4. When detecting new changes in the NFD CRD, the nfd-master should verify the signature in the CRD by using the public key of the nfd-workers.
```
PiotrProkop, line 28: Is the signature tied somehow to the content of the Custom Resource? What if nfd-worker do the initial advertising of features sign  the Custom Resource and then attacker/operator changes the content of the object would it make the old signature invalid?

enhancements/1186-spiffe-integration/README.md
```diff
@@ -0,0 +1,54 @@
+# KEP-1186: Spiffe integration in NFD
```
PiotrProkop, line 1: Please change CRD to CR in the shole KEP and specify what object is suppose to be signed.

enhancements/1186-spiffe-integration/README.md
AhmedGrati: Yes, the signature depends on the spec of the NodeFeature spec. Therefore, if an attacker changes the content of the object, the old signature would be invalid and changes would not be propagated to the actual node(s).

enhancements/1186-spiffe-integration/README.md
AhmedGrati: Done!

AhmedGrati:
ping @marquiz

enhancements/1186-spiffe-integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+### Risks and Mitigations
+- Workloads communication with Spire-agent and Spire-server is done through a Unix socket, which should be mounted from the host to workloads, which is not really recommended from a security perspective.
+
+## Design Details
+- Create a Spiffe utility structure in `pkg/utils` that would be responsible for exposing the following functions:
+    - `NewSpiffeClient`: returns a new `workloadapi.Client` which allows us to communicate with the spire-server.
+    - `SignData`: returns a signature of a given spec.
+    - `VerifyDataSignature`: checks the validity of the given data signature.
+- In the nfd-worker, before creating/updating a NodeFeature CR we should update its CR and add the signature in the annotations section
```
marquiz, line 48: Here we should describe in more detail what data actually get's hashed in the signature. Would be the Spec part of the object, right(?)

enhancements/1186-spiffe-integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+- Some tweaks that need to be done to the cluster in order to enable Spiffe that should be mentioned in docs. Example of Minikube configuration tweak: https://spiffe.io/docs/latest/try/getting-started-k8s/#considerations-when-using-minikube
+
+
+### Risks and Mitigations
+- Workloads communication with Spire-agent and Spire-server is done through a Unix socket, which should be mounted from the host to workloads, which is not really recommended from a security perspective.
+
+## Design Details
+- Create a Spiffe utility structure in `pkg/utils` that would be responsible for exposing the following functions:
+    - `NewSpiffeClient`: returns a new `workloadapi.Client` which allows us to communicate with the spire-server.
+    - `SignData`: returns a signature of a given spec.
```
marquiz, line 46: There probably are good utility/library functions for getting the signature(?) Do we need to put any details about that (e.g. hashing/signing algorithms?).

enhancements/1186-spiffe-integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Proposal
+<img src="./nfd-spiffe.png" style="display: block; margin-left: auto; margin-right: auto;">
+As shown in the figure above, the proposed solution involves:
+
+1. Installation of Spiffe which is composed of Spire-agents as daemonset, and Spire-server as statefulset. Spiffe installation should be available in Helm as well as Kustomize.
+
+2. The cluster administration (who intends to use NFD), should initialize the Spire-server with the specific Spiffe Verifiable Identity Document ID a.k.a SVID. In fact, we need to have only one SVID shared between all of the nfd-worker workloads. Initialization commands should be provided in docs.
+
+3. After features discovery, nfd-workers need to get their private key from Spiffe and sign the NodeFeature CR spec with their private keys. The signature should be put in the annotations section of the NodeFeature CR. e.g.: `annotations.signature`
```
marquiz, line 26: Private key would be the "private key associated with the SPIFFE ID", right? Should we spell this out(?)

enhancements/1186-spiffe-integration/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+4. When detecting new changes in the NodeFeature CR, the nfd-master should verify the signature in the CR by using the public key of the nfd-workers.
```
marquiz, line 28: The nfd-master cannot verify the node name of the NodeFeature, right? That is, the nfd-master cannot verify that the NodeFeature that is targeting node A was actually created by a nfd-worker (or some other daemon) running on node A. IOW, a malicious worker on node B could spoof lables on node A(?) If so, we could possibly mention that as a limitation.

marquiz (commented):
Big thanks @AhmedGrati for the design doc. To me this looks pretty much implementable. Just a few minor comments/questions below.

@ArangoGutierrez wanna take a look?

enhancements/1186-spiffe-integration/README.md
AhmedGrati: Hmm, I think we'll hash the whole spec object. wdyt?

marquiz: We cannot hash everything as e.g. in `metadata` there are fields manage/changed by the server (like timestamps, `resourceVersion` etc). We could has part of the metadata, though, like name, labels and annotations etc. `apiVersion` and `kind` we probably don't need to hash 😅 We don't have status but if we do in the future that's something to think about (hash it separately as a separate sub-resource or include it in the one and only hash) – the question here is which player in the game would be the one responsible for updating status.

AhmedGrati: What I meant with spec is the `.spec` so neither `.metadata` nor `.status` would be included in the hash.

marquiz: Ah, sorry I (mis-)read you message a bit hastily. Yes, definitely the whole `.Spec` content. We might want to hash part of the metadata, too. That would prevent easy replay-attacks where you just copy-paste the spec part (and it's signature) into a new object

AhmedGrati: We can add hashing the `metadata.name`, `metadata.namespace`, `metadata.labels`, and `metadata.annotations`. wdyt?

marquiz: I think that would be good

AhmedGrati:
ping @marquiz @ArangoGutierrez

ArangoGutierrez (approved):
Anything that is aimed towards making NFD more secure is always welcomed. 
/lgtm

PR #1446, opened 2023-11-01
Add Kubectl NFD plugin 

### LINKED ISSUE #1384: Debug tool for NodeFeatureRules

Implement a helper cmdline tool for validating/testing/debugging NodeFeatureRules. For example, subcommands for

- validate: stricter validation for NodeFeatureRule objects than what the CRD openapi validation can provide
- test (with debugging output):
  - locally run NodeFeatureRule(s) against NodeFeature(s)
  - run local NodeFeatureRule(s) against a cluster
  - remote debugging: for a selected node-name, run NodeFeatureRules in a cluster against NodeFeatures in the cluster

**Why is this needed**:

Make it easier for users to develop/test/debug NodeFeatureRules

ArangoGutierrez:
Fixes #1384 

This initial implementation of `kubectl-nfd` adds 3 verification cases:
- Validate a NodeFeatureRule file 
- Run a NodeFeatureRule against a local NodeFeature file
- Run a NodeFeatureRule against a cluster

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: all test templates yamls
+.PHONY: all test templates yamls build build-%
 .FORCE:
 
 GO_CMD ?= go
@@ -90,12 +90,17 @@ IMAGE_BUILD_ARGS_MINIMAL = --target minimal \
 
 all: image
 
-build:
-	@mkdir -p bin
-	$(GO_CMD) build -v -o bin $(BUILD_FLAGS) ./cmd/...
+BUILD_BINARIES := nfd-master nfd-worker nfd-topology-updater nfd-gc kubectl-nfd
 
-install:
-	$(GO_CMD) install -v $(BUILD_FLAGS) ./cmd/...
+build-%:
+	$(GO_CMD) build -v -o bin $(BUILD_FLAGS) ./cmd/$*
+
+build:	$(foreach bin, $(BUILD_BINARIES), build-$(bin))
+
+install-%:
+	$(GO_CMD) install -v $(BUILD_FLAGS) ./cmd/$*
+
+install:	$(foreach bin, $(BUILD_BINARIES), install-$(bin))
 
 image: yamls
 	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
--- a/cmd/kubectl-nfd/subcmd/dryrun.go
+++ b/cmd/kubectl-nfd/subcmd/dryrun.go
@@ -1,0 +1,59 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package subcmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	kubectlnfd "sigs.k8s.io/node-feature-discovery/pkg/kubectl-nfd"
+)
+
+var dryrunCmd = &cobra.Command{
+	Use:   "dryrun",
+	Short: "Process a NodeFeatureRule file against a NodeFeature file",
+	Long:  `Process a NodeFeatureRule file against a local NodeFeature file to dry run the rule against a node before applying it to a cluster`,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Printf("Evaluating NodeFeatureRule %q against NodeFeature %q\n", nodefeaturerule, nodefeature)
+		err := kubectlnfd.DryRun(nodefeaturerule, nodefeature)
+		if len(err) > 0 {
+			fmt.Printf("NodeFeatureRule %q is not valid for NodeFeature %q\n", nodefeaturerule, nodefeature)
+			for _, e := range err {
+				cmd.PrintErrln(e)
+			}
+			// Return non-zero exit code to indicate failure
+			os.Exit(1)
+		}
+		fmt.Printf("NodeFeatureRule %q is valid for NodeFeature %q\n", nodefeaturerule, nodefeature)
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(dryrunCmd)
+
+	dryrunCmd.Flags().StringVarP(&nodefeaturerule, "nodefeaturerule-file", "f", "", "Path to the NodeFeatureRule file to validate")
+	dryrunCmd.Flags().StringVarP(&nodefeature, "nodefeature-file", "n", "", "Path to the NodeFeature file to validate against")
+	err := dryrunCmd.MarkFlagRequired("nodefeaturerule-file")
+	if err != nil {
+		panic(err)
+	}
+	err = dryrunCmd.MarkFlagRequired("nodefeature-file")
+	if err != nil {
+		panic(err)
+	}
+}
--- a/cmd/kubectl-nfd/subcmd/root.go
+++ b/cmd/kubectl-nfd/subcmd/root.go
@@ -1,0 +1,54 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package subcmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var (
+	// Path to the NodeFeatureRule file to validate
+	nodefeaturerule string
+	// Path to the NodeFeature file to run against the NodeFeatureRule
+	nodefeature string
+	// Node to validate against
+	node string
+	// kubeconfig file to use
+	kubeconfig string
+)
+
+// RootCmd represents the base command when called without any subcommands
+var RootCmd = &cobra.Command{
+	Use:   "kubectl-nfd",
+	Short: "NFD kubectl plugin",
+	Long: `kubectl plugin for NFD
+	Debug tool to validate/dryrun/test NodeFeatureRules
+	for more information see: 
+	https://kubernetes-sigs.github.io/node-feature-discovery/v0.14/usage/customization-guide.html#nodefeaturerule-custom-resource`,
+}
+
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	if err := RootCmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
--- a/cmd/kubectl-nfd/subcmd/test.go
+++ b/cmd/kubectl-nfd/subcmd/test.go
@@ -1,0 +1,56 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package subcmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+	kubectlnfd "sigs.k8s.io/node-feature-discovery/pkg/kubectl-nfd"
+)
+
+var testCmd = &cobra.Command{
+	Use:   "test",
+	Short: "Test a NodeFeatureRule file against a Node",
+	Long:  `Test a NodeFeatureRule file against a Node to ensure it is valid before applying it to a cluster`,
+	Run: func(cmd *cobra.Command, args []string) {
+		fmt.Printf("Evaluating NodeFeatureRule against Node %s\n", node)
+		err := kubectlnfd.Test(nodefeaturerule, node, kubeconfig)
+		if len(err) > 0 {
+			fmt.Printf("NodeFeatureRule is not valid for Node %s\n", node)
+			for _, e := range err {
+				cmd.PrintErrln(e)
+			}
+			// Return non-zero exit code to indicate failure
+			os.Exit(1)
+		}
+		fmt.Printf("NodeFeatureRule is valid for Node %s\n", node)
+	},
+}
+
+func init() {
+	RootCmd.AddCommand(testCmd)
+
+	testCmd.Flags().StringVarP(&nodefeaturerule, "nodefeaturerule-file", "f", "", "Path to the NodeFeatureRule file to validate")
+	testCmd.Flags().StringVarP(&node, "nodename", "n", "", "Node to validate against")
+	testCmd.Flags().StringVarP(&kubeconfig, "kubeconfig", "k", "", "kubeconfig file to use")
+	err := testCmd.MarkFlagRequired("nodefeaturerule-file")
+	if err != nil {
+		panic(err)
+	}
+}
--- a/docs/usage/kubectl-plugin.md
+++ b/docs/usage/kubectl-plugin.md
@@ -1,0 +1,70 @@
+---
+title: "Kubectl plugin"
+layout: default
+sort: 10
+---
+
+# Kubectl plugin
+{: .no_toc}
+
+## Table of contents
+{: .no_toc .text-delta}
+
+1. TOC
+{:toc}
+
+---
+
+> ***Developer Preview*** This feature is currently in developer preview and
+> subject to change. It is not recommended to use it in production
+> environments.
+
+## Overview
+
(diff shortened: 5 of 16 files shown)


ArangoGutierrez:
TODO: Documentation

ArangoGutierrez:
Example:
```bash
❯ go run cmd/kubectl-nfd/main.go test -f examples/nodefeaturerule-2.yaml -s node-feature-discovery -k /Users/eduardoa/.kube/config
Processing rule:  my sample labels rule
Processing rule:  my sample extendedResources rule
***     Labels  ***
my-sample-feature-1=true
kubernetes.io/my-sample-feature-2=true
        -Validation errors-
namespace kubernetes.io is not allowed

***     Extended Resources      ***
vendor.io/static=123
vendor.io/dynamic=6
```

```bash
❯ go run cmd/kubectl-nfd/main.go  validate -f examples/nodefeaturerule-2.yaml
Validating rule:  my sample labels rule
Validating rule:  my sample extendedResources rule
namespace kubernetes.io is not allowed
```

Makefile
```diff
@@ -90,10 +90,24 @@ all: image
 
 build:
 	@mkdir -p bin
-	$(GO_CMD) build -v -o bin $(BUILD_FLAGS) ./cmd/...
+	$(GO_CMD) build -v -o bin $(BUILD_FLAGS) ./cmd/nfd-gc
+	$(GO_CMD) build -v -o bin $(BUILD_FLAGS) ./cmd/nfd-worker
+	$(GO_CMD) build -v -o bin $(BUILD_FLAGS) ./cmd/nfd-master
+	$(GO_CMD) build -v -o bin $(BUILD_FLAGS) ./cmd/nfd-topology-updater
```
marquiz, line 96: Maybe you could use a variable to simplify and re-use also in `install:`target below

```
BUILD_BINARIES := nfd-master nfd-worker nfd-topology-updater nfd-gc

build:
    $(GO_CMD) build -v -o bin $(BUILD_FLAGS) $(foreach bin, $(BUILD_BINARIES), ./cmd/$(bin))
```

Makefile
```diff
@@ -90,10 +90,24 @@ all: image
...
 
 install:
-	$(GO_CMD) install -v $(BUILD_FLAGS) ./cmd/...
+	$(GO_CMD) install -v $(BUILD_FLAGS) ./cmd/nfd-gc
+	$(GO_CMD) install -v $(BUILD_FLAGS) ./cmd/nfd-worker
+	$(GO_CMD) install -v $(BUILD_FLAGS) ./cmd/nfd-master
+	$(GO_CMD) install -v $(BUILD_FLAGS) ./cmd/nfd-topology-updater
+
+build-kubectl-plugin:
```
marquiz, line 104: How about rename this to more generic `build-tools` or smth?

Or alternatively we could make this a generic `build-%` target like

```
build-%:
    $(GO_CMD) build -v -o bin ./cmd/$*
```

Then you could also use (alternative to my previous comment):

```
BUILD_BINARIES := nfd-master nfd-worker nfd-topology-updater nfd-gc

build: $(foreach bin, $(BUILD_BINARIES), build-$(bin)):
```

marquiz (commented):
Thanks @ArangoGutierrez. looks cool! Some notes about the Makefile below.

Other than that I didn't have time to look at the code but one immediate suggestion was to share the validation code between nfd-master and the kubelet plugin. E.g. refactor the code and put the validation functions inside `pkg/apis/nfd/validation` for example. WDYT?

Makefile
ArangoGutierrez: that's a better idea, yup.

cmd/kubectl-nfd/subcmd/root.go
```diff
@@ -0,0 +1,56 @@
...
+	// namespace for NodeFeatures
+	namespace string
+)
+
+// RootCmd represents the base command when called without any subcommands
+var RootCmd = &cobra.Command{
+	Use:   "kubectl-nfd",
+	Short: "NFD kubectl plugin",
+	Long: `kubectl plugin for NFD
+	Debbug tool to validate/dryrun/test NodeFeatureRules
```
PiotrProkop, line 44: typo s/Debbug/Debug

pkg/kubectl-nfd/dryrun.go
```diff
@@ -0,0 +1,199 @@
...
+		for k, v := range labels {
+			fmt.Printf("%s=%s\n", k, v)
+		}
+		if len(labelValidation) > 0 {
+			fmt.Println("\t-Validation errors-")
+			for _, err := range labelValidation {
+				fmt.Println(err)
+			}
+		}
+		fmt.Println()
```
PiotrProkop, line 137: ?

pkg/kubectl-nfd/validate.go
```diff
@@ -0,0 +1,248 @@
...
+func validateMatchAny(matchAny []nfdv1alpha1.MatchAnyElem) []error {
+	var validationErr []error
+
+	for _, matcher := range matchAny {
+		for _, match := range matcher.MatchFeatures {
+			nameSplit := strings.SplitN(match.Feature, ".", 2)
+			if len(nameSplit) != 2 {
+				validationErr = append(validationErr, fmt.Errorf("invalid feature name %v (not <domain>.<feature>), cannot be used for templating", match.Feature))
+			}
+		}
```
PiotrProkop, line 90: You can move it to separate func

pkg/kubectl-nfd/dryrun.go
```diff
@@ -0,0 +1,199 @@
...
+	if len(taints) > 0 {
+		taintValidation := validateTaints(taints)
+		fmt.Println("***\tTaints\t***")
+		for _, taint := range taints {
+			fmt.Println(taint)
+		}
+		if len(taintValidation) > 0 {
+			fmt.Println("\t-Validation errors-")
+			for _, err := range taintValidation {
+				fmt.Println(err)
```
PiotrProkop, line 120: Did you mean to append `err` to `validationErr` here or `Println` is on purpose?

pkg/kubectl-nfd/test.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return []error{fmt.Errorf("error reading NodeFeatureRule file: %v", err)}
+	}
+
+	err = yaml.Unmarshal(nfrFile, &nfr)
+	if err != nil {
+		return []error{fmt.Errorf("error reading NodeFeatureRule file: %v", err)}
+	}
+
+	validationErr = append(validationErr, processNodeFeatureRule(nfr, *nf)...)
```
PiotrProkop, line 83: You can add func to pretty print `validationErr` so it'll be easier to consume :)

pkg/kubectl-nfd/dryrun.go
ArangoGutierrez: a BIG TODO here is to pretty print the results, this empty line is just to make things readable

pkg/kubectl-nfd/dryrun.go
ArangoGutierrez: is on purpose here

ArangoGutierrez:
TODO: Pretty print

pkg/apis/nfd/validate/validate.go
```diff
@@ -0,0 +1,177 @@
...
+
+	for _, taint := range taints {
+		ns, _ := splitNs(taint.Key)
+		// if ns is empty, it means that the rule is not allowed
+		if ns == "" {
+			validationErr = append(validationErr, fmt.Errorf("unprefixed keys are disallowed"))
+		}
+		if ns != nfdv1alpha1.TaintNs && !strings.HasSuffix(ns, nfdv1alpha1.TaintSubNsSuffix) {
+			// If the namespace is denied, rule is invalid
+			if ns == "kubernetes.io" || strings.HasSuffix(ns, ".kubernetes.io") {
```
PiotrProkop, line 122: You are using this if statement couple of times maybe make it a  `func` e.g. `IsValidNamespace` or `IsAllowedNamespace`?

cmd/kubectl-nfd/subcmd/dryrun.go
```diff
@@ -0,0 +1,51 @@
...
+
+var dryrunCmd = &cobra.Command{
+	Use:   "dryrun",
+	Short: "Process a NodeFeatureRule file against a NodeFeature file",
+	Long:  `Process a NodeFeatureRule file against a local NodeFeature file to dry run the rule against a node before applying it to a cluster`,
+	Run: func(cmd *cobra.Command, args []string) {
+		err := kubectlnfd.DryRun(nodefeaturerule, nodefeature)
+		if len(err) > 0 {
+			for _, e := range err {
+				cmd.PrintErrln(e)
```
PiotrProkop, line 32: I think if validation fails it should also return non zero exit code, so it can be used in CI etc

ArangoGutierrez: good idea, I was not thinking in CI, just manual

ArangoGutierrez:
@marquiz I did your recommendation to move to `/pkg/apis/v1alpha/validate` 
And today I woke up thinking, should we make it an interface or something to be more re-usable? maybe attached to the CRD, so we can do something like

```go
nfdv1alpha.nodefeaturerue.validate().Labels(labels map(string)string)
```

Or that would be a bad idea?

marquiz:
> maybe attached to the CRD, so we can do something like

I think this is a bit against the K8s API conventions/guidelines (which I think probably makes sense to follow) to couple implementation with the API definition. I know there's some functionality like the rule processing/matching in the `pkg/apis/nfd/v1alpha1` package directly but in retrospect I'd probably do it differently (and might think about moving it away from there for coherence)

ArangoGutierrez:
> > maybe attached to the CRD, so we can do something like
> 
> I think this is a bit against the K8s API conventions/guidelines (which I think probably makes sense to follow) to couple implementation with the API definition. I know there's some functionality like the rule processing/matching in the `pkg/apis/nfd/v1alpha1` package directly but in retrospect I'd probably do it differently (and might think about moving it away from there for coherence)

Sounds like a TODO for me :)

pkg/apis/nfd/validate/validate.go
ArangoGutierrez: I thought about that, but is a one liner, and if you look the variables are different. reusing with a Func will look really similar, since the func would have to have 2 entries and return a bool.

ArangoGutierrez:
@marquiz milestone v0.15?

marquiz:
> @marquiz milestone v0.15?

Definitely. One thing that's still missing is the documentation

/milestone v0.15

pkg/kubectl-nfd/dryrun.go
```diff
@@ -0,0 +1,199 @@
...
+			}
+		}
+	}
+
+	if len(labels) > 0 {
+		labelValidation := validate.Labels(labels)
```
marquiz, line 127: There's one peculiarity related to the labels, annotations and extended resources: the default prefix will be appended to non-namespaced keys by nfd-master. I'm not entirely sure how this should be handled 🤔 Maybe we should move that part to the rule processing (so that the rule output will always have a prefix).

pkg/kubectl-nfd/validate.go
```diff
@@ -0,0 +1,75 @@
...
+
+		// Validate LabelsTemplate
+		validationErr = append(validationErr, validate.Template(rule.LabelsTemplate)...)
+
+		// Validate matchFeatures
+		validationErr = append(validationErr, validate.MatchFeatures(rule.MatchFeatures)...)
+
+		// Validate matchAny
+		validationErr = append(validationErr, validate.MatchAny(rule.MatchAny)...)
+	}
```
marquiz, line 72: For `LabelsTemplate` and `VarsTemplate` we could check that the template is parseable

marquiz (commented):
Nice work. I think it would be very good to share the same validation code between the kubectl plugin and nfd-master. To both reduce code duplication and prevent diverging (to two slightly different validation schemes). I'd suggest to split the api/validation package in a separate PR. Refactoring nfd-master to use this new package. Then when we're happy with that (and get it merged) get the kubectl plugin in. This is a big PR so it would make also review easier

pkg/kubectl-nfd/dryrun.go
ArangoGutierrez: marking as solved by #1455

pkg/kubectl-nfd/validate.go
ArangoGutierrez: marking as solved by #1455

docs/usage/customization-guide.md
```diff
@@ -185,6 +185,37 @@ to specify taints in the NodeFeatureRule object.
 > not tolerate the taint are evicted immediately from the node including the
 > nfd-worker pod.
 
+### kubectl plugin
```
marquiz, line 188: I'd suggest to have the kubelet plugin as a separate file (and thus a main entry shown on the left sidebar on the docs site, under Usage). Then we can just reference that page from here. WDYT?

docs/usage/customization-guide.md
```diff
@@ -185,6 +185,37 @@ to specify taints in the NodeFeatureRule object.
...
+
+The `kubectl` plugin `kubectl nfd` can be used to validate/dryrun and test
+NodeFeatureRule objects. It can be installed with the following command:
+
+```bash
+make build-kubectl-nfd
+KUBECTL_PATH=/usr/local/bin/
+mv ./bin/kubectl-nfd ${KUBECTL_PATH}
+```
```
marquiz, line 197: Add instructions to do git clone (and other steps)

Just a note that we can probably start with a "developer preview" and go with this but going forward we want to have this as one of the release assets

pkg/kubectl-nfd/test.go
```diff
@@ -0,0 +1,122 @@
...
+
+	return validationErr
```
marquiz, line 85: Looks like used first time here so just

```suggestion
	return processNodeFeatureRule(nfr, *nf)
```

cmd/kubectl-nfd/subcmd/test.go
```diff
@@ -0,0 +1,57 @@
...
+
+var testCmd = &cobra.Command{
+	Use:   "test",
+	Short: "Test a NodeFeatureRule file against a Node",
+	Long:  `Test a NodeFeatureRule file against a Node to ensure it is valid before applying it to a cluster`,
+	Run: func(cmd *cobra.Command, args []string) {
+		err := kubectlnfd.Test(nodefeaturerule, node, kubeconfig, namespace)
+		if len(err) > 0 {
+			for _, e := range err {
+				cmd.PrintErrln(e)
```
marquiz, line 34: Should we print a bit more context, e.g. which node the NFR is evaluated on?

pkg/kubectl-nfd/validate.go
```diff
@@ -0,0 +1,78 @@
...
+		// Validate Rule Name
+		if rule.Name == "" {
+			validationErr = append(validationErr, fmt.Errorf("rule name cannot be empty"))
+		}
+
+		// Validate Annotations
+		validationErr = append(validationErr, validate.Annotations(rule.Annotations)...)
+
+		// Validate labels
+		validationErr = append(validationErr, validate.Labels(rule.Labels)...)
```
marquiz, line 56: I think we should handle the dynamic values of labels and ERs here. If value is in the format `@<part-1>.<part-2>` (or just starts with `@` to make it simple) just replace with a dummy value "0" or smth

marquiz (commented):
Fist full pass through the PR...

examples/nodefeature.yaml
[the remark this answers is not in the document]
ArangoGutierrez: is on purpose so people can run dryrun's using the `kubectl-nfd` plugin , thanks for the feedback will update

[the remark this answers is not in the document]
ArangoGutierrez: I added an example, like 60 on that file `docs/usage/kubectl-plugin.md`

docs/usage/kubectl-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+
+```bash
+kubectl get -n node-feature-discovery nodefeature <nodename> -o yaml > <nodefeature.yaml>
+kubectl nfd dryrun -f <nodefeaturerule.yaml> -n <nodefeature.yaml>
+```
+
+Or you can use the example NodeFeature file(it is a minimal NodeFeature file):
+
+```bash
+$ kubectl nfd dryrun -f nodefeaturerule.yaml -n ./examples/nodefeature.yaml
```
marquiz, line 64: Sorry, dir missing for nfr:

```suggestion
$ kubectl nfd dryrun -f examples/nodefeaturerule.yaml -n examples/nodefeature.yaml
```

examples/nodefeature.yaml
```diff
@@ -0,0 +1,113 @@
...
+    flags:
+      cpu.cpuid:
+        elements:
+          SHA1: {}
+          SHA2: {}
+          SHA3: {}
+          SHA512: {}
+      kernel.loadedmodule:
+        elements:
+          auth_rpcgss: {}
```
marquiz, line 72:
```suggestion
          auth_rpcgss: {}
          dummy: {}
```

examples/nodefeature.yaml
```diff
@@ -0,0 +1,113 @@
...
+      cpu.rdt:
+        elements: {}
+      cpu.security:
+        elements: {}
+      cpu.sst:
+        elements: {}
+      cpu.topology:
+        elements:
+          hardware_multithreading: "false"
+      kernel.selinux:
```
marquiz, line 33:
```suggestion
      kernel.config:
        elements:
          DUMMY: m
          NET: "y"
          X86: "y"
      kernel.selinux:
```

docs/usage/kubectl-plugin.md
```diff
@@ -0,0 +1,68 @@
...
+Processing rule:  my sample rule
+Evaluating NodeFeatureRule "examples/nodefeaturerule.yaml" against NodeFeature "examples/nodefeature.yaml"
+NodeFeatureRule "examples/nodefeaturerule.yaml" is valid for NodeFeature "examples/nodefeature.yaml"
```
marquiz, line 67: Missing directory for the nfr. Also, with my other suggestions the output is richer

```suggestion
$ kubectl nfd dryrun -f examples/nodefeaturerule.yaml -n examples/nodefeature.yaml
Processing rule:  my sample rule
***	Labels	***
vendor.io/my-sample-feature=true
Evaluating NodeFeatureRule "examples/nodefeaturerule.yaml" against NodeFeature "examples/nodefeature.yaml"
NodeFeatureRule "examples/nodefeaturerule.yaml" is valid for NodeFeature "examples/nodefeature.yaml"
```

marquiz (commented):
Make the example nfr and nf match

marquiz (approved):
Thanks @ArangoGutierrez for the contribution! Awesome expansion of nfd territory.
/lgtm

PR #1460, opened 2023-11-10
test/e2e: increase timeout for waiting node status

marquiz:
In some occasions the node status (capacity) takes a lot of time to update. Increase the timeout on extended resource tests. Revert the default timeout back to 10s.


ArangoGutierrez (approved):
/lgtm

PR #1461, opened 2023-11-10
Option to stop implicitly adding default prefix to names

marquiz:
Add new `autoDefaultNs` (default is `true`) config option to nfd-master.
Setting the config option to false stops NFD from automatically adding
the `feature.node.kubernetes.io` prefix to labels, annotations and
extended resources. Taints are not affected as for them no prefix is
automatically added. The user-visible part of enabling the option change
is that NodeFeatureRules, local feature files, hooks and configuration
of the "custom" may need to be altereda (if the auto-prefixing is
relied on).

For now, the config option defaults to `true`, meaning no change in
default behavior. However, the intent is to change the default to
`false` in a future release, deprecating the option and eventually
removing it (forcing it to `false`).

The goal of stopping doing "auto-prefixing" is to simplify the operation
(of nfd and users). Make the naming more straightforward and easier to
understand and debug (kind of WYSIWYG), eliminating peculiar corner
cases:

1. Make validation simpler and unambiguous
2. Remove "overloading" of names, i.e. the mapping two values to the
   same actual name. E.g. previously something like
    ```yaml
      labels:
        feature.node.kubernetes.io/foo: bar
        foo: baz
    ```
   Could actually result in node label:
    ```
     feature.node.kubernetes.io/foo: baz
    ```
3. Make the processing/usagee of the `rule.matched` and `local.labels`
   feature in NodeFeatureRules unambiguous and more understadable. E.g.
   previously you could have node label
   `feature.node.kubernetes.io/local-foo: bar` but in the NodeFeatureRule
   you'd need to use the unprefixed name `local-foo` or the fully
   prefixed name, depending on what was specified in the feature file (or
   hook) on the node(s).

> NOTE: setting autoDefaultNs to false is a breaking change for users who
> rely on automatic prefixing with the default `feature.node.kubernetes.io/`
> namespace. NodeFeatureRules, feature files, hooks and custom rules
> (configuration of the "custom" source of nfd-worker) will need to be
> altered.  Unprefixed labels, annoations and extended resources will be
> denied by nfd-master.

### DIFF
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -81,7 +81,7 @@ spec:
             vendor: "acme"
   # Labels to be created
   labels:
-    vendor-feature.enabled: "true"
+    vendor.io/feature.enabled: "true"
 ```
 
 The object targets node named `node-1`. It lists two "flag type" features under
@@ -92,8 +92,7 @@ labels but they will be used as input when the
 [`NodeFeatureRule`](#nodefeaturerule-custom-resource) objects are evaluated.
 
 In addition, the example requests directly the
-`feature.node.kubenernetes.io/vendor-feature.enabled=true` node label to be
-created.
+`vendor.io/feature.enabled=true` node label to be created.
 
 The `nfd.node.kubernetes.io/node-name=<node-name>` must be in place for each
 NodeFeature object as NFD uses it to determine the node which it is targeting.
@@ -130,7 +129,7 @@ spec:
   rules:
     - name: "my sample rule"
       labels:
-        "my-sample-feature": "true"
+        "feature.node.kubernetes.io/my-sample-feature": "true"
       matchFeatures:
         - feature: kernel.loadedmodule
           matchExpressions:
@@ -141,7 +140,7 @@ spec:
 ```
 
 It specifies one rule which creates node label
-`feature.node.kubenernetes.io/my-sample-feature=true` if both of the following
+`feature.node.kubernetes.io/my-sample-feature=true` if both of the following
 conditions are true (`matchFeatures` implements a logical AND over the
 matchers):
 
@@ -216,9 +215,9 @@ having the following contents (or alternatively a shell script
 following stdout output):
 
 ```plaintext
-my-feature.1
-my-feature.2=myvalue
-my.namespace/my-feature.3=456
+feature.node.kubernetes.io/my-feature.1
+feature.node.kubernetes.io/my-feature.2=myvalue
+vendor.io/my-feature.3=456
 ```
 
 This will translate into the following node labels:
@@ -226,7 +225,7 @@ This will translate into the following node labels:
 ```yaml
 feature.node.kubernetes.io/my-feature.1: "true"
 feature.node.kubernetes.io/my-feature.2: "myvalue"
-my.namespace/my-feature.3: "456"
+vendor.io/my-feature.3: "456"
 ```
 
 ### Feature files
@@ -277,12 +276,12 @@ key-value pairs, separated by newlines:
 
 ```plaintext
 # This is a comment
-<name>[=<value>]
+<key>[=<value>]
 ```
 
 The label value defaults to `true`, if not specified.
 
-Label namespace may be specified with `<namespace>/<name>[=<value>]`.
+Label namespace must be specified with `<namespace>/<name>[=<value>]`.
 
 > **NOTE:** The feature file size limit it 64kB. The feature file will be
 > ignored if the size limit is exceeded.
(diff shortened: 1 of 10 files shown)


marquiz:
Request for comments/thoughts

/hold

docs/usage/customization-guide.md
```diff
@@ -81,7 +81,7 @@ spec:
             vendor: "acme"
   # Labels to be created
   labels:
-    vendor-feature.enabled: "true"
+    vendor.io/vendor-feature.enabled: "true"
```
ArangoGutierrez, line 84:
```suggestion
    vendor.io/feature.enabled: "true"
```

Like this is better maybe?

ArangoGutierrez (changes requested):
We discussed this online, I am in (lgtm)
just one nit

docs/usage/customization-guide.md
ArangoGutierrez: 👋🏽

ArangoGutierrez:
@zvonkok @yevgeny-shnaidman @PiotrProkop Any comments?

pkg/nfd-master/nfd-master.go
```diff
@@ -566,6 +567,9 @@ func (m *nfdMaster) filterFeatureLabel(name, value string, features *nfdv1alpha1
 	if ns != nfdv1alpha1.FeatureLabelNs && ns != nfdv1alpha1.ProfileLabelNs &&
 		!strings.HasSuffix(ns, nfdv1alpha1.FeatureLabelSubNsSuffix) && !strings.HasSuffix(ns, nfdv1alpha1.ProfileLabelSubNsSuffix) {
 		// If the namespace is denied, and not present in the extraLabelNs, label will be ignored
+		if ns == "" {
```
PiotrProkop, line 570: you can move this check before the big if statement

pkg/nfd-master/nfd-master.go
```diff
@@ -812,6 +817,9 @@ func filterExtendedResource(name, value string, features *nfdv1alpha1.Features)
 	// Check if given NS is allowed
 	ns, _ := splitNs(name)
 	if ns != nfdv1alpha1.ExtendedResourceNs && !strings.HasSuffix(ns, nfdv1alpha1.ExtendedResourceSubNsSuffix) {
+		if ns == "" {
```
PiotrProkop, line 820: you can move this check just after `splitNs(name)`

docs/usage/customization-guide.md
marquiz: Better, changed as suggested

pkg/nfd-master/nfd-master.go
marquiz: Maybe this is more readable, changed as suggested

pkg/nfd-master/nfd-master.go
marquiz: Maybe this is more readable, changed as suggested

yevgeny-shnaidman:
@marquiz does autoDefaultNs option means that it also supports the previous Rules/hooks configuration? If not that this commit is very problematic, since all of our customers will certainly complain regarding backwards compatibility

marquiz:
> @marquiz does autoDefaultNs option means that it also supports the previous Rules/hooks configuration? If not that this commit is very problematic, since all of our customers will certainly complain regarding backwards compatibility

Thank you for the feedback @yevgeny-shnaidman. Yes, this config option is effectively a "feature gate", reverting to the old behavior.

One possibility would be to actually have this change as "alpha" default-disabled (i.e. `autoDefaultNs=true`) in v0.15 and turn it on by default in a future release. And now looking at this it might be the right thing to do.

What I was/am unsure if almost anybody really rely on the default namespacing of features i.e. have non-namespaced labels in their hook/feature files or rule configuration. Anything I've seen out there is always something like `vendor.com/feature-x=foo` (and not `feature-x=foo` which will turn to `feature.node.kubernetes.io/feature-x=foo`). One extra motivation of this change would be to eventually allow creating non-namespaced labels, too.

ArangoGutierrez:
> > @marquiz does autoDefaultNs option means that it also supports the previous Rules/hooks configuration? If not that this commit is very problematic, since all of our customers will certainly complain regarding backwards compatibility
> 
> Thank you for the feedback @yevgeny-shnaidman. Yes, this config option is effectively a "feature gate", reverting to the old behavior.
> 
> One possibility would be to actually have this change as "alpha" default-disabled (i.e. `autoDefaultNs=true`) in v0.15 and turn it on by default in a future release. And now looking at this it might be the right thing to do.
> 
> What I was/am unsure if almost anybody really rely on the default namespacing of features i.e. have non-namespaced labels in their hook/feature files or rule configuration. Anything I've seen out there is always something like `vendor.com/feature-x=foo` (and not `feature-x=foo` which will turn to `feature.node.kubernetes.io/feature-x=foo`). One extra motivation of this change would be to eventually allow creating non-namespaced labels, too.

unhold?

marquiz:
> unhold?

I'll probably change this feature to default-off so it'll be a non-breaking change in v0.15. Then warn about the users about the upcoming default-on change in a future release.

marquiz:
/retitle Option to stop implicitly adding default prefix to names

marquiz:
Changed the autoDefaultNs default to true, i.e. not changing the existing behavior. The idea is that the default will be changed to false in a future release. PR title and description updated accordingly.

ArangoGutierrez:
> Changed the autoDefaultNs default to true, i.e. not changing the existing behavior. The idea is that the default will be changed to false in a future release. PR title and description updated accordingly.

unhold?

marquiz:
Updated. Depends on #1470 and #1471

marquiz:
Unoholding this as it's now effectively an "alpha" feature, turned off by default

ArangoGutierrez (approved):
/lgtm

PR #1468, opened 2023-11-22
apis/nfd: fix multiple matcher terms targeting the same feature

marquiz:
Fix NodeFeatureRule templating in cases where multiple matchFeatures terms are targeting the same feature. Previously, only matched feature elements from the last matcher terms were used as the input to the template. However, the input should contain all matched elements from all matcher terms.

For example, consider the example rule snippet below:

```yaml
  ...
  labelsTemplate: |
    {{ range .pci.device }}vendor.io/pci-device.{{ .class }}-{{ .device }}=exists
    {{ end }}
  matchFeatures:
    - feature: pci.device
      matchExpressions:
        class: {op: InRegexp, value: ["^03"]}
        vendor: {op: In, value: ["1234"]}
    - feature: pci.device
      matchExpressions:
        class: {op: InRegexp, value: ["^12"]}
```

This rule matches if both a pci device of class `03` from vendor `1234` exists and a pci device of class `12` (from any vendor) exists. Previously, the template would only generate labels from the devices in class `12` (as that's the last term). With this patch the template creates device labels from devices in both classes `03` and `12`.


ArangoGutierrez (approved):
/lgtm

PR #1473, opened 2023-11-24
Reproducible output from expression matching

marquiz:
Fix flakyness of unit tests by adding back the sorting of matched feature elements that was unadvisedly removed in
63c22551df210143c8297f96c65afbafbea8f049. This might help debugging some corner cases in real-life scenarios (when using templating), too.


ArangoGutierrez:
Third time is the charm 

/test pull-node-feature-discovery-verify-master

ArangoGutierrez (approved):
/lgtm

PR #1474, opened 2023-11-24
Use T.Run in expression unit tests

marquiz:
The "better way" of running test cases, get e.g. better output in case of errors.

Also drop some unneeded type definitions from the tests.


ArangoGutierrez (approved):
/lgtm

PR #1475, opened 2023-11-28
docs: correct description of no-publish for topology-updater

marquiz:
The nfd-topology-updater does not do "feature discovery" and does not connect to nfd-master.


ArangoGutierrez (approved):
/lgtm

PR #1476, opened 2023-11-28
docs/tls: update cert-manager deployment instructions

marquiz:
Point to latest version of cert-manager. Let's keep this up-to-date even if TLS support is deprecated and going away.


ArangoGutierrez (approved):
soon to go away :)
/lgtm

PR #1477, opened 2023-11-29
apis/nfd: drop the private template caching fields

marquiz:
Drop the private fields – that were supposed to be used for caching parsed templates – from the Rule type. Keep the API typedefs cleaner and simpler. Moreover, the caching was not even used in practice, effectively complicating code without any benefit: the way the types are used in nfd-master creates a local copy of Rule type storing the cached template in the copy, wasting it from any future users.

There are also other possible caveats in caching like we tried to do it. For example the objects returned by the api lister are supposed to be treated as read-only - in particular if we would be to modify them there should at least be proper locking in place as nfd-master potentially processes the same rule (the same Go object) in parallel for multiple nodes. If any optimization like this will be pursued it should be done properly, probably with private type(s) at the consumer's end, not contaminating the API types.


marquiz:
Rebased

ArangoGutierrez (approved):
/lgtm

PR #1479, opened 2023-11-30
source/custom: add internal rule api

marquiz:
1. Add new internal "API" for the 
    Add internal API for the nfd-worker custom feature source rule
    configuration. This API has already diverged from the NFD
    NodeFeatureRule API in that annotations, extended resources or taints
    are not supported. This patch basically copies the Rule type (and it's
    sub-types) from the nfdv1alpha1 package. It also adds conversion
    functions from the internal rule API to the "external" nfdv1alpha1 API.
    This is done to use the same rule matching functionality (from the
    nfdv1alpha1 package).

    One notable remark is that the feature source rule config supports some
    custom formatting (short forms, multi-type fields) that relies on
    special json/yaml unmarshalling functions that are better to nuke from
    the nfdv1alpha1 package (in another patch). These (legacy) syntax
    specialities are most probably used by nobody but let's keep them as
    they're already there.  Unit tests to cover the custom json
    unmarshalling are now added.

3. Use the new internal api for nfd-worker config parsing
Change the custom feature source of nfd-worker to use the newly added
internal config API for its own configuration. It now uses the internal
types for json/yaml unmarshalling but converts them to external
nfdv1alpha1 API to do the actual rule matching as the internal API does
not duplicate that functionality.

4. Drop custom unmarshaller functions from the nfdv1alpha1 API


marquiz:
Let's wait for #1480 at least
/hold

ArangoGutierrez:
> Let's wait for #1480 at least /hold

#1480 is merged
/unhold

marquiz:
Rebased

source/custom/api/types.go
```diff
@@ -0,0 +1,143 @@
+/*
+Copyright 2021 The Kubernetes Authors.
```
ArangoGutierrez, line 2:
```suggestion
Copyright 2023 The Kubernetes Authors.
```

source/custom/api/expression.go
```diff
@@ -0,0 +1,218 @@
...
+// newMatchExpression returns a new MatchExpression instance.
+func newMatchExpression(op MatchOp, values ...string) *MatchExpression {
+	return &MatchExpression{
+		Op:    op,
+		Value: values,
+	}
+}
+
+// Validate validates the expression.
+func (m *MatchExpression) Validate() error {
```
ArangoGutierrez, line 50: can we leverage the new Validate PKG?

ArangoGutierrez (changes requested):

source/custom/api/types.go
marquiz: Thx, fixed

source/custom/api/expression.go
marquiz: Hmm, not for this internal API.

But that could be possible further change/enhancement on top of this PR. It'd be a slightly bigger change than initially thought.

source/custom/api/types.go
```diff
@@ -0,0 +1,151 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package api
+
+// Rule defines a rule for node customization such as labeling.
+type Rule struct {
```
ArangoGutierrez, line 20: i don't like having duplicated cope like this. but for now it's ok.
I think for v0.16 I want to work on moving Rule type outside the v1alpha1 folder/pkg to clean the api pkg and also to have all that functionality in a more centralized location

ArangoGutierrez (approved):
/lgtm

source/custom/api/types.go
marquiz: Regarding duplication, I cannot think of a better alternative. The two APIs (the CRD API and the custom source configuration) have diverged. In terms of typedefs the custom source config is a subset of the CRD API, but in terms of "syntax", it's a superset as it supports these int-or-string-or-bool kind of fields and some historical shortforms etc

We cannot move the typedef outside the nfdv1alpha1 package as that's part of the API (or we technically *could* import it from anywhere but I think that's a bad idea because that other package then essentially becomes a part of the API...). For the "functionality", i.e. the implementation of Rule evaluation/processing, I agree that we want to move it out of the API package.

PR #1481, opened 2023-12-01
apis/nfd: drop creation helper functions

marquiz:
Drop the creation helper functions as one step in an effort to tidy up the api package. These functions were not much used outside unit tests anyway, the static rules of the nfd-worker custom feature source being the only exception (and if those happened to be invalid we'd catch that e.g. in the e2e-tests).


marquiz:
Unit tests fail until #1480 is in

ArangoGutierrez (approved):
/lgtm

marquiz:
Rebased.
ping @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #1482, opened 2023-12-01
apis/nfd: drop the private regexp caching field

marquiz:
Drop the private field for caching parsed regexp from the MatchExpression type. This tidies up the API type definition and not so tied with particular implementation details. The change also elimiates potential concurrency problems as no locking is in place in the API types.

If caching will be desired in the future, it's better to do it properly in a separate package, not directly in the API types.

Ref: #1477


ArangoGutierrez (approved):
v1alpha2 is around the corner :)
/lgtm

Issue #1423, opened 2023-10-20
KEP: NodeFeatureGroup API (CRD)


ArangoGutierrez:
## Summary

The Kubernetes cluster object doesn't expose all available features in a programmatic way.

When working in a MultiCluster environment (example [kcp](https://github.com/kcp-dev/kcp), [hypershift](https://github.com/openshift/hypershift) ) the central control plane can not access all the available features on each cluster, making it hard to take scheduling and management decisions.

The Node-Feature-Discovery does a good work for exposing a per-node basis feature inventory but querying each cluster in a per-node basis can be a network intensive task.  Various use cases have been identified where having a cluster inventory would facilitate operations at the Cluster management level. 

This KEP proposes NFD to expose an inventory of available features in the cluster via a new API (CRD)

### Goals

- make the information about all clusters easy to query via a centralised API
- expose cluster wide features currently not reported by NFD in a per node scenario, e.g Network config

### Non-Goals

- change existing behaviour at the node level
- To be a MultiCluster management tool, this API is to expose NFD discovered features via a single API (CRD)

#### Story 1

As a platform engineer, I want to known the available features on each cluster registered on my network to be able to make optimal, platform specific, scheduling decisions.

#### Story 2

As a System-Admin I want a single API to know the available features of each cluster on the network.

resource allocations.

### CRD API

```go
// NodeFeatureGroup resource holds the features discovered for all nodes in a
// cluster.
// +kubebuilder:object:root=true
// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
// +genclient
type NodeFeatureGroup struct {
	metav1.TypeMeta   `json:",inline"`
	metav1.ObjectMeta `json:"metadata,omitempty"`

	Spec ClusterFeatureSpec `json:"spec"`
}

// NodeFeatureGroupSpec describes a ClusterFeature object.
type NodeFeatureGroupSpec struct {
	// FeatureGroup is a set of grouped objects by specific features
	// +optional
	FeatureGroup []FeatureGroupSpec `json:"FeatureGroupSpec"`
	// Features is the set of cluster wide features that are not reported at the node level.
	// +optional
	ClusterFeatures []ClusterFeatures`json:"clusterFeatures"`
}
```

zanetworker:
Do you mind clarifying on `make the information about all clusters easy to query via a centralised API`?

[...]

PR #1487, opened 2023-12-01
Add NodeFeatureGroup API

ArangoGutierrez:
Fixes: #1423


pkg/apis/nfd/v1alpha1/nodefeaturegroup_types.go
```diff
@@ -0,0 +1,91 @@
...
+
+// NodeFeatureGroupSpec describes a NodeFeatureGroup object.
+type NodeFeatureGroupSpec struct {
+	// FeatureGroupRules is a list of feature group rules.
+	// Feature group rules are used to group nodes into feature groups.
+	// +optional
+	FeatureGroup `json:"FeatureGroup"`
+	// Features is the set of features that are shared by all nodes in the feature group
+	// +optional
+	Features map[string]string `json:"Features"`
```
AhmedGrati, line 43: also here:
```suggestion
	Features map[string]string `json:"features"`
```

test/e2e/data/nodefeaturegrouprule-1.yaml
```diff
@@ -0,0 +1,11 @@
...
+kind: NodeFeatureGroupRule
+metadata:
+  name: e2e-test-1
+spec:
+  featureGroupRules:
+    - name: "e2e-attribute-test-1"
+      matchFeatures:
+        - feature: kernel.version
+          matchExpressions:
+            major: {op: Exists}
```
AhmedGrati, line 11: Missing EOF

AhmedGrati (commented):
Great job @ArangoGutierrez! I left some comments below. Apart from that, should we mark the `enable-nodefeaturegroup-api` flag as experimental, or make it disabled by default? And can we support enabling/disabling it using configuration files?

docs/reference/master-commandline-reference.md
```diff
@@ -181,6 +181,22 @@ Example:
...
+[NodeFeatureGroup](../usage/custom-resources.md#nodefeaturegroup) CRD API for
+working with NodeFeatureGroups.
+
+Default: false
+
+Example:
+
+```bash
+nfd-master -enable-nodefeaturegroup-api=false
+```
```
AhmedGrati, line 198: Since the default value is false, I suggest:
```suggestion
```bash
nfd-master -enable-nodefeaturegroup-api=true
```
```

File: pkg/apis/nfd/v1alpha1/rule.go [hunk 4]
```diff
@@ -108,6 +108,41 @@ func (r *Rule) Execute(features *Features) (RuleOutput, error) {
 	return ret, nil
 }
 
+// Evaluate the rule against a set of input features, and return true if the
+// rule matches.
+func (r *Rule) Evaluate(features *Features) (bool, error) {
```
AhmedGrati, line 113: Can we add a unit test for this function?

deployment/helm/node-feature-discovery/templates/master.yaml
```diff
@@ -70,6 +70,9 @@ spec:
             {{- if .Values.master.instance | empty | not }}
             - "-instance={{ .Values.master.instance }}"
             {{- end }}
+            {{- if .Values.enableNodeFeatureGroup }}
```
AhmedGrati, line 73:
```suggestion
            {{- if .Values.enableNodeFeatureGroupApi }}
```

AhmedGrati (commented):
We're almost there, I left some comments below.

docs/usage/custom-resources.md
```diff
@@ -86,6 +86,54 @@ See the
 [Customization guide](customization-guide.md#node-feature-rule-custom-resource)
 for full documentation of the NodeFeatureRule resource and its usage.
 
+## NodeFeatureGroup
```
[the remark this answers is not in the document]
adrianchiris: ack, i have no objection with sticking to current approach. thx for giving this a good thinking.

nf.yml
```diff
@@ -0,0 +1,2979 @@
...
+items:
+- apiVersion: nfd.k8s-sigs.io/v1alpha1
+  kind: NodeFeature
+  metadata:
+    annotations:
+      nfd.node.kubernetes.io/worker.version: v0.15.0-devel-128-gaa0f5969-dirty
+    creationTimestamp: "2023-12-01T15:53:12Z"
+    generation: 1
+    labels:
+      nfd.node.kubernetes.io/node-name: minikube
```
AhmedGrati, line 11: I think that we should remove this file.

AhmedGrati (approved):
Great job @ArangoGutierrez, LGTM!

pkg/apis/nfd/v1alpha1/nodefeaturegroup_types.go
```diff
@@ -0,0 +1,91 @@
...
+	// +optional
+	FeatureGroup `json:"featureGroup"`
+	// Features is the set of features that are shared by all nodes in the feature group
+	// +optional
+	Features map[string]string `json:"features"`
+}
+
+type FeatureGroup struct {
+	// Name is the name of the feature group
+	Name string `json:"name"`
```
marquiz, line 48: This looks redundant, same as ObjectMeta.Name, right?

pkg/apis/nfd/v1alpha1/nodefeaturegroup_types.go
```diff
@@ -0,0 +1,91 @@
...
+
+	Spec NodeFeatureGroupSpec `json:"spec"`
+}
+
+// NodeFeatureGroupSpec describes a NodeFeatureGroup object.
+type NodeFeatureGroupSpec struct {
+	// FeatureGroupRules is a list of feature group rules.
+	// Feature group rules are used to group nodes into feature groups.
+	// +optional
+	FeatureGroup `json:"featureGroup"`
```
marquiz, line 40: To me this looks like a redundant hierarchy level. Just have 

```suggestion
	Nodes []string `json:"nodes"`
```

docs/usage/custom-resources.md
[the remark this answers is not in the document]
marquiz: Hmm, now reading this comment again, I really started to like the idea 😅 Re @AhmedGrati's [comment](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1487#discussion_r1413871762) about diverging APIs I think we could handle that by introducing `NodeFeatureGroupRule` API if there is a reason/need to specify a new API. For the time being I have hard time coming up with something else in the NodeFeatureGroupRule than matching node features. If we need to match something from the NodeSpec (or NodeStatus) for example we could just extend the NodeFeatureRule API – or not necessarily even the API but what is available for matching.

Using a single API would also have the benefit that a user could label the node(s) and create NodeFeatureGroup object with a single NodeFeatureRule if she wants to have both labels and NFG. Thoughts?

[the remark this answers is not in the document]
ArangoGutierrez: Hope everyone got time to think about this over the weekend 

cc @zvonkok thoughts?

pkg/nfd-master/nfd-master.go
```diff
@@ -267,7 +275,7 @@ func (m *nfdMaster) Run() error {
 		defer m.Stop()
 	}
 
-	// Run gRPC server
+	//ss Run gRPC server
```
lianghao208, line 278: redundant

pkg/nfd-master/nfd-master.go
```diff
@@ -37,7 +37,10 @@ import (
 	"google.golang.org/grpc/health"
 	"google.golang.org/grpc/health/grpc_health_v1"
 	"google.golang.org/grpc/peer"
+
```
lianghao208, line 40: no need for the new line here

docs/usage/custom-resources.md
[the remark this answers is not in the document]
marquiz: With more thinking and chatting I'd probably go with a separate CRD, but with some changes. At least have a single CRD – put Rule in Spec and list of nodes in Status (https://github.com/kubernetes-sigs/node-feature-discovery/pull/1487#discussion_r1420239469).

One clear benefit of a separate CRD is the RBAC. You can give users the right to create node groups but not NodeFeatureRules (which labels nodes etc).

Later on, we could add the capability create NodeFeatureGroups from NodeFeatureRule (the idea that @adrianchiris presented) if needed.

pkg/apis/nfd/v1alpha1/nodefeaturegroup_types.go
```diff
@@ -0,0 +1,91 @@
...
+type NodeFeatureGroupRule struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec NodeFeatureGroupRuleSpec `json:"spec"`
+}
+
+// NodeFeatureGroupRuleSpec describes a NodeFeatureGroupRule.
+type NodeFeatureGroupRuleSpec struct {
+	FeatureGroupRules []Rule `json:"featureGroupRules"`
```
marquiz, line 80: If we go with the separate CRD, we should specify a separate Rule type for it. Many of the fields (like `labels`, `taints` etc) are not applicable here and should be rejected

pkg/apis/nfd/v1alpha1/const.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2022 The Kubernetes Authors.
+Copyright 2021 The Kubernetes Authors.
```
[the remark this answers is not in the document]
marquiz: Not resolved

marquiz:
This will be the big thing in v0.16 ⭐️

ArangoGutierrez:
@adrianchiris @AhmedGrati  PTAL

pkg/nfd-master/nfd-api-controller.go
```diff
@@ -113,7 +121,29 @@ func newNfdController(config *restclient.Config, nfdApiControllerOptions nfdApiC
...
+			c.updateNodeFeatureGroup(nfg.Name)
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			nfg := newObj.(*nfdv1alpha1.NodeFeatureGroup)
+			klog.V(2).InfoS("NodeFeatureGroup updated", "nodeFeatureGroup", klog.KObj(nfg))
+			c.updateNodeFeatureGroup(nfg.Name)
+		},
+		DeleteFunc: func(obj interface{}) {
+			nfg := obj.(*nfdv1alpha1.NodeFeatureGroup)
+			klog.V(2).InfoS("NodeFeatureGroup deleted", "nodeFeatureGroup", klog.KObj(nfg))
```
AhmedGrati, line 141: Shouldn't we call `updateNodeFeatureGroup` here?

test/e2e/utils/crd.go
```diff
@@ -245,3 +245,37 @@ func init() {
 	// Register k8s scheme to be able to create CRDs
 	_ = apiextensionsv1.AddToScheme(scheme.Scheme)
 }
+
+// CreateNodeFeatureGroupRuleFromFile creates a NodeFeatureGroupRule object from a given file located under test data directory.
```
AhmedGrati, line 249:
```suggestion
// CreateNodeFeatureGroupRulesFromFile creates a NodeFeatureGroupRule object from a given file located under test data directory.
```

AhmedGrati (commented):
Good job @ArangoGutierrez. Left some comments below!

marquiz:
@ArangoGutierrez please rebase and address review comments. I'll take a full pass after that

pkg/apis/nfd/v1alpha1/nodefeaturerule/rule.go
```diff
@@ -108,6 +108,41 @@ func Execute(r *nfdv1alpha1.Rule, features *nfdv1alpha1.Features) (RuleOutput, e
...
 }
 
+// Execute the GroupRule against a set of input features, and return true if the
+// rule matches.
+func ExecuteGroupRule(r *nfdv1alpha1.GroupRule, features *nfdv1alpha1.Features) (bool, error) {
+	matched := false
+	if len(r.MatchAny) > 0 {
+		// Logical OR over the matchAny matchers
+		for _, matcher := range r.MatchAny {
+			if isMatch, matches, err := evaluateMatchAnyElem(&matcher, features); err != nil {
```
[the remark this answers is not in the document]
ArangoGutierrez: This could be a housekeeping issue as a followup for this PR, I agree, we could just return the array and error.

ArangoGutierrez:
> @ArangoGutierrez please rebase and address review comments. I'll take a full pass after that

Rebased! 

Thanks @AhmedGrati for your comments

pkg/apis/nfd/v1alpha1/nodefeaturerule/rule.go
```diff
@@ -108,6 +108,41 @@ func Execute(r *nfdv1alpha1.Rule, features *nfdv1alpha1.Features) (RuleOutput, e
 	return ret, nil
 }
 
+// ExecuteGroupRule executes the GroupRule against a set of input features, and return true if the
+// rule matches.
+func ExecuteGroupRules(r *nfdv1alpha1.GroupRule, features *nfdv1alpha1.Features) (bool, error) {
```
marquiz, line 113:
```suggestion
func ExecuteGroupRule(r *nfdv1alpha1.GroupRule, features *nfdv1alpha1.Features) (bool, error) {
```

pkg/apis/nfd/v1alpha1/nodefeaturerule/rule.go
```diff
@@ -108,6 +108,41 @@ func Execute(r *nfdv1alpha1.Rule, features *nfdv1alpha1.Features) (RuleOutput, e
...
+		if isMatch, _, err := evaluateFeatureMatcher(&r.MatchFeatures, features); err != nil {
+			return false, err
+		} else if !isMatch {
+			klog.V(2).InfoS("rule did not match", "ruleName", r.Name)
+			return false, nil
+		}
+	}
+
+	klog.V(2).InfoS("rule matched", "ruleName", r.Name)
+	return true, nil
```
marquiz, line 143: We have the `vars` and `varsTemplate` fields in the NFG type. We should handle (and test) those as well. Or alternatively drop them from the type definition

pkg/apis/nfd/v1alpha1/nodefeaturerule/rule.go
marquiz, line 1: We need to have unit tests for the new functionality

test/e2e/utils/crd.go
```diff
@@ -123,6 +123,21 @@ func CreateNodeFeatureRulesFromFile(ctx context.Context, cli nfdclientset.Interf
 	return nil
 }
 
+// CreateNodeFeatureGroupRulesFromFile creates a NodeFeatureGroupRule object from a given file located under test data directory.
+func CreateNodeFeatureGroupRulesFromFile(ctx context.Context, cli nfdclientset.Interface, namespace, filename string) error {
```
marquiz, line 127:
```suggestion
func CreateNodeFeatureGroupsFromFile(ctx context.Context, cli nfdclientset.Interface, namespace, filename string) error {
```

test/e2e/utils/rbac.go
```diff
@@ -176,6 +176,11 @@ func createClusterRoleMaster(ctx context.Context, cs clientset.Interface) (*rbac
 				Resources: []string{"nodefeatures", "nodefeaturerules"},
 				Verbs:     []string{"get", "list", "watch"},
 			},
+			{
+				APIGroups: []string{"nfd.k8s-sigs.io"},
+				Resources: []string{"nodefeaturegroups", "nodefeaturegrouprules"},
```
marquiz, line 181:
```suggestion
				Resources: []string{"nodefeaturegroups"},
```

pkg/nfd-master/nfr-updater-pool.go
marquiz, line 1: I think this should not be renamed. It's updating nodes -> node-updater-pool. It's not just NFR but NFs, too

marquiz (commented):
I did a partial review. Didn't have time for the actual controller implementation, yet.

marquiz:
The more I look at this, I feel that the NFG updates should be wired into the existing node updater pool (instead of creating a parallel one). This way nfd-master follows the `-nfd-api-parallelism` flag (instead of doubling on that).

It also should greatly reduce the stress on Kubernetes apiserver as evaluating an NFG basically requires fetching the same objects from the apiserver as any node update (get node object and all related NRs). Just need to refactor the nfd-master code so that everything is re-used, e.g. just modify/rename `nfdAPIUpdateOneNode` to handle NFG, too. 

One special consideration is that now there are multiple parallel worker threads (as node-updater-pool workers) acting/racing on the same NFG object, trying to update its status. Maybe the easiest way to handle this would be through internal locking (have a per-NFG lock in nfd-master).

I'd be happy to hear others' thoughts on this, too. @PiotrProkop @adrianchiris @ahmedThresh

PiotrProkop:
> > As for locking if we would use patch instead of update maybe we can just implement some retries when resource version mismatch happens.
> 
> Reliably deleting a specific item from a list isn't possible with patching AFAIU.

Yeah, I haven't thought about removing  😞

cmd/nfd-master/main.go
```diff
@@ -135,6 +135,8 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 	flagset.BoolVar(&args.EnableNodeFeatureApi, "enable-nodefeature-api", true,
 		"Enable the NodeFeature CRD API for receiving node features. This will automatically disable the gRPC communication."+
 			" DEPRECATED: will be removed in a future release along with the deprecated gRPC API.")
+	flagset.BoolVar(&args.EnableNodeFeatureGroupApi, "enable-nodefeaturegroup-api", false,
+		"Enable the NodeFeatureGroup CRD API for receiving cluster feature rules.")
```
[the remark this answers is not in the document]
ArangoGutierrez: I think is better to go for `alpha=disabled-by-default and turn it on in the next release.`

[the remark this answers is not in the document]
ArangoGutierrez: >Another consideration is that should we introduce smth like kubernetes feature gates (like -feature-gates NodeFeatureGroupApi=true) to control this kind of new features instead of a somewhat arbitrary -enable-x or -disable-y flags. WDYT?

This sounds like a nice followup PR on it's own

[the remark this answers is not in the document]
marquiz: The more I look at this I think we should just add the `-feature-gates` flag and make this the first user of it. Either add it in a separate PR or this one. WDYT?

docs/usage/custom-resources.md
```diff
@@ -96,6 +96,52 @@ with the
...
+    nodes:
+    - node-1
+    - node-2
+    - node-3
+```
+
+NodeFeatureGroupRule is an NFD-specific custom resource that is designed for
+grouping Nodes by matching features rules. NFD-Master watches for
+NodeFeatureGroupRules objects in the cluster and creates a NodeFeatureGroup
+object for each NodeFeatureRule.
```
marquiz, line 126: This is not correct wrt. to the latest design. There is only one type of CRD object (NFG)

examples/nodefeaturerule.yaml
```diff
@@ -29,3 +29,15 @@ spec:
         - feature: kernel.version
           matchExpressions:
             major: { op: Exists }
+---
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureRule
```
marquiz, line 34: Is this relevant to NFG?

pkg/nfd-gc/nfd-gc.go
marquiz, line 1: I think we don't need nfd-gc anymore, with only single CRD. So these change s can be reverted. Right?

test/e2e/utils/crd.go
```diff
@@ -123,6 +123,21 @@ func CreateNodeFeatureRulesFromFile(ctx context.Context, cli nfdclientset.Interf
 	return nil
 }
 
+// CreateNodeFeatureGroupsFromFile creates a NodeFeatureGroupRule object from a given file located under test data directory.
```
marquiz, line 126: No rule, anymore

```suggestion
// CreateNodeFeatureGroupsFromFile creates a NodeFeatureGroup object from a given file located under test data directory.
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -873,6 +886,64 @@ core:
 				})
 			})
 
+			// Test NodeFeatureGroupRule
```
marquiz, line 889:
```suggestion
			// Test NodeFeatureGroup
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -873,6 +886,64 @@ core:
...
+			Context("and NodeFeatureGroupRules objects deployed", func() {
```
marquiz, line 890:
```suggestion
			Context("and NodeFeatureGroup objects deployed", func() {
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -873,6 +886,64 @@ core:
...
+				BeforeEach(func(ctx context.Context) {
+					// We need a NodeFeature from the node, can't be a fake one
+					if !useNodeFeatureApi {
+						Skip("NodeFeature API not enabled")
+					}
+				})
+				It("custom NodeFeatureGroup from the NodeFeatureGroupRule rules should be created", func(ctx context.Context) {
```
marquiz, line 897: Doesn't look right, e.g.

```suggestion
				It("NodeFeatureGroup should be updated", func(ctx context.Context) {
```

??

test/e2e/node_feature_discovery_test.go
```diff
@@ -873,6 +886,64 @@ core:
...
+					workerDS := testds.NFDWorker(podSpecOpts...)
+					workerDS, err = f.ClientSet.AppsV1().DaemonSets(f.Namespace.Name).Create(ctx, workerDS, metav1.CreateOptions{})
+					Expect(err).NotTo(HaveOccurred())
+
+					By("Waiting for worker daemonset pods to be ready")
+					Expect(testpod.WaitForReady(ctx, f.ClientSet, f.Namespace.Name, workerDS.Spec.Template.Labels["name"], 2)).NotTo(HaveOccurred())
+
+					checkNodeFeatureObject(ctx, targetNodeName)
+
+					By("Creating NodeFeatureGroupRules #1")
```
marquiz, line 924:
```suggestion
					By("Creating NodeFeatureGroup #1")
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -873,6 +886,64 @@ core:
...
+					Expect(testutils.CreateNodeFeatureGroupsFromFile(ctx, nfdClient, f.Namespace.Name, "nodefeaturegroup-1.yaml")).NotTo(HaveOccurred())
+
+					By("Verifying NodeFeatureGroup from NodeFeatureRules #1")
```
marquiz, line 927:
```suggestion
					By("Verifying NodeFeatureGroup #1")
```

pkg/nfd-master/updater-pool.go
```diff
@@ -85,16 +87,48 @@ func (u *nodeUpdaterPool) runNodeUpdater(queue workqueue.RateLimitingInterface)
 	u.wg.Done()
 }
 
-func (u *nodeUpdaterPool) start(parallelism int) {
+func (u *updaterPool) processNodeFeatureGroupUpdateRequest(ngfQueue workqueue.RateLimitingInterface) bool {
```
marquiz, line 90: Rebase on the latest master and replicate that we now have in `processNodeUpdateRequest()`. On top of my head at least:

- `cli` (kubernetes client) is an arg to this method
- check existence of NFG and pass cli and that object as args to `u.nfdMaster.nfdAPIUpdateNodeFeatureGroup`
- on errors retry indefinitely (unless notfound)

pkg/nfd-master/nfd-master.go
```diff
@@ -825,6 +846,80 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
...
+func (m *nfdMaster) nfdAPIUpdateNodeFeatureGroup(nodeFeatureGroupName string) error {
+	klog.V(1).InfoS("processing of nodeFeatureGroup", "nodeFeatureGroup", nodeFeatureGroupName)
+	nodeGroupValidator := make(map[string]bool)
+	var err error
+
+	kubeconfig, err := utils.GetKubeconfig(m.args.Kubeconfig)
+	if err != nil {
+		return err
+	}
+	nfdClient := nfdclientset.NewForConfigOrDie(kubeconfig)
```
marquiz, line 872: We should get the cli as an arg to the function

pkg/nfd-master/nfd-master.go
```diff
@@ -825,6 +846,80 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
...
+	}
+
+	// Execute rules and create matching groups
+	nodes := make([]string, 0)
+	for _, rule := range nodeFeatureGroup.Spec.Rules {
+		for _, feature := range nodeFeatures {
+			match, err := nodefeaturerule.ExecuteGroupRule(&rule, &feature.Spec.Features)
+			if err != nil {
+				return fmt.Errorf("failed to evaluate rule %q: %w", rule.Name, err)
+			}
```
marquiz, line 891: Hmm, this is different from how we handle NodeFeatureRules. We should follow the same logic here. Basically process every rule only once, against a union all NodeFeatures. I.e. merge all NodeFeatures (of this node) into one mega `Features` struct and give that as input to the rule. See how `nfdAPIUpdateOneNode()` does it. Maybe it would make sense to split out the merging logic into a common function

api/nfd/v1alpha1/types.go
```diff
@@ -131,6 +131,54 @@ type NodeFeatureRuleSpec struct {
 	Rules []Rule `json:"rules"`
 }
 
+// NodeFeatureGroup resource holds Node pools by featureGroup
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=nfg
```
marquiz, line 136: Hmm, I think we should handle status as a separate subresource

```suggestion
// +kubebuilder:resource:scope=Namespaced,shortName=nfg
// +kubebuilder:subresource:status
```

Then nfd-master will only update the status subresource and not touch the other parts

marquiz (commented):
Thanks @ArangoGutierrez for the update. Getting closer. Some comments below.

In addition to the code changes, I think the commit message should be a bit more verbose for a big change like this 😇 Explain what is done and why

ArangoGutierrez:
Depends on https://github.com/kubernetes-sigs/node-feature-discovery/pull/1681

pkg/nfd-master/nfd-master.go
ArangoGutierrez: done

pkg/nfd-master/updater-pool.go
ArangoGutierrez: done

test/e2e/node_feature_discovery_test.go
```diff
@@ -873,6 +887,74 @@ core:
...
+						group, err := nfdClient.NfdV1alpha1().NodeFeatureGroups(f.Namespace.Name).Get(ctx, "e2e-test-1", metav1.GetOptions{})
+						if err != nil {
+							return false
+						}
+						return areEqualLists(group.Status.Nodes, expectedGroup.Status.Nodes)
+					}, 5*time.Minute, 5*time.Second).Should(BeTrue())
+
+					By("Deleting nfd-worker daemonset")
+					err = f.ClientSet.AppsV1().DaemonSets(f.Namespace.Name).Delete(ctx, workerDS.Name, metav1.DeleteOptions{})
+					Expect(err).NotTo(HaveOccurred())
```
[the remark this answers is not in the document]
ArangoGutierrez: I removed this section

test/e2e/node_feature_discovery_test.go
```diff
@@ -873,6 +887,74 @@ core:
 				})
 			})
 
+			// Test NodeFeatureGroups
+			Context("and NodeFeatureGroups objects deployed", func() {
```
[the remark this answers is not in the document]
ArangoGutierrez: done

pkg/nfd-master/nfd-master.go
```diff
@@ -825,6 +846,80 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
 	return nil
 }
 
+func (m *nfdMaster) nfdAPIUpdateAllNodeFeatureGroups() error {
+	klog.V(1).InfoS("updating All NodeFeatureGroups, will process all nodeFeatureGroupRules")
+
+	nodeFeatureGroups, err := m.nfdController.featureGroupLister.List(k8sLabels.Everything())
+	if err != nil {
+		return fmt.Errorf("failed to get nodeFeatureGroupRules resources: %w", err)
```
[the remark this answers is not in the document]
ArangoGutierrez: done

pkg/nfd-master/nfd-master.go
```diff
@@ -489,12 +498,24 @@ func (m *nfdMaster) nfdAPIUpdateHandler() {
...
+				}
+			} else {
+				for nodeFeatureGroupName := range nodeFeatureGroup {
+					m.updaterPool.addNodeFeatureGroup(nodeFeatureGroupName)
 				}
 			}
 
 			// Reset "work queue" and timer
 			updateAll = errUpdateAll
+			updateAllNodeFeatureGroups = false
```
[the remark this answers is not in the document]
ArangoGutierrez: done

pkg/nfd-master/nfd-api-controller.go
```diff
@@ -91,8 +100,8 @@ func newNfdController(config *restclient.Config, nfdApiControllerOptions nfdApiC
 	}
 
 	// Add informer for NodeFeatureRule objects
-	ruleInformer := informerFactory.Nfd().V1alpha1().NodeFeatureRules()
-	if _, err := ruleInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+	nodeRuleInformer := informerFactory.Nfd().V1alpha1().NodeFeatureRules()
```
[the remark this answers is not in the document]
ArangoGutierrez: done

pkg/nfd-master/nfd-api-controller.go
```diff
@@ -149,9 +174,29 @@ func getNodeNameForObj(obj metav1.Object) (string, error) {
 	return nodeName, nil
 }
 
+func (c *nfdController) updateOneNode(typ string, obj metav1.Object) {
```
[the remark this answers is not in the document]
ArangoGutierrez: Just to accommodate the new funcs in order, no other reason

docs/reference/feature-gates.md
```diff
@@ -25,3 +26,11 @@ When enabled, NFD will register the Node Feature API with the Kubernetes API
 server. The Node Feature API is used to expose node-specific hardware and
 software features to the Kubernetes scheduler. The Node Feature API is a beta
 feature and is enabled by default.
+
+## NodeFeatureGroupAPI
+
+The `NodeFeatureGroupAPI` feature gate enables the Node Feature Group API.
+When enabled, NFD will register the Node Feature Group API with the Kubernetes
+API server. The Node Feature Group API is used to create nodeGroups of nodes
+that share common hardware and software features. The Node Feature Group API
```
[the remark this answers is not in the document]
ArangoGutierrez: done

pkg/nfd-master/nfd-master.go
ArangoGutierrez: I revisited this func, and this is different from the `nfdAPIUpdateOneNode` yes, but we are visiting every rule just once, the thing is that for each rule we need to see all nodes to check if they are fit for the rule and add them to the Node that will become a nodeFeatureGroup in the end.

pkg/nfd-master/updater-pool.go
```diff
@@ -85,16 +90,55 @@ func (u *nodeUpdaterPool) runNodeUpdater(queue workqueue.RateLimitingInterface)
...
+	if nfg, err = getNodeFeatureGroup(cli, u.nfdMaster.namespace, nfgName.(string)); apierrors.IsNotFound(err) {
+		klog.InfoS("NodeFeatureGroup not found, skip update", "NodeFeatureGroupName", nfgName)
+	} else if err := u.nfdMaster.nfdAPIUpdateNodeFeatureGroup(u.nfdMaster.nfdClient, nfg); err != nil {
+		if ngfQueue.NumRequeues(nfgName) < 15 {
+			klog.InfoS("retrying NodeFeatureGroup update")
+			ngfQueue.AddRateLimited(nfgName)
+			return true
+		} else {
+			klog.ErrorS(err, "failed to update NodeFeatureGroup")
+		}
```
marquiz, line 114: Let's retry indefinitely (like we do for nodes), or wdyt(?) Also, a bit more verbose log messages

```suggestion
		if ngfQueue.NumRequeues(nfgName) < 15 {
			klog.InfoS("retrying NodeFeatureGroup update", "nodeFeatureGroup", klog.KObj(nfg), "lastError", err)
		} else {
			klog.ErrorS(err, "failed to update NodeFeatureGroup, queueing for retry", "nodeFeatureGroup", klog.KObj(nfg), "lastError", err, "numRetries", n)
		}
		ngfQueue.AddRateLimited(nfgName)
		return true
```

I think it could be also just one error message, instead of info and err (separated by a rather arbitrary `< 15`), if you like.

pkg/nfd-master/nfd-master.go
marquiz: Well, we're visiting every rule once per each node. But the features of the node must be a union of all the NodeFeature objects of that node. Think a about a rule that checks that (in pseudocode) `featureA == "foo" && featureB == "bar"`. Now, the end result must be the same, IMO, independent whether `featureA` and `featureB` are specified in one NodeFeature object or two separate ones. You might want to write rules that combine NFD's builtin features with some 3rd party extension and that's not possible if you don't merge all features of a node into a union. Plus, importantly, it's also about consistency, treat features the same way whether we're processing NodeFeatureRules or NodeFeatureGroup (rules).

api/nfd/v1alpha1/types.go
marquiz: You changed this, right? Can be closed as resolved(?)

marquiz (commented):
Thanks for the update @ArangoGutierrez. I think we're pretty close. I had one comment below and one unresolved comment about merging the node features when processing NFGs.

In addition to that we'd need a bit more documentation
- describe NFD in `docs/usage/custom-resources.md` (I think this was in some earlier iteration of the PR but has been lost)
- a new section in `docs/usage/customization-guide.md` describing the usage (with an example) of NFG

EDIT: forgot to add that RBAC rules for nfd-master to be able to get/list/watch NFGs and update the status (subresource) of NFG is needed

pkg/nfd-master/nfd-master.go
ArangoGutierrez: We a re visiting all nodes for each rule*

pkg/nfd-master/nfd-master.go
```diff
@@ -825,6 +868,92 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
...
+		}
+		// After the nfd namespace, sort objects by their name
+		if objs[i].Name != objs[j].Name {
+			return objs[i].Name < objs[j].Name
+		}
+		// Objects with the same name are sorted by their namespace
+		return objs[i].Namespace < objs[j].Namespace
+	})
+
+	klog.V(1).InfoS("processing of nodeFeatureGroup", "nodeFeatureGroup", nodeFeatureGroup.Name)
```
marquiz, line 915:
```suggestion
	klog.V(1).InfoS("processing of nodeFeatureGroup", "nodeFeatureGroup", klog.KObj(nodeFeatureGroup))
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -1044,3 +1122,20 @@ func getNode(nodes []corev1.Node, nodeName string) corev1.Node {
...
+		set1[elem] = struct{}{}
+	}
+
+	set2 := make(map[string]struct{})
+	for _, elem := range list2 {
+		set2[elem] = struct{}{}
+	}
+
+	// Check if sets are equal
+	return reflect.DeepEqual(set1, set2)
```
marquiz, line 1140: You could use the sets package from apimachinery (`k8s.io/apimachinery/pkg/util/sets`)

```suggestion
	set1 := sets.NewString(list1...)
	set2 := sets.NewString(list2...)
	return set1.Equal(set2)
```

pkg/nfd-master/nfd-master.go
ArangoGutierrez: Now I got what you meant, my traveling brain was melted. will address

test/e2e/node_feature_discovery_test.go
ArangoGutierrez: done

pkg/nfd-master/nfd-master.go
```diff
@@ -792,16 +838,14 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
 		return objs[i].Namespace < objs[j].Namespace
 	})
 
-	klog.V(1).InfoS("processing of node initiated by NodeFeature API", "nodeName", node.Name)
-
-	features := nfdv1alpha1.NewNodeFeatureSpec()
+	klog.V(1).InfoS("processing of node initiated by", "API", api, "nodeName", nodeName)
```
marquiz, line 841: I think this is confusing log message. 

```suggestion
```

Also drop the `api` arg from the method

pkg/nfd-master/nfd-master.go
```diff
@@ -758,21 +801,24 @@ func (m *nfdMaster) nfdAPIUpdateAllNodes() error {
...
+			Name: nodeName,
+		},
 	}
 
-	sel := k8sLabels.SelectorFromSet(k8sLabels.Set{nfdv1alpha1.NodeFeatureObjNodeNameLabel: node.Name})
+	sel := k8sLabels.SelectorFromSet(k8sLabels.Set{nfdv1alpha1.NodeFeatureObjNodeNameLabel: nodeName})
 	objs, err := m.nfdController.featureLister.List(sel)
 	if err != nil {
-		return fmt.Errorf("failed to get NodeFeature resources for node %q: %w", node.Name, err)
+		return nodeFeatures, fmt.Errorf("failed to get NodeFeature resources for node %q: %w", nodeName, err)
```
marquiz, line 821: We should also somehow distinguish nodes that do not have any nodefeatures so that we don't try to process nodes that don't have any features discovered. E.g. by returning `nil` or so. WDYT?

cmd/nfd-master/main.go
```diff
@@ -111,7 +111,9 @@ func main() {
 
 	// Get new NfdMaster instance
 	args.GrpcHealthPort = GrpcHealthPort
-	instance, err := master.NewNfdMaster(master.WithArgs(args))
+	instance, err := master.NewNfdMaster(
+		master.WithArgs(args),
+	)
```
marquiz, line 116: nit: kinda noise, irrelevant. Suggest to keep it as it was

```suggestion
	instance, err := master.NewNfdMaster(master.WithArgs(args))
```

examples/nodefeaturegroup.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureGroup
+metadata:
+  name: node-feature-group-1
+spec:
+  featureGroupRules:
+    - name: "e2e-attribute-test-1"
```
marquiz, line 7: nit: sync with what we have in docs

```suggestion
metadata:
  name: node-feature-group-example
spec:
  featureGroupRules:
    - name: "kernel version"
```

pkg/nfd-master/nfd-master.go
```diff
@@ -813,19 +857,120 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
...
 		return err
 	}
 
 	return nil
 }
 
+func (m *nfdMaster) nfdAPIUpdateAllNodeFeatureGroups() error {
+	klog.V(1).InfoS("updating all NodeFeatureGroups")
+
+	nodeFeatureGroupsList, err := m.nfdClient.NfdV1alpha1().NodeFeatureGroups(m.namespace).List(context.TODO(), metav1.ListOptions{})
```
marquiz, line 893: Hmm, couldn't/shouldn't we use the lister from the controller to do this? We're already watching/listing/caching all NFGs there so let's use that instead of putting extra burden on the API server(?)

pkg/nfd-master/nfd-master.go
```diff
@@ -792,16 +838,14 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
 		return objs[i].Namespace < objs[j].Namespace
 	})
 
-	klog.V(1).InfoS("processing of node initiated by NodeFeature API", "nodeName", node.Name)
-
-	features := nfdv1alpha1.NewNodeFeatureSpec()
+	klog.V(1).InfoS("processing of node initiated by", "nodeName", nodeName)
```
marquiz, line 841: Hmm, this log message looks pretty useless to me (and this new version looks confusing). Prolly dates back to some experiments with the NodeFeatureAPI. I think we could just ditch this msg in this PR

```suggestion
```

pkg/nfd-master/nfd-master.go
```diff
@@ -813,19 +857,120 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
...
+		}
+	} else {
+		klog.V(2).InfoS("no NodeFeatureGroup objects found")
+	}
+
+	return nil
+}
+
+func (m *nfdMaster) nfdAPIUpdateNodeFeatureGroup(nfdClient *nfdclientset.Clientset, nodeFeatureGroup *nfdv1alpha1.NodeFeatureGroup) error {
+	klog.V(1).InfoS("updating NodeFeatureGroup", "nodeFeatureGroup", klog.KObj(nodeFeatureGroup))
```
marquiz, line 910: This feels a bit redundant as later we log if the object was updated or (with V(1)) if there were no changes. If we want to keep this, increase the verbosity and change the wording a bit(?)

```suggestion
```

OR

```suggestion
	klog.V(2).InfoS("evaluating NodeFeatureGroup", "nodeFeatureGroup", klog.KObj(nodeFeatureGroup))
```

pkg/nfd-master/nfd-master.go
```diff
@@ -813,19 +857,120 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
...
+	}
+
+	// Execute rules and create matching groups
+	nodePool := make([]string, 0)
+	nodeGroupValidator := make(map[string]bool)
+	for _, rule := range nodeFeatureGroup.Spec.Rules {
+		for _, feature := range nodeFeaturesList {
+			match, err := nodefeaturerule.ExecuteGroupRule(&rule, &feature.Spec.Features)
+			if err != nil {
+				return fmt.Errorf("failed to evaluate rule %q: %w", rule.Name, err)
```
marquiz, line 941: Could we log the node name for which this failed?

pkg/nfd-master/nfd-master.go
```diff
@@ -758,21 +797,29 @@ func (m *nfdMaster) nfdAPIUpdateAllNodes() error {
...
 	}
 
 	return nil
 }
 
-func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.Node) error {
-	if m.nfdController == nil || m.nfdController.featureLister == nil {
-		return nil
+// mergeNodeFeatures merges the features of the given NodeFeature objects into a single NodeFeatureSpec.
+func (m *nfdMaster) mergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeFeature, error) {
```
marquiz, line 807: It's so easy to nit about code written by others 🙄 But now reading this the function name is a bit misleading to me.

```suggestion
func (m *nfdMaster) getMergedNodeFeatures(nodeName string) (*nfdv1alpha1.NodeFeature, error) {
```

**OR**

```suggestion
func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeFeature, error) {
```

pkg/nfd-master/nfd-master.go
```diff
@@ -813,19 +856,121 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
...
+
+	// Execute rules and create matching groups
+	nodePool := make([]string, 0)
+	nodeGroupValidator := make(map[string]bool)
+	for _, rule := range nodeFeatureGroup.Spec.Rules {
+		for _, feature := range nodeFeaturesList {
+			match, err := nodefeaturerule.ExecuteGroupRule(&rule, &feature.Spec.Features)
+			if err != nil {
+				klog.ErrorS(err, "failed to evaluate rule", "ruleName", rule.Name)
+				return fmt.Errorf("failed to evaluate rule %q: %w", rule.Name, err)
```
marquiz, line 941: Just log the error, don't return so that we don't end up in infinite re-try loop 

```suggestion
				continue
```

pkg/nfd-master/nfd-master.go
```diff
@@ -758,21 +797,29 @@ func (m *nfdMaster) nfdAPIUpdateAllNodes() error {
...
 }
 
-func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.Node) error {
-	if m.nfdController == nil || m.nfdController.featureLister == nil {
-		return nil
+// mergeNodeFeatures merges the features of the given NodeFeature objects into a single NodeFeatureSpec.
+func (m *nfdMaster) mergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeFeature, error) {
+	nodeFeatures := &nfdv1alpha1.NodeFeature{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: nodeName,
```
marquiz, line 810: I think piggy-backing on the Name field is fine here. You could add a comment to the function description ("docstring") that the function returns a NodeFeature object where the name field is set to the node name. WDYT?

pkg/nfd-master/nfd-master.go
```diff
@@ -813,19 +856,121 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
...
+	}
+
+	// Execute rules and create matching groups
+	nodePool := make([]string, 0)
+	nodeGroupValidator := make(map[string]bool)
+	for _, rule := range nodeFeatureGroup.Spec.Rules {
+		for _, feature := range nodeFeaturesList {
+			match, err := nodefeaturerule.ExecuteGroupRule(&rule, &feature.Spec.Features)
+			if err != nil {
+				klog.ErrorS(err, "failed to evaluate rule", "ruleName", rule.Name)
```
marquiz, line 940: I'd log the node name, too.

```suggestion
				klog.ErrorS(err, "failed to evaluate rule", "ruleName", rule.Name, "nodeName", feature.Name)
```

marquiz (commented):
Thanks @ArangoGutierrez for the update. Possibly the last round of comments for this one 🤞

test/e2e/node_feature_discovery_test.go
```diff
@@ -873,6 +887,70 @@ core:
...
+					Expect(err).NotTo(HaveOccurred())
+
+					targetNodes := make([]string, 0)
+					for _, node := range nodes {
+						checkNodeFeatureObject(ctx, node.Name)
+						if !Expect(node.Name).ToNot(BeEmpty(), "No suitable worker node found") {
+							continue
+						}
+						targetNodes = append(targetNodes, node.Name)
+					}
```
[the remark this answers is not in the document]
ArangoGutierrez: fixed, the idea is to create a list of node names. new implementation is simpler

pkg/nfd-master/nfd-master.go
```diff
@@ -758,21 +797,30 @@ func (m *nfdMaster) nfdAPIUpdateAllNodes() error {
...
 	}
 
 	return nil
 }
 
-func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.Node) error {
-	if m.nfdController == nil || m.nfdController.featureLister == nil {
-		return nil
+// getAndMergeNodeFeatures merges the features of the given NodeFeature objects into a single NodeFeatureSpec.
+// It returns a NodeFeature object with the merged features for the given node.
```
marquiz, line 807: nit: this now looks like saying the same thing twice
```suggestion
// getAndMergeNodeFeatures merges the NodeFeature objects of the given node into a single NodeFeatureSpec.
// The Name field of the returned NodeFeatureSpec contains the node name.
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -873,6 +887,62 @@ core:
...
+
+					By("Creating NodeFeatureGroups #1")
+					Expect(testutils.CreateNodeFeatureGroupsFromFile(ctx, nfdClient, f.Namespace.Name, "nodefeaturegroup-1.yaml")).NotTo(HaveOccurred())
+
+					By("Verifying NodeFeatureGroups #1")
+					targetNodes := make([]string, 0)
+					for _, node := range nodes {
+						if !Expect(node.Name).ToNot(BeEmpty(), "No suitable worker node found") {
+							continue
+						}
```
marquiz, line 927: I still don't get this. In which case would `node.Name` be empty? Why just not `targetNodes = nodes` -> drop the targetNodes var and just use nodes?

ArangoGutierrez: NFG.Status.Nodes is a list of strings, so we only need the name, but you are right, I can trim it a bit

api/nfd/v1alpha1/types.go
```diff
@@ -131,6 +131,55 @@ type NodeFeatureRuleSpec struct {
...
+}
+
+// NodeFeatureGroupSpec describes a NodeFeatureGroup object.
+type NodeFeatureGroupSpec struct {
+	Rules []GroupRule `json:"featureGroupRules"`
+}
+
+type NodeFeatureGroupStatus struct {
+	// Nodes is a list of nodes in the cluster that match the featureGroupRules
+	// +optional
```
marquiz, line 155: Small suggestion for improvement (we shouldn't have duplicate entries here)

```suggestion
	// +optional
	// +listType=set
```

marquiz (approved):
@ArangoGutierrez kudos for the monumental effort on this feature. With the latest changes I believe and hope we came up with a extensible future proof API for the node feature groups. I hope people find this useful 🚀

/lgtm

PR #1489, opened 2023-12-05
Makefile: add env var controls to make test targets configurable

ArangoGutierrez:


AhmedGrati (approved):

marquiz:
Thanks @ArangoGutierrez for bringing this up. When/if bringing support for this finer-grained control on what to test I think a better approach would really be to enable [Ginkgo labels](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#spec-labels). We could start with simple top/high level labels like `master`, `worker`, `topology-updater` and `gc` for example and then add more fine grained labels later on (like `nfr`, `nf`, `nrt`, `crd-api`, `grpc-api` or whatever). WDYT?

ArangoGutierrez:
> Thanks @ArangoGutierrez for bringing this up. When/if bringing support for this finer-grained control on what to test I think a better approach would really be to enable [Ginkgo labels](https://onsi.github.io/ginkgo/MIGRATING_TO_V2#spec-labels). We could start with simple top/high level labels like `master`, `worker`, `topology-updater` and `gc` for example and then add more fine grained labels later on (like `nfr`, `nf`, `nrt`, `crd-api`, `grpc-api` or whatever). WDYT?

How about now?

Makefile
```diff
@@ -44,6 +44,9 @@ K8S_NAMESPACE ?= node-feature-discovery
 
 OPENSHIFT ?=
 
+RUN_FULL_E2E ?= true
```
marquiz, line 47: I'd suggest to prefix the name with E2E, like the other corresponding e2e-related variables, `E2E_TEST_ALL_IMAGES` or `E2E_TEST_FULL_IMAGE` or so... If we want we could also make this disabled by default (and enable it only in `scripts/test-infra/test-e2e.sh`) making local testing nicer – this could be done in a separate PR, too.

I think it would also make sense to move this a few lines downwards where the other `E2E_*` variables are specified.

marquiz (commented):
Nice update 👍 A few thoughts on the new version below

test/e2e/node_feature_discovery_test.go
```diff
@@ -177,7 +177,7 @@ func cleanupCRs(ctx context.Context, cli *nfdclient.Clientset, namespace string)
 }
 
 // Actual test suite
-var _ = SIGDescribe("NFD master and worker", func() {
+var _ = Describe("[k8s-sig/node-feature-discovery]", Label("e2e-test", "master-and-worker"), func() {
```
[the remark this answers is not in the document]
marquiz: Yeah, a bit contradictory 🙄 Could be a follow-up, too. Just that I think separate labels for master and worker would be more useful than `master-and-worker`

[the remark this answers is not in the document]
ArangoGutierrez: I agree, but as you said, will make this PR bigger, let's try to keep it small

Makefile
```diff
@@ -56,10 +56,14 @@ HOSTMOUNT_PREFIX ?= /
 KUBECONFIG ?= ${HOME}/.kube/config
 E2E_TEST_CONFIG ?=
 E2E_PULL_IF_NOT_PRESENT ?= false
+E2E_TEST_FULL_IMAGE ?= false
```
marquiz, line 59: Please enable this in `scripts/test-infra/test-e2e.sh` (as commented [here](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1489#discussion_r1419200447))

Makefile
```diff
@@ -183,24 +187,26 @@ helm-lint:
...
-	$(GO_CMD) test -covermode=atomic -coverprofile=coverage.out ./cmd/... ./pkg/... ./source/...
+	$(GO_CMD) test $(TEST_FLAGS) ./cmd/... ./pkg/... ./source/...
 
 e2e-test:
 	@if [ -z ${KUBECONFIG} ]; then echo "[ERR] KUBECONFIG missing, must be defined"; exit 1; fi
 	$(GO_CMD) test -v ./test/e2e/ -args -nfd.repo=$(IMAGE_REPO) -nfd.tag=$(IMAGE_TAG_NAME) \
 	    -kubeconfig=$(KUBECONFIG) \
 	    -nfd.e2e-config=$(E2E_TEST_CONFIG) \
 	    -nfd.pull-if-not-present=$(E2E_PULL_IF_NOT_PRESENT) \
-	    -ginkgo.focus="\[kubernetes-sigs\]" \
```
marquiz, line 194: We should keep the ginkgo focus in place. We don't want to accidentally start running kubernetes e2e tests. We import kubernetes e2e so the test specs are imported, too. We might (want to) change that, i.e. not import k8s e2e but just it's e2e framework but that's a subject of other PR(s)

Makefile
```diff
@@ -56,10 +56,14 @@ HOSTMOUNT_PREFIX ?= /
 KUBECONFIG ?= ${HOME}/.kube/config
 E2E_TEST_CONFIG ?=
 E2E_PULL_IF_NOT_PRESENT ?= false
+E2E_TEST_FULL_IMAGE ?= false
+E2E_GINKGO_LABEL_FILTER ?= "e2e-test"
```
marquiz, line 60: If/when we use ginkgo focus we could keep this empty by default

test/e2e/nfd_gc_test.go
```diff
@@ -38,7 +38,7 @@ import (
 )
 
 // Actual test suite
-var _ = SIGDescribe("NFD GC", func() {
+var _ = NFDDescribe(Label("e2e-test", "nfd-gc"), func() {
```
marquiz, line 41: I think we could drop the `e2e-test` label. Or then specify it in the NFDDescribe helper

```suggestion
var _ = NFDDescribe(Label("nfd-gc"), func() {
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -177,7 +177,7 @@ func cleanupCRs(ctx context.Context, cli *nfdclient.Clientset, namespace string)
 }
 
 // Actual test suite
-var _ = SIGDescribe("NFD master and worker", func() {
+var _ = NFDDescribe(Label("e2e-test", "master-and-worker"), func() {
```
marquiz, line 180: You added `nfd-worker` below so just use `nfd-master` here
```suggestion
var _ = NFDDescribe(Label("nfd-master"), func() {
```

test/e2e/framework.go
```diff
@@ -18,7 +18,6 @@ package e2e
 
 import "github.com/onsi/ginkgo/v2"
 
-// SIGDescribe annotates the test with the SIG label.
-func SIGDescribe(text string, body func()) bool {
-	return ginkgo.Describe("[kubernetes-sigs] "+text, body)
+func NFDDescribe(args ...interface{}) bool {
+	return ginkgo.Describe("[k8s-sigs/node-feature-discovery]", args...)
```
marquiz, line 22: If we want a top-level ginkgo label covering all nfd e2e tests we can specify it here to avoid repeating it all over

```suggestion
	return ginkgo.Describe("[k8s-sigs/node-feature-discovery]", append(args, ginkgo.Label("nfd-all"))...)
```

Makefile
```diff
@@ -56,10 +56,14 @@ HOSTMOUNT_PREFIX ?= /
 KUBECONFIG ?= ${HOME}/.kube/config
 E2E_TEST_CONFIG ?=
 E2E_PULL_IF_NOT_PRESENT ?= false
+E2E_TEST_FULL_IMAGE ?= false
+E2E_GINKGO_LABEL_FILTER ?=
 
 BUILD_FLAGS = -tags osusergo,netgo \
               -ldflags "-s -w -extldflags=-static -X sigs.k8s.io/node-feature-discovery/pkg/version.version=$(VERSION) -X sigs.k8s.io/node-feature-discovery/pkg/utils/hostpath.pathPrefix=$(HOSTMOUNT_PREFIX)"
 
+TEST_FLAGS = -covermode=atomic -coverprofile=coverage.out
```
marquiz, line 65: Unrelated 
```suggestion
```

docs/developer-guide/index.md
```diff
@@ -164,6 +164,17 @@ cluster you need to specify the kubeconfig to be used:
...
+There are several environment variables that can be used to customize the
+e2e-tests:
+
+| Variable                   | Description                                                       | Default value |
+| -------------------------- | ----------------------------------------------------------------- | ------------- |
+| KUBECONFIG                | Kubeconfig for running e2e-tests                                  | *empty* |
+| E2E_TEST_CONFIG           | Parameterization file of e2e-tests (see [example][e2e-config-sample]) | *empty* |
+| E2E_PULL_IF_NOT_PRESENT  | Set ImagePullPolicy to IfNotPresent during e2e-tests | false |
+| E2E_TEST_FULL_IMAGE     | Run e2e-test also against the Full Image | false |
+| E2E_GINKGO_LABEL_FILTER | Ginkgo label filter to use for running e2e tests | *emptys* |
```
marquiz, line 176: nit: fix the table alignment – there is clearly effort to have it nicely aligned so make it so

Another suggestion: it's nice to have the test related variables in a separate table here -> I suggest to remove the E2E_* variables from the other table

docs/developer-guide/index.md
```diff
@@ -121,10 +121,7 @@ makefile overrides.
 | IMAGE_REGISTRY             | Container image registry to use                                   | registry.k8s.io/nfd |
 | IMAGE_TAG_NAME             | Container image tag name                                          | &lt;nfd version&gt; |
 | IMAGE_EXTRA_TAG_NAMES      | Additional container image tag(s) to create when building image   | *empty* |
-| K8S_NAMESPACE              | nfd-master and nfd-worker namespace                               | node-feature-discovery |
-| KUBECONFIG                 | Kubeconfig for running e2e-tests                                  | *empty* |
-| E2E_TEST_CONFIG            | Parameterization file of e2e-tests (see [example][e2e-config-sample]) | *empty* |
-| E2E_PULL_IF_NOT_PRESENT    | True-ish value makes the image pull policy IfNotPresent (to be used only in e2e tests) | false |
+| K8S_NAMESPACE              | nfd-master and nfd-worker namespace                               | node-feature-discovery |s
```
marquiz, line 124:
```suggestion
| K8S_NAMESPACE              | nfd-master and nfd-worker namespace                               | node-feature-discovery |
```

marquiz (approved):
Thanks @ArangoGutierrez for the PR. This makes running/debugging/developing e2e tests locally sooo much nicer.
/lgtm

PR #1490, opened 2023-12-07
Change label system-os_release.VERSION_ID to system-os_release.VERSION_ID.full

zvlb:
In this PR, I changed the label name from `feature.node.kubernetes.io/system-os_release.VERSION_ID` to `feature.node.kubernetes.io/system-os_release.VERSION_ID.full`.

The issue arises from the fact that I am collecting logs from the pods to Elasticsearch. When gathering meta-information about the pod, all node labels are also collected. As a result, the following labels are gathered:
```
"feature.node.kubernetes.io/system-os_release.VERSION_ID": "22.04",
"feature.node.kubernetes.io/system-os_release.VERSION_ID.major": "22",
"feature.node.kubernetes.io/system-os_release.VERSION_ID.minor": "04"
```

However, attempting to write this information to the Elasticsearch index will result in an error because the field `feature.node.kubernetes.io/system-os_release.VERSION_ID` simultaneously serves as a string with the value `22.04` and as an object with nested `major` and `minor` fields. Elasticsearch will reject such logs with a mapping error.

To fix this issue, I changed the label name from `feature.node.kubernetes.io/system-os_release.VERSION_ID` to `feature.node.kubernetes.io/system-os_release.VERSION_ID.full` (similar to what is done, for example, in the kernel source).

P.S. I build and test it in my K8s:
```bash
$ kubectl get no -o json | jq '.items[].metadata.labels' | grep "system-os_release"
  "feature.node.kubernetes.io/system-os_release.ID": "ubuntu",
  "feature.node.kubernetes.io/system-os_release.VERSION_ID.full": "22.04",
  "feature.node.kubernetes.io/system-os_release.VERSION_ID.major": "22",
  "feature.node.kubernetes.io/system-os_release.VERSION_ID.minor": "04",
  ...
```

### DIFF
--- a/docs/usage/features.md
+++ b/docs/usage/features.md
@@ -227,7 +227,7 @@ instructions.
 | Feature                                 | Value  | Description                                                 |
 | --------------------------------------- | ------ | ----------------------------------------------------------- |
 | **`system-os_release.ID`**              | string | Operating system identifier                                  |
-| **`system-os_release.VERSION_ID`**      | string | Operating system version identifier (e.g. '6.7')            |
+| **`system-os_release.VERSION_ID.full`** | string | Operating system version identifier (e.g. '6.7')            |
 | **`system-os_release.VERSION_ID.major`**| string | First component of the OS version id (e.g. '6')             |
 | **`system-os_release.VERSION_ID.minor`**| string | Second component of the OS version id (e.g. '7')            |
 
--- a/source/system/system.go
+++ b/source/system/system.go
@@ -91,7 +91,8 @@ func (s *systemSource) Discover() error {
 	} else {
 		s.features.Attributes[OsReleaseFeature] = nfdv1alpha1.NewAttributeFeatures(release)
 
-		if v, ok := release["VERSION_ID.full"]; ok {
+		if v, ok := release["VERSION_ID"]; ok {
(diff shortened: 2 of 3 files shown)


zvlb:
I don't know who is using label `feature.node.kubernetes.io/system-os_release.VERSION_ID` in their operations. However, if my PR is accepted, it is likely that users may experience issues.

I can make it so that the label `system-os_release.VERSION_ID.full` is used instead of `system-os_release.VERSION_ID` if a certain flag is present. 
However, I don't want to implement this without confirmation. Additionally, I need a name for the flag)

marquiz:
> However, if my PR is accepted, it is likely that users may experience issues.

Yes, we shouldn't break existing users without a notice and transition period. IF we were  to do this, we should deprecate the old label and support it for a few releases (see https://kubernetes-sigs.github.io/node-feature-discovery/v0.14/reference/versions.html#feature-labels).

Some alternative approaches:

1. You disable the problematic label(s) locally with your NFD configuration e.g. with the `labelWhitelist` config option or just disabling the whole `system` feature source (removing all `system-os_release.*` labels) if you don't need that info.
2. Instead of renaming the label make the `system` feature source configurable, e.g. select which fields from `/etc/os-release` to publish as labels

PR #1491, opened 2023-12-08
nfd-worker: set owner reference in NodeFeature objects

marquiz:
This patch creates a owner-dependent relationship between the nfd-worker pod and the NodeFeature object that it creates. With this change the orphaned NodeFeature object(s) gets automatically garbage-collected when the nfd-worker pod goes away, without the need for manual clean-up actions.


ArangoGutierrez (approved):
/lgtm

PR #1492, opened 2023-12-08
deployment/kustomize: drop default-combined overlay

marquiz:
The "combined" overlay, deploying nfd-master and nfd-worker in the same pod (with a daemonset) doesn't make sense anymore as we have enabled NodeFeature API. There is no direct communication between nfd-master and nfd-worker anymore, Moreover, the combined deployment can be seen as broken as there is one NodeFeature controller (i.e. nfd-master) on each node, causing them to race against each other, all processing all NodeFeature objects.


ArangoGutierrez (approved):
Never used nor understood this combined option .
/lgtm

PR #1494, opened 2023-12-08
deployment/kustomize: drop nfd-master service

marquiz:
Not needed anymore as we're not relying on gRPC anymore.


ArangoGutierrez (approved):
/lgtm

PR #1497, opened 2023-12-08
feat: add cpu socket count in `cpu.topology`

### LINKED ISSUE #1383: Add the CPU Socket information

The CPU Sockets number, for the bare metal worker nodes (otherwise, virtualized worker nodes' sockets count is just emulated number).

**Why is this needed**:

On some scenarios, the number of CPU Sockets is required for the license calculation perspective.
For example, take a look https://access.redhat.com/discussions/480953 where the 3rd party applications are licensed based on the number of CPU Sockets, where Kubernetes cluster (OpenShift, to be specific) runs on the bare metal systems.

Given that in **some** scenarios it's [possible to discover the number of cpu sockets](https://stackoverflow.com/questions/59158453/how-to-programmatically-find-the-number-of-cpu-sockets-and-core-per-socket-in-li), tools available like https://github.com/ryran/xsos, I believe such information shall be exposed, with the caveat that in some other conditions such value is either not available, or just represents the virtual number of sockets. 

There was a related issue closed https://github.com/kubernetes-sigs/node-feature-discovery/issues/784 where there was a comment https://github.com/kubernetes-sigs/node-feature-discovery/issues/784#issuecomment-1069554415 from @marquiz  that the node labels might not properly reflect the cpu related attributes in some cases, but wondering whether if we narrowed down the set of scenarios to support, it would make sense to proceed?

Other way of asking - if there were PR submitted, to introduce number of CPU Sockets, what are the things to consider, or perhaps such PR would be rejected right at the beginning?

AhmedGrati:
Resolves #1383

### DIFF
--- a/source/cpu/cpu.go
+++ b/source/cpu/cpu.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/klog/v2"
@@ -276,7 +277,36 @@ func discoverTopology() map[string]string {
 		features["hardware_multithreading"] = strconv.FormatBool(ht)
 	}
 
+	if socketCount, err := getCPUSocketCount(); err != nil {
+		klog.ErrorS(err, "failed to get sockets count")
+	} else {
+		features["socket_count"] = strconv.FormatInt(socketCount, 10)
+	}
+
 	return features
+}
+
+func getCPUSocketCount() (int64, error) {
+	files, err := os.ReadDir(hostpath.SysfsDir.Path("bus/cpu/devices"))
+	if err != nil {
+		return 0, err
+	}
+
+	uniquePhysicalIDs := sets.NewString()
+
+	for _, file := range files {
+		// Try to read physical_package_id from topology
+		physicalID, err := os.ReadFile(hostpath.SysfsDir.Path("bus/cpu/devices", file.Name(), "topology/physical_package_id"))
+		if err != nil {
+			return 0, err
+		}
+		id := strings.TrimSpace(string(physicalID))
+		if err != nil {
+			return 0, err
+		}
+		uniquePhysicalIDs.Insert(id)
+	}
+	return int64(uniquePhysicalIDs.Len()), nil
 }
 
 // Check if any (online) CPUs have thread siblings
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -867,6 +867,7 @@ The following features are available for matching:
 |                  |              | **`bf.enabled`** | bool | `true` if Intel SST-BF (Intel Speed Select Technology - Base frequency) has been enabled, otherwise does not exist |
 | **`cpu.topology`** | attribute  |          |            | CPU topology related features |
 | | |          **`hardware_multithreading`** | bool       | Hardware multithreading, such as Intel HTT, is enabled |
+| | |          **`socket_count`**            | int        | Number of CPU Sockets |
 | **`cpu.coprocessor`** | attribute |        |            | CPU Coprocessor related features |
 | | |          **`nx_gzip`**                 | bool       | Nest Accelerator GZIP support is enabled |
 | **`kernel.config`** | attribute |          |            | Kernel configuration options |


marquiz (commented):
Thanks for working on this @AhmedGrati.

Instead of grepping /proc/cpuinfo (and mounting /proc from the host) we should get the information from sysfs. For example, iterate over cpu devices and read their `topology/physical_package_id`. Number of unique ids is the number of sockets. I think we already loop over all cpus when checking for smt so we should be able to hook the new functionality there.

AhmedGrati:
Thanks @marquiz @PiotrProkop. PTAL.

source/cpu/cpu.go
```diff
@@ -276,9 +282,38 @@ func discoverTopology() map[string]string {
...
+	if socketCount, err := getCPUSocketCount(); err != nil {
+		klog.ErrorS(err, "failed to get sockets count")
+	} else {
+		features["socket_count"] = strconv.FormatInt(socketCount, 10)
+	}
+
 	return features
 }
 
+func getCPUSocketCount() (int64, error) {
```
marquiz, line 294: Thanks @AhmedGrati for the update.

One note that this `getCPUSocketCount()` looks very similar to `haveThreadSiblings()`. I think an improvement could be to refactor the code by combining these both e.g. directly inside `discoverTopology()`. That could also be a subsequent PR. WDYT?

AhmedGrati: I can do it in the same PR, no problem from my side.

marquiz: Up to you, really. It'd be nice to merge this soon, though 😅

One difference in combining these into one func is probably the error handling. If any (file read) error is encountered then no "topology" features are returned (vs. with the current implementation socket_count can be returned even if smt detection fails). But I think this is pretty theoretical (or not significant) as there should be little to none practical cases where thread_siblings could be read but the physical_package_id not (or vise versa).

AhmedGrati: Got it, let's merge this then, and I'll open another PR for the combination.

source/cpu/cpu.go
```diff
@@ -276,9 +282,38 @@ func discoverTopology() map[string]string {
...
+
+	uniquePhysicalIDs := make(map[int]bool)
+
+	for _, file := range files {
+		// Try to read physical_package_id from topology
+		physicalID, err := os.ReadFile(hostpath.SysfsDir.Path("bus/cpu/devices", file.Name(), "topology/physical_package_id"))
+		if err != nil {
+			return 0, err
+		}
+		id, err := strconv.Atoi(strings.TrimSuffix(string(physicalID), "\n"))
```
marquiz, line 308: We don't actually need to convert the value to integer type, I think. Then, we could use TrimSpace to simplify a bit (trima all leading and trailing whitespace).

Thus, how about?


```suggestion
		id, err := strings.TrimSpace(string(physicalID))
```

source/cpu/cpu.go
```diff
@@ -276,9 +282,38 @@ func discoverTopology() map[string]string {
...
 	return features
 }
 
+func getCPUSocketCount() (int64, error) {
+	files, err := os.ReadDir(hostpath.SysfsDir.Path("bus/cpu/devices"))
+	if err != nil {
+		return 0, err
+	}
+
+	uniquePhysicalIDs := make(map[int]bool)
```
marquiz, line 300: nit: maybe just `map[int]struct{}` because we don't really use the value for anything.

```suggestion
	uniquePhysicalIDs := make(map[int]struct{})
```

Alternatively, I think, we could start using the `k8s.io/apimachinery/pkg/util/sets` (e.g. `sets.NewInt()` or `sets.NewString()` in this case) package in NFD as we already depend on apimachinery.

marquiz (approved):
👍 Two suggestions below

source/cpu/cpu.go
AhmedGrati: that makes more sense.

source/cpu/cpu.go
```diff
@@ -197,6 +198,11 @@ func (s *cpuSource) GetLabels() (source.FeatureLabels, error) {
 		labels["hardware_multithreading"] = v
 	}
 
+	// CPU Socket Count
+	if v, ok := features.Attributes[TopologyFeature].Elements["socket_count"]; ok {
+		labels["socket_count"] = v
+	}
```
marquiz, line 204: I think we could start with having this only as a raw feature, no label generated by default (https://github.com/kubernetes-sigs/node-feature-discovery/issues/1383#issuecomment-1737335857).

```suggestion
```

marquiz (approved):
Thanks @AhmedGrati 👍

We can add the label later if there is high demand for that

ArangoGutierrez (approved):
/lgtm

PR #1500, opened 2023-12-12
Makefile: fix e2e-testing of the full image

marquiz:
Fix syntax error in the shell code. Also enable verbosity to print out the shell script being run.


ArangoGutierrez (approved):
/lgtm

PR #1501, opened 2023-12-12
nfd-master: remove default denied ns from config

marquiz:
These are now handled by the validate package. If we have them here in nfd-master, the default namespace (feature.node.kubernetes.io) gets denied.


ArangoGutierrez (approved):
/lgtm

PR #1502, opened 2023-12-12
Use non-exp maps package

marquiz:
The maps package became available as a standard non-experimental package in Go 1.21.


ArangoGutierrez (approved):
/lgtm

PR #1503, opened 2023-12-13
Use generics for maps and slices

marquiz:


pkg/apis/nfd/v1alpha1/expression.go
```diff
@@ -24,6 +24,7 @@ import (
 	"strconv"
 	"strings"
 
+	"golang.org/x/exp/maps"
```
ArangoGutierrez, line 27: I am confused, didn't we just got rid of this pkg? 
https://github.com/kubernetes-sigs/node-feature-discovery/pull/1502

marquiz: Indeed, we did 😇 But the non-exp `maps` package doesn't provide the `Keys()` and `Values()` funcs which we use here (and elsewhere in this PR

ArangoGutierrez (approved):
/lgtm

PR #1516, opened 2023-12-21
docs: fix malformed hyperlink syntax

marquiz:


ArangoGutierrez (approved):
why mdlint didn't caught this? interesting
/lgtm

PR #1518, opened 2023-12-21
docs: second fix to the prometheus kustomize overlay name

marquiz:


docs/deployment/metrics.md
```diff
@@ -35,7 +35,7 @@ The exposed metrics are
 ## Kustomize
 
 To deploy NFD with metrics enabled using kustomize, you can use the
-[Metrics Overlay](kustomize.md#metrics).
+[prometheus overlay](kustomize.md#metrics).
```
ArangoGutierrez, line 38: lowercase?

ArangoGutierrez (changes requested):

docs/deployment/metrics.md
marquiz: The name of the overlay is in lower case. This is not a heading so I think the term "prometheus overlay" should be in lowercase

ArangoGutierrez (approved):
/lgtm

PR #1532, opened 2023-12-29
helm: add post-delete hook that cleans up the node

marquiz:
This patch adds a post-delete hook to the Helm chart that runs "nfd-master --prune" in the cluster. This cleans up the node of labels, annotations, taints and extended resources that were created by NFD.


ArangoGutierrez (approved):
/lgtm

Issue #1380, opened 2023-09-27
Configurable restrictions on CRs


marquiz:
**What would you like to be added**:

I would like to be able to configure restrictions on NodeFeatures and NodeFeatureRule objects. E.g.
1. filtering of which NodeFeature objects to watch (e.g. per K8s namespace)
2. prevent labels from 3rd party NodeFeatures (just "raw features" allowed)
3. prevent overwriting of labels
4. prevent overwriting of "raw features"
5. set the max number of labels/ERs/taints that can be created, per CR and in total

**Why is this needed**:

Improved control on what is allowed on the cluster

ArangoGutierrez:
V0.15 or v0.16?

marquiz:
> V0.15 or v0.16?

Depends on people's time, I guess 😇 We could have something small already in v0.15

AhmedGrati:
Two questions here:
- regarding setting the max number of taints, labels, and ERs: If the requested number exceeds the max, should we skip the nodes' update or update the nodes with a subset of labels? e.g. if the max is 10 labels and the request has 12 labels, we only take the first 10 labels.
- In the filtering of which NodeFeature to watch, I'm currently thinking of passing a regex of `<namespace>/<name>`, so users can have the most flexibility (with a default of `*/*`. Any objections here?

wdyt @marquiz @ArangoGutierrez ?

marquiz:
Almost there
/remove-lifecycle rotten

PR #1540, opened 2024-01-15
[WIP] feat: configure CR restrictions

AhmedGrati:
POC of #1380
Things TBD:
- [ ] unit and e2e tests
- [ ] Docs

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -171,11 +171,12 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 	args.Klog = klogutils.InitKlogFlags(flagset)
 
 	overrides := &master.ConfigOverrideArgs{
-		LabelWhiteList: &utils.RegexpVal{},
-		DenyLabelNs:    &utils.StringSetVal{},
-		ExtraLabelNs:   &utils.StringSetVal{},
-		ResourceLabels: &utils.StringSetVal{},
-		ResyncPeriod:   &utils.DurationVal{Duration: time.Duration(1) * time.Hour},
+		LabelWhiteList:    &utils.RegexpVal{},
+		DenyLabelNs:       &utils.StringSetVal{},
+		ExtraLabelNs:      &utils.StringSetVal{},
+		ResourceLabels:    &utils.StringSetVal{},
+		AllowedNamespaces: &utils.StringSliceVal{},
+		ResyncPeriod:      &utils.DurationVal{Duration: time.Duration(1) * time.Hour},
 	}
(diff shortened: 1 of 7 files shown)


marquiz:
Thanks @AhmedGrati for working on this. Will look at it later
/assign

cmd/nfd-master/main.go
```diff
@@ -72,6 +72,18 @@ func main() {
...
+		case "max-taints-per-cr":
+			args.Overrides.MaxTaintsPerCR = overrides.MaxTaintsPerCR
+		case "max-extended-resources-per-cr":
+			args.Overrides.MaxExtendedResourcesPerCR = overrides.MaxExtendedResourcesPerCR
+		case "allowed-namespaces":
+			args.Overrides.AllowedNamespaces = overrides.AllowedNamespaces
+		case "deny-node-feature-labels":
+			args.Overrides.DenyNodeFeatureLabels = overrides.DenyNodeFeatureLabels
+		case "overwrite-labels":
+			args.Overrides.OverwriteLabels = overrides.OverwriteLabels
```
marquiz, line 86: I think we could have these as config file options only, if you want to keep things simple. E.g. the leader election parameters are config-file-only so this is asymmetric, already. WDYT?

pkg/nfd-master/nfd-api-controller.go
```diff
@@ -47,13 +48,15 @@ type nfdController struct {
 type nfdApiControllerOptions struct {
 	DisableNodeFeature bool
 	ResyncPeriod       time.Duration
+	AllowedNamespaces  []string
```
marquiz, line 51: Hmm, looking at this now I think we could do better (more k8s-like). Could we use [metav1.LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.29/#labelselector-v1-meta) directly. It would make it very flexible to pick up the namespaces to watch. The name of the config option would be something like `nodeFeatureNamespaceSelector`.  WDYT?

marquiz (commented):
Thanks you @AhmedGrati for working on this feature. Some comments below

pkg/nfd-master/nfd-api-controller.go
AhmedThresh: If I got this correctly, users would need to pass an array of maps to specify selectors, right?

marquiz: Or even use the metav1.LabelSelector type directly?

AhmedThresh:
@marquiz I think we can close this.

TessaIO:
ping @marquiz, can we close this?

marquiz:
Superseded by #1592 
/close

PR #1545, opened 2024-01-16
test/e2e: Get node object in cleanupNode() only once

ozhuraki:


marquiz (commented):
Thanks @ozhuraki for the patch. However, the current implementation is deliberately written as is. It's basically to handle/reconcile races (somebody changes the node between our Get and Update). We need to (re-)get up-to-date version of the node object. We want the latest version of all the fields we're about to modify, plus the api-server requires the `metadata.resourceVersion` to be right to prevent races.

So unfortunately this is a NACK

ozhuraki:
@marquiz

Thanks for the explanation.

Running e2e I didn't see such issues, per your judgement.

In case you think such might be needed, I can reopen, closing it.

marquiz:
> Running e2e I didn't see such issues, per your judgement.

Yeah, normally this shouldn't happen as you typically run the tests in isolation, with no other agent modifying the nodes at the same time. So this safe-measure is not about e2e-tests racing with itself but other party.

/close

PR #1553, opened 2024-01-18
hack/generate: patch auto-generated deepcopy functions

marquiz:
Temporary (hopefully) fix for deepcopy functions that code-generator doesn't get quite right.


ArangoGutierrez (approved):
crossing fingers this won't last long
/lgtm

PR #1554, opened 2024-01-22
topology-updater: initialize properly with -no-publish

marquiz:
We need to parse kubeconfig (and initialize the apihelper) even with -no-publish as the PodResourcesScanner accesses the k8s API even if we're not publishing/updating NRTs.


PiotrProkop (approved):
/lgtm

Issue #1556, opened 2024-01-22
`nfd-worker` errors reading `speed` parameter for paravirtualized devices


mantramantra12:
**What happened**:
The `nfd-worker` pods on my cluster get this error every time they perform feature discovery:

```
"failed to read net iface attribute" err="read /host-sys/class/net/ifb0/speed: invalid argument" attributeName="speed"
```

I believe this is because the devices are paravirtual and so the `speed` parameter is nonsensical in this context. The file still exists however I think because the kernel puts it there.

**What you expected to happen**:
Feature discovery completes without error

**How to reproduce it (as minimally and precisely as possible)**:
Run feature discovery on a node with paravirtual devices connected

**Anything else we need to know?**:
N/A

**Environment**:
- Kubernetes version (use `kubectl version`): 1.27.7
- Cloud provider or hardware configuration: Proxmox ubuntu VM
- OS (e.g: `cat /etc/os-release`): Ubuntu 22.04.3 LTS
- Kernel (e.g. `uname -a`): Linux <hostname> 5.15.0-91-generic 101-Ubuntu SMP Tue Nov 14 13:30:08 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

marquiz:
Fixed by #1557 
/close

weikinhuang:
I'm still seeing this for the localhost interface

```
E0321 23:00:33.543500       1 network.go:152] "failed to read net iface attribute" err="read /host-sys/class/net/lo/speed: invalid argument" attributeName="speed"
```

marquiz:
@weikinhuang which version of nfd are you running? The fix hasn't been released, yet. It'll be in the upcoming v0.16 release

weikinhuang:
@marquiz oh, I thought the fix was in 0.15.2. I am on 0.15.3. If that's the case thanks for letting me know!

PR #1557, opened 2024-01-22
Prevent `nfd-worker` erroring when reading attributes from paravirtual devices

mantramantra12:
PR for issue described here: https://github.com/kubernetes-sigs/node-feature-discovery/issues/1556

### DIFF
--- a/source/network/network.go
+++ b/source/network/network.go
@@ -148,7 +148,7 @@ func readIfaceInfo(path string, attrFiles []string) nfdv1alpha1.InstanceFeature 
 	for _, attrFile := range attrFiles {
 		data, err := os.ReadFile(filepath.Join(path, attrFile))
 		if err != nil {
-			if !os.IsNotExist(err) {
+			if !os.IsNotExist(err) && !strings.Contains(err.Error(), "invalid argument") {
 				klog.ErrorS(err, "failed to read net iface attribute", "attributeName", attrFile)
 			}
 			continue


mantramantra12:
If the response from the os when trying to read a device file contains 'invalid argument' then the device likely doesn't support that attribute and the file has been included erroneously by the kernel. This extra line catches this case and prevents nfd-worker from reporting an error when encountering paravirtual devices that don't support the 'speed' attribute, for example

source/network/network.go
```diff
@@ -148,7 +148,7 @@ func readIfaceInfo(path string, attrFiles []string) nfdv1alpha1.InstanceFeature
 	for _, attrFile := range attrFiles {
 		data, err := os.ReadFile(filepath.Join(path, attrFile))
 		if err != nil {
-			if !os.IsNotExist(err) {
+			if !os.IsNotExist(err) && !strings.Contains(err.Error(), "invalid argument") {
```
marquiz, line 151: Thanks @mantramantra12 for the contribution. I'd suggest to use `errors.Is()` to check the error code directly instead of relying on string comparison

```suggestion
			if !os.IsNotExist(err) && !errors.Is(err, syscall.EINVAL) {
```

/ok-to-test

mantramantra12: I agree this is much better - didn't know about `syscall.EINVAL`

marquiz (approved):
Thanks @mantramantra12 

/lgtm

Issue #1559, opened 2024-01-24
nfd-topology-updater missing pods-fingerprint in nrt object


leemingeer:
**What happened**:
i have deployed nfd-topology-updater with `kubectl apply -k deployment/overlays/topologyupdater/` to provide noderesourcetopology for [Topology aware scheduler plugin](https://github.com/kubernetes-sigs/scheduler-plugins/tree/master/kep/119-node-resource-topology-aware-scheduling/README.md)

the plugin report "nrtcache: missing NodeTopology podset fingerprint data" logID="sync1706085940168",  that is because the NRT missing nodeTopologyPodsFingerprint which is added by nfd-topology-updater while it missing startup parameters `pods-fingerprint` which is false by default, we should set to true explicitly.
 
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

PR #1560, opened 2024-01-24
nfd-topology-updater add pods fingerprint by default

leemingeer:
Fixes:#1559


marquiz (commented):
Thanks @leemingeer for the patch.

If we want to change the default, we should change the default setting of the cmdline flag (`cmd/nfd-topology-updater/main.go`), adapt the Helm chart, if needed, and also change the documentation (cmdline reference, Helm parameters).

/cc @PiotrProkop @ffromani

ffromani:
for context, we set the `-pods-fingerprint` defaulting to `false` to maximize the backward compatibility. Besides a minimal extra computational cost, there's no any issue ever reporte concrerning this functionality, so switching the default to true is fine to me, it's a good time to turn on by default.

marquiz:
Thanks @ffromani, sounds plausible to me. I'm good with changing the default

leemingeer:
> Thanks @leemingeer for the patch.
> 
> If we want to change the default, we should change the default setting of the cmdline flag (`cmd/nfd-topology-updater/main.go`), adapt the Helm chart, if needed, and also change the documentation (cmdline reference, Helm parameters).
> 
> /cc @PiotrProkop @ffromani

thanks for your suggestions, i'v committed new patch. podSetFingerprint is already set to true in the values file and i believe there is no need to do chart adaption, right?

PiotrProkop:
@leemingeer you need to modify helm chart so one can disable this feature, cause right now It will be always enabled 😄 
https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/deployment/helm/node-feature-discovery/templates/topologyupdater.yaml#L69

marquiz (commented):
@leemingeer thanks for the update, looks good.

Took a quick look at the Helm chart and we need to adapt that, too, e.g. [topology-updater template](https://github.com/kubernetes-sigs/node-feature-discovery/blob/f3e07aeb91f87b880f1b041ae5b8d552b5be9bd8/deployment/helm/node-feature-discovery/templates/topologyupdater.yaml#L69), the `values.yaml` and the also the parameter un docs/deployment/helm

go.mod
```diff
@@ -218,6 +218,7 @@ replace (
 	k8s.io/controller-manager => k8s.io/controller-manager v0.29.0
 	k8s.io/cri-api => k8s.io/cri-api v0.29.0
 	k8s.io/csi-translation-lib => k8s.io/csi-translation-lib v0.29.0
+	k8s.io/endpointslice => k8s.io/endpointslice v0.29.0
```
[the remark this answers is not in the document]
leemingeer: done

leemingeer:
[@PiotrProkop](https://github.com/PiotrProkop) [@marquiz](https://github.com/marquiz) yes, you are right, thanks.
the updated patch has committed, please review.

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -483,7 +483,7 @@ topologyUpdater:
   annotations: {}
   daemonsetAnnotations: {}
   affinity: {}
-  podSetFingerprint: true
+  disablePodSetFingerprint:
```
marquiz, line 486: I didn't realize pod fingerprint has already been enabled by default in the Helm deployment, ever since the feature was introduced in NFD v0.13.0 🤦‍♂️. The documentation was incorrect. Just keep this as is

```suggestion
  podSetFingerprint: true
```

And generally, we should not break existing users, unless there's a very good reason. Plus, we've made a promise not to break without deprecating first ([deprecation policy](https://kubernetes-sigs.github.io/node-feature-discovery/v0.15/reference/versions.html#configuration-options)).

leemingeer: this means we don't need to provide option which disable PodSetFingerprint  in the chart, because it is already enabled in previous chart versions?

leemingeer: done

leemingeer: done

deployment/helm/node-feature-discovery/templates/topologyupdater.yaml
```diff
@@ -66,8 +66,8 @@ spec:
           - "-key-file=/etc/kubernetes/node-feature-discovery/certs/tls.key"
           - "-cert-file=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
           {{- end }}
-          {{- if .Values.topologyUpdater.podSetFingerprint }}
-          - "-pods-fingerprint"
+          {{- if .Values.topologyUpdater.podSetFingerprint | empty }}
```
marquiz, line 69: nit: I'm not sure if there's a functional difference but but using `not` would feel more correct for boolean args

```suggestion
          {{- if not .Values.topologyUpdater.podSetFingerprint }}
```

leemingeer: yes, there's no functional difference, not condition is more clear
done

marquiz (approved):
Thanks @leemingeer! Looks good to me

/assign @PiotrProkop

PR #1561, opened 2024-01-24
Drop pkg/apihelper

### LINKED ISSUE #1547: Drop pkg/apihelper

Get rid of the `pkg/apihelper` package (and the related mocks) and replace it with direct usage of client(set) libraries.

This also means that the nfd-master unit tests need a major overhaul. Which is actually good as in their current form they are almost unmaintainable. E.g. error messages from the mock clients are practically indecipherable. Also, the current pattern of verifying that some specific set of k8s api calls, in particular order, with exact parameters, are made is very impractical if not pointless, and makes refactoring/modifying certain parts of the code very painful (and these are practically covered by the e2e tests).

**Why is this needed**:

Simplify code. Also uniformity accross the code base: many of the apiserver access (e.g. NFD API, NRT CRs) is already done directly, without resorting to apihelper.

marquiz:
This PR removes the apihelper package which was an unnecessary internal interface (and complication) for accessing the Kubernetes API. The change is split into multiple preliminary steps (PRs) to make the review easier.

1. [x] #1562
This change is part of an effort to remove the pkg/apihelper package.
GetKubeconfig is useful helper functionality shared accross the codebase
so move it into a "safe" location.
2. [x] #1566
Stop using apihelper for getting the noderesourcetopology-api client.
3. [x] #1567
Stop using pkg/apihelper for accessing the Kubernetes API. Modify unit
tests to use the fake kubernetes client instead of mocked apihelper
interface.
4. [x] #1568
5. [x] #1570
    Implement some of frequently used helper functions inpackage.

    This patch also contains big changes to the nfd-master unit tests. Much
    of this is about migrating from the mocked apihelper interface to fake
    kubernetes client that provides a bit more apiserver'ish functionality.
    At the same time there is quite a bit of renaming in the tests,
    shortening and unifying naming and getting rid of the extensive usage of
    "mock" everywhere.

Fixes #1547


marquiz:
Depends on #1562
/hold

marquiz:
Still on hold, next piece split into #1566

marquiz:
Only #1567 pending <-- ping @PiotrProkop

marquiz:
All dependent PRs merged, there's now only a single commit
/unhold

PR #1565, opened 2024-01-25
deployment/helm: don't deploy topology-updater conf unnecessarily

marquiz:
Only deploy the topology-updater config if topology-updater itself (the daemon) is deployed.


PiotrProkop (approved):
/lgtm

PR #1566, opened 2024-01-25
topology-updater: get topology api client directly

marquiz:
Stop using apihelper for getting the noderesourcetopology-api client.

Next separate, self-contained step of dropping the apihelper package (i.e. this PR is part of #1561).


PiotrProkop (approved):
/lgtm

PR #1568, opened 2024-01-25
pkg/utils: move JsonPatch from pkg/apihelper

marquiz:
Part of the effort of dropping apihelper package (#1561)


AhmedGrati (approved):
LGTM!

PR #1570, opened 2024-01-26
nfd-master: ditch apihelper

marquiz:
Implement some of frequently used helper functions inpackage.

This patch also contains big changes to the nfd-master unit tests. Much of this is about migrating from the mocked apihelper interface to fake kubernetes client that provides a bit more apiserver'ish functionality. At the same time there is quite a bit of renaming in the tests, shortening and unifying naming and getting rid of the extensive usage of "mock" everywhere.

This is the last self-contained piece to split out from #1561

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -19,6 +19,7 @@ package nfdmaster
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"maps"
 	"net"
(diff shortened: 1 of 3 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -1440,3 +1404,26 @@ func (m *nfdMaster) filterFeatureAnnotations(annotations map[string]string) map[
...
+
+func (m *nfdMaster) patchNode(nodeName string, patches []utils.JsonPatch, subresources ...string) error {
+	if len(patches) > 0 {
+		data, err := json.Marshal(patches)
+		if err == nil {
+			_, err = m.k8sClient.CoreV1().Nodes().Patch(context.TODO(), nodeName, types.JSONPatchType, data, metav1.PatchOptions{})
+		}
+		return err
+	}
+	return nil
```
AhmedGrati, line 1424: Maybe we can do this, wdyt?
```suggestion
	if len(patches) == 0 {
	   return nil
	}
        data, err := json.Marshal(patches)
        if err == nil {
	        _, err = m.k8sClient.CoreV1().Nodes().Patch(context.TODO(), nodeName, types.JSONPatchType, data, metav1.PatchOptions{})
        }
        return err
```

AhmedGrati (commented):
Good job @marquiz! Only one comment left.

pkg/nfd-master/nfd-master.go
marquiz: Definitely, looks better this way. I was just blindly copying the function... Now changed as suggested

PR #1571, opened 2024-01-26
nfd-master: fix node status patching

marquiz:
Correctly patch the "status" subresource. This got broken when refactoring the code in 7a050e7cf9c2e8f21c2f3ddda0a9745a45de472c and wasn't even catched by the unit tests as the fake kubernetes client doesn't handle subresources as the real apiserver does.


marquiz:
Fortunately we have extensive e2e-tests 🙄
/assign @ArangoGutierrez @AhmedGrati

PR #1574, opened 2024-01-30
source/system: Add reading vendor information

### LINKED ISSUE #1057: Add support for hardware vendor detection
I think it would be good to allow NFD to discover the underlying hardware that the system is running on.  This can be useful in bare metal environments where you may have a system that's running HP hardware, or Dell hardware and you would need to run specific things on specific systems

**Why is this needed**:
In my use case, let's say I want to run `dellhw_exporter` only on systems which use Dell hardware, I don't have a way with NFD to detect this out of the box.

ozhuraki:
Add reading vendor information from /sys/devices/virtual/dmi/id/sys_vendor

Closes #1057

### DIFF
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -979,6 +979,8 @@ The following features are available for matching:
 |                  |              | **`<sysfs-attribute>`** | string | Sysfs network interface attribute, available attributes: `dax`, `rotational`, `nr_zones`, `zoned` |
 | **`system.osrelease`** | attribute |       |            | System identification data from `/etc/os-release` |
 |                  |              | **`<parameter>`** | string | One parameter from `/etc/os-release` |
+| **`system.dmiid`** | attribute |       |            | DMI identification data from `/sys/devices/virtual/dmi/id/` |
+|                  |              | **`sys_vendor`** | string | Vendor name from `/sys/devices/virtual/dmi/id/sys_vendor` |
 | **`system.name`** | attribute   |          |            | System name information |
 |                  |              | **`nodename`** | string | Name of the kubernetes node object |
 | **`usb.device`** | instance     |          |            | USB devices present in the system |
--- a/source/system/system.go
+++ b/source/system/system.go
@@ -43,6 +43,7 @@ const Name = "system"
 const (
 	OsReleaseFeature = "osrelease"
 	NameFeature      = "name"
+	DmiIdFeature     = "dmiid"
 )
 
 // systemSource implements the FeatureSource and LabelSource interfaces.
@@ -101,6 +102,22 @@ func (s *systemSource) Discover() error {
 		}
 	}
 
+	// Get DMI ID attributes
+	dmiIDAttributeNames := []string{"sys_vendor"}
+	dmiAttrs := make(map[string]string)
+	for _, name := range dmiIDAttributeNames {
+		val, err := getDmiIDAttribute(name)
+		if err != nil {
+			klog.ErrorS(err, "failed to get DMI entry", "attributeName", name)
+		} else {
+			dmiAttrs[name] = val
+		}
+	}
+
+	if len(dmiAttrs) > 0 {
+		s.features.Attributes[DmiIdFeature] = nfdv1alpha1.NewAttributeFeatures(dmiAttrs)
+	}
+
 	klog.V(3).InfoS("discovered features", "featureSource", s.Name(), "features", utils.DelayedDumper(s.features))
 
 	return nil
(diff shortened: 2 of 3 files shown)


source/system/system.go
```diff
@@ -153,6 +162,16 @@ func splitVersion(version string) map[string]string {
 	return components
 }
 
+// Read /sys/devices/virtual/dmi/id/sys_vendor
+func getVendor() (string, error) {
+	vendor, err := os.ReadFile("/sys/devices/virtual/dmi/id/sys_vendor")
```
AhmedGrati, line 167: Here we can replace `/sys` with the variable `hostpath.SysfsDir`

AhmedGrati (commented):
Thanks @ozhuraki! One comment left below.

source/system/system.go
ozhuraki: @AhmedGrati
Updated, thanks!

docs/usage/customization-guide.md
```diff
@@ -976,6 +976,8 @@ The following features are available for matching:
 |                  |              | **`<sysfs-attribute>`** | string | Sysfs network interface attribute, available attributes: `dax`, `rotational`, `nr_zones`, `zoned` |
 | **`system.osrelease`** | attribute |       |            | System identification data from `/etc/os-release` |
 |                  |              | **`<parameter>`** | string | One parameter from `/etc/os-release` |
+| **`system.vendor`** | attribute |       |            | System vendor identification data from `/sys/devices/virtual/dmi/id/` |
+|                  |              | **`name`** | string | Vendor name from `/sys/devices/virtual/dmi/id/sys_vendor` |
```
marquiz, line 980: We could just expose the "attributes" as is, not apply any name mapping. Like, say, `storage.block` above. This would then become just `sys_vendor`. This would also prevent any possibility of name clashes in the future. WDYT?

source/system/system.go
```diff
@@ -153,6 +163,16 @@ func splitVersion(version string) map[string]string {
 	return components
 }
 
+// Read /sys/devices/virtual/dmi/id/sys_vendor
+func getVendor() (string, error) {
```
marquiz, line 167: Just a suggestion to make this generic for exposing other attributes

```suggestion
func getDmiIdAttribute(name string) (string, error) {
```

marquiz (commented):
Thanks @ozhuraki! A few quick comments below.

Need to think if there are other meaningful attributes other than `sys_vendor` to expose. But that can also be a follow-up PR.

docs/usage/customization-guide.md
ozhuraki: >This would then become just sys_vendor. This would also prevent any possibility of name clashes in the future. WDYT?

This makes sense, thanks! Updated so.

source/system/system.go
ozhuraki: Thanks, makes sense too, updated.

source/system/system.go
```diff
@@ -101,6 +102,15 @@ func (s *systemSource) Discover() error {
...
 	}
 
+	// Get vendor information
+	sys_vendor, err := getDmiIdAttribute("sys_vendor")
+	if err != nil {
+		klog.ErrorS(err, "failed to get sys_vendor")
+	} else {
+		s.features.Attributes[DmiIdFeature] = nfdv1alpha1.NewAttributeFeatures(nil)
+		s.features.Attributes[DmiIdFeature].Elements["sys_vendor"] = sys_vendor
+	}
```
marquiz, line 112: Now that we're on this, we could make it a bit more generic so that adding new dmi-id attributes in the future is super simple (with little risk of breaking anything existing).

For example something like this (or specifying one additional helper func to do this):

```suggestion
	dmiIdAttributeNames := []string{"sys_vendor"}
	dmiAttrs := map[string]string
	for name := range dmiIdAttributeNames {
		if val, err := getDmiIdAttribute(name); err {
			klog.ErrorS(err, "failed to get DMI entry", "attributeName", name)
		} else {
			dmiAttrs[name] = val
		}
	}
	s.features.Attributes[DmiIdFeature] = nfdv1alpha1.NewAttributeFeatures(dmiAttrs)
```

WDYT?

marquiz (commented):
Thanks @ozhuraki for the update. Looks good to me. I think we could merge this as is but added one suggestion below

source/system/system.go
ozhuraki: @marquiz
Thanks, makes sense, updated.

marquiz (approved):
Nice job @ozhuraki 
/assign @PiotrProkop

source/system/system.go
```diff
@@ -153,6 +170,16 @@ func splitVersion(version string) map[string]string {
 	return components
 }
 
+// Read /sys/devices/virtual/dmi/id attribute
+func getDmiIdAttribute(name string) (string, error) {
```
ArangoGutierrez, line 174: Usually is a good practice to use ID
`getDmiIdAttribute` --> `getDmiIDAttribute`
But is not a blocker for this patch

source/system/system.go
```diff
@@ -101,6 +102,22 @@ func (s *systemSource) Discover() error {
 		}
 	}
 
+	// Get DMI ID attributes
+	dmiIdAttributeNames := []string{"sys_vendor"}
```
ArangoGutierrez, line 106: Same ID vs Id comment

source/system/system.go
ozhuraki: @ArangoGutierrez
Thanks! Updated both

ArangoGutierrez (approved):
/lgtm

PR #1580, opened 2024-02-09
scripts/test-infra: bump golangci-lint to v1.56.1

marquiz:


ArangoGutierrez (approved):
Latest an greatest 
/lgtm

PR #1585, opened 2024-02-13
add swap support in nfd

### LINKED ISSUE #1584: Swap support
I am working on https://github.com/kubernetes/enhancements/issues/2400 which adds swap support. I think that this project could be a good fit for labeling nodes if they have swap enabled.
**Why is this needed**:
Swap is an important characteristic of a node and it is usually turned off. We are introducing ways for heterogenous nodes for swap so we should have a label for a swap enabled node.

I think the implementation could just check `swapon` and verify that swap is enabled on the node.

kannon92:
Fixes #1584

### DIFF
--- a/deployment/components/common/worker-mounts.yaml
+++ b/deployment/components/common/worker-mounts.yaml
@@ -4,6 +4,9 @@
   - name: host-boot
     hostPath:
       path: "/boot"
+  - name: host-proc-swaps
+    hostPath:
+      path: "/proc/swaps"
   - name: host-os-release
     hostPath:
       path: "/etc/os-release"
@@ -37,6 +40,9 @@
     readOnly: true
   - name: host-sys
     mountPath: "/host-sys"
+    readOnly: true
+  - name: host-proc-swaps
+    mountPath: "/host-proc/swaps"
     readOnly: true
   - name: host-usr-lib
     mountPath: "/host-usr/lib"
--- a/deployment/helm/node-feature-discovery/templates/worker.yaml
+++ b/deployment/helm/node-feature-discovery/templates/worker.yaml
@@ -87,6 +87,9 @@ spec:
         - name: host-lib
           mountPath: "/host-lib"
           readOnly: true
+        - name: host-proc-swaps
+          mountPath: "/host-proc/swaps"
+          readOnly: true
         {{- if .Values.worker.mountUsrSrc }}
         - name: host-usr-src
           mountPath: "/host-usr/src"
@@ -122,6 +125,9 @@ spec:
         - name: host-lib
           hostPath:
             path: "/lib"
+        - name: host-proc/swaps
+          hostPath:
+            path: "/proc/swaps"
         {{- if .Values.worker.mountUsrSrc }}
         - name: host-usr-src
           hostPath:
--- a/source/memory/memory.go
+++ b/source/memory/memory.go
@@ -40,6 +40,9 @@ const NvFeature = "nv"
 // NumaFeature is the name of the feature set that holds all NUMA related features.
 const NumaFeature = "numa"
 
+// SwapFeature is the name of the feature set that holds all Swap related features
+const SwapFeature = "swap"
+
 // memorySource implements the FeatureSource and LabelSource interfaces.
 type memorySource struct {
 	features *nfdv1alpha1.Features
@@ -68,6 +71,11 @@ func (s *memorySource) GetLabels() (source.FeatureLabels, error) {
 		labels["numa"] = true
 	}
 
+	// Swap
+	if isSwap, ok := features.Attributes[SwapFeature].Elements["enabled"]; ok && isSwap == "true" {
+		labels["swap"] = true
+	}
+
 	// NVDIMM
 	if len(features.Instances[NvFeature].Elements) > 0 {
 		labels["nv.present"] = true
@@ -93,6 +101,13 @@ func (s *memorySource) Discover() error {
 		s.features.Attributes[NumaFeature] = nfdv1alpha1.AttributeFeatureSet{Elements: numa}
 	}
 
+	// Detect Swap
+	if swap, err := detectSwap(); err != nil {
+		klog.ErrorS(err, "failed to detect Swap nodes")
+	} else {
+		s.features.Attributes[SwapFeature] = nfdv1alpha1.AttributeFeatureSet{Elements: swap}
+	}
+
 	// Detect NVDIMM
 	if nv, err := detectNv(); err != nil {
 		klog.ErrorS(err, "failed to detect nvdimm devices")
@@ -111,6 +126,20 @@ func (s *memorySource) GetFeatures() *nfdv1alpha1.Features {
 		s.features = nfdv1alpha1.NewFeatures()
 	}
 	return s.features
+}
+
+// detectSwap detects Swap node information
+func detectSwap() (map[string]string, error) {
+	procBasePath := hostpath.ProcDir.Path("swaps")
+	lines, err := getNumberOfLinesFromFile(procBasePath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read swaps file: %w", err)
+	}
+	// /proc/swaps has a header row
+	// If there is more than a header then we assume we have swap.
+	return map[string]string{
+		"enabled": strconv.FormatBool(lines > 1),
+	}, nil
 }
 
 // detectNuma detects NUMA node information
@@ -164,6 +193,14 @@ func readNdDeviceInfo(path string) nfdv1alpha1.InstanceFeature {
 		attrs[attrName] = strings.TrimSpace(string(data))
 	}
 	return *nfdv1alpha1.NewInstanceFeature(attrs)
+}
+
+func getNumberOfLinesFromFile(path string) (int, error) {
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return 0, err
+	}
+	return len(strings.Split(string(data), "\n")), nil
 }
 
 func init() {
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -966,6 +966,8 @@ The following features are available for matching:
 | **`memory.numa`**  | attribute  |          |            | NUMA nodes |
 |                  |              | **`is_numa`** | bool  | `true` if NUMA architecture, `false` otherwise |
 |                  |              | **`node_count`** | int | Number of NUMA nodes |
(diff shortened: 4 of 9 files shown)


source/memory/memory.go
```diff
@@ -113,6 +129,29 @@ func (s *memorySource) GetFeatures() *nfdv1alpha1.Features {
...
 }
 
+// detectSwap detects Swap node information
+func detectSwap() (map[string]string, error) {
+	procBasePath := hostpath.ProcDir.Path("swaps")
+	file, err := os.Open(procBasePath)
+	defer file.Close()
+	if err != nil {
+		return nil, err
+	}
```
AhmedThresh, line 139: I think CI is yelling because of this:
```suggestion
	if err != nil {
		return nil, err
	}
	defer file.Close()
```

kannon92: addressed.

deployment/helm/node-feature-discovery/templates/worker.yaml
```diff
@@ -122,6 +122,9 @@ spec:
         - name: host-lib
           hostPath:
             path: "/lib"
+        - name: host-proc-swap
```
marquiz, line 125: Suggest a small change to naming:

```suggestion
        - name: host-proc-swaps
```

You also need to add this in volumeMounts of the container.

deployment/components/common/worker-mounts.yaml
```diff
@@ -4,6 +4,9 @@
   - name: host-boot
     hostPath:
       path: "/boot"
+  - name: host-swap
```
marquiz, line 7: This is different name than what is used in the volumeMounts below. I also suggest to use the same naming than in the Helm chart 

```suggestion
  - name: host-proc-swaps
```

source/memory/memory.go
```diff
@@ -113,6 +129,34 @@ func (s *memorySource) GetFeatures() *nfdv1alpha1.Features {
 	return s.features
 }
 
+// detectSwap detects Swap node information
+func detectSwap() (map[string]string, error) {
+	procBasePath := hostpath.ProcDir.Path("swaps")
+	file, err := os.Open(procBasePath)
+	defer func() {
```
marquiz, line 136: Only defer close if the open succeeded -> move this under the error checking below?

source/memory/memory.go
```diff
@@ -113,6 +129,34 @@ func (s *memorySource) GetFeatures() *nfdv1alpha1.Features {
...
+		if err != nil {
+			klog.ErrorS(err, "unable to close swaps file")
+		}
+	}()
+	if err != nil {
+		return nil, err
+	}
+
+	// Read the CSV data
+	reader := csv.NewReader(file)
```
marquiz, line 147: Hmm, what if we just simple&stupid `os.ReadFile(procBasePath)` and count the lines? We do not look at the data so this would simplify the code.

source/memory/memory.go
```diff
@@ -113,6 +129,34 @@ func (s *memorySource) GetFeatures() *nfdv1alpha1.Features {
...
+	reader.FieldsPerRecord = -1 // Allow variable number of fields
+	data, err := reader.ReadAll()
+	if err != nil {
+		return nil, err
+	}
+	// /proc/swaps has a header row
+	// If there is more than a header then we assume we have swap.
+	return map[string]string{
+		"is_swap": strconv.FormatBool(len(data[0]) > 1),
```
marquiz, line 156: How about "enabled"? Then the feature to look for would be `memory.swap.enable` which sounds more logical to me than `memory.swap.is_swap`. WDYT?

marquiz (commented):
Thanks @kannon92 for the patch. Some quick notes below

deployment/helm/node-feature-discovery/templates/worker.yaml
kannon92: updated. both name and added it to volumeMounts.

source/memory/memory.go
kannon92: not sure this is needed with the simplication.

source/memory/memory.go
kannon92: yea that is a good idea. updated.

source/memory/memory.go
kannon92: good idea. updated.

AhmedThresh:
@kannon92 I think that you should mention the swap feature in the `docs/usage/features.md`.

AhmedThresh (approved):
@kannon92 Thanks for your efforts!

kannon92:
@marquiz @AhmedThresh I am new to this project so I don't know if I should have e2e tests or unit tests covering these changes.

AhmedThresh:
> @marquiz @AhmedThresh I am new to this project so I don't know if I should have e2e tests or unit tests covering these changes.

These kinds of features generally are not included in the e2e tests AFAIK. However, I see that we can implement a unit test for the `getNumberOfLinesFromFile` function. wdyt?

marquiz:
> I think that you should mention the swap feature in the `docs/usage/features.md`.

Good point. In addition, document it in the table of "available features" in `docs/usage/customization-guide.md`, too.

Also, a nit about the commit message title (ref https://www.kubernetes.dev/docs/guide/pull-requests/#commit-message-guidelines): start with a capital letter, e.g. `Add swap support`, or alternatively provide context like `source/network: detect swap` or smth

marquiz:
> I am new to this project so I don't know if I should have e2e tests or unit tests covering these changes.

@kannon92 very good question. We have been lacking on unit testing of the feature detection 🙃 It'd be good to add a simple unit test as @AhmedThresh suggested.

I've long had in my backlog to add more comprehensive unit tests for each source/feature that rely on scanning host fs. Have a fake rootfs and run feature discovery against that. But you can skip this (unless you want to work on it) as there is not infrastructure or prior art, yet (I have a rough poc for the pci source, maybe I clean that up and submit as a pr...)

kannon92:
> > I am new to this project so I don't know if I should have e2e tests or unit tests covering these changes.
> 
> @kannon92 very good question. We have been lacking on unit testing of the feature detection 🙃 It'd be good to add a simple unit test as @AhmedThresh suggested.
> 
> I've long had in my backlog to add more comprehensive unit tests for each source/feature that rely on scanning host fs. Have a fake rootfs and run feature discovery against that. But you can skip this (unless you want to work on it) as there is not infrastructure or prior art, yet (I have a rough poc for the pci source, maybe I clean that up and submit as a pr...)

So in some gigs, I have been told that any unit test involving file io is not really a true unit..

Some options for me could be to add some mock data for /proc/swaps and confirm the two cases (header versus swap partition).

Or we could go with gomocks?

marquiz:
> So in some gigs, I have been told that any unit test involving file io is not really a true unit..

Perhaps so when looking at textbook definitions. Maybe we can then call it integration test or something else. But in our case where the functionality is very much about reading data from the filesystem using file io as the "mock" is pretty handy and gives good coverage, IMO.

Some more complicated parsing deserves "classical" unit testing for sure (I took a stab at that in #1588, something that had been lying in my unfinished devel backlog for a while).

> Or we could go with gomocks?

TBH, I don't see much sense in wiring in some mocked filesystem interface over just having mocked data in fs. I'll leave this for you to consider, do whatever makes sense. 

FYI, I cleaned up my pci testing work and submitted #1589. It uses mocked sysfs on disk. Not classical unit test but I think makes sense there

AhmedThresh (approved):

kannon92:
updated with your comments.

deployment/components/common/worker-mounts.yaml
```diff
@@ -38,6 +41,9 @@
   - name: host-sys
     mountPath: "/host-sys"
     readOnly: true
+  - name: host-proc-swaps
+    mountPath: "/host-swaps"
```
marquiz, line 45: Not right

```suggestion
    mountPath: "/host-proc/swaps"
```

deployment/helm/node-feature-discovery/templates/worker.yaml
```diff
@@ -87,6 +87,9 @@ spec:
         - name: host-lib
           mountPath: "/host-lib"
           readOnly: true
+        - name: host-proc-swaps
+          mountPath: "/host-swaps"
```
marquiz, line 91:
```suggestion
          mountPath: "/host-proc/swaps"
```

deployment/helm/node-feature-discovery/templates/worker.yaml
```diff
@@ -122,6 +125,9 @@ spec:
         - name: host-lib
           hostPath:
             path: "/lib"
+        - name: host-proc-swaps
+          hostPath:
+            path: "/proc/swaps"
```
marquiz, line 130: We could (or should) also start using types here

```suggestion
            path: "/proc/swaps"
            type: File
```
WDYT?

kannon92: I don't really have a preference here. My main thing is that this could be done in one go for all the other types.

marquiz: That's fine. Probably cleaner to do it in one go for all

deployment/helm/node-feature-discovery/templates/worker.yaml
kannon92: updated.

deployment/components/common/worker-mounts.yaml
kannon92: updated.

marquiz: @kannon92 this wasn't updated

kannon92: sorry missed that one.

deployment/helm/node-feature-discovery/templates/worker.yaml
```diff
@@ -87,6 +87,9 @@ spec:
         - name: host-lib
           mountPath: "/host-lib"
           readOnly: true
+        - name: host-proc-swaps
+W2AQ          mountPath: "/host-proc/swaps"
```
AhmedThresh, line 91: small nit here

marquiz (approved):
Thanks @kannon92 for the work. Looks good to me now 👍

deployment/helm/node-feature-discovery/templates/worker.yaml
kannon92: yikes. sorry about that.

kannon92:
ping @ArangoGutierrez 

Swap KEP was merged into 1.30 k8s so it'd be nice to see this progress.

marquiz:
ping @ArangoGutierrez

PR #1587, opened 2024-02-15
helm: add priorityClassName option

ArangoGutierrez:


AhmedThresh:
@ArangoGutierrez Wouldn't be useful if we add the same config for Kustomize also?

ArangoGutierrez:
> @ArangoGutierrez Wouldn't be useful if we add the same config for Kustomize also?

the thing is, in kustomize we don't have a clear way of passing input as with helm values file. I prefer to keep kustomize as is for now

marquiz (commented):
Thanks @ArangoGutierrez. Makes sense

I think we should use the same priority class for nfd-gc and nfd-topology-updater, too(?)

Suggestion regarding the commit message: You could prefix the title with the "helm:" context as with the PR title
/retitle helm: add priorityClassName option

ArangoGutierrez:
@marquiz can we add this to v0.16?

marquiz (approved):
Looks good to me

marquiz:
> can we add this to v0.16?

Yeah, sure it will be there, once lgtm'd 😄

AhmedThresh (approved):

PR #1592, opened 2024-02-16
feat/nfd-master: configure CR restrictions

### LINKED ISSUE #1380: Configurable restrictions on CRs

AhmedThresh:
Resolves #1380 
Features implemented:
- NodeFeatures Namespace selection by using selectors.
- Max number of Taints, Labels, and ERs that can be generated by a single CR.
- Enable/Disable overriding labels.
- Enable/Disable NodeFeatures labels.

Result of e2e tests:
Screenshot from 2024-02-16 15-44-25


AhmedThresh:
ping @marquiz

pkg/nfd-master/nfd-master.go
```diff
@@ -840,13 +869,24 @@ func (m *nfdMaster) refreshNodeFeatures(nodeName string, labels map[string]strin
 	maps.Copy(extendedResources, crExtendedResources)
 	extendedResources = m.filterExtendedResources(features, extendedResources)
 
+	if m.config.Restrictions.MaxExtendedResourcesPerCR > 0 && len(extendedResources) > m.config.Restrictions.MaxExtendedResourcesPerCR {
```
marquiz, line 872: Hmm, I started to think that maybe we could use these settings for disabling extended resources, annotations (or even labels). Or is this a bad idea 🧐 The default would then be -1, meaing "no limit", 0 would in practice mean disable. Thoughts?

marquiz (commented):
Thanks for the update @AhmedThresh

/ok-to-test

pkg/nfd-master/nfd-master.go
AhmedThresh: I think that's better option to start with as an experimental feature and then we can gather insights later on. It would as simple as turning on/off labels, annotations, etc.... But it would give less flexibility/granular control on the behavior of the master compared to the current implementation. wdyt?

TessaIO:
ping @marquiz. Any more thoughts on this?

pkg/nfd-master/nfd-master.go
```diff
@@ -580,6 +598,11 @@ func (m *nfdMaster) filterFeatureLabels(labels Labels, features *nfdv1alpha1.Fea
 		}
 	}
 
+	if m.config.Restrictions.MaxLabelsPerCR > 0 && len(outLabels) > m.config.Restrictions.MaxLabelsPerCR {
```
ArangoGutierrez, line 601: Ummm, what if instead of skipping we trim the `outLabels` to the MAX number? If the user took time to specify a max number, they expect a min number of labels, not 0. debugging 0 vs "missing" labels is a diff scenario.

pkg/nfd-master/nfd-master.go
ArangoGutierrez: I agree with Markus here

ArangoGutierrez (changes requested):

pkg/nfd-master/nfd-master.go
TessaIO: I think it's an area for discussion here, since we already discussed in the comments section https://github.com/kubernetes-sigs/node-feature-discovery/issues/1380#issuecomment-1884484022
wdyt @marquiz

marquiz: IMO it should be "transactional" i.e. all or nothing. If we trim to MAX if user asks for MAX+1 it's sorta random what gets left out. Plus very hard to spot the problem if e.g. only one of the labels is missing.

pkg/nfd-master/nfd-master.go
marquiz: Sorry @AhmedThresh I didn't quite understand your comment/suggestion. One consideration is that we cannot easily (or should not) change the defaults after this gets merged. So we should decide now if we want to do 0=disable

TessaIO: fair enough, and for values different than 0 and -1 should they be ignored? e.g.: we don't apply any sort of filtering right?

AhmedThresh: @marquiz looking more into this, I think we should also change the name of these restrictions to enable/disable labels or whatever. We can do this using feature gates. wdyt?

pkg/nfd-master/nfd-master.go
marquiz: Hmm, looking at this now it's not "PerCR" but just maximum number of labels. So this (and the others) should be renamed.

```suggestion
	if m.config.Restrictions.MaxLabels > 0 && len(outLabels) > m.config.Restrictions.MaxLabels {
```

pkg/nfd-master/nfd-master.go
marquiz: @AhmedThresh what do you mean. Change the setting from `MaxExtendedResources: int` to `DisableExtendedResources: bool`? Or adding a separate setting for enabling/disabling, in addition to the max limit?

marquiz: Oh, and feature gates are not for this kind of control. They are supposed to be used for maturizing new/experimental features, eventually each feature gate getting locked to true and finally removed.

Anyway, continuing my previous comment, I think you can turn `Max*` settings to`Disable*` if you want. We can add the max settings later if needed/we want

pkg/nfd-master/nfd-master.go
```diff
@@ -71,6 +71,16 @@ type ExtendedResources map[string]string
 // Annotations are used for NFD-related node metadata
 type Annotations map[string]string
 
+// Restrictions contains the restrictions on the NF and NFR Crs
+type Restrictions struct {
+	NodeFeatureNamespaceSelector *metav1.LabelSelector
+	MaxLabelsPerCR               int
```
marquiz, line 77: For completeness, let's add a corresponding setting to annotations, too

pkg/nfd-master/nfd-master.go
```diff
@@ -71,6 +71,16 @@ type ExtendedResources map[string]string
...
+	MaxTaintsPerCR               int
+	MaxExtendedResourcesPerCR    int
+	DenyNodeFeatureLabels        bool
+	OverwriteLabels              bool
```
marquiz, line 81: Suggesting a slight rename and change in functionality. If we add this, I think we should treat labels, annotations, taints and ERs the same way. I.e. allow overwriting previous values or then deny.

```suggestion
	AllowOverwrite              bool
```

WDYT?

pkg/nfd-master/nfd-master.go
AhmedThresh: Yes, I meant that we should use Disable/Enable instead of Max. Okay then, there's no need for feature gates.

AhmedThresh:
Thanks for the review @marquiz and @ArangoGutierrez. I'll continue working on this ASAP.

pkg/nfd-master/nfd-master.go
marquiz: OK. Go ahead with disable/enable if/when that seems better. I also think that's a good initial implementation.

marquiz:
ping @TessaIO, do you think you have time to work on this for v0.16? No pressure, we can defer to v0.17, just asking 😄

AhmedThresh:
@marquiz I'll do my best to polish this by the end of the week :)

TessaIO:
@marquiz I'll do my best to polish this by the end of the week :)

ArangoGutierrez:
Ummm @TessaIO I think once this PR is ready for merge, we should close it, and you should re open it under [TessaIO](https://github.com/TessaIO) to add more contributions to TessaIO and not merge this temporal Account AhmedThressh

AhmedThresh:
ping @marquiz @ArangoGutierrez

ArangoGutierrez:
One nit comment (besides the pages tests failing....) is that we have 3 commits for 4 major changes 

- NodeFeatures Namespace selection by using selectors.
- Max number of Taints, Labels, and ERs that can be generated by a single CR.
- Enable/Disable overriding labels.
- Enable/Disable NodeFeatures labels.

it would be good to accommodate it, so each commit is self contained for those 4 changes you describe

TessaIO:
> One nit comment (besides the pages tests failing....) is that we have 3 commits for 4 major changes
> 
> * NodeFeatures Namespace selection by using selectors.
> * Max number of Taints, Labels, and ERs that can be generated by a single CR.
> * Enable/Disable overriding labels.
> * Enable/Disable NodeFeatures labels.
> 
> it would be good to accommodate it, so each commit is self contained for those 4 changes you describe

Thanks @ArangoGutierrez. That actually depends on how you view the feature. from my POV, I see it as implementation -> deployment -> documentation. But I don't mind creating 4 commits each for a separate feature instead.
Also, I couldn't find the reason for pages test failure. Netlify didn't show anything. Any clue about that?

TessaIO:
ping @marquiz @ArangoGutierrez

TessaIO:
@ArangoGutierrez @marquiz can you spot what's wrong here?
I can build on my local machine and golangci-lint works without any error

marquiz:
Hmm, looks like some glitch 🤨
/retest

TessaIO:
@marquiz rebased but still same error

marquiz:
> @marquiz rebased but still same error

Yeah, you need to fix the build failure 😉 The function prototype(s)/args have changed a bit so you need to adjust the impl in the patch

TessaIO:
Screenshot from 2024-05-02 23-54-22

The problem is that the build command works on my workstation

marquiz:
> The problem is that the build command works on my workstation

I don't know what you're testing locally but the version in this PR cannot be built

ArangoGutierrez:
@AhmedThresh / @TessaIO Looks like the current version of this PR doesn't build / pass the basic CI

AhmedThresh:
@ArangoGutierrez @marquiz Thanks. It Should be good now. 
For commits, I will improve them in the new PR that I will open with @TessaIO account

pkg/nfd-master/nfd-master.go
```diff
@@ -883,13 +914,24 @@ func (m *nfdMaster) refreshNodeFeatures(cli k8sclient.Interface, node *corev1.No
...
 
 	// Taints
 	var taints []corev1.Taint
 	if m.config.EnableTaints {
-		taints = filterTaints(crTaints)
+		taints = m.filterTaints(crTaints)
+	}
+
+	// If we deny node feature labels, we'll empty the labels variable
+	if m.config.Restrictions.DenyNodeFeatureLabels {
```
marquiz, line 932: This setting should only disable labels from NodeFeature objects. Not all labels. And more specifically, it should (I think) disable labels from __3rd party__ NodeFeature objects. That is, leave the built-in labels from nfd-worker intact but deny 3rd party extensions from creating node labels directly. Built-in labels can be disabled separately with nfd-worker config if desired. WDYT?

So, this should be done in `nfdAPIUpdateOneNode()` where the NF objects are handled.

pkg/nfd-master/nfd-master.go
```diff
@@ -883,13 +914,24 @@ func (m *nfdMaster) refreshNodeFeatures(cli k8sclient.Interface, node *corev1.No
 	maps.Copy(extendedResources, crExtendedResources)
 	extendedResources = m.filterExtendedResources(features, extendedResources)
 
+	if m.config.Restrictions.DisableExtendedResources {
+		klog.InfoS("maximum number taints per CR restriction is applied, skipping")
+		extendedResources = map[string]string{}
+	}
```
marquiz, line 920: Maybe move this to `filterExtendedResources()`, similarly to labels and annotations?

Also, adjust the logging

```suggestion
	if len(extendedResources) > 0 && m.config.Restrictions.DisableExtendedResources {
		klog.V(2).InfoS("extended resources are disabled in configuration (restrictions.disableExtendedResources=true)")
		extendedResources = map[string]string{}
	}
```

pkg/nfd-master/nfd-master.go
```diff
@@ -1425,6 +1469,10 @@ func (m *nfdMaster) filterFeatureAnnotations(annotations map[string]string) map[
 
 		outAnnotations[annotation] = value
 	}
+	if m.config.Restrictions.DisableAnnotations {
+		outAnnotations = map[string]string{}
+	}
```
marquiz, line 1474: I think this is fine. But let's be consistent, either add logging (similarly to labels and ERs) like:

```suggestion
	if len(outAnnotations) > 0 && m.config.Restrictions.DisableAnnotations {
		klog.V(2).InfoS("node annotations are disabled in configuration (restrictions.disableAnnotations=true)")
		outAnnotations = map[string]string{}
	}
```

Or then remove logging from labels and ERs. WDYT?

pkg/nfd-master/nfd-master.go
```diff
@@ -1161,7 +1203,7 @@ func createPatches(removeKeys []string, oldItems map[string]string, newItems map
 	// Determine items to add or replace
 	for key, newVal := range newItems {
 		if oldVal, ok := oldItems[key]; ok {
-			if newVal != oldVal {
+			if newVal != oldVal && overwrite {
```
marquiz, line 1206: Heh, I was thinking some quite different meaning for the allowOverwrite setting –  but, what you're implementing here is actually pretty neat 😅

However, I don't think this now works quite as expected as values effectively become consts. NFD cannot change labels (or annotations or ERs) that it previously created. However, it definitely should be able to do so (if something in the system changes we want to reflect that in the labels).

So, I think the interpretation/logic should be so that `allowOverwrite=false` denies NFD from overwriting labels/annotations/ERs/taints that it does not own. Basically:

```golang
if oldVal, ok := oldItems[key]; ok {
    if newVal != oldVal && (!removeKeys.Has(key) || overwrite) {
        patches = append(patches, utils.NewJsonPatch("replace", jsonPath, key, newVal))
    }
}
```

Also, `removeKeys` is a REALLY bad name. Maybe change it to `ownedKeys` or something and change it to `sets.String`.

Thoughts?

marquiz (commented):
Wow, thanks for the update (and thx for working on this in the first place) @TessaIO. Some comments below. We're getting there, one step at a time.

TessaIO:
Thanks for the review @marquiz. I will address your comments ASAP.
also, @ArangoGutierrez feel free to take a look at this.

pkg/nfd-master/nfd-api-controller.go
```diff
@@ -55,9 +62,11 @@ func init() {
 
 func newNfdController(config *restclient.Config, nfdApiControllerOptions nfdApiControllerOptions) (*nfdController, error) {
 	c := &nfdController{
-		stopChan:           make(chan struct{}),
-		updateAllNodesChan: make(chan struct{}, 1),
-		updateOneNodeChan:  make(chan string),
+		stopChan:                     make(chan struct{}, 1),
```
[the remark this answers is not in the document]
TessaIO: Rebase issues 😅

pkg/nfd-master/nfd-master.go
```diff
@@ -72,6 +72,17 @@ type ExtendedResources map[string]string
 // Annotations are used for NFD-related node metadata
 type Annotations map[string]string
 
+// Restrictions contains the restrictions on the NF and NFR Crs
+type Restrictions struct {
+	NodeFeatureNamespaceSelector *metav1.LabelSelector
+	DisableLabels                bool
+	DisableTaints                bool
```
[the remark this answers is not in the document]
TessaIO: Yes, I will delete this one to ensure backwards compatibility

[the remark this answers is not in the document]
TessaIO: wdyt?

pkg/nfd-master/nfd-master.go
TessaIO: Yes, that makes sense. At first, I thought that NFD could change labels/annotations/ERs/taints that it previously created. Thanks for the information. I will update that.

TessaIO: I would suggest changing the ownedKeys to a set of strings in another PR, to make this one more atomic. wdyt?

docs/reference/master-configuration-reference.md
```diff
@@ -344,3 +344,114 @@ Comma-separated list of `pattern=N` settings for file-filtered logging.
...
+The following options specify the restrictions that can be applied by the
+nfd-master on the deployed Custom Resources in the cluster.
+
+### restrictions.nodeFeatureNamespaceSelector
+
+The `nodeFeatureNamespaceSelector` option specifies the NodeFeatures namespaces
+to watch, which can be selected by using `metav1.LabelSelector` as a type for
+this option.
+
+Default: all namespaces are selected to be watched.
```
marquiz, line 359: nit: better follow the style of other sections in this doc

```suggestion
this option. An empty value selects all namespaces to be watched.

Default: Default: *empty*
```

pkg/nfd-master/nfd-master.go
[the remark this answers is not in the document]
marquiz: Yeah, let's remove this one instead (also to not break existing users)

docs/reference/master-configuration-reference.md
```diff
@@ -344,3 +344,114 @@ Comma-separated list of `pattern=N` settings for file-filtered logging.
...
+
+```yaml
+restrictions:
+  maxExtendedResourcesPerCR: true
+```
+
+### restrictions.disableAnnotations
+
+The `disableAnnotations` option specifies whether to generate annotations
+from NodeFeature CR or not.
```
marquiz, line 420: nit:

```suggestion
The `disableAnnotations` option controls whether to allow creation of node annotations
from NodeFeatureRule CR or not.
```

docs/reference/master-configuration-reference.md
```diff
@@ -344,3 +344,114 @@ Comma-separated list of `pattern=N` settings for file-filtered logging.
...
+
+```yaml
+restrictions:
+  disableTaints: true
+```
+
+### restrictions.disableExtendedResources
+
+The `disableExtendedResources` option specifies whether to generate taints
+from NodeFeature CR or not.
```
marquiz, line 406: nit:

```suggestion
The `disableExtendedResources` option controls whether to allow creation of node extended resources
from NodeFeatureRule CR or not.
```

docs/reference/master-configuration-reference.md
```diff
@@ -344,3 +344,114 @@ Comma-separated list of `pattern=N` settings for file-filtered logging.
...
+      - key: "kubernetes.io/metadata.name"
+        operator: "In"
+        values:
+          - "node-feature-discovery"
+```
+
+### restrictions.disableLabels
+
+The `disableLabels` option specifies whether to generate labels
+from NodeFeature CR or not.
```
marquiz, line 378: nit:

```suggestion
The `disableLabels` option controls whether to allow creation of node labels
from NodeFeature and NodeFeatureRule CRs or not.
```

docs/reference/master-configuration-reference.md
```diff
@@ -344,3 +344,114 @@ Comma-separated list of `pattern=N` settings for file-filtered logging.
...
+
+```yaml
+restrictions:
+  disableAnnotations: true
+```
+
+### restrictions.allowOverwrite
+
+The `allowOverwrite` option specifies whether to overwrite existing
+labels, annotations, and extended resources if there's an overlap, or not.
```
marquiz, line 434:
```suggestion
The `allowOverwrite` option controls whether NFD is allowed to overwrite and take over management of existing
node labels, annotations, and extended resources. Labels, annotations and extended resources created by NFD itself are not affected (overwrite cannot be disabled). NFD tracks the labels, annotations and extended resources that it manages with specific [node annotations](../get-started/introduction.md#node-annotations).
```

docs/reference/master-configuration-reference.md
```diff
@@ -344,3 +344,114 @@ Comma-separated list of `pattern=N` settings for file-filtered logging.
...
+
+```yaml
+restrictions:
+  overwriteLabels: false
+```
+
+### restrictions.denyNodeFeatureLabels
+
+The `denyNodeFeatureLabels` option specifies whether to deny labels from NodeFeature
+objects or not.
```
marquiz, line 448:
```suggestion
The `denyNodeFeatureLabels` option specifies whether to deny labels from 3rd party NodeFeature
objects or not. NodeFeature objects created by nfd-worker are not affected.
```

marquiz (commented):
A few suggestions/nits on the docs

AhmedThresh:
Thanks for the review @marquiz. PTAL!

TessaIO:
ping @marquiz

marquiz:
> ping @marquiz

Thanks @TessaIO, will take a look in a bit. I was busy with the v0.16 release, plus other stuff 😇

pkg/nfd-master/nfd-api-controller.go
```diff
@@ -45,12 +46,17 @@ type nfdController struct {
 	updateOneNodeChan              chan string
 	updateAllNodeFeatureGroupsChan chan struct{}
 	updateNodeFeatureGroupChan     chan string
+
+	k8sClient       k8sclient.Interface
```
marquiz, line 50: Unused?

pkg/nfd-master/nfd-api-controller.go
```diff
@@ -66,6 +72,14 @@ func newNfdController(config *restclient.Config, nfdApiControllerOptions nfdApiC
 		updateNodeFeatureGroupChan:     make(chan string),
 	}
 
+	labelMap, err := metav1.LabelSelectorAsSelector(nfdApiControllerOptions.NodeFeatureNamespaceSelector)
+	if err != nil {
+		klog.ErrorS(err, "failed to convert label selector to map", "selector", nfdApiControllerOptions.NodeFeatureNamespaceSelector)
+		return nil, err
+	}
+
+	c.namespaceLister = NewNamespaceLister(nfdApiControllerOptions.K8sClient, labelMap)
```
marquiz, line 81: One possible optimization would be that only start the namespace lister if `opts.NodeFeatureNamespaceSelector != nil`. Then short circuit `isNamespaceSelected()` to return true if `c.namespaceLister == nil`, and skip stop() in `c.stop()`. Doesn't need to be in this PR but you could possibly add a TODO comment. WDYT?

docs/reference/master-configuration-reference.md
```diff
@@ -344,3 +344,118 @@ Comma-separated list of `pattern=N` settings for file-filtered logging.
...
+Default: false
+
+Example:
+
+```yaml
+restrictions:
+  disableLabels: true
+```
+
+### restrictions.disableTaints
```
marquiz, line 389: disableTaints is gone, right?

```suggestion
```

pkg/nfd-master/nfd-master.go
```diff
@@ -716,7 +740,7 @@ func getDynamicValue(value string, features *nfdv1alpha1.Features) (string, erro
 	return element, nil
 }
 
-func filterTaints(taints []corev1.Taint) []corev1.Taint {
+func (m *nfdMaster) filterTaints(taints []corev1.Taint) []corev1.Taint {
```
marquiz, line 743: nit: no need to touch this now(?)

```suggestion
func filterTaints(taints []corev1.Taint) []corev1.Taint {
```

TessaIO:
@marquiz are u planning to give this a look once again? Maybe I need to polish some things it's been a while since I worked on it.

test/e2e/node_feature_discovery_test.go
```diff
@@ -1064,6 +1064,229 @@ resyncPeriod: "1s"
...
+					}
+					eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchLabels(expectedLabels, nodes))
+
+					By("Deleting NodeFeature object")
+					err = nfdClient.NfdV1alpha1().NodeFeatures(f.Namespace.Name).Delete(ctx, nodeFeatures[0], metav1.DeleteOptions{})
+					Expect(err).NotTo(HaveOccurred())
+				})
+			})
+
+			Context("max labels, taints restrictions should be respected", func() {
```
marquiz, line 1113: The comment is outdated, there is no "max" setting, just disable/enable(?)

test/e2e/node_feature_discovery_test.go
```diff
@@ -1064,6 +1064,229 @@ resyncPeriod: "1s"
...
+
+					eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchLabels(expectedLabels, nodes))
+
+					By("Deleting NodeFeatureRule #3")
+					err = nfdClient.NfdV1alpha1().NodeFeatureRules().Delete(ctx, "e2e-test-3", metav1.DeleteOptions{})
+					Expect(err).NotTo(HaveOccurred())
+				})
+			})
+
+			Context("max extended resources restriction should be respected", func() {
```
marquiz, line 1172: same here

```suggestion
			Context("disble extended resources restriction should be respected", func() {
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -1064,6 +1064,229 @@ resyncPeriod: "1s"
...
+					nodeFeatures, err := testutils.CreateOrUpdateNodeFeaturesFromFile(ctx, nfdClient, "nodefeature-1.yaml", f.Namespace.Name, targetNodeName)
+					Expect(err).NotTo(HaveOccurred())
+
+					By("Verifying node labels from NodeFeature object #1 are not created")
+					// No labels should be created since the f.Namespace is not in the selected Namespaces
+					expectedLabels := map[string]k8sLabels{
+						targetNodeName: {},
+						"*":            {},
+					}
+					eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchLabels(expectedLabels, nodes))
```
marquiz, line 1105: Hmm, is this kinda racy as this would/could fall through even if the labels et al would eventually be created by NFD, but it hadn't yet created them at the time of this check?

test/e2e/node_feature_discovery_test.go
```diff
@@ -1064,6 +1064,229 @@ resyncPeriod: "1s"
...
+					By("Creating NodeFeatureRules #3")
+					Expect(testutils.CreateNodeFeatureRulesFromFile(ctx, nfdClient, "nodefeaturerule-3.yaml")).NotTo(HaveOccurred())
+
+					By("Verifying node taints and annotation from NodeFeatureRules #3")
+					expectedLabels := map[string]k8sLabels{
+						targetNodeName: {},
+						"*":            {},
+					}
+
+					eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchLabels(expectedLabels, nodes))
```
marquiz, line 1164: Should we first check that the taints (or at least one taint) was created before checking labels? So that we're sure than NFD actually did process the NFR and update the node?

test/e2e/node_feature_discovery_test.go
```diff
@@ -1064,6 +1064,229 @@ resyncPeriod: "1s"
...
+					targetNodeName := nodes[0].Name
+					Expect(targetNodeName).ToNot(BeEmpty(), "No suitable worker node found")
+
+					By("Creating nfd-worker config")
+					cm := testutils.NewConfigMap("nfd-worker-conf", "nfd-worker.conf", `
+core:
+  sleepInterval: "1s"
+  featureSources: ["fake"]
+  labelSources: []
+`)
```
marquiz, line 1139: Do we actually need the worker config (the sleep interval). Would make the test a lot simpler to use just `SpecWithContainerExtraArgs("-label-sources", "fake")`. Also, I think we don't want to disable all label sources (enable fake) so that the worker tries to create some node labels.

test/e2e/node_feature_discovery_test.go
```diff
@@ -1064,6 +1064,229 @@ resyncPeriod: "1s"
...
+				})
+				It("Nothing should be created", func(ctx context.Context) {
+					nodes, err := getNonControlPlaneNodes(ctx, f.ClientSet)
+					Expect(err).NotTo(HaveOccurred())
+
+					targetNodeName := nodes[0].Name
+					Expect(targetNodeName).ToNot(BeEmpty(), "No suitable worker node found")
+
+					By("Creating nfd-worker config")
+					cm := testutils.NewConfigMap("nfd-worker-conf", "nfd-worker.conf", `
```
marquiz, line 1192: Also here I'm not sure we need the worker config(?)

test/e2e/node_feature_discovery_test.go
```diff
@@ -1064,6 +1064,229 @@ resyncPeriod: "1s"
...
+						targetNodeName: {},
+						"*":            {},
+					}
+					expectedCapacity := map[string]corev1.ResourceList{
+						targetNodeName: {},
+						"*":            {},
+					}
+
+					By("Creating NodeFeatureRules #4")
+					Expect(testutils.CreateNodeFeatureRulesFromFile(ctx, nfdClient, "nodefeaturerule-4.yaml")).NotTo(HaveOccurred())
```
marquiz, line 1222: Also here, should we pick a NodeFeatureRule that tries to create some ERs and e.g. annotations -> we could verify that the annotations are created but the ERs are not. Maybe we should have one dedicated NFR with one rule that creates all, labels, annotations and ERs. Then for each of these test cases verify that the others are created but the one disabled is not. WDYT?

marquiz (commented):
Wow, good job @TessaIO 🚀 A few small comments on the code and e2e tests but we're almost there

marquiz:
ping @TessaIO

TessaIO:
Thanks for the review @marquiz. Gonna update the PR over the weekend!

pkg/nfd-master/nfd-api-controller.go
AhmedThresh: makes sense, thanks for the suggestion

AhmedThresh:
@marquiz I addressed some comments, I'll address the other ones ASAP.

AhmedThresh:
Rebasing on this PR is becoming harder each time. @marquiz Would it bother you if I close this and open a new PR from @TessaIO account with everything fixed there?

marquiz:
> Rebasing on this PR is becoming harder each time. @marquiz Would it bother you if I close this and open a new PR from @TessaIO account with everything fixed there?

Sorry, I don't quite get how a new PR would help rebasing? Or something releated to AhmedThresh vs TessaIO?

If possible, I'd prefer keeping it in this same PR for now (as we're almost done). The e2e tests should be easy to rebase, basically removing one level of indentation. The `nfdTestSuite` closure function was lately removed as part of removing support for gRPC.

test/e2e/node_feature_discovery_test.go
AhmedThresh: Should we add a sleep to wait, or what do you suggest to mitigate this racy case?

test/e2e/node_feature_discovery_test.go
AhmedThresh: It's hard to see what test case you are referencing here, but for the context `disable labels restrictions should be respected` I think this is already done, right?

test/e2e/node_feature_discovery_test.go
AhmedThresh: yes that makes sense.

test/e2e/node_feature_discovery_test.go
marquiz: I think I was referring to the `By("Verifying node labels from NodeFeature object #1 are not created")` test case

test/e2e/node_feature_discovery_test.go
marquiz: Wouldn't [this](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1592#discussion_r1668783915)? Check that thing X (e.g. annotations) was created and then check that thing Y (e.g. labels) wasn't

marquiz:
@AhmedThresh @TessaIO thanks for the udpates. The e2e tests are still failing. I haven't looked at the details.

Maybe it would be simpler/easier to create a new dedicated NFR (under `test/e2e/data`) for these new tests (ref https://github.com/kubernetes-sigs/node-feature-discovery/pull/1592#discussion_r1668783915). WDYT?

TessaIO:
> @AhmedThresh @TessaIO thanks for the udpates. The e2e tests are still failing. I haven't looked at the details.
> 
> Maybe it would be simpler/easier to create a new dedicated NFR (under `test/e2e/data`) for these new tests (ref [#1592 (comment)](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1592#discussion_r1668783915)). WDYT?

Yes sure, I'm trying it to make it work with the current setup then I will create one NFR for all.

AhmedThresh:
@marquiz PTAL

marquiz:
Wow, thanks @AhmedThresh @tessa, will review tomorrow. Looking forward to get this in for v0.17

marquiz (commented):
Thanks for the update. Some thoughts on the e2e below. I think we could simplify quite a bit there (and still cover the new functionality)

test/e2e/node_feature_discovery_test.go
```diff
@@ -1018,6 +1017,305 @@ resyncPeriod: "1s"
...
+				nodeFeatures, err := testutils.CreateOrUpdateNodeFeaturesFromFile(ctx, nfdClient, "nodefeature-1.yaml", f.Namespace.Name, targetNodeName)
+				Expect(err).NotTo(HaveOccurred())
+
+				By("Verifying node labels from NodeFeature object #1 are not created")
+				// No labels should be created since the f.Namespace is not in the selected Namespaces
+				expectedLabels := map[string]k8sLabels{
+					targetNodeName: {},
+					"*":            {},
+				}
+				eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchLabels(expectedLabels, nodes))
```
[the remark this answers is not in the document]
TessaIO: The issue is with the last verification because when a NodeFeature is not selected in the current implementation, we don't do anything. We don't act as if it was deleted, therefore labels won't be removed even tho the NodeFeature is not selected anymore.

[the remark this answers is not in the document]
TessaIO: We can do it the other way around, were start from an initial state where we have no labels, we create a non-selected NodeFeature, and then we patch the namespace with specific labels to make NodeFeature selectable and therefore we can see labels. WDYT?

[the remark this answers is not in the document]
marquiz: > The issue is with the last verification because when a NodeFeature is not selected in the current implementation, we don't do anything. We don't act as if it was deleted, therefore labels won't be removed even tho the NodeFeature is not selected anymore.

Yup, you're right. That's why I suggested running nfd-master with a resync-period of 1s -> forces full evaluation of CRDs every second. When the NodeFeature gets "un-selected" the related labels should be removed.



> We can do it the other way around, were start from an initial state where we have no labels, we create a non-selected NodeFeature, and then we patch the namespace with specific labels to make NodeFeature selectable and therefore we can see labels. WDYT?

This also needs the quick resync-period in order to mitigate the "we do nothing" problem. But again, I think this has the same raciness problem, hard to be entirely sure when the un-selected NF was processed. So I think the way I suggested is safer because you have clear triggers to watch when nfd-master acted. WDYT?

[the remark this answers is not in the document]
TessaIO: > When the NodeFeature gets "un-selected" the related labels should be removed.

This is the part missing in the implementation. If you check the current code, if it's unselected we simply log that it's not selected and we continue. There's no removing the labels in that case. Do you think that it should be added? 
I think filtering of namespaces should happen at the level of  `getAndMergeNodeFeatures`  function. thoughts?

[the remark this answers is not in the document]
marquiz: Ach, exactly :man_facepalming: That is a key part missing. We really need to do the filtering there (too). If we only do the filtering in the event functions it doesn't prevent nfd-master to using all the NFs

test/e2e/node_feature_discovery_test.go
```diff
@@ -1018,6 +1017,267 @@ resyncPeriod: "1s"
...
+				By("Verifying node taints and labels from NodeFeatureRules #6")
+				expectedTaints := map[string][]corev1.Taint{
+					"*": {},
+				}
+				eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchTaints(expectedTaints, nodes))
+
+				By("Verifying node labels from NodeFeature object #6 are not created")
+				expectedLabels := map[string]k8sLabels{
+					targetNodeName: {},
+					"*":            {},
```
marquiz, line 1272: Hmm, shouldn't we actually get all the labels, annotations and ERs from nodefeaturerule-6?

pkg/nfd-master/nfd-master.go
```diff
@@ -876,6 +907,12 @@ func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.No
 		return fmt.Errorf("failed to merge NodeFeature objects for node %q: %w", node.Name, err)
 	}
 
+	// If we deny node feature labels, we'll empty the labels variable
+	if m.config.Restrictions.DenyNodeFeatureLabels {
+		klog.V(2).InfoS("node feature labels are disabled in configuration (restrictions.denyNodeFeatureLabels=true)")
+		nodeFeatures.Labels = map[string]string{}
```
marquiz, line 913: As per documentation (in `docs/reference/master-configuration-reference.md`) we should disable/deny labels from 3rd party NFs. So the one created by nfd-worker should be ok. We need to move this check to `getAndMergeNodeFeatures()`. Then, how to determine the NF created by nfd-worker: I think the logic would be namespace==our namespace AND name==nodeName.

Thoughts?

pkg/nfd-master/nfd-master.go
```diff
@@ -817,41 +841,56 @@ func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeF
...
 		// Merge in features
 		//
 		// NOTE: changing the rule api to support handle multiple objects instead
 		// of merging would probably perform better with lot less data to copy.
-		features := objs[0].Spec.DeepCopy()
+		features := filteredObjs[0].Spec.DeepCopy()
 		if !nfdfeatures.NFDFeatureGate.Enabled(nfdfeatures.DisableAutoPrefix) && m.config.AutoDefaultNs {
-			features.Labels = addNsToMapKeys(features.Labels, nfdv1alpha1.FeatureLabelNs)
+			if m.config.Restrictions.DenyNodeFeatureLabels && !m.isThirdPartyNodeFeature(*filteredObjs[0], nodeName, m.namespace) {
+				klog.V(2).InfoS("node feature labels are disabled in configuration (restrictions.denyNodeFeatureLabels=true)")
```
marquiz, line 881: This shouldn't depend on `NFDFeatureGate.Enabled(nfdfeatures.DisableAutoPrefix)`

We should skip the NF independent on whether that feature gate is enable or disabled

test/e2e/node_feature_discovery_test.go
```diff
@@ -1018,6 +1017,287 @@ resyncPeriod: "1s"
...
+				Expect(testutils.CreateNodeFeatureRulesFromFile(ctx, nfdClient, "nodefeaturerule-6.yaml")).NotTo(HaveOccurred())
+
+				By("Verifying node taints and labels from NodeFeatureRules #6")
+				expectedTaints := map[string][]corev1.Taint{
+					"*": {},
+				}
+				eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchTaints(expectedTaints, nodes))
+
+				expectedAnnotations := map[string]k8sAnnotations{
+					targetNodeName: {
```
marquiz, line 1269: Shouldn't we get this for every node? I.e. 

```suggestion
					"*": {
```

And ditch the targetnode? The targetnode would only be meaningful if you also deployed a NF. Then the target node should get labels from the NF but not from the NFR. WDYT?

pkg/nfd-master/nfd-master.go
```diff
@@ -817,41 +841,57 @@ func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeF
...
 		// Merge in features
 		//
 		// NOTE: changing the rule api to support handle multiple objects instead
 		// of merging would probably perform better with lot less data to copy.
-		features := objs[0].Spec.DeepCopy()
-		if !nfdfeatures.NFDFeatureGate.Enabled(nfdfeatures.DisableAutoPrefix) && m.config.AutoDefaultNs {
-			features.Labels = addNsToMapKeys(features.Labels, nfdv1alpha1.FeatureLabelNs)
+		features := filteredObjs[0].Spec.DeepCopy()
+
+		if m.config.Restrictions.DenyNodeFeatureLabels && !m.isThirdPartyNodeFeature(*filteredObjs[0], nodeName, m.namespace) {
```
marquiz, line 880: This still only affects the 'DisableAutoPrefix` feature gate, not if the NF is skipped or not

pkg/nfd-master/nfd-master.go
```diff
@@ -817,41 +841,57 @@ func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeF
...
-		for _, o := range objs[1:] {
+		for _, o := range filteredObjs[1:] {
 			s := o.Spec.DeepCopy()
 			if !nfdfeatures.NFDFeatureGate.Enabled(nfdfeatures.DisableAutoPrefix) && m.config.AutoDefaultNs {
-				s.Labels = addNsToMapKeys(s.Labels, nfdv1alpha1.FeatureLabelNs)
+				if m.config.Restrictions.DenyNodeFeatureLabels && !m.isThirdPartyNodeFeature(*filteredObjs[0], nodeName, m.namespace) {
+					klog.V(2).InfoS("node feature labels are disabled in configuration (restrictions.denyNodeFeatureLabels=true)")
+				} else {
+					s.Labels = addNsToMapKeys(s.Labels, nfdv1alpha1.FeatureLabelNs)
+				}
```
marquiz, line 894: Looking at this, I think the whole function needs to be refactored. The whole outer loop needs to be rewritten (and we shouldn't need to `if len(filteredObjs) > 0`). For example, something like (haven't tested):

```golang
features := &nfdv1alpha1.NodeFeatureSpec{}

for _, o := range filteredObjs {
	if m.config.Restrictions.DenyNodeFeatureLabels && m.isThirdPartyNodeFeature(o, nodeName, m.namespace) {
		klog.V(2).InfoS("node feature labels are disabled in configuration (restrictions.denyNodeFeatureLabels=true)")
		continue
	}
	s := o.Spec.DeepCopy()

	if !nfdfeatures.NFDFeatureGate.Enabled(nfdfeatures.DisableAutoPrefix) && m.config.AutoDefaultNs {
		s.Labels = addNsToMapKeys(s.Labels, nfdv1alpha1.FeatureLabelNs)
	}

	s.MergeInto(features)
}

return &nfdv1alpha1.NodeFeature{
		Name: nodeName,
	},
	Spec: *features
}, nil
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -1018,6 +1017,284 @@ resyncPeriod: "1s"
...
+					"*": {
+						"e2e.feature.node.kubernetes.io/restricted-er-1": resourcev1.MustParse("2"),
+					},
+				}
+				eventuallyNonControlPlaneNodes(ctx, f.ClientSet).WithTimeout(1 * time.Minute).Should(MatchCapacity(expectedCapacity, nodes))
+
+				By("Verifying node labels from NodeFeature object #6 are not created")
+				expectedLabels := map[string]k8sLabels{
+					"*": {
+						"e2e.feature.node.kubernetes.io/restricted-label-1": "true",
```
marquiz, line 1288: There has been so much devilish details in this feature that it would be good to verify that the non-3rd-party NF labels ARE still created. Either by deploying nfd-worker or faking a NF.

You could also just add a note and we could add it in a separate PR. WDYT?

test/e2e/node_feature_discovery_test.go
```diff
@@ -1018,6 +1017,305 @@ resyncPeriod: "1s"
...
+`)
+				_, err := f.ClientSet.CoreV1().ConfigMaps(f.Namespace.Name).Create(ctx, cm, metav1.CreateOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("No labels should be created", func(ctx context.Context) {
+				nodes, err := getNonControlPlaneNodes(ctx, f.ClientSet)
+				Expect(err).NotTo(HaveOccurred())
+
+				targetNodeName := nodes[0].Name
+				Expect(targetNodeName).ToNot(BeEmpty(), "No suitable worker node found")
```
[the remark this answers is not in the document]
marquiz: How about this?

pkg/nfd-master/nfd-master.go
```diff
@@ -817,42 +841,63 @@ func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeF
...
-			features.Labels = addNsToMapKeys(features.Labels, nfdv1alpha1.FeatureLabelNs)
+		features := filteredObjs[0].Spec.DeepCopy()
+
+		if m.config.Restrictions.DenyNodeFeatureLabels && m.isThirdPartyNodeFeature(*filteredObjs[0], nodeName, m.namespace) {
+			klog.V(2).InfoS("node feature labels are disabled in configuration (restrictions.denyNodeFeatureLabels=true)")
+			features.Labels = nil
+		} else {
+			if !nfdfeatures.NFDFeatureGate.Enabled(nfdfeatures.DisableAutoPrefix) && m.config.AutoDefaultNs {
+				features.Labels = addNsToMapKeys(features.Labels, nfdv1alpha1.FeatureLabelNs)
+			}
```
marquiz, line 886: Note: the else is not necessary but 

```suggestion
		if !nfdfeatures.NFDFeatureGate.Enabled(nfdfeatures.DisableAutoPrefix) && m.config.AutoDefaultNs {
			features.Labels = addNsToMapKeys(features.Labels, nfdv1alpha1.FeatureLabelNs)
		}
```

pkg/nfd-master/nfd-master.go
```diff
@@ -817,42 +841,63 @@ func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeF
...
 		}
-		for _, o := range objs[1:] {
+
+		for _, o := range filteredObjs[1:] {
 			s := o.Spec.DeepCopy()
+			if m.config.Restrictions.DenyNodeFeatureLabels && m.isThirdPartyNodeFeature(*o, nodeName, m.namespace) {
+				klog.V(2).InfoS("node feature labels are disabled in configuration (restrictions.denyNodeFeatureLabels=true)")
+				s.Labels = nil
+				continue
```
marquiz, line 894: We shouldn't skip the whole NF, just set labels to nil.

```suggestion
```

Why didn't the e2e tests catch this? :thinking: We should have a NFR there to verify that the features are still present. You can also just add a note about this there for now.

test/e2e/node_feature_discovery_test.go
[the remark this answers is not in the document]
TessaIO: Ah yes, I forgot about

TessaIO:
ping @marquiz

docs/reference/master-configuration-reference.md
```diff
@@ -338,3 +338,104 @@ Comma-separated list of `pattern=N` settings for file-filtered logging.
 Default: *empty*
 
 Run-time configurable: yes
+
+## restrictions
+
+The following options specify the restrictions that can be applied by the
```
marquiz, line 344: One last comment. Should we call this **EXPERIMENTAL** in this release(?) There have been so many subtle details during the review process that we've certainly missed something :blush: WDYT?

TessaIO: Yes, please. I will create another issue to cover those unseen details. If you have some bullet points in mind you can leave them here.

TessaIO:
@marquiz PTAL

marquiz (approved):
Thanks @TessaIO for this monumental effort with a lot of twists, turns and surprises :100: :rocket: I'm ready to merge this now.

We have some TODO items to improve the e2e-tests and will address those later

I'll let @ArangoGutierrez to take a final look. E.g. check the docs if you understand what we're trying to do here :stuck_out_tongue_winking_eye: 
/assign @ArangoGutierrez

TessaIO:
> Thanks @TessaIO for this monumental effort with a lot of twists, turns and surprises 💯 🚀 I'm ready to merge this now.

Thanks for your thorough and precise review! 🚀

TessaIO:
kind reminder @ArangoGutierrez @marquiz

ArangoGutierrez (approved):
/lgtm

marquiz:
Ach, finally :rocket: :relaxed:

PR #1594, opened 2024-02-19
source/cpu: fix build tags on rdt discovery

marquiz:
Our current code only builds/works on linux.


xvilo:
> I wonder is there people out there spinning up kube clusters on those AWS Apple Silicon nodes? should we add some support for Apple Kube Clusters???? /lgtm

Might be late to the party, but yes! Home lab, M1 Mac Mini with Asahi linux :)

PR #1595, opened 2024-02-20
nfd-master: check if node exists before trying update

marquiz:
Make the node-updater-pool worker fail fast (and not retry updates) if a node does not exist.

### DIFF
--- a/pkg/nfd-master/node-updater-pool.go
+++ b/pkg/nfd-master/node-updater-pool.go
@@ -21,6 +21,7 @@ import (
 	"time"
 
 	"golang.org/x/time/rate"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 )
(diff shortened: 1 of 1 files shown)


pkg/nfd-master/node-updater-pool.go
```diff
@@ -41,15 +42,20 @@ func newNodeUpdaterPool(nfdMaster *nfdMaster) *nodeUpdaterPool {
...
 
 	defer queue.Done(nodeName)
 
 	nodeUpdateRequests.Inc()
-	if err := u.nfdMaster.nfdAPIUpdateOneNode(nodeName.(string)); err != nil {
+
+	// Check if node exists
+	if _, err := u.nfdMaster.getNode(nodeName); apierrors.IsNotFound(err) {
+		klog.InfoS("node not found, skip update", "nodeName", nodeName)
+	} else if err := u.nfdMaster.nfdAPIUpdateOneNode(nodeName); err != nil {
```
AhmedThresh, line 58: See https://github.com/kubernetes-sigs/node-feature-discovery/pull/1596#discussion_r1495919998

PR #1596, opened 2024-02-20
nfd-master: retry node updates indefinitely

marquiz:
Treat node updates like a reconciliation loop. Keep trying on node
update as long as it fails. Node update permafailing likely indicates a
bug in the nfd code (there should be no reason for it to fail forever)
and it's better to clearly see it in the logs/metrics rather than giving
up after a few retries.

Depends on #1595

### DIFF
--- a/pkg/nfd-master/node-updater-pool.go
+++ b/pkg/nfd-master/node-updater-pool.go
@@ -56,14 +56,15 @@ func (u *nodeUpdaterPool) processNodeUpdateRequest(queue workqueue.RateLimitingI
 	if _, err := u.nfdMaster.getNode(nodeName); apierrors.IsNotFound(err) {
 		klog.InfoS("node not found, skip update", "nodeName", nodeName)
 	} else if err := u.nfdMaster.nfdAPIUpdateOneNode(nodeName); err != nil {
-		if queue.NumRequeues(nodeName) < 15 {
-			klog.InfoS("retrying node update", "nodeName", nodeName, "lastError", err)
-			queue.AddRateLimited(nodeName)
-			return true
+		if n := queue.NumRequeues(nodeName); n < 15 {
+			klog.InfoS("retrying node update", "nodeName", nodeName, "lastError", err, "numRetries", n)
 		} else {
-			klog.ErrorS(err, "failed to update node", "nodeName", nodeName)
+			klog.ErrorS(err, "node update failed, queuing for retry ", "nodeName", nodeName, "numRetries", n)
+			// Count only long-failing attempts
 			nodeUpdateFailures.Inc()
 		}
+		queue.AddRateLimited(nodeName)
+		return true
 	}
 	queue.Forget(nodeName)
 	return true


marquiz:
Pending #1595
/hold

/assign @ArangoGutierrez
ping @AhmedThresh

pkg/nfd-master/node-updater-pool.go
```diff
@@ -41,23 +42,29 @@ func newNodeUpdaterPool(nfdMaster *nfdMaster) *nodeUpdaterPool {
 }
 
 func (u *nodeUpdaterPool) processNodeUpdateRequest(queue workqueue.RateLimitingInterface) bool {
-	nodeName, quit := queue.Get()
+	n, quit := queue.Get()
+	nodeName := n.(string)
 	if quit {
 		return false
 	}
```
AhmedThresh, line 49: I think string casting should be moved after the quit check because according to the queue package:
```go
if len(q.queue) == 0 {
	// We must be shutting down.
	return nil, true
}
```
We risk having nil casting error
```suggestion
	if quit {
		return false
	}
	nodeName := n.(string)
```

AhmedThresh (commented):
Thanks @marquiz for working on this. one comment is left below!

pkg/nfd-master/node-updater-pool.go
```diff
@@ -41,23 +42,29 @@ func newNodeUpdaterPool(nfdMaster *nfdMaster) *nodeUpdaterPool {
...
-	if err := u.nfdMaster.nfdAPIUpdateOneNode(nodeName.(string)); err != nil {
-		if queue.NumRequeues(nodeName) < 15 {
-			klog.InfoS("retrying node update", "nodeName", nodeName, "lastError", err)
-			queue.AddRateLimited(nodeName)
-			return true
+
+	// Check if node exists
+	if _, err := u.nfdMaster.getNode(nodeName); apierrors.IsNotFound(err) {
+		klog.InfoS("node not found, skip update", "nodeName", nodeName)
+	} else if err := u.nfdMaster.nfdAPIUpdateOneNode(nodeName); err != nil {
```
AhmedThresh, line 58: Looking into this also, I see that `nfdAPIUpdateOneNode` calls the `getNode`  function implicitly (by invoking `refreshNodeFeatures` and `updateNodeObject` functions) which makes us query the API-server twice.
what about this?
```suggestion
	if err := u.nfdMaster.nfdAPIUpdateOneNode(nodeName); apiErrors.IsNotFound(err) != nil {
```

pkg/nfd-master/node-updater-pool.go
marquiz: Yes, this is fixed in #1595. I'll rebase after that gets in

pkg/nfd-master/node-updater-pool.go
marquiz: I thought about that, too. However, I decided it's safer this way as we're also querying other api objects (NodeFeatures and NodeFeatureRules, and maybe soon even NodeFeatureGroups) so we cannot be sure which object was it that returned this error (especially after the code evolves).

AhmedThresh: Should we then update the ´nfdAPIUpdateOneNode´ to accept the actual node structure instead of the name, to avoid re-querying the api-server?

marquiz: > Should we then update the nfdAPIUpdateOneNode to accept the actual node structure instead of the name, to avoid re-querying the api-server?

I considered also this option but decided not to do this (yet) as the lower node update functions are wired to the gRPC API, too (from which we only get the node name, not the node object). I think it's cleaner and easier do refactoring of this all after the gRPC API is dropped.

TessaIO: Okay then, should be good from my side!

TessaIO (approved):
Thanks @marquiz, LGTM!

marquiz:
Thanks @TessaIO, check #1595

marquiz:
Rebased. #1595 merged
/unhold

fmuyassarov (commented):
/lgtm

Issue #1597, opened 2024-02-21
In the Helm chart, add in a way to customize the Issuer used for the NFD certs


allenmun197:
**What would you like to be added**:  I would like to see the NFD helm chart have some way of bringing a custom Issuer or ClusterIssuer as the CA for the NFD certs.

**Why is this needed**:  This is useful because we have enterprise policies which do not allow self-signed certs in the cluster and have an approved ClusterIssuer to use.  This feature should be useful for anyone looking to centralize the certs being issued in the cluster.

PR #1598, opened 2024-02-21
helm: add ability to use a custom issuer

allenmun197:
This PR adds in the ability to use a custom Issuer when creating the certs for NFD in the helm chart.  This is useful when there are cluster/enterprise policies that deny creating a self-signed CA.

Fixes https://github.com/kubernetes-sigs/node-feature-discovery/issues/1597


deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -529,7 +529,10 @@ gc:
 # need to manually, or otherwise, provision the TLS certs as secrets
 tls:
   enable: false
-  certManager: false
+  certManager:
+    enable: false
+    issuerKind:
+    issuerName:
```
marquiz, line 535: We shouldn't break existing users


```suggestion
  certManager: false
  certManageIssuerKind:
  certManageIssuerName:
```

OR

```suggestion
  certManager: false
  certManagerCertificate:
    issuerKind:
    issuerName:
```

marquiz (commented):
Thanks @allenmun197. I think we could take the patch, even though all of this is deprecated functionality. In more detail: note that the gRPC communication (and everything related, like TLS and cert-manager support) has been deprecated since NFD v0.13, disabled since NFD v0.14 and likely to be dropped in NFD v0.17. So, unless you specifically enable gRPC with `enableNodeFeatureApi: true` the TLS/cert-manager settings will not be used.

When adding new helm params, we also need to document those in `docs/deployment/helm.md`.

allenmun197:
@marquiz 

Ok, I moved the fields into `certManagerCertificate` as suggested.  Also added the documentation for the new helm fields in `docs/deployment/helm.md`.  I also added an additional check to make sure if `.tls.certManagerCertificate.issuerKind` is set but `.tls.certManagerCertificate.issuerName` is not set, it will not use `.tls.certManagerCertificate.issuerKind` and just default to `Issuer`.  This fixes an issue when you have a values file that looks like
```
tls:
  certManager: enable
  certManagerCertificate:
    #issuerName: example-issuer
    issuerKind: ClusterIssuer
```

deployment/helm/node-feature-discovery/templates/cert-manager-certs.yaml
```diff
@@ -60,8 +60,8 @@ spec:
   dnsNames:
   - {{ include "node-feature-discovery.fullname" . }}-topology-updater.{{ include "node-feature-discovery.namespace" .  }}.svc.cluster.local
   issuerRef:
-    name: nfd-ca-issuer
-    kind: Issuer
+    name: {{ default "nfd-ca-issuer" .Values.tls.certManagerCertificate.issuerName }}
+    kind: {{ default "Issuer" ( kindIs "invalid" .Values.tls.certManagerCertificate.issuerName | ternary "" .Values.tls.certManagerCertificate.issuerKind ) }}
```
marquiz, line 64: This is quite cryptic to read, at least for me 😇

How about a bit longer but more explicit:

```sugggestion
    {{- if and .Values.tls.certManagerCertificate.issuerName .Values.tls.certManagerCertificate.issuerKind }}
    kind: {{ .Values.tls.certManagerCertificate.issuerKind }}
    {{- else }}
    kind: Issuer
    {{- end }}
```

allenmun197: Got it.  Updated the PR to use more simplified logic with regards to the issuer kind.

marquiz (approved):
Thanks @allenmun197. I haven't tested this but looks good to me 👍 
/assign @ArangoGutierrez

allenmun197:
@marquiz @ArangoGutierrez 

Is this PR still in review?

PR #1600, opened 2024-02-27
Move NFD api to a separate go mod

ArangoGutierrez:
The idea behind this patch is to make it easier for third party apps to import NFD API.


Dockerfile
```diff
@@ -7,6 +7,7 @@ FROM ${BUILDER_IMAGE} as builder
 
 # Get (cache) deps in a separate layer
 COPY go.mod go.sum /go/node-feature-discovery/
+COPY api/ /go/node-feature-discovery/api/
```
marquiz, line 10: Just the go.mod?

api/go.mod
```diff
@@ -0,0 +1,63 @@
...
+
+require (
+	github.com/gogo/protobuf v1.3.2
+	github.com/stretchr/testify v1.8.4
+	golang.org/x/exp v0.0.0-20240222234643-814bf88cf225
+	k8s.io/api v0.29.2
+	k8s.io/apimachinery v0.29.2
+	k8s.io/client-go v0.29.2
+	k8s.io/klog/v2 v2.120.1
+	sigs.k8s.io/node-feature-discovery v0.15.1
```
marquiz, line 13: We shouldn't depend on the main pkg -> some refactoring needed

marquiz (commented):
Thanks @ArangoGutierrez. Maybe we should put it under `api/nfd/v1alpha1`?

Dockerfile
```diff
@@ -10,11 +10,11 @@ COPY go.mod go.sum /go/node-feature-discovery/
 
 WORKDIR /go/node-feature-discovery
 
-RUN go mod download
-
 # Do actual build
 COPY . /go/node-feature-discovery
 
+RUN go mod download
+
```
marquiz, line 17: Doing this after the COPY nullifies the whole point of pre-loading go modules (for caching). Can't you copy the go.{mod,sum} of the new module above, similarly to the main module?

go.mod
```diff
@@ -20,22 +19,23 @@ require (
 	github.com/spf13/cobra v1.8.0
 	github.com/stretchr/testify v1.8.4
 	github.com/vektra/errors v0.0.0-20140903201135-c64d83aba85a
-	golang.org/x/exp v0.0.0-20240112132812-db7319d0e0e3
-	golang.org/x/net v0.20.0
+	golang.org/x/exp v0.0.0-20240222234643-814bf88cf225
```
marquiz, line 22: If/when doing any updates on dependencies, please submit a separate PR.

go.mod
marquiz, line 1: Should(n't) we add a replace directive to use the api/nfd module from the local copy?

go.mod
ArangoGutierrez: was not intentional, result of `go mod tidy`

go.mod
ArangoGutierrez: look at the last line of replace

hack/generate.sh
```diff
@@ -26,14 +26,12 @@ ${K8S_CODE_GENERATOR}/generate-groups.sh client,informer,lister \
     "nfd:v1alpha1" --output-base=. \
     --go-header-file hack/boilerplate.go.txt
 
-rm -rf pkg/generated
-
-mv sigs.k8s.io/node-feature-discovery/pkg/generated pkg/
+rm -rf api/v1alpha1/{clientset,informers,listers}
```
marquiz, line 29: Shouldn't this be like

```suggestion
rm -rf api/generated
```

ArangoGutierrez: Good catch!

ArangoGutierrez:
After https://github.com/kubernetes-sigs/node-feature-discovery/pull/1605, we can restart the conversation here

ArangoGutierrez:
@marquiz PTAL

ArangoGutierrez:
@marquiz rebased PTAL

ArangoGutierrez:
@vsoch  this helps towards the goal of serving your use case.

Issue #1603, opened 2024-03-05
hooks does not work as expect


wpeng102:
**What happened**:
Hooks can not add label to the node.

In the following code block, NFD wants to merge `labelsFromHooks` to `labelsFromFiles`, but in it uses `labelsFromHooks` incorrectly in Line 167
https://github.com/kubernetes-sigs/node-feature-discovery/blob/b984db0a163af61240d5dbd0913fc137e5d1ecfa/source/local/local.go#L162-L168

PR #1604, opened 2024-03-05
fix hook issue

wpeng102:
Fix issue: https://github.com/kubernetes-sigs/node-feature-discovery/issues/1603

Root cause:
In the following code block, NFD wants to merge `labelsFromHooks` to `labelsFromFiles`, but in it uses `labelsFromHooks` incorrectly in Line 167
https://github.com/kubernetes-sigs/node-feature-discovery/blob/b984db0a163af61240d5dbd0913fc137e5d1ecfa/source/local/local.go#L162-L168


marquiz (approved):
Thanks @wpeng102 for the fix

/cherry-pick release-0.15

PR #1605, opened 2024-03-08
Update generate scripts to use latest code_gen functions

ArangoGutierrez:


ArangoGutierrez:
Ha!  doesn't need rebase :) 
@marquiz PTAL

hack/update_codegen.sh
```diff
@@ -0,0 +1,69 @@
...
+  --output-base "${NFD_ROOT}" \
+  --boilerplate ${NFD_ROOT}/hack/boilerplate.go.txt
+
+kube::codegen::gen_client \
+  --input-pkg-root sigs.k8s.io/node-feature-discovery/pkg/apis \
+  --output-pkg-root sigs.k8s.io/node-feature-discovery/pkg/generated \
+  --output-base "${NFD_ROOT}" \
+  --boilerplate ${NFD_ROOT}/hack/boilerplate.go.txt \
+  --with-watch \
+  --with-applyconfig
```
marquiz, line 62: Do we need/want this?

marquiz (commented):
Thanks @ArangoGutierrez.

Some notes below. Plus the PR needs a rebase.

Also, the commit message is misleading (and short). It's not updating the codegen tools but updating/revamping the generator script.

marquiz:
> Ha! doesn't need rebase :)

OK, so it seems. However, could you rebase so that it becomes cleaerer what this PR actually changes

Dockerfile_generator
```diff
@@ -21,3 +21,7 @@ RUN curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v25.3
 
 # Expect to be working with nfd
 WORKDIR /go/node-feature-discovery
+
+# We need to set the /go/node-feature-discovery directory as a safe directory.
+# This allows git commands to run in the container.
```
marquiz, line 26: Maybe add a note that this is only for MacOs?

pkg/apis/nfd/v1alpha1/types.go
```diff
@@ -31,11 +31,12 @@ type NodeFeatureList struct {
 	Items []NodeFeature `json:"items"`
 }
 
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
marquiz, line 35: Why were these moved? If they need to be moved then the kubebuilder stuff, too?

pkg/apis/nfd/v1alpha1/types.go
```diff
@@ -111,13 +112,14 @@ type NodeFeatureRuleList struct {
 	Items []NodeFeatureRule `json:"items"`
 }
 
+// +genclient
+// +genclient:nonNamespaced
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
marquiz, line 117: Same here, do we need to move these?

pkg/generated/applyconfiguration/internal/internal.go
marquiz, line 1: I think all the applyconfiguration files should be dropped from the PR now that `--with-applyconfig` was dropped from the codegen args. Right?

marquiz (commented):
Looks better and more confined, thanks for the update. A few more comments below. 

Also, please update the commit message (https://github.com/kubernetes-sigs/node-feature-discovery/pull/1605#pullrequestreview-1927344426)

ArangoGutierrez:
/retitle Update generate scripts to use latest code_gen functions

marquiz (approved):
/lgtm

PR #1609, opened 2024-03-11
nfd-worker: Add liveness probe

### LINKED ISSUE #1549: Liveness probe for nfd-worker

Add a liveness and/or readiness probe, e.g. a gRPC health server to nfd-worker.

**Why is this needed**:

More eyes on the well-being of nfd-worker instances.

ozhuraki:
Closes #1549

### DIFF
--- a/cmd/nfd-worker/main.go
+++ b/cmd/nfd-worker/main.go
@@ -32,7 +32,8 @@ import (
 
 const (
 	// ProgramName is the canonical name of this program
-	ProgramName = "nfd-worker"
+	ProgramName    = "nfd-worker"
+	GrpcHealthPort = 8082
 )
 
 func main() {
@@ -81,6 +82,7 @@ func main() {
 	utils.ConfigureGrpcKlog()
 
 	// Get new NfdWorker instance
+	args.GrpcHealthPort = GrpcHealthPort
 	instance, err := worker.NewNfdWorker(args)
 	if err != nil {
 		klog.ErrorS(err, "failed to initialize NfdWorker instance")
--- a/deployment/base/worker-daemonset/worker-daemonset.yaml
+++ b/deployment/base/worker-daemonset/worker-daemonset.yaml
@@ -19,6 +19,17 @@ spec:
         - name: nfd-worker
           image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
           imagePullPolicy: Always
+          livenessProbe:
+            grpc:
+              port: 8082
+            initialDelaySeconds: 10
+            periodSeconds: 10
+          readinessProbe:
+            grpc:
+              port: 8082
+            initialDelaySeconds: 5
+            periodSeconds: 10
+            failureThreshold: 10
           command:
             - "nfd-worker"
           args:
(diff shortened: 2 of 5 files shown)


cmd/nfd-worker/main.go
```diff
@@ -73,6 +74,7 @@ func main() {
 	utils.ConfigureGrpcKlog()
 
 	// Get new NfdWorker instance
+	args.GrpcHealthPort = GrpcHealthPort
```
marquiz, line 77: Just checkin', is this similar to how it's currently handled in nfd-master (i.e. not user configurable)? If so, it's good, we can add command line flag to both nfd-master and nfd-worker later, if needed.

deployment/base/worker-daemonset/worker-daemonset.yaml
marquiz, line 1: We need to do the corresponding additions in the Helm chart, too

marquiz (commented):
Thanks @ozhuraki for the PR. Two comments below. Almost ready to get in, imo

/ok-to-test

marquiz (approved):
Thanks @ozhuraki, looks good to me
/assign @ArangoGutierrez

marquiz:
@ozhuraki please rebase

ozhuraki:
@marquiz
Rebased, please take a look

marquiz (commented):
Thanks @ozhuraki 
/assign @TessaIO

deployment/helm/node-feature-discovery/templates/worker.yaml
```diff
@@ -43,6 +43,17 @@ spec:
...
+          grpc:
+            port: 8082
+          initialDelaySeconds: 10
+          periodSeconds: 10
+        readinessProbe:
+          grpc:
+            port: 8082
+          initialDelaySeconds: 5
+          periodSeconds: 10
+          failureThreshold: 10
```
fmuyassarov, line 56: since you added here, probably we could it in https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/deployment/helm/node-feature-discovery/values.yaml as well but have it commented out for the sake of extra documentation. Similar to https://github.com/kubernetes-sigs/node-feature-discovery/blob/7df0f17f6838183f67f5f6b662fb8715e598a7de/deployment/helm/node-feature-discovery/values.yaml#L97

fmuyassarov (commented):
Looks good to me. A single nit comment in line.

deployment/helm/node-feature-discovery/templates/worker.yaml
marquiz: @fmuyassarov what do you suggest? Add all `livenessProbe` and `readinessProbe` settings in values.yaml? I don't think this makes sense as NFD meaningfully only supports this one type of health probe.

If we want to add configurability then I think it would be best to do in a separate PR, enabling nfd-master and nfd-worker in a similar way.

fmuyassarov: > @fmuyassarov what do you suggest? Add all `livenessProbe` and `readinessProbe` settings in values.yaml? I don't think this makes sense as NFD meaningfully only supports this one type of health probe.

By one type, I guess you meant gRPC type probe. My point was that to add whatever is here as commented out (since we don't expect probes to be configured by users) in values.yaml for the sake of extra documentation. But it is not a big deal and can be exempt from adding.

ozhuraki: @fmuyassarov
Added with a comment, please take a look

fmuyassarov (commented):
/lgtm

PR #1612, opened 2024-03-13
nfd-master: mark the -crd-controller flag as deprecated

marquiz:
Plan the removal of the -crd-controller flag along with the gRPC API. This flag does not make much sense after that as all communication with nfd-worker is based on CRDs - with the CRD controller disabled nfd-master is virtually a functionless stub.


ArangoGutierrez (approved):
/lgtm

Issue #1614, opened 2024-03-14
NFD Master memory leak


tatodorov:
**What happened**:
Node Feature Discovery Master continuously utilizes more memory and never release it.
I am running NVIDIA GPU Operator 23.9.2 and Node Feature Discovery 0.14.4 on Kubernetes 1.24.6.
Over a period of 1 hour I observe how the NFD master reaches 2 GB of memory.
I had to set a memory limit, since couple of times it exhausted the entire memory of the host.
Also, I configured NFD GC to run garbage collection every minute.
However, this didn't lead to release of memory.
Currently, I removed the GPU Operator, NFD Workers and NFD GC.
The memory usage of NFD Master keeps increasing.
Every minute, I can see the following in NFD Master's log:
```bash
I0314 12:21:30.594297       1 nfd-master.go:280] "reloading configuration"
I0314 12:21:30.594616       1 nfd-master.go:1214] "configuration file parsed" path="/etc/kubernetes/node-feature-discovery/nfd-master.conf"
I0314 12:21:30.594887       1 nfd-master.go:1274] "configuration successfully updated" configuration=<
        DenyLabelNs: {}
        EnableTaints: false
        ExtraLabelNs:
          nvidia.com: {}
        Klog: {}
        LabelWhiteList: {}
        LeaderElection:
          LeaseDuration:
            Duration: 15000000000
          RenewDeadline:
            Duration: 10000000000
          RetryPeriod:
            Duration: 2000000000
        NfdApiParallelism: 10
        NoPublish: false
        ResourceLabels: {}
        ResyncPeriod:
          Duration: 3600000000000
 >
I0314 12:21:30.594905       1 nfd-master.go:287] "stopping the nfd api controller"
I0314 12:21:30.594915       1 nfd-master.go:1338] "starting the nfd api controller"
I0314 12:21:30.595201       1 node-updater-pool.go:106] "stopping the NFD master node updater pool"
I0314 12:21:30.596167       1 node-updater-pool.go:81] "starting the NFD master node updater pool" parallelism=10
I0314 12:21:31.931613       1 nfd-master.go:694] "will process all nodes in the cluster"
```

This is the content of the ConfigMap mounted to NFD Master:
```bash
data:
  nfd-master.conf: |-
    extraLabelNs:
    - nvidia.com
```

**What you expected to happen**:
NFD Master to have a steady memory usage.

**How to reproduce it (as minimally and precisely as possible)**:

[...]

PR #1615, opened 2024-03-14
nfd-master: fix memory leak in nfd api-controller

marquiz:
Fixes a memory leak that happened when stopping (and then re-starting) the nfd api controller. The stop channel was not used properly which caused the underlying informer to keep on running.


marquiz:
Refs #1614 

/assign @ArangoGutierrez 
/cherry-pick release-0.15
/cherry-pick release-0.14

ArangoGutierrez (approved):
/lgtm

PR #1620, opened 2024-03-14
Use close to signal stop channedl in worker and topology-updater

ArangoGutierrez:
The stop channel was not being closed properly


marquiz (approved):
Thanks @ArangoGutierrez. Very good simplification (and robustification) of the code 👍
/lgtm

marquiz:
/retitle Use close to signal stop channedl in worker and topology-updater

PR #1623, opened 2024-03-14
Add FeatureGate framework to handle new features

ArangoGutierrez:


pkg/utils/featuregate.go
```diff
@@ -0,0 +1,349 @@
...
+
+const (
+	flagName = "feature-gates"
+)
+
+var (
+	NFDMutableFeatureGate MutableFeatureGate = NewFeatureGate()
+
+	// NFDFeatureGate is a shared global FeatureGate.
+	// Top-level commands/options setup that needs to modify this feature gate should use DefaultMutableFeatureGate.
```
marquiz, line 43: DefaultMutableFeatureGate??

pkg/utils/featuregate.go
```diff
@@ -0,0 +1,349 @@
...
+	NFDFeatureGate FeatureGate = NFDMutableFeatureGate
```
marquiz, line 44: Would the `pkg/features` package be a better place for these global variables.WDYT? I feel the util package is for re-usable functions.

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -10,7 +10,8 @@ nameOverride: ""
 fullnameOverride: ""
 namespaceOverride: ""
 
-enableNodeFeatureApi: true
```
marquiz, line 13: I think we don't necessarily need to remove this. We could do a favor (and break some existing users) by adding a small snippet in `deployment/helm/node-feature-discovery/templates/_helpers.tpl` that sets `featureGates.NodeFeatureApi` based on this.

We should change the default, though


```
enableNodeFeatureApi: null
```

and only act if `enableNodeFeatureApi` is non-null.

WDYT?

marquiz (commented):
Really cool @ArangoGutierrez. Some comments below.

Also, I think it would be good to document the original source of the featuregate code, e.g. in the commit message.

docs/reference/feature-gates.md
marquiz, line 1: One more suggestion: could you add a table similar to https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/#feature-gates-for-alpha-or-beta-features?

pkg/utils/featuregate.go
```diff
@@ -0,0 +1,349 @@
...
+import (
+	"flag"
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"sync"
+	"sync/atomic"
+
+	"github.com/spf13/pflag"
```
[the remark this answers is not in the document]
ArangoGutierrez: good catch, it was a missing thing

scripts/test-infra/logcheck.conf
```diff
@@ -1 +1,2 @@
 -structured sigs.k8s.io/node-feature-discovery/pkg/utils/grpc_log.go
+-structured sigs.k8s.io/node-feature-discovery/pkg/utils/featuregate/featuregate.go
```
TessaIO, line 2: Nice and clean work @ArangoGutierrez! Small nit here: can you add EOF?

deployment/helm/node-feature-discovery/templates/_helpers.tpl
marquiz, line 1: Sorry, I don't think this works (especially when more feature gates are added). My previous comment on this was bad. I somehow thought that you could modify "values.yaml" with the helpers.tpl (i.e. set the value of `featureGates.NodeFeatureApi` from the tpl) which is not the case, of course. You'd better just revert the Helm chart as it was in the previous version. There will be a slight inconvenience for the existing users (if somebody still has NodeFeatureAPI disabled) but I think there's not nice way around that (and that stuff is deprecated anyway)

marquiz (commented):
Looks cool @ArangoGutierrez. Some more feedback but we'll get there hopefully after this round

pkg/utils/featuregate/featuregate.go
```diff
@@ -0,0 +1,339 @@
...
+		if !ok {
+			return fmt.Errorf("unrecognized feature gate: %s", k)
+		}
+		if featureSpec.LockToDefault && featureSpec.Default != v {
+			return fmt.Errorf("cannot set feature gate %v to %v, feature is locked to %v", k, v, featureSpec.Default)
+		}
+		enabled[k] = v
+
+		if featureSpec.PreRelease == Deprecated {
+			klog.Warningf("Setting deprecated feature gate %s=%t. It will be removed in a future release.", k, v)
```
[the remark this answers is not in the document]
marquiz: I see no change here. Plz also change the other Warningf's

deployment/helm/node-feature-discovery/templates/master.yaml
```diff
@@ -111,6 +110,10 @@ spec:
             - "-key-file=/etc/kubernetes/node-feature-discovery/certs/tls.key"
             - "-cert-file=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
             {{- end }}
+            # Go over featureGate and add the feature-gate flag
+            {{- range $key, $value := .Values.featureGate }}
+            - "-feature-gate={{ $key }}={{ $value }}"
```
marquiz, line 115:
```suggestion
            - "-feature-gates={{ $key }}={{ $value }}"
```

deployment/helm/node-feature-discovery/templates/worker.yaml
```diff
@@ -61,14 +61,17 @@ spec:
...
-        {{- end }}
+{{- end }}
 {{- if .Values.tls.enable }}
         - "-ca-file=/etc/kubernetes/node-feature-discovery/certs/ca.crt"
         - "-key-file=/etc/kubernetes/node-feature-discovery/certs/tls.key"
         - "-cert-file=/etc/kubernetes/node-feature-discovery/certs/tls.crt"
+{{- end }}
+# Go over featureGate and add the feature-gate flag
+{{- range $key, $value := .Values.featureGate }}
+        - "-feature-gate={{ $key }}={{ $value }}"
```
marquiz, line 74:
```suggestion
        - "-feature-gates={{ $key }}={{ $value }}"
```

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -10,7 +10,8 @@ nameOverride: ""
...
+featureGate:
```
marquiz: Oh, one more suggestion:

```suggestion
featureGates:
```

TessaIO (approved):
Really like that, thanks @ArangoGutierrez!

marquiz (approved):
I like this, too @ArangoGutierrez. Finally some order here instead of the `-enable-X` / `-disable-y` flag mess.
/lgtm

docs/deployment/helm.md
```diff
@@ -98,7 +98,7 @@ Chart parameters are available.
...
 | `tls.certManager.certManagerCertificate.issuerKind` | string |           | Specifies on what kind of issuer is used, can be either ClusterIssuer or Issuer (default).  Requires `tls.certManager.certManagerCertificate.issuerName` to be set.  **NOTE**: this parameter is related to the deprecated gRPC API and will be removed with it in a future release |
-| `enableNodeFeatureApi`| bool   | true                                    | Enable the [NodeFeature](../usage/custom-resources.md#nodefeature) CRD API for communicating node features. This will automatically disable the gRPC communication. **NOTE**: this parameter is related to the deprecated gRPC API and will be removed with it in a future release    |
+| `featureGate.NodeFeatureAPI`| bool   | true                                    | Enable the [NodeFeature](../usage/custom-resources.md#nodefeature) CRD API for communicating node features. This will automatically disable the gRPC communication. **NOTE**: this parameter is related to the deprecated gRPC API and will be removed with it in a future release    |
```
marquiz, line 101:
```suggestion
| `featureGates.NodeFeatureAPI`| bool   | true                                    | Enable the [NodeFeature](../usage/custom-resources.md#nodefeature) CRD API for communicating node features. This will automatically disable the gRPC communication. **NOTE**: this parameter is related to the deprecated gRPC API and will be removed with it in a future release    |
```

marquiz (changes requested):

marquiz (approved):
Now 🚀
/lgtm

PR #1631, opened 2024-03-16
chore/deployment: add resources requests and limits for helm and Kustomize

### LINKED ISSUE #1616: Specify resource limits (and requests) in NFD deployment files

Refs #1614 

We should specify resource requests (cpu and memory requests and limits) in the NFD deployment files. I.e. both kustomize and Helm chart.

**Why is this needed**:

Best practices.

TessaIO:
Resolves #1616 

_**Note: for the moment all of the resource requests and limits for all components are the same (I don't have full visibility on what is needed especially in prod environments). I'd be happy if you want me to change them.**_


docs/deployment/helm.md
```diff
@@ -132,7 +132,7 @@ API's you need to install the prometheus operator in your cluster.
 | `master.service.type`      | string  | ClusterIP                               | NFD master service type. **NOTE**: this parameter is related to the deprecated gRPC API and will be removed with it in a future release |
 | `master.service.port`      | integer | 8080                                    | NFD master service port. **NOTE**: this parameter is related to the deprecated gRPC API and will be removed with it in a future release |
 | `master.resources`         | dict    | {}                                      | NFD master pod [resources management](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/) |
-| `master.nodeSelector`      | dict    | {}                                      | NFD master pod [node selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) |
+| `master.nodeSelector`      | dict    | {limits: {cpu: 300m, memory: 1Gi}, requests: {cpu: 100m, memory: 128Mi}} | NFD master pod [node selector](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector) |
```
fmuyassarov, line 135: perhaps this should be `master.resources` ?

fmuyassarov: nit: maybe it is just a personal taste, but it is seems it is easier to read if you could have two columns for `x.resources.limits` and `x.resources.requests` instead of merging them into a single column.

TessaIO: > perhaps this should be `master.resources` ?

akhh forgot to check this, yes it should be resources

TessaIO: > nit: maybe it is just a personal taste, but it is seems it is easier to read if you could have two columns for `x.resources.limits` and `x.resources.requests` instead of merging them into a single column.

Yes I think it should be better, thanks.

marquiz (commented):
Thanks @TessaIO for the PR!

Below some comments about the resource requests/limits. Let's try to base them on some rational basis.

deployment/base/master/master-deployment.yaml
```diff
@@ -21,6 +21,13 @@ spec:
         - name: nfd-master
           image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
           imagePullPolicy: Always
+          resources:
+            limits:
+              cpu: 300m
+              memory: 1Gi
+            requests:
+              cpu: 100m
+              memory: 128Mi
```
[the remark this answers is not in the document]
PiotrProkop: How `top` calculates memory? I've just checked my prod's cluster Grafana with 558 NodeFeatures objects and `nfd-master` uses 156MiB. I calculate memory usage as `sum(container_memory_working_set_bytes)`. So even if memory usage grows linearly with the number of nodes in the cluster it should be 320MiB for 1000 nodes.

deployment/base/topologyupdater-daemonset/topologyupdater-daemonset.yaml
```diff
@@ -22,6 +22,13 @@ spec:
           command:
             - "nfd-topology-updater"
           args: []
+          resources:
+            limits:
+              cpu: 300m
+              memory: 1Gi
+            requests:
+              cpu: 100m
+              memory: 128Mi
```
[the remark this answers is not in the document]
PiotrProkop: We use:
```
          resources:
            requests:
              cpu: 50m
              memory: 40Mi
            limits:
              cpu: 100m
              memory: 60Mi
```

deployment/base/worker-daemonset/worker-daemonset.yaml
```diff
@@ -21,6 +21,13 @@ spec:
           imagePullPolicy: Always
           command:
             - "nfd-worker"
+          resources:
+            limits:
+              cpu: 300m
+              memory: 1Gi
+            requests:
+              cpu: 100m
+              memory: 128Mi
```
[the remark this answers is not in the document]
PiotrProkop: I agree we haven't seen a major memory consumption from `nfd-worker` process, from my experience it's memory usage doesn't exceeds `20Mi`

marquiz (approved):
Looks good to me.

PTAL @PiotrProkop

deployment/base/master/master-deployment.yaml
```diff
@@ -21,6 +21,13 @@ spec:
         - name: nfd-master
           image: gcr.io/k8s-staging-nfd/node-feature-discovery:master
           imagePullPolicy: Always
+          resources:
+            limits:
+              cpu: 300m
+              memory: 4Gi
+            requests:
+              cpu: 100m
+              memory: 128Mi
```
cmontemuino, line 30: Setting `limits.memory: 4Gi` means you're ok with an OOM Kill at that number. Hence, the expectation is for the master pod to requesting more memory.
Setting a lower `requests.memory` is not helping at all. Let's say you get an OOM Kill, then the scheduler spawns a new pod in a node with enough memory according to the requests, but not enough for the limits. This is perfectly fine for the scheduler, but your pod will die again.
Because of the above and because memory cannot be returned to the pool (after GC for example), this is why the recommendation is to set limits equal to requests for memory. You only help the scheduler this way.

marquiz: Thanks @cmontemuino for the feedback. I think one reason to set limits is to really oom kill the pod if it get's crazy (with memory leaks like #1614). It might be better to set it to 8Gi than 4Gi, though. The memory consumption very heavily depends on the number of nodes in the cluster (on the ballpark of ~1.5MB per node or so, it seems). We probablty want the default deployment to work in a wide variety of clusters so we cannot set requests to 4Gi or 8Gi as it's total waste (or cannot be satisfied) in smaller clusters and lowering limits to 128Mi will cause problems in bigger clusters. Thoughts?

PiotrProkop: In my clusters I  am setting limits 25% above actual usage. As I prefer to be alerted about OOMs in my control plane quicker than later so It's easier to triage issues wrt to resource starving but I understand that other can disagree with this approach 😄

cmontemuino: @marquiz -- I'm **not** suggesting to left limits unset. What I'm saying is to use the same value for **both** `resources.limits.memory` and `resources.requests.memory`.
Having them with different values won't help because memory is not an "elastic" resource.

marquiz: @cmontemuino as I explained I don't think we can do that. We don't want to set requests=limits=4Gi as that will hurt/block small clusters and similarly we don't want to set requests=limits=128Mi (or smth on that range) because we'll get oom'ed in bigger clusters.

I think we should document the resource requests/limits better, though.

TessaIO: > I think we should document the resource requests/limits better, though.

Any suggestions?

cmontemuino: > @cmontemuino as I explained I don't think we can do that. We don't want to set requests=limits=4Gi as that will hurt/block small clusters and similarly we don't want to set requests=limits=128Mi (or smth on that range) because we'll get oom'ed in bigger clusters.
> 
> I think we should document the resource requests/limits better, though.

@marquiz -- got it 👍 

Thus it becomes a documentation issue. It must be super clear what implications are there when having different limits and requests for memory. Otherwise OOM Kill will start happening for smaller clusters; and likely non-availability for clusters with memory-pressure scenarios.

Would you like me to try documenting it in a PR?

marquiz: > Would you like me to try documenting it in a PR?

@cmontemuino please do, thanks for volunteering, really appreciated 👍

ArangoGutierrez:
I would like the LGTM to come from @PiotrProkop  on this PR

PR #1633, opened 2024-03-19
apis/nfd/validate: loosen validation of feature annotations

marquiz:
Don't require that the annotation value must conform to the (strict)
requirements of label values. In the Kubernetes API annotation values do
not have other restrictions than that the total size (keys and values)
of _all_ annotations combined of an object must not exceed 256kB.

This patch sets a maximum size limit of 1kB for the value of a single
feature annotation created by NFD. This limit is rather arbitrary but
should be enough for the NFD usage scenarios (until proven wrong).

Also add license header to the test.go file and fix one bug in
MatchFeature validation.

### DIFF
--- a/pkg/apis/nfd/validate/validate.go
+++ b/pkg/apis/nfd/validate/validate.go
@@ -57,7 +57,7 @@ func MatchFeatures(matchFeature nfdv1alpha1.FeatureMatcher) []error {
 	var validationErr []error
 
 	for _, match := range matchFeature {
-		nameSplit := strings.SplitN(match.Feature, ".", 2)
+		nameSplit := strings.Split(match.Feature, ".")
 		if len(nameSplit) != 2 {
 			validationErr = append(validationErr, fmt.Errorf("invalid feature name %v (not <domain>.<feature>), cannot be used for templating", match.Feature))
 		}
--- a/pkg/apis/nfd/validate/validate.go
+++ b/pkg/apis/nfd/validate/validate.go
@@ -155,8 +155,8 @@ func Annotation(key, value string) error {
 	}
 
 	// Validate annotation value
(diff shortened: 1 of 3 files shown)


pkg/apis/nfd/validate/validate_test.go
```diff
@@ -1,10 +1,27 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
ArangoGutierrez, line 2: 24?

ArangoGutierrez (changes requested):
Just a small question

pkg/apis/nfd/validate/validate_test.go
marquiz: The file was created in 2023 (by you, I think 😅) but the license header was missing. So I put the year of the original contribution here. Makes sense?

marquiz:
Rebased. @ArangoGutierrez PTAL

marquiz:
ping @ArangoGutierrez

pkg/apis/nfd/validate/validate.go
```diff
@@ -79,6 +80,13 @@ func Template(labelsTemplate string) []error {
 	return validationErr
 }
 
+func sortErrors(errs []error) []error {
```
PiotrProkop, line 83: shouldn't we just invoke this func in tests? what's the value in sorting errors by char len?

marquiz: It doesn't sort by length but alphabetically.

Sorting can be done in tests, yes. I moved it there.

marquiz:
PR updated. @PiotrProkop PTAL

marquiz:
Thanks @PiotrProkop for the review!

PR #1634, opened 2024-03-20
topology-updater: Set APIVersion, Kind in the OwnerReference explicitly

ozhuraki:
APIVersion and Kind are empty in the returned namespace object and need to be set explicitly.


marquiz (approved):
Thanks @ozhuraki for the fix.

Looks like the typemeta fields are empty when getting objects.
/ok-to-test
/lgtm

/cc @PiotrProkop

PR #1636, opened 2024-03-21
test/e2e: Increase the timeouts to get tests to pass reliably

ozhuraki:
A few tests run unreliably due to timeouts that became to small:

- Increase the timeout in eventuallyNonControlPlaneNodes() to 1 minute
- Increase the timeout in GetNodeTopology() to 2 minutes

### DIFF
--- a/test/e2e/gomega.go
+++ b/test/e2e/gomega.go
@@ -39,7 +39,7 @@ type k8sAnnotations map[string]string
 func eventuallyNonControlPlaneNodes(ctx context.Context, cli clientset.Interface) AsyncAssertion {
 	return Eventually(func(g Gomega, ctx context.Context) ([]corev1.Node, error) {
 		return getNonControlPlaneNodes(ctx, cli)
-	}).WithPolling(1 * time.Second).WithTimeout(10 * time.Second).WithContext(ctx)
+	}).WithPolling(1 * time.Second).WithTimeout(60 * time.Second).WithContext(ctx)
 }
 
 // MatchLabels returns a specialized Gomega matcher for checking if a list of
--- a/test/e2e/topology_updater_test.go
+++ b/test/e2e/topology_updater_test.go
@@ -421,7 +421,7 @@ excludeList:
 					}
 				}
 				return memoryFound
-			}, 1*time.Minute, 10*time.Second).Should(BeFalse())
+			}, 2*time.Minute, 10*time.Second).Should(BeFalse())
 		})
 	})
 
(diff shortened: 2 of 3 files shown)


test/e2e/gomega.go
```diff
@@ -39,7 +39,7 @@ type k8sAnnotations map[string]string
 func eventuallyNonControlPlaneNodes(ctx context.Context, cli clientset.Interface) AsyncAssertion {
 	return Eventually(func(g Gomega, ctx context.Context) ([]corev1.Node, error) {
 		return getNonControlPlaneNodes(ctx, cli)
-	}).WithPolling(1 * time.Second).WithTimeout(10 * time.Second).WithContext(ctx)
+	}).WithPolling(1 * time.Second).WithTimeout(60 * time.Second).WithContext(ctx)
```
marquiz, line 42: What kind of environment are you running the tests? Are you seeing all the tests failing or just some specific one? The reason I'm asking is that it really shouldn't take nfd(-master) a minute to update the nodes. Or is it some very big cluster you're having. It would be better to first try to investigate why it's taking so long.

ozhuraki: @marquiz 
> What kind of environment are you running the tests?
Just plain kubernetes on bare metal, 1 node.
> Are you seeing all the tests failing or just some specific one?
A 2-3 specific ones fail occasionaly on these timeouts. 10 seconds here seems to be too little.

Actually seeing the same in the one that is failing in e2e dashboard: "[FAILED] [10.485 seconds]".

marquiz: > Just plain kubernetes on bare metal, 1 node.

We need to investigate that.



> Actually seeing the same in the one that is failing in e2e dashboard: "[FAILED] [10.485 seconds]".

That error ([here](https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/postsubmit-node-feature-discovery-e2e-test/1770698579086675968)) has nothing to do with the 10s timeout. The test failed bc for some reason nfd-master pod failed (it just happened to take ~10sec for the test until it failed)

ArangoGutierrez: Agree with Marquiz, NFD tests should all run under 10s, instead of increasing the timeout, it indicates a bug or something we need to fix

ArangoGutierrez:
@ozhuraki after the last couple of weeks (with new changes getting into NFD) does the test still fails in your environment? 
I think this PR should be turned into a BUG repo / Issue, given that we agree on not increasing the test timeouts

ozhuraki:
@ArangoGutierrez
> after the last couple of weeks (with new changes getting into NFD) does the test still fails in your environment?

Occasionally yes. It might be that my environment is slower.

> I think this PR should be turned into a BUG repo / Issue, given that we agree on not increasing the test timeouts

Yes, let's do so, closing it now. I will still check, if needed, I will create an issue.

PR #1639, opened 2024-03-22
chore/deploy: make interval property in PodMonitor configurable

### LINKED ISSUE #1638: Make `interval` property in PodMonitor configurable

To [interval value in PodMonitor's `spec.podMetricsEndpoints`](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/deployment/helm/node-feature-discovery/templates/prometheus.yaml#L15) configurable from `values.yaml`.

**Why is this needed**:

Being able to fine tune the interval. Current value might be too aggressive or not, depending on the cluster, but also specific needs for monitoring.

This should be a very simple feature to implement, with no impact in current installations.

**I'm open to contribute if required**.

TessaIO:
Resolves #1638

### DIFF
--- a/deployment/helm/node-feature-discovery/values.yaml
+++ b/deployment/helm/node-feature-discovery/values.yaml
@@ -537,4 +537,5 @@ tls:
 
 prometheus:
   enable: false
+  scrapeInterval: 10s
   labels: {}
--- a/deployment/helm/node-feature-discovery/templates/prometheus.yaml
+++ b/deployment/helm/node-feature-discovery/templates/prometheus.yaml
@@ -12,7 +12,7 @@ metadata:
 spec:
   podMetricsEndpoints:
     - honorLabels: true
-      interval: 10s
+      interval: {{ .Values.prometheus.scrapeInterval }}
       path: /metrics
       port: metrics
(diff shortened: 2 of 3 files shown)


deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -537,4 +537,5 @@ tls:
 
 prometheus:
   enable: false
+  interval: 10s
```
marquiz, line 540: I'd suggest a small clarification on what interval are we configuring

```suggestion
  scrapeInterval: 10s
```

An alternative could be making deepening the hierarchy like `prometheus.podmonitor.interval` but I don't like that much, especially as we don't have other podmonitor settings

marquiz (commented):
Thx @TessaIO. Small suggestion regarding naming.

We also need to document the new setting in `docs/deployment/helm.md`

/cc @cmontemuino

marquiz (approved):
Thanks @TessaIO, looks good to me, now

TessaIO:
ping @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

Issue #1640, opened 2024-03-27
Node Feature Discovery Master random crashes


tatodorov:
**What happened**:
Ocasionally, NFD Master crashes with the following error:
```bash
fatal error: concurrent map read and map write
```
Followed by a go routine dump.

<details>
  <summary>Go routine dump</summary>

```bash
goroutine 358442 [running]:
k8s.io/apimachinery/pkg/runtime.(*Scheme).ObjectKinds(0xc0004439d0, {0x24401f8?, 0xc003598c60})
        /go/pkg/mod/k8s.io/apimachinery@v0.28.4/pkg/runtime/scheme.go:267 +0x1ca
k8s.io/apimachinery/pkg/runtime.(*parameterCodec).EncodeParameters(0xc0003f8900, {0x24401f8, 0xc003598c60}, {{0x2148806, 0xf}, {0x213da3e, 0x8}})
        /go/pkg/mod/k8s.io/apimachinery@v0.28.4/pkg/runtime/codec.go:190 +0x6e
k8s.io/client-go/rest.(*Request).SpecificallyVersionedParams(0xc000afa6c0, {0x24401f8?, 0xc003598c60?}, {0x2440450?, 0xc0003f8900?}, {{0x2148806?, 0x11?}, {0x213da3e?, 0xc0016559a0?}})
        /go/pkg/mod/k8s.io/client-go@v0.28.4/rest/request.go:377 +0x83
k8s.io/client-go/rest.(*Request).VersionedParams(...)
        /go/pkg/mod/k8s.io/client-go@v0.28.4/rest/request.go:370
sigs.k8s.io/node-feature-discovery/pkg/generated/clientset/versioned/typed/nfd/v1alpha1.(*nodeFeatures).List(0xc0003cef00, {0x2453fe8, 0xc00012e008}, {{{0x0, 0x0}, {0x0, 0x0}}, {0x0, 0x0}, {0x0, ...}, ...})
        /go/node-feature-discovery/pkg/generated/clientset/versioned/typed/nfd/v1alpha1/nodefeature.go:89 +0x145
sigs.k8s.io/node-feature-discovery/pkg/generated/informers/externalversions/nfd/v1alpha1.NewFilteredNodeFeatureInformer.func1({{{0x0, 0x0}, {0x0, 0x0}}, {0x0, 0x0}, {0x0, 0x0}, 0x0, 0x0, ...})
[...]

PR #1641, opened 2024-03-27
nfd-master: do nfd API scheme registration in an init function

marquiz:
Prevents (rare) races on nfd-master reconfigurartion. Previously the scheme was registered at nfd API controller creation/startup time. This caused a race with some lister/informer goroutines of the previous (stoppped) controller still running and accessing (reading) the sceme while we were updating (writing) it.


marquiz:
This should fix #1640

/assign @ArangoGutierrez 
/cc @tatodorov

PiotrProkop (approved):
/lgtm

PR #1642, opened 2024-03-27
nfd-master: protect node updater pool queueing with a lock

marquiz:
Prevents races when (re-)starting the queue. There are no reports on issues related to this (and I haven't come up with any actual failure path in the current code) but better to be safe and follow the best practices.


ArangoGutierrez (approved):
/lgtm

PR #1652, opened 2024-04-03
nfd-master: get node object only once when updating node

marquiz:
Prevent excess queries of node objects from the Kubernetes apiserver. This significantly speeds up node updates (and reduces the load on the apiserver) as the client-side throttling (which is good) does not bite us that hard.

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -503,7 +503,7 @@ func (m *nfdMaster) prune() error {
 		klog.InfoS("pruning node...", "nodeName", node.Name)
 
 		// Prune labels and extended resources
-		err := m.updateNodeObject(node.Name, Labels{}, Annotations{}, ExtendedResources{}, []corev1.Taint{})
+		err := m.updateNodeObject(&node, Labels{}, Annotations{}, ExtendedResources{}, []corev1.Taint{})
 		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -905,11 +905,11 @@ func (m *nfdMaster) setTaints(taints []corev1.Taint, nodeName string) error {
 	}
 
 	if taintsUpdated {
-		err = controller.PatchNodeTaints(context.TODO(), m.k8sClient, nodeName, node, newNode)
+		err := controller.PatchNodeTaints(context.TODO(), m.k8sClient, node.Name, node, newNode)
```
TessaIO, line 908: nit
```suggestion
		err = controller.PatchNodeTaints(context.TODO(), m.k8sClient, node.Name, node, newNode)
```

marquiz: It didn't really matter. I prefer to use the narrowest scope for variables, though. Changed this (and another similar place) to use the `if err := ...` construct

@TessaIO PTAL

TessaIO: Yes, I got it, I thought it would be better to be consistent.

TessaIO:
Thanks for this improvement @marquiz 
/lgtm

PR #1653, opened 2024-04-04
nfd-master: use separate k8s api clients for each updater

marquiz:
Sharing the same client between updater threads virtually serializes access, in practice making the effective parallelism close to 1.

With this patch, in my bench cluster of 300 nodes, the time taken by updating all nodes drops from ~2 minutes to ~12 seconds (with the default parallelism of 10 node updater threads). This demonstrates the 10-fold increased parallelism from ~1 to 10.

There might be other solutions that could be explored, e.g. caching nodes with an indexer/lister but otoh nfd doesn't necessarily need/want to watch every little change in each node. We only need to get the node when something in our own CRDs change (we don't react to any changes in the node object itself). Using multiple clients was the most obvious choice to solve the problem for now.

### DIFF
--- a/pkg/nfd-master/node-updater-pool.go
+++ b/pkg/nfd-master/node-updater-pool.go
@@ -22,6 +22,7 @@ import (
 
 	"golang.org/x/time/rate"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	k8sclient "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 )
@@ -41,7 +42,7 @@ func newNodeUpdaterPool(nfdMaster *nfdMaster) *nodeUpdaterPool {
 	}
 }
 
-func (u *nodeUpdaterPool) processNodeUpdateRequest(queue workqueue.RateLimitingInterface) bool {
+func (u *nodeUpdaterPool) processNodeUpdateRequest(cli k8sclient.Interface, queue workqueue.RateLimitingInterface) bool {
 	n, quit := queue.Get()
 	if quit {
 		return false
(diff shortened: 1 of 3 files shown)


marquiz:
Depends on #1652 
/hold

/assign @ArangoGutierrez @TessaIO

marquiz:
Depends on #1655

marquiz:
#1655 merged
/unhold

Should not need rebase 🤨

pkg/nfd-master/node-updater-pool.go
```diff
@@ -71,7 +72,13 @@ func (u *nodeUpdaterPool) processNodeUpdateRequest(queue workqueue.RateLimitingI
 }
 
 func (u *nodeUpdaterPool) runNodeUpdater(queue workqueue.RateLimitingInterface) {
-	for u.processNodeUpdateRequest(queue) {
+	// For tests, re-use the api client from nfd-master
+	cli := u.nfdMaster.k8sClient
+	if u.nfdMaster.kubeconfig != nil {
```
ArangoGutierrez, line 77: I am thinking, should we do the if first and defer to an `else` in case `u.nfdMaster.kubeconfig == nil`

ArangoGutierrez (changes requested):
Just one nit/question

pkg/nfd-master/node-updater-pool.go
marquiz: This was a short version to initialize the `cli` variable. An alternative would be like:

```golang
	var cli k8sclient.Interface
	if u.nfdMaster.kubeconfig != nil {
		// For normal execution, initialize a separate api client for each updater
		cli = k8sclient.NewForConfigOrDie(u.nfdMaster.kubeconfig)
	} else {
		// For tests, re-use the api client from nfd-master
		cli := u.nfdMaster.k8sClient
	}
```

I'm ok with either

ArangoGutierrez: I like it, not a blocker

ArangoGutierrez (approved):

pkg/nfd-master/node-updater-pool.go
marquiz: I changed the code to what I suggested

ArangoGutierrez (approved):
/lgtm

PR #1654, opened 2024-04-04
Support KUBERNETES_MASTER env variable

marquiz:
Support (insecure) connections to the Kubernetes API server by setting the KUBERNETES_MASTER environment variable. This is how e.g. kubectl behaves and also matches the error message(s) from client-go package. This patch is targeted for local testing and development.

### DIFF
--- a/pkg/utils/kubernetes.go
+++ b/pkg/utils/kubernetes.go
@@ -49,8 +49,5 @@ func GetKubernetesNamespace() string {
 
 // GetKubeconfig returns the kubeconfig for the cluster
 func GetKubeconfig(path string) (*restclient.Config, error) {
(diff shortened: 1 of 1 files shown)


ArangoGutierrez (changes requested):
If the patch is " This patch is targeted for local testing and development." 
Let's document this behaviour

marquiz:
> If the patch is " This patch is targeted for local testing and development."
> Let's document this behaviour

Hmm, I start to dislike this approach myself 🙄 This was basically added to make the unit tests in #1655 to pass. I try to think another approach to achieve this 
/hold

marquiz:
After sleeping over this 👎
/close

PR #1655, opened 2024-04-04
nfd-master: parse kubeconfig even with NoPublish set

marquiz:
Don't try to be too smart when kubeconfig is needed. In practice, the nfd-master really doesn't work anymore (with the NodeFeature API enabled) without a kubeconfig set. This patch fixes crashes happening when NoPublish is enabled, e.g. in listing all nodes in the nfd api handler and in getting single node objects in the node updater pool.


marquiz:
Depends on #1664
/hold

marquiz:
Now depends on #1658

ArangoGutierrez (approved):
/lgtm

marquiz:
#1658 was merged
/unhold

PR #1656, opened 2024-04-05
Tidy up usage of channels for signaling

marquiz:
This started as a small effort to simplify the usage of "ready" channel in nfd-master. It extended into a wider simplification/unification of the channel usage.

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -149,7 +149,7 @@ type nfdMaster struct {
 	server          *grpc.Server
 	healthServer    *grpc.Server
 	stop            chan struct{}
-	ready           chan bool
+	ready           chan struct{}
 	k8sClient       k8sclient.Interface
 	nodeUpdaterPool *nodeUpdaterPool
 	deniedNs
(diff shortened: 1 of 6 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -475,11 +474,8 @@ func (m *nfdMaster) Stop() {
 // Wait until NfdMaster is able able to accept connections.
 func (m *nfdMaster) WaitForReady(timeout time.Duration) bool {
 	select {
-	case ready, ok := <-m.ready:
-		// Ready if the flag is true or the channel has been closed
-		if ready || !ok {
-			return true
-		}
+	case <-m.ready:
+		return true
```
ArangoGutierrez, line 478: I think we need to re write the entire func, the last return will never be used

ArangoGutierrez (changes requested):

pkg/nfd-master/nfd-master.go
marquiz: > I think we need to re write the entire func, the last return will never be used

True, I modified the func to eliminate unreachable parts

ArangoGutierrez: pardon my ignorance, should this `select` be inside an empty `for` ?

ArangoGutierrez (changes requested):

pkg/nfd-master/nfd-master.go
marquiz: No, the select blocks until the channel is closed or the timer expoires

PR #1657, opened 2024-04-05
nfd-master: prevent crash on empty config struct

marquiz:
Change the handling of LabelWhiteList config option to use a pointer to detect when the option is unset. This doesn't fix any detected crash but is merely general improvement and stabilization, serving easier testing.

Also, use the regexp type from the core libs for the config struct - dropping the unmasrhalling code for our custom regexp type - as the core regexp now implements unmarshaller itself.


ArangoGutierrez (approved):
/lgtm

PR #1658, opened 2024-04-05
nfd-master: implement opts for modifying NfdMaster instance

marquiz:
This provides a more controlled way for setting up the NfdMaster instance for testing.


ArangoGutierrez (approved):
/lgtm

PR #1660, opened 2024-04-08
nfd-master: stop node-updater pool before reconfiguring api-controller

marquiz:
Prevents potential race between node-updater pool and the api-controller when re-configuring nfd-master. Reconfiguration causes a new api-controller instance to be created so nfd api lister might change in the midst of processing a node update (if the pool was running). No actual issues related to this have been identified but races (like this) should still be avoided.

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -350,6 +350,11 @@ func (m *nfdMaster) Run() error {
 				return err
 			}
 
+			// Stop the nodeUpdaterPool so that no node updates are underway
+			// while we reconfigure the NFD API controller (including the
+			// listers) below
+			m.nodeUpdaterPool.stop()
+
 			// restart NFD API controller
 			if m.nfdController != nil {
 				klog.InfoS("stopping the nfd api controller")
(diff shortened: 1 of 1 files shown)


TessaIO (approved):

pkg/nfd-master/nfd-master.go
```diff
@@ -350,6 +350,11 @@ func (m *nfdMaster) Run() error {
 				return err
 			}
 
+			// Stop the nodeupdater pool so that no node updates are underway
```
ArangoGutierrez, line 353:
```suggestion
			// Stop the nodeUpdaterPool so that no node updates are underway
```

pkg/nfd-master/nfd-master.go
```diff
@@ -361,13 +366,13 @@ func (m *nfdMaster) Run() error {
 					return nil
 				}
 			}
+			// Restart the node updater pool
```
ArangoGutierrez, line 369:
```suggestion
			// Restart the nodeUpdaterPool
```

ArangoGutierrez (changes requested):

marquiz:
ping @ArangoGutierrez

marquiz:
ping @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #1662, opened 2024-04-09
Fix Make generate  

ArangoGutierrez:


api/nfd/v1alpha1/types.go
```diff
@@ -112,7 +112,7 @@ type NodeFeatureRuleList struct {
 }
 
 // NodeFeatureRule resource specifies a configuration for feature-based
-// customization of node objects, such as node labeling.
+// customization of node objects, such as node labeling. LOOSEME
```
marquiz, line 115: LOOSEME?

.gitignore
```diff
@@ -3,3 +3,4 @@ demo/helper-scripts/*.pdf
 demo/helper-scripts/*.log
 /kustomization.yaml
 coverage.out
+vendor/
```
marquiz, line 6: I think it's better not to have this in gitignore as we don't ship with the vendored deps. With this change an out-of-date vendor/ directory get's ignored and easily causes unexpected build results

go.mod
```diff
@@ -3,206 +3,203 @@ module sigs.k8s.io/node-feature-discovery
 go 1.22
 
 require (
+	cloud.google.com/go/compute v1.23.3
+	cloud.google.com/go/compute/metadata v0.2.3
+	github.com/Azure/azure-sdk-for-go v68.0.0+incompatible
```
marquiz, line 8: We shouldn't depend on this stuff directly. I don't believe any changes to go.mod would be needed in this PR

hack/update_codegen.sh
```diff
@@ -18,51 +18,56 @@ set -o errexit
...
+  rm -rf ${TMP_VENDOR_DIR}
+  rm sigs.k8s.io
+}
+
+trap cleanup EXIT
 GO_CMD=${1:-go}
 NFD_ROOT=$(realpath $(dirname ${BASH_SOURCE[0]})/..)
 
-go mod vendor
+"${GO_CMD}" mod vendor
```
marquiz, line 33: Remind me: we need the `vendor/` for running `go generate`? And thus can't just do `go mod vendor -o "$TMP_VENDOR"`?

api/nfd/v1alpha1/types.go
ArangoGutierrez: ups that was my hack to check things were being updated properly

.gitignore
ArangoGutierrez: ok

hack/update_codegen.sh
ArangoGutierrez: `"${GO_CMD}" generate ./cmd/... ./pkg/... ./source/...` fails if things are not in `vendor`

marquiz (approved):
/lgtm

PR #1664, opened 2024-04-16
Add Status to NodeFeature CRD

### LINKED ISSUE #1071: Status for NodeFeature CRD

Add status field to the NodeFeature CRD. It could indicate if and when the NodeFeature was applied, any errors happened etc.

**Why is this needed**:

Better visibility to users what is going on.

ozhuraki:
Closes #1071


pkg/nfd-worker/nfd-worker.go
```diff
@@ -762,10 +762,14 @@ func (m *nfdWorker) updateNodeFeatureObject(labels Labels) error {
...
 		nfrUpdated.Spec = nfdv1alpha1.NodeFeatureSpec{
-			Features: *features,
+ 			Features: *features,
 			Labels:   labels,
 		}
-
+		currentTime := time.Now()
+		msg := fmt.Sprintf("Updated: %q", currentTime)
+		nfrUpdated.Status = nfdv1alpha1.NodeFeatureStatus{
+			Status: []string{ msg },
```
TessaIO, line 771: If we're going to store only one string, why are we storing it in an array here?

TessaIO (commented):
Thanks for your work @ozhuraki. I actually think that we can store other useful information in the worker NFR status, rather than only updated at, like number of features/labels discovered, number of sources that had failures when doing discovery. I don't have concrete use cases here but this is my opinion for now.

pkg/nfd-worker/nfd-worker.go
ozhuraki: @TessaIO

Thanks  for the input. Added:

- number of features
- number of labels discovered
- number of sources that had failures

Please take a look.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -762,10 +765,19 @@ func (m *nfdWorker) updateNodeFeatureObject(labels Labels) error {
...
 
+		currentTime := time.Now()
+		updated := fmt.Sprintf("Updated: %q", currentTime)
+		nFeatures := fmt.Sprintf("Number of feature sources: %q", len(m.featureSources))
+		nFeatureErrors := fmt.Sprintf("Number of sources that had failuers: %q", m.featureSourceErrors)
+		nLabels := fmt.Sprintf("Number of label sources: %q", len(m.labelSources))
+
+		nfrUpdated.Status = nfdv1alpha1.NodeFeatureStatus{
+			Status: []string{ updated, nFeatures, nFeatureErrors, nLabels },
+		}
```
TessaIO, line 780: I think we should put status details each in a different field and not in the same string. What I mean here is that we would have the following:
```
status:
   updatedAt: xxxx
   numberOfFeatures: xxxx
   numberOfLabels: xxxx
```
This is how it's usually done. wdyt @ozhuraki 
Thoughts on this @marquiz @ArangoGutierrez

ozhuraki: @TessaIO
> I think we should put status details each in a different field and not in the same string

Could be done this way too, let's wait for more input.

TessaIO: If I get your implementation correctly, the status in the Kubernetes object would be like this:
```
status: 123456789 2 3 5
```
It's really hard for the user to interpret this, that's why I suggested that we put them in separate, yet descriptive fields.

PiotrProkop: I agree that having a list of strings in an API is not very user friendly, how someone will refer to those fields in programmatic way? Will we sort the conditions etc.
It should be concrete fields with appropriate types, like integer for failure numbers etc.

ozhuraki:
@TessaIO @PiotrProkop

Thanks for the input. Updated, please take a look.

ozhuraki:
@PiotrProkop

Thanks for the review, updated, please take a look.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -765,7 +768,12 @@ func (m *nfdWorker) updateNodeFeatureObject(labels Labels) error {
 			Features: *features,
 			Labels:   labels,
 		}
-
+		nfrUpdated.Status = nfdv1alpha1.NodeFeatureStatus{
+			LastAppliedAt:         metav1.Time{Time: time.Now()},
```
PiotrProkop, line 772: I would probably go with UTC here.

api/nfd/v1alpha1/types.go
```diff
@@ -53,6 +54,22 @@ type NodeFeatureSpec struct {
 	Labels map[string]string `json:"labels"`
 }
 
+// Status of a NodeFeature object.
+type NodeFeatureStatus struct {
+	// Annotations to create if the rule matches.
```
PiotrProkop, line 59: The comment is slightly off.

api/nfd/v1alpha1/types.go
```diff
@@ -53,6 +54,22 @@ type NodeFeatureSpec struct {
...
+	// +optional
+	LastAppliedAt metav1.Time `json:"lastAppliedAt,omitempty"`
+	// +optional
+	// Number of features discovered.
+	NumberOfFeatures int `json:"numberOfFeatures,omitempty"`
+	// +optional
+	// Number of errors during feature discovery.
```
PiotrProkop, line 66: s/during/during last

ozhuraki:
@PiotrProkop
Thanks, updated.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -218,9 +219,11 @@ func (w *nfdWorker) startGrpcHealthServer(errChan chan<- error) error {
 // Run feature discovery.
 func (w *nfdWorker) runFeatureDiscovery() error {
 	discoveryStart := time.Now()
+	w.featureSourceErrors = 0
```
TessaIO, line 222: I think this can be more meaningful as a variable name
```suggestion
	w.numberOffeatureSourceErrors = 0
```

TessaIO:
@ozhuraki Are you planning to add some e2e tests for this feature?

ozhuraki:
@TessaIO
OK, I will try to add e2e test

ozhuraki:
@TessaIO

Thanks, added e2e tests, please take a look.

test/e2e/node_feature_discovery_test.go
```diff
@@ -500,6 +500,17 @@ var _ = NFDDescribe(Label("nfd-master"), func() {
 					}
 					eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchLabels(expectedLabels, nodes))
 
+					By("Verifying the CRD status data")
+
```
ArangoGutierrez, line 504:
```suggestion
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -500,6 +500,17 @@ var _ = NFDDescribe(Label("nfd-master"), func() {
 					}
 					eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchLabels(expectedLabels, nodes))
 
+					By("Verifying the CRD status data")
```
ArangoGutierrez, line 503:
```suggestion
					By("Verifying the CRD status")
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -500,6 +500,17 @@ var _ = NFDDescribe(Label("nfd-master"), func() {
...
+
+					nf, err := nfdClient.NfdV1alpha1().NodeFeatures(f.Namespace.Name).Get(ctx, targetNodeName, metav1.GetOptions{})
+					Expect(err).NotTo(HaveOccurred())
+
+					labelsMatch := len(expectedLabels[targetNodeName]) == nf.Status.NumberOfLabels
+
```
ArangoGutierrez, line 509:
```suggestion
```

test/e2e/node_feature_discovery_test.go
```diff
@@ -500,6 +500,17 @@ var _ = NFDDescribe(Label("nfd-master"), func() {
 					}
 					eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchLabels(expectedLabels, nodes))
 
+					By("Verifying the CRD status data")
+
+					nf, err := nfdClient.NfdV1alpha1().NodeFeatures(f.Namespace.Name).Get(ctx, targetNodeName, metav1.GetOptions{})
+					Expect(err).NotTo(HaveOccurred())
+
+					labelsMatch := len(expectedLabels[targetNodeName]) == nf.Status.NumberOfLabels
```
ArangoGutierrez, line 508: This is simply checking a number of labels, but not the `NodeFeatureStatus` , please re-write to have a expected `NodeFeatureStatus` and check that the same is being returned

ArangoGutierrez (changes requested):

ozhuraki:
@ArangoGutierrez

Thanks! Updated, please take a look

ArangoGutierrez:
> @ArangoGutierrez
> 
> Thanks! Updated, please take a look

@ozhuraki could you rebase, sorry it took me so long, I have been swamped by other stuff

api/nfd/v1alpha1/types.go
```diff
@@ -41,6 +41,7 @@ type NodeFeature struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec NodeFeatureSpec `json:"spec"`
+	Status NodeFeatureStatus `json:"status"`
```
ArangoGutierrez, line 44: let's have status as a subresource, similar to `NodeFeatureGroup`

api/nfd/v1alpha1/types.go
```diff
@@ -53,6 +54,22 @@ type NodeFeatureSpec struct {
...
+type NodeFeatureStatus struct {
+	// UTC time when the NodeFeature object was last updated.
+	// +optional
+	LastAppliedAt metav1.Time `json:"lastAppliedAt,omitempty"`
+	// +optional
+	// Number of features discovered.
+	NumberOfFeatures int `json:"numberOfFeatures,omitempty"`
+	// +optional
+	// Number of errors during last feature discovery.
+	NumberOfFeatureErrors int `json:"numberOfFeatureErrors,omitempty"`
```
ArangoGutierrez, line 67: this sounds more like a metric to me, I don't know what others think, I would not have this here

pkg/nfd-worker/nfd-worker.go
```diff
@@ -121,18 +121,19 @@ type ConfigOverrideArgs struct {
...
+	clientConn                  *grpc.ClientConn
+	configFilePath              string
+	config                      *NFDConfig
+	kubernetesNamespace         string
+	grpcClient                  pb.LabelerClient
+	healthServer                *grpc.Server
+	nfdClient                   *nfdclient.Clientset
+	stop                        chan struct{} // channel for signaling stop
+	featureSources              []source.FeatureSource
+	numberOfFeatureSourceErrors int
```
ArangoGutierrez, line 135: I think I'll prefer this to be exposed as a metric, but not in the scope of this PR

ArangoGutierrez (changes requested):

ozhuraki:
@ArangoGutierrez

Sorry for a delay, missed your coments and thanks for the input.

Converted status to a subresource, please take a look.

api/nfd/v1alpha1/types.go
```diff
@@ -55,6 +56,37 @@ type NodeFeatureSpec struct {
 	Labels map[string]string `json:"labels"`
 }
 
+// Status of a NodeFeature object
```
ArangoGutierrez, line 59:
```suggestion
//  NodeFeatureStatus holds the status of a NodeFeature object
```

ArangoGutierrez (changes requested):
/ok-to-test

luisdavim:
Though having a status on these objects is good, we have to keep in mind that they can be quite big and this could make the problem described in https://github.com/kubernetes-sigs/node-feature-discovery/issues/2026 worse.

PR #1669, opened 2024-04-23
api/nfd: use varargs in the NewInstanceFeatures helper

marquiz:
Make usage of this helper function more flexible.


ArangoGutierrez (approved):
/lgtm

PR #1670, opened 2024-04-23
apis/nfd: no error on ops that never match

marquiz:
Return false (i.e. "did not match") but no error when evaluating a match expression against a "flag" type feature (which don't have any associated value, just the name) if a MatchOp that never matches is used.

This is preparation for supporting multi-type features, i.e. one feature, like "cpu.cpuid", having e.g. "flag" and "attribute" type features.


ArangoGutierrez (approved):
/lgtm

PR #1671, opened 2024-04-24
apis/nfd: allow different types of features of the same name

marquiz:
This patch changes the handling of NodeFeatureRules so that one feature name (say "cpu.cpuid") can hold different types of features (flags, attributes and/or instances). Requiring features to choose one single type has not been a limitation of the API itself (and there has been no validation on this) but an implementation decision.

The new evalutation logic of match expressions is such that "flags" and "attributes" are basically evaluated as an union - they are both maps but "flags" just don't have any value associated with the key. However, "instances" are handled separately as that is basically an array of maps and needs to be evaluated in a different way (loop over the array of instances and evaluate expressions against the attributes of each). Because of this difference care must be taken if mixing "instance" features with "flag" and/or "attribute" features.

Note that the API types or their validation is not changed - just the implementation of how the NodeFeatureRules are evaluated.


marquiz:
Cherry-picked #1672 here to make this buildable

marquiz:
Rebased. Test coverage increased.

/assign @ArangoGutierrez

marquiz:
ping @ArangoGutierrez

pkg/apis/nfd/nodefeaturerule/expression.go
```diff
@@ -337,23 +337,152 @@ func MatchGetValues(m *nfdv1alpha1.MatchExpressionSet, values map[string]string)
...
-	v, err := MatchGetInstances(m, instances)
-	return len(v) > 0, err
+	isMatch, _, err := MatchGetInstances(m, instances)
+	return isMatch, err
 }
 
 // MatchGetInstances evaluates the MatchExpressionSet against a set of instance
 // features, each of which is an individual set of key-value pairs
 // (attributes). A slice containing all matching instances is returned. An
 // empty (non-nil) slice is returned if no matching instances were found.
```
ArangoGutierrez, line 347: doc needs to be updated

ArangoGutierrez (changes requested):

pkg/apis/nfd/nodefeaturerule/expression.go
marquiz: You refer to the return values? Fixed that part.

ArangoGutierrez (approved):
/lgtm

PR #1673, opened 2024-04-24
cpu: advertise AVX10 version

marquiz:
Add new cpuid label "feature.node.kubernetes.io/cpu-cpuid.AVX10_VERSION" that advertises the supported version of AVX10 vector ISA. Correspondingly, the patch adds AVX10_VERSION to the "cpu.cpuid" feature for NodeFeatureRules to consume.

This makes cpu.cpuid on amd64 architecture a "multi-type" feature in that it contains "flags" and potentially also "attributes" (the only cpuid attribute so far is the AVX10_VERSION).


marquiz:
#1671 merged
/unhold

ArangoGutierrez (approved):
/lgtm

PR #1678, opened 2024-04-25
cloudbuild.yaml: change machine type to e1-highcpu-32

marquiz:
Mitigate the big increase of build time after enabling armv7 builds.


ArangoGutierrez (approved):
/lgtm

PR #1679, opened 2024-04-25
Dependabot: Add proper dependabot config file

ArangoGutierrez:


.github/dependabot.yml
```diff
@@ -0,0 +1,24 @@
...
+    groups:
+      k8sio:
+        patterns:
+        - k8s.io/*
+
+  - package-ecosystem: "github-actions"
+    target-branch: master
+    directory: "/"
+    schedule:
+      interval: "daily"
```
TessaIO, line 24: Should we make this also weekly and on Sunday?

ArangoGutierrez: This doesn't change that often as go deps. so catching a new version of an action is ok

ArangoGutierrez:
> Thanks @ArangoGutierrez, this is a great improvement 👍 Two comments below

Ready PTAL @marquiz

.github/dependabot.yml
marquiz, line 24: I think we want the labels here, too, right

```suggestion
      interval: "daily"
    labels:
    - dependencies
    - ok-to-test
```

marquiz (approved):
/lgtm

PR #1680, opened 2024-04-26
test/e2e: stop importing kubernetes test/e2e

marquiz:
Don't import the kubernetes tests/e2e "root" package (we still use the test/e2e/framework). We only used the simple e2e runner function from there so copy that over to the nfd test/e2e package. This change removes a lot of dependencies speeding up builds.


ArangoGutierrez (approved):
/lgtm

TessaIO (approved):
LGTM

PR #1684, opened 2024-04-26
build: specify buildx builder name everywhere

marquiz:
Fixes scenarios where there are other builders, too.


ArangoGutierrez (approved):
/lgtm

PR #1685, opened 2024-04-26
cloudbuild.yaml: downgrade machine type to e2-highcpu-8

marquiz:
The higher tier builder didn't give any advantage, especially after armv7 builds were disabled. Let's revert back to highcpu-8.


ArangoGutierrez (approved):
/lgtm

PR #1694, opened 2024-05-02
source/kernel: silence misleading error on selinux detection

marquiz:
Silence a irrelevant error that was printed in case /sys/fs/selinux/ directory exists but selinuxfs is not mounted (i.e. the dir is empty).


ArangoGutierrez (approved):
PING @PiotrProkop

PR #1700, opened 2024-05-06
nfd-gc: Remove stale NRT objects

### LINKED ISSUE #1586: Improve nfd-gc on NodeResourceTopology objects

We could possibly improve the behavior of nfd-gc by removing stale NRT objects (whose creator pod does not exist anymore). The `ownerReference` field shouldn't be used as that creates warnings (see [k8s garbage-collection](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#:~:text=In%20v1.20%2B%2C%20if%20the,selector%3Dreason%3DOwnerRefInvalidNamespace.)). We could use an nfd-specific annotation (like `nfd.node.kubernetes.io/owner-pod: <namespace>/<pod-uuid>`). The nfd-gc would delete the object if the pod doesn't exist. Possibly acting only if `<namespace>` is the same where nfd-gc itself is running (so that it doesn't need cluster-wide access to get/list all pods in all namespaces).

This would be a separate improvement in addiotion to #1551 (which didn't play out quite as I initially thought).

**Why is this needed**:

Better management of NRT objects. Potentially fewer stale objects in the cluster.

ozhuraki:
Remove stale NRT objects whose creator pod does not exist anymore.

Closes #1586

### DIFF
--- a/pkg/nfd-gc/nfd-gc.go
+++ b/pkg/nfd-gc/nfd-gc.go
@@ -19,6 +19,7 @@ package nfdgarbagecollector
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	topologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
@@ -34,6 +35,7 @@ import (
 	"k8s.io/klog/v2"
 
 	nfdv1alpha1 "sigs.k8s.io/node-feature-discovery/api/nfd/v1alpha1"
+	"sigs.k8s.io/node-feature-discovery/pkg/nfd-topology-updater"
 	"sigs.k8s.io/node-feature-discovery/pkg/utils"
 	"sigs.k8s.io/node-feature-discovery/pkg/version"
 )
@@ -42,6 +44,7 @@ var (
 	gvrNF   = nfdv1alpha1.SchemeGroupVersion.WithResource("nodefeatures")
 	gvrNRT  = topologyv1alpha2.SchemeGroupVersion.WithResource("noderesourcetopologies")
 	gvrNode = corev1.SchemeGroupVersion.WithResource("nodes")
+	gvrPod  = corev1.SchemeGroupVersion.WithResource("pods")
 )
 
 // Args are the command line arguments
@@ -61,6 +64,8 @@ type nfdGarbageCollector struct {
 	stopChan chan struct{}
 	client   metadataclient.Interface
 	factory  metadatainformer.SharedInformerFactory
+	// gcNRTs holds owner-pod and namespace information for detecting stale NRT objects
+	gcNRTs map[string]string
 }
 
 func New(args *Args) (NfdGarbageCollector, error) {
@@ -76,6 +81,7 @@ func New(args *Args) (NfdGarbageCollector, error) {
 		stopChan: make(chan struct{}),
 		client:   cli,
 		factory:  metadatainformer.NewSharedInformerFactory(cli, 0),
+		gcNRTs:   make(map[string]string),
 	}, nil
 }
 
@@ -190,7 +196,30 @@ func (n *nfdGarbageCollector) garbageCollect() {
 
 	// Handle NodeResourceTopology objects
 	listAndHandle(gvrNRT, func(meta metav1.PartialObjectMetadata) {
-		if !nodeNames.Has(meta.Name) {
+		deleteNRT := false
+
+		if objRef, ok := meta.Annotations[nfdtopologyupdater.NRTOwnerPodAnnotation]; ok {
+			if s := strings.Split(objRef, "/"); len(s) == 2 {
+				pod := s[1]
+				_, err := n.client.Resource(gvrPod).Get(context.TODO(), pod, metav1.GetOptions{})
+				if errors.IsNotFound(err) {
+					if val, ok := n.gcNRTs[meta.Name]; !ok {
+						n.gcNRTs[meta.Name] = objRef
+					} else {
+						if val != objRef {
+							n.gcNRTs[meta.Name] = objRef
+						} else {
+							delete(n.gcNRTs, meta.Name)
+							deleteNRT = true
+						}
+					}
+				} else if err != nil {
+					klog.ErrorS(err, "failed to get Pod object")
+				}
+			}
+		}
+
+		if !nodeNames.Has(meta.Name) || deleteNRT {
 			n.deleteNRT(meta.Name)
 		}
 	})
--- a/pkg/nfd-topology-updater/nfd-topology-updater.go
+++ b/pkg/nfd-topology-updater/nfd-topology-updater.go
@@ -52,6 +52,7 @@ const (
 	TopologyManagerPolicyAttributeName = "topologyManagerPolicy"
 	// TopologyManagerScopeAttributeName represents an attribute which defines Topology Manager Policy Scope
 	TopologyManagerScopeAttributeName = "topologyManagerScope"
+	NRTOwnerPodAnnotation             = "nfd.node.kubernetes.io/owner-pod"
 )
 
 // Args are the command line arguments
@@ -294,6 +295,7 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 		nrtNew := v1alpha2.NodeResourceTopology{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:            w.nodeName,
+				Annotations:     map[string]string{NRTOwnerPodAnnotation: fmt.Sprintf("%s/%s", w.kubernetesNamespace, os.Getenv("POD_NAME"))},
 				OwnerReferences: w.ownerRefs,
 			},
 			Zones:      zoneInfo,
@@ -317,6 +319,7 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 	nrtMutated := nrt.DeepCopy()
 	nrtMutated.Zones = zoneInfo
 	nrtMutated.OwnerReferences = w.ownerRefs
+	nrtMutated.Annotations[NRTOwnerPodAnnotation] = fmt.Sprintf("%s/%s", w.kubernetesNamespace, os.Getenv("POD_NAME"))
 
 	attributes := scanResponse.Attributes
 
--- a/deployment/helm/node-feature-discovery/templates/topologyupdater.yaml
+++ b/deployment/helm/node-feature-discovery/templates/topologyupdater.yaml
@@ -57,6 +57,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: status.hostIP
+        - name: POD_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.name
       {{- with .Values.topologyUpdater.extraEnvs }}
         {{- toYaml . | nindent 8 }}
       {{- end}}
--- a/pkg/nfd-gc/nfd-gc_test.go
+++ b/pkg/nfd-gc/nfd-gc_test.go
@@ -36,7 +36,7 @@ import (
 
 func TestNRTGC(t *testing.T) {
 	Convey("When theres is old NRT ", t, func() {
-		gc := newMockGC(nil, []string{"node1"})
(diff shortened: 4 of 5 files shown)


pkg/nfd-gc/nfd-gc.go
```diff
@@ -146,6 +146,16 @@ func (n *nfdGarbageCollector) garbageCollect() {
...
 	}
 
+	pods, err := n.factory.Core().V1().Pods().Lister().List(labels.Everything())
+	if err != nil {
+		klog.ErrorS(err, "failed to list Pod objects")
+		return
+	}
+	podNames := sets.NewString()
+	for _, pod := range pods {
+		nodeNames.Insert(pod.Name)
```
PiotrProkop, line 156: you are inserting into `nodeNames` ?

pkg/nfd-gc/nfd-gc.go
```diff
@@ -146,6 +146,16 @@ func (n *nfdGarbageCollector) garbageCollect() {
 		nodeNames.Insert(node.Name)
 	}
 
+	pods, err := n.factory.Core().V1().Pods().Lister().List(labels.Everything())
```
PiotrProkop, line 149: are you listing all Pods in the cluster?

pkg/nfd-gc/nfd-gc.go
```diff
@@ -175,6 +185,9 @@ func (n *nfdGarbageCollector) garbageCollect() {
 			if !nodeNames.Has(nrt.Name) {
 				n.deleteNRT(nrt.Name)
 			}
+			if !podNames.Has(nrt.Labels["pod-name"]) {
```
PiotrProkop, line 188: why `pod-name` value? use const?

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -294,6 +294,7 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 		nrtNew := v1alpha2.NodeResourceTopology{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:            w.nodeName,
+				Labels:          map[string]string{"owner-pod": os.Getenv("POD_NAME")},
```
ArangoGutierrez, line 297: Let's use https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/types.go#L295 that is the to-go way to define ownership instead of a custom label

ArangoGutierrez (changes requested):

pkg/nfd-topology-updater/nfd-topology-updater.go
PiotrProkop, line 297: why `owner-pod` ? use const? Pod names are only unique in the namespace not across whole K8s cluster.

ArangoGutierrez: My bad as @PiotrProkop commented to my offline, you can not set OwnerReference to a Cluster scoped object

ozhuraki:
@PiotrProkop @ArangoGutierrez

Thanks! Updated, please take a look

ozhuraki:
@PiotrProkop @ArangoGutierrez

Updated unit tests, please take a look

PiotrProkop:
@ozhuraki @ArangoGutierrez one question, if we upgrade NFD-gc on existing cluster with this feature, it would remove all NRTs in the cluster cause none of them, has the `owner-pod` labels? Is it acceptable? Even if NRTs would be recreated after some time automatically it can lead to  some downtime for components leveraging NRTs like topo-aware-scheduler.
/cc @ffromani

ozhuraki:
@PiotrProkop
Thanks for the input. Yes, this would happen.

I updated it so that NRT is garbage collected only when it has label, please take a look.

pkg/nfd-gc/nfd-gc.go
```diff
@@ -37,6 +38,10 @@ import (
 	"sigs.k8s.io/node-feature-discovery/pkg/version"
 )
 
+const (
+	OwnerPod = "owner-pod"
```
marquiz, line 42: I think the label should have a prefix

```suggestion
	NRTOwnerPodLabel = "nfd.node.kubernetes.io/owner-pod"
```

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -294,6 +295,8 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 		nrtNew := v1alpha2.NodeResourceTopology{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:            w.nodeName,
+				Namespace:       w.kubernetesNamespace,
```
marquiz, line 298: NRT is is cluster-scoped so it doesn't have a namespace

```suggestion
```

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -294,6 +295,8 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
+				Labels:          map[string]string{nfdgc.OwnerPod: os.Getenv("POD_NAME")},
```
marquiz, line 299: I think we should include the namespace in the label (and then check it in nfd-gc)

```suggestion
				Labels:          map[string]string{nfdgc.OwnerPod: fmt.Sprintf("%s/%s", w.kubernetesNamespace, os.Getenv("POD_NAME"))},
```

marquiz (commented):
Thanks @ozhuraki for the PR. A few comments from me below.

@PiotrProkop PTAL

ozhuraki:
@marquiz 
Thanks! Updated, please take a look

pkg/nfd-gc/nfd-gc.go
```diff
@@ -172,7 +180,26 @@ func (n *nfdGarbageCollector) garbageCollect() {
...
 	} else {
 		for _, nrt := range nrts.Items {
+			delete := false
+
 			if !nodeNames.Has(nrt.Name) {
+				delete = true
+			}
+
+			if label, ok := nrt.Labels[NRTOwnerPodLabel]; ok {
+				if s := strings.Split(string(label), "/"); len(s) == 2 {
```
marquiz, line 190: nit: no need for type conversion

```suggestion
				if s := strings.Split(label, "/"); len(s) == 2 {
```

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -294,6 +295,7 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 		nrtNew := v1alpha2.NodeResourceTopology{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:            w.nodeName,
+				Labels:          map[string]string{nfdgc.NRTOwnerPodLabel: fmt.Sprintf("%s/%s", w.kubernetesNamespace, os.Getenv("POD_NAME"))},
```
marquiz, line 298: We should also set the label when updating the NRT. Otherwise we get into trouble with pod restarts/updates (when the name changes) because gc doesn't find the pod and GCs the NRT object.

pkg/nfd-gc/nfd-gc.go
```diff
@@ -37,6 +39,10 @@ import (
 	"sigs.k8s.io/node-feature-discovery/pkg/version"
 )
 
+const (
+	NRTOwnerPodLabel = "nfd.node.kubernetes.io/owner-pod"
```
marquiz, line 43: nit: I think the "owner" of the label name is the nfd-topology-updater so this would be better to move in `pkg/nfd-topology-updater/nfd-topology-updater.go`. Makes sense?

pkg/nfd-gc/nfd-gc.go
```diff
@@ -172,7 +180,26 @@ func (n *nfdGarbageCollector) garbageCollect() {
...
+
 			if !nodeNames.Has(nrt.Name) {
+				delete = true
+			}
+
+			if label, ok := nrt.Labels[NRTOwnerPodLabel]; ok {
+				if s := strings.Split(string(label), "/"); len(s) == 2 {
+					ns := s[0]
+					pod := s[1]
+					_, err := n.k8sClient.CoreV1().Pods(ns).Get(context.TODO(), pod, metav1.GetOptions{})
```
marquiz, line 193: Hmm, I think there's a window for nasty race condition in case of (topology-updater) pod restarts:

a) topology-updater pod is deleted
b) nfd-gc kicks in and doesn't see the pod (even if it's temporarily gone)
c) nfd-gc GCs the NRT
d) -> problems 

One possible solution would be to have keep a list (e.g. `map[string]string`, key is the NRT name, value is the pod ns/name) of stale NRTs. On the first GC round, if the topology-updater pod is not found mark the NRT on this list. On the next GC round, if the pod is still not found (and the pod ns/name hasn't changed) delete the NRT. WDYT?

ping @PiotrProkop @ffromani

marquiz (commented):
Thanks for the update @ozhuraki. I identified a few possible race issues but those shouldn't be too hard to solve.

pkg/nfd-gc/nfd-gc.go
PiotrProkop: I think we'll end up with this race condition every time when nfd-topology-updater will be upgraded in the cluster, as the name of the pod(created by daemonset) will change on the node.
Another idea would be to somehow check if `nfd-topology-updater` is being handled by some controller(via OwnerReference) and just check if this controller still exists? 
I am still debating if I like this idea or not as it may make the code much more complicated and I don't know if it's worth it 😄 even though 99% of the deployment should create `nfd-topology-updater` via daemon-sets we can't guarantee that, so we would need to also write the code for this 1% of deployments.

marquiz: With what I suggested the race should be solved. Don't GC immediately but wait for two GC periods. If the version (or pod template spec in the DaemonSet whatever) is updated the new pod will update the owner pod label -> the NRT would not be GC'd, even if the new pod would be (temporarily) gone, too. Only when the owner pod (marked by the special label, we could use annotation, too) is gone for two consecutive GC periods the NRT would be GC'd.

This method doesn't depend on how the topology-updater is deployed (daemonset, deployment, custom controller) as the "owner ref" label is set by the topology-updater pod.

ozhuraki:
@marquiz @PiotrProkop

Thanks for the help here and apologies, missed your comments.

Updated per your suggestion to solve the race and rebased, please take a look

pkg/nfd-gc/nfd-gc.go
```diff
@@ -190,7 +198,27 @@ func (n *nfdGarbageCollector) garbageCollect() {
...
+
+		if label, ok := meta.Annotations[nfdtopologyupdater.NRTOwnerPodAnnotation]; ok {
+			if s := strings.Split(label, "_"); len(s) == 2 {
+				ns := s[0]
+				pod := s[1]
+				_, err := n.k8sClient.CoreV1().Pods(ns).Get(context.TODO(), pod, metav1.GetOptions{})
+				if errors.IsNotFound(err) {
+					if _, ok := gcNRTs[meta.Name]; !ok {
+						gcNRTs[meta.Name] = label
+					} else {
```
marquiz, line 211: Wouldn't we need to check the value as well? `val == label`. If the pod name (or namespace) has changed, DON'T delete the NRT.

I just realized that the `label` is somewhat misleading name for the variable. How about e.g. `objRef` or `objKey` or smth?

pkg/nfd-gc/nfd-gc.go
```diff
@@ -57,10 +62,11 @@ type NfdGarbageCollector interface {
...
 type nfdGarbageCollector struct {
-	args     *Args
-	stopChan chan struct{}
-	client   metadataclient.Interface
-	factory  metadatainformer.SharedInformerFactory
+	args      *Args
+	stopChan  chan struct{}
+	client    metadataclient.Interface
+	factory   metadatainformer.SharedInformerFactory
+	k8sClient k8sclient.Interface
```
marquiz, line 69: We shouldn't need a separate client but just reuse the metadataclient above (we're not interested in the content of the podspec)

```suggestion
```

ozhuraki:
@marquiz

Thanks for the useful input. Updated, please take a look

PR #1707, opened 2024-05-15
nfd-master: add DisableAutoPrefix feature gate

marquiz:
Now that we have support for feature gates deprecate the autoDefaultNs
config option of nfd-master and replace it with a new alpha feature gate
DisableAutoPrefix (defaults to false). Using a feature gate to handle
and communicate these kind of changes, where the default behavior is
intended to be changed in a future release, feels much more natural than
using random flags/options.

The combined logic of the feature gate and the config option is a
logical OR over disabling auto-prefixing. That is, auto-prefixing is
disabled if either the feature gate or the config options is used set to
disable it:

```
                         | DisableAutoPrefix (feature gate)
                         | false | true
    -------------------- | --------------------------------
    autoDefaultNs   true |  ON   | OFF
    (config opt)   false |  OFF  | OFF
```

### DIFF
--- a/docs/reference/feature-gates.md
+++ b/docs/reference/feature-gates.md
@@ -16,7 +16,8 @@ The feature gates are set using the `-feature-gates` command line flag or
 
 | Name                  | Default | Stage  | Since   | Until  |
 | --------------------- | ------- | ------ | ------- | ------ |
-| `NodeFeatureAPI`      | true    | Beta   | V0.14   |   |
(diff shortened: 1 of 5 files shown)


TessaIO (approved):
Thanks @marquiz. LGTM

marquiz:
Unit tests fixed
/assign @TessaIO

marquiz:
Fixes in documentation

PR #1708, opened 2024-05-15
Re-add -enable-nodefeature-api cmdline flag

marquiz:
Bring back the -enable-nodefeature-api command line flag and the corresponding enableNodeFeatureApi helm config value that were removed without deprecation when the NodeFeatureAPI feature gate was introduced. The thinking behind this change is to not break existing users (without warning) unless totally unavoidable. Now the -enable-nodefeature-api flag is marked as deprecated and slated for removal in NFD v0.17.

The NodeFeatureAPI feature gate and the -enable-nodefeature-api flag work together so that the NodeFeature API is disabled (gRPC is enabled, instead) if either of them is set to false.

This patch selectively reverts parts of
06c4733bc5c5c3881edd18875114e91ef922fe2f.


marquiz:
Rebased. @ArangoGutierrez PTAL

ArangoGutierrez (approved):
We were a bit aggressive on the feature gates PR 
/lgtm

PR #1713, opened 2024-05-23
nfd-worker: improved log when creating NodeFeature object

marquiz:
Don't log an empty NodeFeature object.


ArangoGutierrez (approved):
/lgtm

PR #1714, opened 2024-05-23
api/nfd: document all undocumented fields in the types

marquiz:


api/nfd/v1alpha1/types.go
```diff
@@ -141,12 +149,17 @@ type NodeFeatureGroup struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   NodeFeatureGroupSpec   `json:"spec"`
+	// Spec defines the rules to be evaluated.
+	Spec NodeFeatureGroupSpec `json:"spec"`
+
+	// Status of the node feature group after the most recent evaluation of the
```
ArangoGutierrez, line 155:
```suggestion
	// Status of the nodeFeatureGroup after the most recent evaluation of the
```
no?

ArangoGutierrez (changes requested):
just one nit

api/nfd/v1alpha1/types.go
marquiz: I just spelled out the natural language version of the purpose of the object 🙂 But I don't have strong opinions. Changed as suggested.

ArangoGutierrez: for node feature we have NodeFeature, so is just to have a similar form

ArangoGutierrez (approved):
/lgtm

PR #1715, opened 2024-05-24
source/cpu: disable AVX10 label

marquiz:
Disable AVX10 as unnecessary as AVX10_LEVEL is better suited for checking AVX10 compatibility. There is not yet any hardware with the feature so disabling it shouldn't cause problems for users.


marquiz:
Depends on #1673 
/hold

/assign @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #1720, opened 2024-05-27
Revert "build(deps): bump actions/checkout from 1 to 4"

marquiz:
Checkout v4 does some strange mangling of tags, basically overwrites our signed/annotated tag with something like:

  /usr/bin/git -c protocol.version=2 fetch --no-tags --prune \
    --no-recurse-submodules origin \
    +2e93c3fd478b6631dd7bda9c6ec66a9f99cafaa8:refs/tags/v0.16.0

Which breaks our gh-pages build on release branches/tags.

This reverts commit 130ef1070d4f90e2a44959a68a576ef11ad93c4f.


ArangoGutierrez (approved):
We need to check why and fix this before next release 
/lgtm

PR #1737, opened 2024-06-07
source/cpu: Detect E/P cores and expose IDs through labels

ozhuraki:
Detect which CPUs are which types of the cores (P-cores or E-cores) and expose IDs through labels.

### DIFF
--- a/source/cpu/cpu.go
+++ b/source/cpu/cpu.go
@@ -194,6 +194,16 @@ func (s *cpuSource) GetLabels() (source.FeatureLabels, error) {
 		labels["coprocessor.nx_gzip"] = v
 	}
 
+	// E cores
+	if v, ok := features.Attributes[TopologyFeature].Elements["e_cores"]; ok {
+		labels["e_cores"] = v
+	}
+
+	// P cores
+	if v, ok := features.Attributes[TopologyFeature].Elements["p_cores"]; ok {
+		labels["p_cores"] = v
+	}
+
 	return labels, nil
 }
 
@@ -259,6 +269,27 @@ func getCPUModel() map[string]string {
 	return cpuModelInfo
 }
 
+func getCPUID(name string) int {
+	id := 0
+	base := 1
+	for idx := len(name) - 1; idx > 0; idx-- {
+		d := name[idx]
+
+		if '0' <= d && d <= '9' {
+			id += base * (int(d) - '0')
+			base *= 10
+		} else {
+			if base > 1 {
+				return id
+			}
+
+			return -1
+		}
+	}
+
+	return -1
+}
+
 func discoverTopology() map[string]string {
 	features := make(map[string]string)
 
@@ -270,8 +301,13 @@ func discoverTopology() map[string]string {
 
 	ht := false
 	uniquePhysicalIDs := sets.NewString()
+	coreIDs := sets.NewString()
+	pCoreIDs := sets.NewString()
+	eCoreIDs := sets.NewString()
 
 	for _, file := range files {
+		coreIDs.Insert(strconv.Itoa(getCPUID(file.Name())))
+
 		// Try to read siblings from topology
 		siblings, err := os.ReadFile(hostpath.SysfsDir.Path("bus/cpu/devices", file.Name(), "topology/thread_siblings_list"))
 		if err != nil {
@@ -298,6 +334,31 @@ func discoverTopology() map[string]string {
 
 	features["hardware_multithreading"] = strconv.FormatBool(ht)
 	features["socket_count"] = strconv.FormatInt(int64(uniquePhysicalIDs.Len()), 10)
+
+	for _, entry := range files {
+
+		// Try to read core_cpus_list from topology
+		cpus, err := os.ReadFile(hostpath.SysfsDir.Path("bus/cpu/devices", entry.Name(), "topology/core_cpus_list"))
+		if err != nil {
+			klog.ErrorS(err, "error reading core_cpus_list file")
+			return map[string]string{}
+		}
+
+		if id := strings.Split(string(cpus), "-"); len(id) == 2 {
+			if _, ok := coreIDs[id[0]]; ok {
+				pCoreIDs.Insert(id[0])
+			}
+			if _, ok := coreIDs[id[1]]; ok {
+				pCoreIDs.Insert(id[1])
+			}
+		} else {
+			id1 := strings.Split(string(id[0]), "\n")
+			eCoreIDs.Insert(id1[0])
+		}
+	}
+
+	features["e_cores"] = strings.Join(eCoreIDs.List(), "_")
+	features["p_cores"] = strings.Join(pCoreIDs.List(), "_")
 
 	return features
 }
--- a/docs/usage/features.md
+++ b/docs/usage/features.md
@@ -48,6 +48,8 @@ feature.node.kubernetes.io/<feature> = <value>
 | ----------------------------------- | ------ | --------------------------------------------------------------------------- |
 | **`cpu-cpuid.<cpuid-flag>`**        | true   | CPU capability is supported. **NOTE:** the capability might be supported but not enabled. |
 | **`cpu-hardware_multithreading`**   | true   | Hardware multithreading, such as Intel HTT, enabled (number of logical CPUs is greater than physical CPUs) |
+| **`cpu-e_cores`**                   | string | List of E core CPU IDs |
+| **`cpu-p_cores`**                   | string | List of P core CPU IDs |
 | **`cpu-coprocessor.nx_gzip`**       | true   | Nest Accelerator for GZIP is supported(Power). |
 | **`cpu-power.sst_bf.enabled`**      | true   | Intel SST-BF ([Intel Speed Select Technology][intel-sst] - Base frequency) enabled |
 | **`cpu-pstate.status`**             | string | The status of the [Intel pstate][intel-pstate] driver when in use and enabled, either 'active' or 'passive'. |


marquiz (commented):
Thanks for the patch @ozhuraki. I have hard time seeing the practical usefulness of this this information. I mean how is it possible to meaningfully consume these properties – I think they definitely should not be exposed as built-in/default node labels.

A more useful strategy could be to add this information as [attributes](https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/blob/f22ff000953d2fb5ea257fe0f6c5b3fc3ff5d950/pkg/apis/topology/v1alpha2/types.go#L64) in the NodeResourceTopology object that the nfd-topology-updater handles. This way, the E/P core information could be consumed by a Kubernetes (topology-aware) scheduler extension.

ping @kad

kad:
I have several concerns regarding this PR:
1. way of detecting cores. We should use `/sys/devices/cpu_*/*cpus` for getting information about P/E cores. not something else.
2. Those subdirs `cpu_*` are present only on hybrid CPUs, so if we have only one subdir `/sys/devices/cpu/`, it means only one type of the cores exists in the package, thus no need to expose those attributes at all. In other words: labels should be exposed only if CPU is hybrid.
3. Even on hybrid CPUs, those labels should not be by default populated, only when configuration requests that. I agree with Markus that those labels are not directly consumable, so extra step to enable them on hybrid CPUs is fine.
4. Exposing it as attributes in topology exporter is fine. There we can put them without configuration option if running on hybrid CPUs.
5. Names of the types of the cores should not be hard-coded, but taken from sysfs. e.g. 

```
$ grep "" /sys/devices/cpu_*/*cpus
/sys/devices/cpu_atom/cpus:16-19
/sys/devices/cpu_core/cpus:0-15
$
```
so, the labels should have "foo..atom" and "foo...core" names. @marquiz please suggest some good prefix instead of "foo..."

ozhuraki:
@marquiz @kad

Thanks for the helpful input. Updated, please take a look

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,39 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
+	cpu_core, err := os.ReadFile(hostpath.SysfsDir.Path("sys/devices/cpu_core", "cpus"))
+	if err != nil {
+		klog.ErrorS(err, "error reading cpu_core/cpus file")
+		return v1alpha2.AttributeList{}
+	}
+
+	attrCore := v1alpha2.AttributeInfo{
+		Name:  "cpu_core",
+		Value: strings.TrimSpace(string(cpu_core)),
+	}
```
marquiz, line 368: nit/suggestion: Looks like identical functionality to the cpu_atom stuff above. Maybe split into a function or do a for loop. WDYT?

ozhuraki:
@marquiz

Thanks for the useful input! Updated, please take a look

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,38 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 	return nil
 }
 
+func getCPUCoreAttribute(coreKind string) (error, v1alpha2.AttributeInfo) {
+	cpu, err := os.ReadFile(hostpath.SysfsDir.Path("sys/devices/" + coreKind, "cpus"))
+	if os.IsNotExist(err) {
+		klog.ErrorS(err, "error reading cpu entry file", "coreKind", coreKind)
+		return err, v1alpha2.AttributeInfo{}
+	}
```
marquiz, line 347: We should **ignore** the error if it's ENOENT. I think we should leave the error handling, or at least logging, to the caller

```suggestion
	if err != nil {
		return v1alpha2.AttributeInfo{}, err
	}
```

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,38 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 	return nil
 }
 
+func getCPUCoreAttribute(coreKind string) (error, v1alpha2.AttributeInfo) {
```
marquiz, line 342: nit: the common convention in go is to return error as the last value. It's so widely adopted that the other way around feels wonky

```suggestion
func getCPUCoreAttribute(coreKind string) (v1alpha2.AttributeInfo, error) {
```

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,38 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
+}
+
+// Dicsover E/P cores
+func discoverCpuCores() v1alpha2.AttributeList {
+	attrList := v1alpha2.AttributeList{}
+
+	err, attrAtom := getCPUCoreAttribute("cpu_atom")
+	if err == nil {
+		attrList = append(attrList, attrAtom)
+	}
```
marquiz, line 364: It would feel better to do the error handling here:
```suggestion
	if err, attrAtom := getCPUCoreAttribute("cpu_atom"); err != nil {
		if !os.IsNotExist(err) {
			klog.ErrorS(err, "error reading cpu entry file")
		}
	} else {
		attrList = append(attrList, attrAtom)
	}
```

A possible optimization would be to read all attrs in a loop to save some LOC

```suggestion
for _, name := range []string{"cpu_atom", "cpu_core"} {
	if err, attr := getCPUCoreAttribute("cpu_atom"); err != nil {
		if !os.IsNotExist(err) {
			klog.ErrorS(err, "error reading cpu entry file")
		}
	} else {
		attrList = append(attrList, attr)
	}
}

```

WDYT?

kad: there is another optimisation that can be taken: not to hardcode "atom" or "core", but detect what is present on the system and build label based on that. Not critical, but cleaner in my opinion.

marquiz: You mean read all `cpu_*` that are found? Why not. Would eliminate the os.IsNotExist check :)

ozhuraki:
@marquiz @kad

Thanks for the help! Updated, please take a look

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,45 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
 }
 
+func getCPUCoreAttribute(entry string) (v1alpha2.AttributeInfo, error) {
+	cpus, err := os.ReadFile(entry)
+	if err != nil {
+		klog.ErrorS(err, "error reading cpu entry file", "entry", entry)
+		return v1alpha2.AttributeInfo{}, err
+	}
+
+	dir, _ := filepath.Split(entry)
```
marquiz, line 349: dir is unused?

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,45 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 	return nil
 }
 
+func getCPUCoreAttribute(entry string) (v1alpha2.AttributeInfo, error) {
```
marquiz, line 342: nit/suggestion: this now looks like a generic function to read any file and return the content as an attribute. I'd suggest to rename the function as such (and change the var names inside the func generic, too)

```suggestion
func readAttributeFromFile(filepath string) (v1alpha2.AttributeInfo, error) {
```

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,45 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
+	cpus, err := os.ReadFile(entry)
+	if err != nil {
+		klog.ErrorS(err, "error reading cpu entry file", "entry", entry)
+		return v1alpha2.AttributeInfo{}, err
```
marquiz, line 346: I'd suggest  to leave the logging to the caller
```suggestion
		return v1alpha2.AttributeInfo{}, err
```

marquiz (commented):
Thanks for the update. A few more comments below.
/ok-to-test
/retitle nfd-topology-updater: Detect E/P cores and expose through attributes

pkg/nfd-topology-updater/nfd-topology-updater.go
ozhuraki: dir is actually used, just one line later

ozhuraki:
@marquiz

Thanks, updated, please take a look

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,46 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
+		return attrList
+	}
+
+	for _, entry := range cpuPaths {
+		attr, err := readAttributeFromFile(entry)
+		if err != nil {
+			klog.ErrorS(err, "error reading cpu entry file", "entry", entry)
+		} else {
+			attrList = append(attrList, attr)
+		}
```
marquiz, line 376: Sorry @ozhuraki on the back'n'forth here (and a bit hasty review last round :roll_eyes: ) but looking at the latest iteration with the for loop I feel we could just ditch the `readAttributeFromFile()`. One additional reason being that getting the name of the attribute is a bit specialized (not take the name of the file but the name of the directory).

So one more suggestion:

```suggestion
		cpus, err := os.ReadFile(entry)
		if err != nil {
			klog.ErrorS(err, "error reading cpu entry file", "entry", entry)
		} else {
			attrList = append(attrList, v1alpha2.AttributeInfo{
				Name:  filepath.Base(filepath.Dir(entry)),
				Value: strings.TrimSpace(string(cpus)),
			}
		}
```

WDYT?

ozhuraki:
@marquiz

Thanks, updated, please take a look

marquiz (approved):
Thanks @ozhuraki for the update, I think we could merge this.
/assign @PiotrProkop 

One general note about documentation that it would be nice to have e.g. a table listing all the information that we expose as attributes in the NRT. Maybe we should create an issue about this(?)

PiotrProkop:
One question and sorry if I misunderstand how hybrid cpus can be consumed in K8s. Wouldn't be more useful to just modify [ResourceAggregator](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/pkg/resourcemonitor/noderesourcesaggregator.go#L93) to advertise new resources for each cpu type in each `Zone` (NUMA node) by mapping generic k8s `cpu` resource to each `cpu_type`? But I may be wrong as I don't know how currently those E/P cores are assigned and consumed in Kubernetes.

ozhuraki:
@PiotrProkop

There's nothing special about it, just one way of exposing such information.

We can also modify ResourceAggregator to advertise new resources for each cpu type too.

I will make an issue and will try to add it with a sepatate PR.

PiotrProkop:
I just don't know how this information provided via attributes can be useful to custom scheduler without mapping to already used cpus. NRT object only exposes info about allocated/free cpus in given Zone without an info which cpu id is used and which is free.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,32 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
 	return nil
 }
 
+// Dicsover E/P cores
```
uniemimu, line 342:
```suggestion
// Discover E/P cores
```

uniemimu (commented):
Found typo

ozhuraki:
@PiotrProkop

Thanks and sorry, missed you comment.

Yes, it's merely exposing an information for particuar kind of cores for workloads that just want particular kind of cores on hybrid cpus.

I made an issue #1964 to expose it through ResourceAggregator (so the custom scheduler could track allocated/free cores) and will try to add such PR too.

@uniemimu

Thanks, updated.

PiotrProkop:
I still feel like this information should be then exposed via labels and just say that given node has `e` or `p` cores without the numbers. And then no special scheduler plugin is needed but I may be wrong on what's the usecase here. I'll ask @ffromani to chime in.

ozhuraki:
@PiotrProkop

We initially had it through labels but later switched to annotations since labels were restrictive on exposing IDs.

I can additionaly expose via labels and just say that given node has E or P, as you suggest.

ozhuraki:
@PiotrProkop

Added labels too, please take a look.

ffromani:
got the ping, will review ASAP

uniemimu (approved):
For me this is fine, but that covers just the looks of it, not the usefulness.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -349,6 +377,9 @@ func (w *nfdTopologyUpdater) updateNRTTopologyManagerInfo(nrt *v1alpha2.NodeReso
 	updateAttributes(&nrt.Attributes, tmAttributes)
 	nrt.TopologyPolicies = deprecatedTopologyPolicies
 
+	attrList := discoverCpuCores()
+	updateAttributes(&nrt.Attributes, attrList)
+
```
ffromani, line 382: PTAL https://github.com/kubernetes-sigs/node-feature-discovery/issues/1964#issuecomment-2556886885 
rather than constrain ourselves to the key/value attribute list, I'd explore the option to add new zones describing the E/P split and relationship

ArangoGutierrez:
@ozhuraki, can you rebase so we can revive this initiative?

source/cpu/cpu.go
```diff
@@ -198,6 +198,16 @@ func (s *cpuSource) GetLabels() (source.FeatureLabels, error) {
 		labels["coprocessor.nx_gzip"] = v
 	}
 
+	_, err := os.ReadFile(hostpath.SysfsDir.Path("sys/devices/cpu_atom/cpus"))
```
ffromani, line 203: do we want to access the filesystem *in this function*? it seems we are only operating in memory up till this change

pkg/nfd-topology-updater/nfd-topology-updater.go
ffromani: ok, I guess we want to keep using attributes. I'm not convinced this is the best approach, but I can live with this

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,32 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
 }
 
+// Discover E/P cores
+func discoverCpuCores() v1alpha2.AttributeList {
+	attrList := v1alpha2.AttributeList{}
+
+	cpusPathGlob := hostpath.SysfsDir.Path("sys/devices/cpu_*/cpus")
+	cpuPaths, err := filepath.Glob(cpusPathGlob)
+	if err != nil {
+		klog.ErrorS(err, "error reading cpu entries", "cpusPathGlob", cpusPathGlob)
```
ffromani, line 322: what if the path doesn't exist? I would expect to be a relatively common occurrence on non-hybrid archs (including older Intel archs). So I think we want
1. to swallow the error if the error `IsNotFound` or similar; can happen, it's ok
2. to complain very loudly if it's an actual read error: this is more likely something serious

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -337,6 +339,32 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
+	}
+
+	for _, entry := range cpuPaths {
+		cpus, err := os.ReadFile(entry)
+		if err != nil {
+			klog.ErrorS(err, "error reading cpu entry file", "entry", entry)
+		} else {
+			attrList = append(attrList, v1alpha2.AttributeInfo{
+				Name:  filepath.Base(filepath.Dir(entry)),
+				Value: strings.TrimSpace(string(cpus)),
```
ffromani, line 333: Attributes are free text, there's no schema.
But as best practice we should strive to have as close as possible to a real schema: the possible names and values should be defined as stricly as possible and as practical. Now: we are passing through the kernel API, which is usually very stable. But still: what are the possible entries for both name and value? let's at least document that so users of this API have a easier time.

ffromani (commented):
Sorry it took me forever to review.
Generally I'm (obviously) fine reporting the E/P core type. Some questions inside, but the overall direction is fine to me. Please consider adding tests (or document the existing tests cover this flow)

ArangoGutierrez:
PING @ozhuraki  / @ffromani

PR #1739, opened 2024-06-10
Fix the problem with starting the master with empty cache

IbirbyZh:
We faced the problem when master deleted some of labels on start. Sometimes he doesn't gets NodeFeatures when they are present in cluster because of empty cache in informer


marquiz:
@IbirbyZh please sign the CLA so that we can get the patch merged

/cherry-pick release-0.16

PR #1755, opened 2024-07-03
Use worker DS OwnerReference for NF's

ArangoGutierrez:


pkg/nfd-worker/nfd-worker.go
```diff
@@ -713,25 +713,21 @@ func (m *nfdWorker) updateNodeFeatureObject(labels Labels) error {
...
-				APIVersion: "v1",
-				Kind:       "Pod",
-				Name:       podName,
-				UID:        types.UID(podUID),
-				Controller: &isTrue,
-			},
-		}
-	} else {
-		klog.InfoS("Cannot set NodeFeature owner reference, POD_NAME and/or POD_UID not specified")
+	client, err := m.getKubeClient()
```
adrianchiris, line 718: should we get owner-ref once on start (eg in main) and pass it to nfdWorker as attribute ?
this will save api call on every update (if such occurs).

ArangoGutierrez: agree

ArangoGutierrez: done

pkg/nfd-worker/nfd-worker.go
```diff
@@ -713,25 +713,21 @@ func (m *nfdWorker) updateNodeFeatureObject(labels Labels) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to get kube client: %w", err)
 	}
 
+	selfPod, err := client.CoreV1().Pods(namespace).Get(context.TODO(), podName, metav1.GetOptions{})
```
tariq1890, line 723: Why not use the downward API? You pass the pod name as an envar to the worker daemonset pod template spec and retrieve it via the envar here

pkg/nfd-worker/nfd-worker.go
```diff
@@ -713,25 +713,21 @@ func (m *nfdWorker) updateNodeFeatureObject(labels Labels) error {
 
 	features := source.GetAllFeatures()
 
-	// Create owner ref
-	ownerRefs := []metav1.OwnerReference{}
+	// Get pod ower reference
```
tariq1890, line 716:
```suggestion
	// Get pod owner reference
```

pkg/nfd-worker/nfd-worker.go
adrianchiris: i see latest version still gets pod owner ref on every call

ArangoGutierrez: I forgot to push? umm let me check

ArangoGutierrez: my bad, I forgot to push the last ammend

adrianchiris (approved):
lgtm

pkg/nfd-worker/nfd-worker.go
```diff
@@ -713,25 +729,6 @@ func (m *nfdWorker) updateNodeFeatureObject(labels Labels) error {
 
 	features := source.GetAllFeatures()
 
-	// Create owner ref
-	ownerRefs := []metav1.OwnerReference{}
-	podName := os.Getenv("POD_NAME")
```
marquiz, line 718: I think we want to keep this owner ref as well. I.e. have both the DS (or whatever there are) and the Pod. NF will be deleted only when all the owners are gone.

ArangoGutierrez: done

pkg/nfd-worker/nfd-worker.go
```diff
@@ -271,6 +273,36 @@ func (w *nfdWorker) Run() error {
...
+	// Add pod owner reference if it exists
+	podUID := os.Getenv("POD_UID")
+	if podName != "" && podUID != "" {
+		isTrue := true
+		ownerReference = append(ownerReference, metav1.OwnerReference{
+			APIVersion: "v1",
+			Kind:       "Pod",
+			Name:       podName,
+			UID:        types.UID(podUID),
+			Controller: &isTrue,
```
marquiz, line 300: Now that we're at it we could just drop this, I think. It doesn't affect the garbage collection and nobody's actually looking at this

```suggestion
			Controller: &isTrue,
```

alternatively, use `ptr.To` from `k8s.io/utils`

```suggestion
			Controller: ptr.To[bool](true),
```

pkg/nfd-worker/nfd-worker.go
```diff
@@ -713,25 +745,6 @@ func (m *nfdWorker) updateNodeFeatureObject(labels Labels) error {
...
-			{
-				APIVersion: "v1",
-				Kind:       "Pod",
-				Name:       podName,
-				UID:        types.UID(podUID),
-				Controller: &isTrue,
-			},
-		}
-	} else {
-		klog.InfoS("Cannot set NodeFeature owner reference, POD_NAME and/or POD_UID not specified")
```
marquiz, line 732: Should/could we preserve this log message (modify slightly that this is about this pod)?

pkg/nfd-worker/nfd-worker.go
```diff
@@ -271,6 +273,36 @@ func (w *nfdWorker) Run() error {
...
+	podName := os.Getenv("POD_NAME")
+	client, err := w.getKubeClient()
+	if err != nil {
+		return fmt.Errorf("failed to get kube client: %w", err)
+	}
+
+	selfPod, err := client.CoreV1().Pods(w.kubernetesNamespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("failed to get pod %q: %w", podName, err)
+	}
```
marquiz, line 286: I'd suggest to not error out – or at least not be this strict. To make it easy to run nfd-worker from the command line for dev purposes.

Either just log an error, or (more preferable?) do the `if podName != ""` like below.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -130,41 +131,83 @@ type nfdWorker struct {
...
 	labelSources        []source.LabelSource
+	ownerReference      []metav1.OwnerReference
 }
 
 // This ticker can represent infinite and normal intervals.
 type infiniteTicker struct {
 	*time.Ticker
 }
 
+// NfdWorkerOption sets properties of the NfdMaster instance.
```
marquiz, line 147:
```suggestion
// NfdWorkerOption sets properties of the NfdWorker instance.
```

pkg/nfd-worker/nfd-worker.go
```diff
@@ -271,6 +314,33 @@ func (w *nfdWorker) Run() error {
...
+	if podName != "" && podUID != "" {
+		ownerReference = append(ownerReference, metav1.OwnerReference{
+			APIVersion: "v1",
+			Kind:       "Pod",
+			Name:       podName,
+			UID:        types.UID(podUID),
+		})
+	} else {
+		klog.InfoS("Cannot append POD ownerReference to NodeFeature, POD_NAME and/or POD_UID not specified")
+	}
```
marquiz, line 340: nit/suggestion: I think this could be made a bit clearer (e.g. both owner refs depend on POD_NAME env). For example like:

```golang
	if podName != "" {
		if selfPod, err := w.k8sClient.CoreV1().Pods(w.kubernetesNamespace).Get(context.TODO(), podName, metav1.GetOptions{}); err != nil {
			klog.ErrorS(err, "failed to get self pod, cannot inherit ownerReference for NodeFeature")
		} else {
			ownerReference = append(ownerReference, selfPod.OwnerReferences...)
		}

		podUID := os.Getenv("POD_UID")
		if podUID != "" {
			ownerReference = append(ownerReference, metav1.OwnerReference{
				APIVersion: "v1",
				Kind:       "Pod",
				Name:       podName,
				UID:        types.UID(podUID),
			})
		} else {
			klog.InfoS("Cannot append POD ownerReference to NodeFeature, POD_UID not specified")
		}
	} else {
		klog.InfoS("Cannot set NodeFeature owner references, POD_NAME not specified")
	}
```

WDYT?

marquiz (commented):
Thanks @ArangoGutierrez for the update. One more round of comments...

pkg/nfd-worker/nfd-worker.go
```diff
@@ -271,6 +314,43 @@ func (w *nfdWorker) Run() error {
 	labelTrigger.Reset(w.config.Core.SleepInterval.Duration)
 	defer labelTrigger.Stop()
 
+	// Create owner ref
+	ownerReference := []metav1.OwnerReference{}
+	// Get pod owner reference
+	podName := os.Getenv("POD_NAME")
+
+	selfPod, err := w.k8sClient.CoreV1().Pods(w.kubernetesNamespace).Get(context.TODO(), podName, metav1.GetOptions{})
```
adrianchiris, line 322: being able to get the owning daemonset to be later set for NodeFeature CR is critical IMO and should not be treated a best effort. especially when we attempt to do it once at start. 

can we fail Run() if error occurred but is not of type IsNotFound ? (which AFAIU is to support dev mode)

personally i prefer to be strict and always fail here and below (when getting pod). adding an option to skip failure via env var or flag (e.g DEV_MODE or --dev) so we have a deterministic behaviour.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -271,6 +314,43 @@ func (w *nfdWorker) Run() error {
...
+	if errors.IsNotFound(err) {
```
adrianchiris, line 323: we should log (and fail imo) if other error happened.

pkg/nfd-worker/nfd-worker.go
ArangoGutierrez: @marquiz ?

pkg/nfd-worker/nfd-worker.go
```diff
@@ -271,6 +314,43 @@ func (w *nfdWorker) Run() error {
...
+		klog.InfoS("Pod not found, cannot inherit ownerReference for NodeFeature")
+	} else if err == nil {
+		ownerReference = append(ownerReference, selfPod.OwnerReferences...)
+	}
```
marquiz, line 327: Duplicate code, this is already handled below.

```suggestion
```

pkg/nfd-worker/nfd-worker.go
marquiz: I think it's fine to bail out in case of an error (any error) occurred.

ArangoGutierrez: how about now?

pkg/nfd-worker/nfd-worker.go
```diff
@@ -761,7 +816,7 @@ func (m *nfdWorker) updateNodeFeatureObject(labels Labels) error {
 		nfrUpdated := nfr.DeepCopy()
 		nfrUpdated.Annotations = map[string]string{nfdv1alpha1.WorkerVersionAnnotation: version.Get()}
 		nfrUpdated.Labels = map[string]string{nfdv1alpha1.NodeFeatureObjNodeNameLabel: nodename}
-		nfrUpdated.OwnerReferences = ownerRefs
+		nfrUpdated.OwnerReferences = m.ownerReference
```
adrianchiris, line 819: @ArangoGutierrez question regarding L825. if no change to NF spec will it still update the object to set owner reference ? just want to make sure that if we update NFD in the cluster (e.g helm upgrade -n <nfd-namespace> ....), owner reference will be added to existing NF CRs.

adrianchiris (approved):
small question otherwise code LGTM

pkg/nfd-worker/nfd-worker.go
ArangoGutierrez: `DeepEqual` evaluates the entire object, so the answer is yes, if this is a new pod, because XYZ reason, we would update the NF with the new ownerRef (if this changed)

marquiz: Yes, it updates the object if anything changes (including the object metadata)

marquiz (approved):
Nice job @ArangoGutierrez. I think this is good to go

/lgtm

PR #1781, opened 2024-07-10
fix: take into consideration possibility of having empty line in swap file

### LINKED ISSUE #1751: Swap misidentified due to trailing newline in file
**What happened**:

While deploying NFD 0.16.0 on a k8s 1.30.2 node (Ubuntu 22.04), I noticed it assigned `memory-swap=true` despite swap not being enabled on this system.

I expected NFD to recognize that no swap device is registered

As far as I can tell, the issue is that `/proc/swaps` ends in a newline character `\n`, while the testdata has no such trailing newline. So any system which has `/proc/swaps` with a trailing newline will be misidentified.

This system is a VM which was initially created with swap which was then removed as part of the k8s provisioning process. `/proc/swaps` contains only the header line and a single `\n` character now.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.30.2
- Cloud provider or hardware configuration: VM on local proxmox instance
- OS (e.g: `cat /etc/os-release`): Ubuntu 22.04
- Kernel (e.g. `uname -a`): 5.15.0-113-generic
- Install tools: 
  - Cluster created with `kubeadm version: &version.Info{Major:"1", Minor:"30", GitVersion:"v1.30.2", GitCommit:"39683505b630ff2121012f3c5b16215a1449d5ed", GitTreeState:"clean", BuildDate:"2024-06-11T20:27:59Z", GoVersion:"go1.22.4", Compiler:"gc", Platform:"linux/amd64"}`
  - Helm: [Terraform provider](https://registry.terraform.io/providers/hashicorp/helm/2.14.0) version 2.14.0

I believe the fix is as simple as ignoring empty lines when checking number of lines in the file, though I won't have time to open a PR until next weekend at the earliest, so someone else is welcome to take this on as well.

TessaIO:
Closes #1751


source/memory/memory.go
```diff
@@ -195,12 +195,19 @@ func readNdDeviceInfo(path string) nfdv1alpha1.InstanceFeature {
...
+func getNumberOfNonEmptyLinesFromFile(path string) (int, error) {
 	data, err := os.ReadFile(path)
 	if err != nil {
 		return 0, err
 	}
-	return len(strings.Split(string(data), "\n")), nil
+	content := strings.Split(string(data), "\n")
+	length := 0
+	for _, line := range content {
+		if line != "" {
```
marquiz, line 206: Maybe a bit over-cautious, but how about trimming any whitespace from the lines, ignoring "empty" lines that only have whitespace, WDYT?

```suggestion
	length := 0
	for _, line := range strings.Split(string(data), "\n") {
		if strings.TrimSpace(line) != "" {
```

TessaIO: yeah, that's a valid point even though I don't expect it to be the case. Gonna update that

marquiz (approved):
Thank you @TessaIO for helping with this!

Issue #1546, opened 2024-01-17
Drop NFD gRPC API


marquiz:
**What would you like to be added**:

Remove the gRPC API that was deprecated in [v0.14.0](https://github.com/kubernetes-sigs/node-feature-discovery/releases/tag/v0.14.0) and replaced by NodeFeature CRD API. This includes code (including all related config and command line options like TLS), tests and documentation.

By our [deprecation policy](https://kubernetes-sigs.github.io/node-feature-discovery/v0.15/reference/versions.html) (I think this would fall under the "configuration options") we could do this already in v0.16 if so desired.

**Why is this needed**:

Drop deprecated feature. Simplify code, tests and documentation.

**Tasks**:

- [x] Flip the feature gate to GA (#1778)
- [x] Drop the `-enable-nodefeature-api` flag (#1780)
- [x] Drop the TLS-related flags (`-ca-file`, `-cert-file`, `-key-file`, `-verify-node-name`)
- [x] Helm: drop support for cert-manager
- [x] Drop the `-port` flag from nfd-master
- [x] Drop the `-server` and `-server-name-override` flags from nfd-worker
- [x] Make the NFD API controller perma-on, drop `-featurerules-controller` and `-crd-controller` flags
- [x] Drop the gRPC API (`pkg/labeler` package)
- [x] Update developer guide in the docs (running nfd-daemons locally, stand-alone, as described there is not possible anymore)
- [x] Drop all documentation related to the gRPC API

marquiz:
Maybe it's time for this in v0.17. Opinions @ArangoGutierrez, @adrianchiris?

ArangoGutierrez:
V0.14 was released on Sep 7, 2023. V0.17 will provably be released around the same time (Pre-kubecon days), that's a full year. 
I think it's ok
/lgtm

marquiz:
I think the checklist in the description has many pretty straightforward tasks
/help

marquiz:
At least the docs is still missing

@ArangoGutierrez could you update the checklist in the description?

/reopen

ArangoGutierrez:
Can we close this as done @marquiz

marquiz:
> Can we close this as done @marquiz

Yes, I think so. Thank you @ArangoGutierrez for your efforts on making this happen

/close

PR #1778, opened 2024-07-10
feature-gates: mark NodeFeatureAPI as GA

marquiz:
The feature gate is locked to true. That is, it is not possible to revert back to the gPRC-based communication which makes the gRPC API ready for removal.

PR #1788, opened 2024-07-12
nfd-worker: change TestRun to use NodeFeature API

marquiz:
Run nfd-worker with NodeFeature API enabled (against a fake apiserver) instead of using the deprecated gRPC (against a nfd-master instance).

Expand the test to verify the features and labels that are advertised as a NodeFeature object.

Refs #1546


marquiz:
Pre-requisite for #1778

ArangoGutierrez (approved):
/lgtm

PR #1789, opened 2024-07-16
[release-0.16] Add revision history limit for master replica and for garbage collector

marquiz:
Signed-off-by: Omer Aplatony <omerap12@gmail.com>
(cherry picked from commit 920306cba8bb1b9cc06b26313ccdf6e02da95561)

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/master.yaml
+++ b/deployment/helm/node-feature-discovery/templates/master.yaml
@@ -13,6 +13,7 @@ metadata:
   {{- end }}
 spec:
   replicas: {{ .Values.master.replicaCount }}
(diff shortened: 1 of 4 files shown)


rouke-broersma:
@marquiz the release notes https://github.com/kubernetes-sigs/node-feature-discovery/releases/tag/v0.16.3 mention that this change is included in the release, however if I extract the chart bundle from the registry locally it looks like the change is not included.

In fact the files in the bundle all have a timestamp from last week instead of yesterday.

image

As you can see here, line 16 should have the revision history limit but it does not:

image

marquiz:
Thanks for reporting this @rouke-broersma. The issue is now fixed

PR #1795, opened 2024-07-17
scripts/test-infra: bump golangci-lint to v1.59.1

marquiz:


ArangoGutierrez (approved):
hope nothing breaks, every lint upgrade brings shame to the existing code ;P 

/lgtm

PR #1782, opened 2024-07-11
Helm: Add revision history limit for master replica

omerap12:
Fixes: #1759 

Tested locally. Output:
```yaml
---
# Source: node-feature-discovery/templates/master.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name:  release-name-node-feature-discovery-master
  namespace: default
  labels:
    helm.sh/chart: node-feature-discovery-0.2.1
    app.kubernetes.io/name: node-feature-discovery
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/version: "master"
    app.kubernetes.io/managed-by: Helm
    role: master
spec:
  replicas: 1
  revisionHistoryLimit: 3
  selector:
    matchLabels:
      app.kubernetes.io/name: node-feature-discovery
      app.kubernetes.io/instance: release-name
      role: master
  template:
    metadata:
      labels:
        app.kubernetes.io/name: node-feature-discovery
        app.kubernetes.io/instance: release-name
        role: master
    spec:
      serviceAccountName: release-name-node-feature-discovery
      enableServiceLinks: false
      securityContext:
        {}
      containers:
        - name: master
          securityContext:
            allowPrivilegeEscalation: false
            capabilities:
              drop:
              - ALL
            readOnlyRootFilesystem: true
            runAsNonRoot: true
          image: "gcr.io/k8s-staging-nfd/node-feature-discovery:master"
          imagePullPolicy: Always
          livenessProbe:
            grpc:
              port: 8082
            initialDelaySeconds: 10
            periodSeconds: 10
          readinessProbe:
            grpc:
              port: 8082
            initialDelaySeconds: 5
            periodSeconds: 10
            failureThreshold: 10
          ports:
          - containerPort: 8080
            name: grpc
          - containerPort: 8081
            name: metrics
          env:
          - name: NODE_NAME
            valueFrom:
              fieldRef:
                fieldPath: spec.nodeName
          command:
            - "nfd-master"
          resources:
            limits:
              memory: 4Gi
            requests:
              cpu: 100m
              memory: 128Mi
          args:
            ## By default, disable crd controller for other than the default instances
            - "-crd-controller=true"
            # Go over featureGates and add the feature-gate flag
            - "-feature-gates=NodeFeatureAPI=true"
            - "-feature-gates=NodeFeatureGroupAPI=false"
            - "-metrics=8081"
          volumeMounts:
            - name: nfd-master-conf
              mountPath: "/etc/kubernetes/node-feature-discovery"
[...]

PR #1797, opened 2024-07-17
Helm: Add revision history limit for worker daemonset

### LINKED ISSUE #1759: Add support for revisionHistoryLimit to Helm Chart
Please add support for `revisionHistoryLimit` in the Helm Chart and/or set a reasonable default such as `3`.

**Why is this needed**:
Systems which implement NFD via a Gitops process do not need 10 copies each `ReplicaSet` and `ControllerRevision`. They would never be used for rollback as the git commit would be reverted instead. Having a little bit of history for comparison can be handy so a default value of `3` is often reasonable for that and for rollbacks for non-gitops.

(The actual default value NFD wants to use I don't care about.  I just want to be able to override it.)

rouke-broersma:
#1782 added support for setting revision history limit for master and gc replicas. This PR adds the same for worker replicas.

Fix #1759

### DIFF
--- a/docs/deployment/helm.md
+++ b/docs/deployment/helm.md
@@ -168,6 +168,7 @@ API's you need to install the prometheus operator in your cluster.
 | `worker.annotations`                | dict   | {}                      | NFD worker pod [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)                                                                                         |
 | `worker.daemonsetAnnotations`       | dict   | {}                      | NFD worker daemonset [annotations](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/)                                                                                   |
 | `worker.args`                       | array  | []                      | Additional [command line arguments](../reference/worker-commandline-reference.md) to pass to nfd-worker                                                                                              |
+| `worker.revisionHistoryLimit`       | integer |                                  | Specify how many old ControllerRevisions for this DaemonSet you want to retain. [revisionHistoryLimit](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/daemon-set-v1/#DaemonSetSpec)          |
 
 ### Topology updater parameters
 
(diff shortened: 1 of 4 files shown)


marquiz (commented):
Thanks @rouke-broersma for this improvement. The daemonsets being missing in #1782 somehow slipped through my (sloppy) review 😇

In addition to nfd-worker, could you also add the setting for the nfd-topology-updater DaemonSet.

rouke-broersma:
Yes I'll do that shortly

docs/deployment/helm.md
```diff
@@ -198,6 +199,7 @@ API's you need to install the prometheus operator in your cluster.
 | `topologyUpdater.podSetFingerprint`           | bool    | true                     | Enables compute and report of pod fingerprint in NRT objects.                                                                                                                                    |
 | `topologyUpdater.kubeletStateDir`             | string  | /var/lib/kubelet         | Specifies kubelet state directory path for watching state and checkpoint files. Empty value disables kubelet state tracking.                                                                     |
 | `topologyUpdater.args`                        | array   | []                       | Additional [command line arguments](../reference/topology-updater-commandline-reference.md) to pass to nfd-topology-updater                                                                      |
+| `topologyUpdater.revisionHistoryLimit`       | integer |                                  | Specify how many old ControllerRevisions for this DaemonSet you want to retain. [revisionHistoryLimit](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/daemon-set-v1/#DaemonSetSpec)          |
```
marquiz, line 202: nit: fix the table formatting (align column with the one(s) above). The tables were just recently formatted/aligned so let's try to keep it that way :)

rouke-broersma: I notice the last table dividers are not aligned, so I assume this is not necessary. I have aligned the others.

marquiz: Thanks @rouke-broersma for the update. The last dividers are not be perfectly aligned and I think that's ok (especially because the text in some of the last columns is super long and it would be impractical to perfectly align the rest with these).

marquiz (approved):
Looks good to me!
/assign @ArangoGutierrez 

/ok-to-test
/cherry-pick release-0.16

marquiz (commented):

Issue #1802, opened 2024-07-19
[v0.16.0] All node labels removed while informer cache failed to sync


ahmetb:
**What happened**:

While migrating from 0.10 to 0.16.0:

* all node feature labels got removed
* `kubectl get nodefeatures -n node-feature-discovery` was unresponsive at the time (likely because the cluster size is 4000 nodes and the NodeFeature CR objects are 130kb each by default)

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**: Run nfd chart by default on a 4000 node cluster.

**Anything else we need to know?**:

There were extensive informer sync errors in `nfd-master` logs (seeming to be timing out after 60s). This is likely because the LIST NodeFeatures is a very expensive call (each object is very large + a lot of Nodes in the cluster).

Attaching logs: [nfd-master.log](https://github.com/user-attachments/files/16316036/redacted.log)

My suspicion is that the `nfd-master` somehow does not wait for informer cache to sync (as the first informer sync error occurs exactly 60s after the process starts) –and it treats lack of response as "empty set of labels" and clearing the labels. (But I'm not familiar with the inner workings of the codebase, it's just a theory.)

💡 We don't see the issue on much smaller clusters. 

💡 We have not yet tried v0.16.2 (release notes mention it fixes a node removal issue, but it's clear what was the root cause there).

**Environment**:
- Kubernetes version (use `kubectl version`): v1.23.17
- Cloud provider or hardware configuration: Bare-metal
- OS (e.g: `cat /etc/os-release`): Not applicable
- Kernel (e.g. `uname -a`): Not applicable
- Install tools: installed via Helm
- Network plugin and version (if this is a network-related bug): Not applicable
- Others: Not applicable

marquiz:
I think this happens when upgrading from NFD v0.13 or earlier. The  specific reason is the NodeFeature CRD API that was introduced in v0.14 (earlier versions of NFD relied on gRPC communication between nfd-master and nfd-worker instances).

What happens (I think) is that nfd-master starts quickly (and syncs NodeFeature cache) as there are no NodeFeature objects in the cluster. And, bc of that it starts removing the labels from nodes (as it sees no features for them published by the nfd-worker instances). At the same time the nfd-worker instances start and start pushing NodeFeature objects to the apiserver -> there are ton of them and the cache throws timeout errors. Eventually things "stabilize" and node labels are back.

[...]

PR #1809, opened 2024-07-25
nfd-master: check nfd api informer cache sync result

marquiz:
Bail out if there were errors in syncing the cache of any resource.


ArangoGutierrez (approved):
We need to add more "chaos" scenarios to our E2E suite
/lgtm

TessaIO:
> We need to add more "chaos" scenarios to our E2E suite /lgtm

@ArangoGutierrez should we create a ticket for that?

ahmetb:
This is about #1802, right?

marquiz:
> This is in response to #1802, right?

Kinda. Inspired by it but not (afaiu) fixing anything you encountered. `WaitForCacheSync` blocks and I'm not entirely sure in which scenarios it returns with caches not synced but we'd better catch those situations without falling through unnoticed.

PR #1813, opened 2024-07-26
nfd-gc: only fetch object metadata

marquiz:
Significantly reduce the apiserver and network load by only listing/getting the object metadata.


marquiz:
Changed to use global (package-wide) vars for GVRs, as suggested in https://github.com/kubernetes-sigs/node-feature-discovery/pull/1815#discussion_r1696721412

@PiotrProkop PTAL

Issue #1806, opened 2024-07-23
Improved readiness probe for nfd-worker


marquiz:
**What would you like to be added**:

Improve the gRPC health server so that it doesn't report ready/live until the feature discovery has succeeded and the NodeFeature CR object has been created/updated.

**Why is this needed**:

Better reflect the actual readiness nfd-worker.

marquiz:
ping @omerap12 :blush:

omerap12:
yeap :) 
/assign

Issue #1548, opened 2024-01-17
Improved liveness probe for nfd-master


marquiz:
**What would you like to be added**:

The implementation of the gRPC health server in nfd-master is very simple and naive, in practice only telling that the nfd-master main loop was started and that the program hasn't exited. This could be improved e.g. by including the state of the metrics server, nfd api controller and node updater pool (i.e. healthcheck fails if any of those goes down).

**Why is this needed**:

More trust on the liveness data.

omerap12:
Hey @marquiz  , maybe I can assist with it :)

marquiz:
> Hey @marquiz , maybe I can assist with it :)

Wow, appreciated

omerap12:
> > Hey @marquiz , maybe I can assist with it :)
> 
> Wow, appreciated

sure why not, seems like a task I can take ( but I may ask a few questions along the way if that's ok )

marquiz:
> I may ask a few questions along the way if that's ok

Definitely. The "design" details in the description are pretty vague and really hastily written

omerap12:
Hi @marquiz, I've started working on issue #1814. Could you take a quick look and let me know if I'm on the right track?

omerap12:
friendly remainder :) @marquiz

omerap12:
> > friendly remainder :) @marquiz
> 
> Thanks @omerap12. I try to find a peaceful slot tomorrow to look at that

Sure, it's just a quick sketch, and I wanted to check if I'm addressing this in a sensible manner. Thanks!

PR #1814, opened 2024-07-26
Issue 1548

omerap12:


pkg/nfd-master/nfd-master.go
```diff
@@ -162,6 +162,64 @@ type nfdMaster struct {
...
+}
+
+func (s *customHealthServer) SetK8sClient(client k8sclient.Interface) {
+	s.k8sclient = client
+}
+
+// Check method for customHealthServer
+func (s *customHealthServer) Check(ctx context.Context, req *grpc_health_v1.HealthCheckRequest) (*grpc_health_v1.HealthCheckResponse, error) {
+	klog.InfoS("Check request received")
+	metricServerStatus, err := s.CheckPods("kube-system", "k8s-app=metrics-server")
```
marquiz, line 178: Oh, sorry for the ambiguity. I wasn't thinking about kubernetes metrics-server but the metrics http server inside nfd-master itself that is started with `utils.CreateMetricsServer()`. We could e.g. just set a health flag to false if the goroutine started there exits.

marquiz (commented):
Thanks @omerap12 for starting to work on this! There was one misconception regarding my description in #1548 (this PR seems to address that, i.e. nfd-master, instead of #1806).

In addition to monitoring the (internal) metrics server we could watch the goroutines of node updater pool (and set health to false if they exit, for whatever reaseon). Then we could think about using `SetWatchErrorHandler.()` to monitor the informers in the nfd api controller.

omerap12:
Hi @marquiz, thanks for the review! I’ll make the necessary code adjustments and have updated the title as well, lol. I’ll keep you posted.

omerap12:
Hey @marquiz,

Here's a summary of what I've done so far:
* I created a custom health server within the NFD master. This server has two boolean fields that indicate the health status of the internal metric server and whether any of the major nodeUpdaterPool functions have panicked.
* The custom health server periodically checks these fields. If either of them is false, the pod's liveness status is set to false.
Looking forward to more instructions! :)

omerap12:
@marquiz , Friendly ping :)

PR #1829, opened 2024-08-08
docs: use jekyll-rtd-theme from a ruby gem

marquiz:
The upstream repo (and the release downloads)
github.com/rundocs/jekyll-rtd-theme has been deleted. This broke our docs generation as the remote theme configuration depended on downloading the release artefact.

This patch changes the docs building to use a Ruby gem instead of the remote theme setting. To complicate matters, the gem has an seemingly incorrect (too strict) version dependency. To mitigate this, we now install bundler-override plugin to ignore this particular dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -246,11 +246,11 @@ poll-images:
 
 site-build:
 	@mkdir -p docs/vendor/bundle
-	$(SITE_BUILD_CMD) sh -c "bundle install && jekyll build $(JEKYLL_OPTS)"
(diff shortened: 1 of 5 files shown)


marquiz:
Finally I was able to make the netlify conf work. @ArangoGutierrez @fmuyassarov PTAL

fmuyassarov (commented):
Good job @marquiz !
/lgtm

PR #1843, opened 2024-08-19
nfd-master: use only unbuffered chans in the nfd api-controller

marquiz:
There's no reason why the "update all" chans should be buffered (while the other are not).


TessaIO (approved):

ArangoGutierrez (approved):
/lgtm

PR #1844, opened 2024-08-19
nfd-master: explicit state variable for the node updater pool

marquiz:


pkg/nfd-master/updater-pool.go
```diff
@@ -158,18 +154,14 @@ func (u *updaterPool) start(parallelism int) {
...
 	u.Lock()
 	defer u.Unlock()
 
-	if u.queue == nil || u.queue.ShuttingDown() {
-		klog.InfoS("the NFD master updater pool is not running.")
-		return
-	}
-
-	if u.nfgQueue == nil || u.nfgQueue.ShuttingDown() {
+	if !u.started {
```
TessaIO, line 164: why don't we use the u.running method?

marquiz: As that's also behind a lock

TessaIO: perfect thanks, didn't see the first lock 😄

PR #1845, opened 2024-08-19
NFD image compatibility proposal

mfranczy:
Proposal to describe image requirements by using NFD labels and OCI artifact.

### DIFF
--- a/enhancements/1845-nfd-image-compatibility/README.md
+++ b/enhancements/1845-nfd-image-compatibility/README.md
@@ -1,0 +1,160 @@
+# KEP-1845: Image Compatibility with NFD
+
+## Summary
+
+Currently, there is no standard solution for describing container image requirements in relation to hardware or operating systems.
+Cloud-native technologies are being adopted by high-demand industries where container compatibility is critical for service performance and cluster preparation.
+This proposal introduces the concept of NFD image compatibility metadata.
+NFD features via NodeFeatureRules CRs can be effectively added to images to specify requirements for a host or operating system.
+
+The document has been prepared based on the experience and progress of the [OCI Image Compatibility working group](https://github.com/opencontainers/wg-image-compatibility/tree/main/docs/proposals).
+
+## Motivation
+
+Image compatibility metadata will help container image authors describe compatibility requirements in a standardized way.
+This metadata will be uploaded with the image to the image registry.
+As a result, container compatibility requirements will become discoverable and programmable, supporting various consumers and use cases where applications require a specific compatible environment.
+
+### Goals
+
+#### Phase 1
+
+- Use existing NFD features via the NodeFeatureRule API to describe container image requirements.
+- Create a new OCI artifact type for compatibility metadata.
+- Allow verification of node compatibilitym including nodes that are not yet part of the k8s cluster.
+- Add or extend the sources with missing features.
+
+#### Phase 2
+
+Phase 2 involves future prediction and shows the general direction.
+After the completion of Phase 1, either this document should be updated, or a new proposal should be created that considers the following points:
+
+- Update or generate pods with appropriate node selectors via a mutation webhook or a scheduler plugin.
+
+### Non-Goals
+
+- Make image compatibility a hard requirement for the NFD installation/usage.
+- Cover applications ABI compatibility.
+
+## Proposal
+
+Build a new NFD client tool with the following initial scope:
+
+- CRUD OCI artifact.
+- Validate nodes based on provided metadata.
+- Run directly on a host which is not part of the Kubernetes cluster, or run as a Kubernetes job on a Kubernetes node.
+
+### Design Details
+
+#### OCI Artifact
+
+[An OCI artifact](https://github.com/opencontainers/image-spec/blob/main/manifest.md#guidelines-for-artifact-usage) should be created to store image compatibility metadata on the image side.  
+The artifact can be connected with an image over [the subject field](https://github.com/opencontainers/distribution-spec/blob/11b8e3fba7d2d7329513d0cff53058243c334858/spec.md#pushing-manifests-with-subject).
+
+##### Manifest
+
+```json
+{
+  "schemaVersion": 2,
+  "mediaType": "application/vnd.oci.image.manifest.v1+json",
+  "artifactType": "application/vnd.k8s.nfd.image-compatibility.v1",
+  "config": {
+    "mediaType": "application/vnd.oci.empty.v1+json",
+    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
+    "size": 2
+  },
+  "layers": [
+    {
+      "mediaType": "application/vnd.k8s.nfd.image-compatibility.spec.v1+yaml",
+      "digest": "sha256:4a47f8ae4c713906618413cb9795824d09eeadf948729e213a1ba11a1e31d052",
+      "size": 1710
+    }
+  ],
+  "subject": {
+    "mediaType": "application/vnd.oci.image.manifest.v1+json",
+    "digest": "sha256:5b0bcabd1ed22e9fb1310cf6c2dec7cdef19f0ad69efa1f392e94a4333501270",
+    "size": 7682
+  },
+  "annotations": {
+    "oci.opencontainers.image.created": "2024-03-27T08:08:08Z"
+  }
+}
+```
+
+##### Artifact Payload (Schema)
+
+- **version** - *string*  
+This REQUIRED property specifies the version of the API being used.
+
+- **compatibilities** - *array of object*  
+This REQUIRED property is a list of compatibility sets.
+
+  - **rules** - *object*  
+  This REQUIRED property is a reference to the spec of [NodeFeatureRule API](https://kubernetes-sigs.github.io/node-feature-discovery/v0.16/usage/custom-resources.html#nodefeaturerule).
+  The spec makes it possible to describe image requirements using the discovered features from NFD sources.
+  For further reading, please review [the documentation](https://kubernetes-sigs.github.io/node-feature-discovery/v0.16/usage/customization-guide.html#nodefeaturerule-custom-resource).
+
+  - **weight** - *int*  
+  This OPTIONAL property specify the [node affinity weight](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity-weight).
+
+  - **tag** - *string*  
+  This OPTIONAL property allows grouping or dividing of compatibility sets.
+
+  - **description** - *string*  
+  This OPTIONAL property is intended for a brief description of a compatibility set.
+
+Example
+
+```yaml
+version: v1alpha1
+compatibilities:
+- tag: "prefered"
+  weight: 10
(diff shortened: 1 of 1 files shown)


enhancements/1845-nfd-image-compatibility/README.md
```diff
@@ -0,0 +1,184 @@
...
+- `"kernel-version.full": ">=4.15,<5.25; >5.30"` - greater or equal to 4.15 and lower than 5.25 or greater than 5.30.
+- `"kernel-version.full": ">=5.15.x-x-generic"` - great or equal to 5.15 generic kernel flavor.
+- `"kernel-version.full": ">=5.15.x-x-generic; >=5.30"` - great or equal to 5.15 generic kernel flavor or greater or equal to any 5.30 kernel version.
+- `"kernel-version.full": "5.15"` - must equal to 5.15.
+
+#### NFD client
+
+A new standalone cmd line utility should be implemented for the NFD project that shares the same functionality as [nfd kubectl plugin](https://nfd.sigs.k8s.io/usage/kubectl-plugin).
+
+Both clients should implemented the following commands:
```
mfranczy, line 166: Maybe it would be better to just have one client?

ChaoyiHuang: it seems that current nfd client should work as the kubectl plugin, because it has to access k8s api server with credential. 

for compatibility, access to registry may need different credential. And for the sceanrio where the node is not part of k8s cluster, providing k8s credential seems unnecessary.

enhancements/1845-nfd-image-compatibility/README.md
```diff
@@ -0,0 +1,184 @@
...
+This makes hard container compatibility requirements discoverable, programmable, and will support different consumers and cover use cases where the application requires a specific compatible environment.
+
+### Goals
+
+#### Phase 1
+
+- Use existing NFD labels (keys) to describe container image requirements.
+- Create a notation that will describe values including list of ranges and literal.
+- Create a new OCI artifact type for compatibility metadata.
+Allow to verify node compatibility - including nodes that are not part of k8s cluster yet.
```
ArangoGutierrez, line 25:
```suggestion
- Allow to verify node compatibility - including nodes that are not part of k8s cluster yet.
```

ArangoGutierrez (commented):
I like where this is going, as with every KEP, let's iterate fast, and release often. Let's try to avoid a long running discussion on this PR

@marquiz 

cc @vsoch

enhancements/1845-nfd-image-compatibility/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+#### NFD client
+
+A new standalone cmd line utility should be implemented for the NFD project that shares the same functionality as [nfd kubectl plugin](https://nfd.sigs.k8s.io/usage/kubectl-plugin).
+
+Both clients should implemented the following commands:
+
+- `validate` -  validate a NodeFeatureRule object (implemented in kubectl plugin).
+- `test` - test a NodeFeatureRule object against a node (implemented in kubectl plugin).
+- `dryrun` - DryRun a NodeFeatureRule object against a NodeFeature file (implemented in kubectl plugin).
```
vsoch, line 170: Can you better explain "dryrun" ? I'm used to seeing it as `--dryrun` to supplement a more clearly obvious action.

vsoch (commented):
I recognize all of this work! Thanks for opening this up @mfranczy - looks good to me. :+1:

enhancements/1845-nfd-image-compatibility/README.md
mfranczy: I am also used to see it as you. I kept it that way to be compatibile with the [kubectl `dryrun` command](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/cmd/kubectl-nfd/subcmd/dryrun.go).

mfranczy:
I am in the process of building nfd client prototype. I reduced the scope with commands. ORAS SDK provides a lot of useful functions so there is no need to introduce much for artifact manipulation (thanks ORAS folks!).

marquiz:
Thanks @mfranczy for the proposal. Just thinking aloud below, not proposing anything or having any answers.

It looks like we need to rely on node labels to be able to integrate this with the kubernetes components (scheduler etc). If we want to do without changes to the kubernetes API or extend the kubernetes components with some plugins looking at NFD CRDs (or similar) or something exotic like that.

What we need to think is the node label management on NFD side. We have a bunch of built-in labels (that get populated by just deploying NFD) but a lot of stuff is only visible in NodeFeatures only. For automatizing the creation NodeFeatureRules (a future, step, yes) we need to understand what we get automatically, from the built-in-labels and what we need custom labels for. We could use namespacing e.g. prefix all image compatibility labels with `oci.feature.node.kubernetes.io/` (or `opencontainers.org/` or similar).

Regarding the expressions, would CEL (or parts of it) be useful there?
https://github.com/google/cel-spec

mfranczy:
@marquiz thanks for looking into it.

> It looks like we need to rely on node labels to be able to integrate this with the kubernetes components (scheduler etc). If we want to do without changes to the kubernetes API or extend the kubernetes components with some plugins looking at NFD CRDs (or similar) or something exotic like that.

I think looking at NFD CRDs would be a good choice for start.

> What we need to think is the node label management on NFD side. We have a bunch of built-in labels (that get populated by just deploying NFD) but a lot of stuff is only visible in NodeFeatures only. For automatizing the creation NodeFeatureRules (a future, step, yes) we need to understand what we get automatically, from the built-in-labels and what we need custom labels for. We could use namespacing e.g. prefix all image compatibility labels with oci.feature.node.kubernetes.io/ (or opencontainers.org/ or similar).

Right, a few missing pieces I already identified, like discovery of runtime configuration (`/proc/sys`) that could be a built-in source only visible to NodeFeatures. Basically for image compatibility there is a lot of things already covered by the project (like kernel modules, config, version, usb and pci devices etc.). Maybe we don't need to introduce a new namespacing for the image compatibility purpose. From my perspective we could extend the built-in functionality and keep the original NFD labels (or features - if the label is not populated to a node) to describe the compatibility. What do you think?

> Regarding the expressions, would CEL (or parts of it) be useful there?
https://github.com/google/cel-spec

Looks very promising, I think we could use it. That's a good idea.

vsoch:
> Considering the danymic situation, extending scheduler with plugin to validate the image compatibility and node features seems to be more viable.

In that context agree - the scheduler would determine compatibility, but then the container runtime doing the pull would do so based on more fine-grained selection for optional performance. It would be scoped to what the scheduler scoped it to (architecture, GPU and devices, etc).

mfranczy:
> > Regarding the expressions, would CEL (or parts of it) be useful there?
> > https://github.com/google/cel-spec
> 
> Looks very promising, I think we could use it. That's a good idea.

@marquiz I will try to implement this in the prototype and will inform if there is anything that stops us to use CEL.

mfranczy:
@marquiz there is nothing that prevents to add extra fields for CEL expressions in the spec.

However, @marquiz @ArangoGutierrez after looking deeply into NodeFeatureRules, I believe we could just use the rules object (or part of it) to describe image requirements in the artifact. It provides everything what we need to describe image requirements, it already allows to use expressions, it's possible to set a custom hierarchy, it operates directly on features and most important we don't need to create any complex translator from the new spec to NFR for the deployment in k8s cluster.

If you're ok with the idea then I would update the proposal to include NFR rules object into compatibility artifact.

marquiz:
> If you're ok with the idea then I would update the proposal to include NFR rules object into compatibility artifact.

In principle, I like this idea. That just means that the Kubernetes needs to be made aware of this stuff which complicates matters in the short term. With this, I also agree with @ChaoyiHuang @vsoch that integrating this in the kubernetes control plane (with new functionality, say a scheduler plugin) is a more viable and maintainable solution for the longer term

enhancements/1845-nfd-image-compatibility/README.md
```diff
@@ -0,0 +1,182 @@
+# KEP-1845: Image Compatibility with NFD
+
+## Summary
+
+Today, there is no standard solution of describing container image requirements against hardware or operating systems.
+Cloud-native is being adapted by highly demanded industries where container compatibility plays a critical role to services in terms of cluster preparation and overall performance.
```
ArangoGutierrez, line 6:
```suggestion
Cloud-native is being adopted by highly demanded industries where container compatibility plays a critical role for services in terms of cluster preparation and overall performance.
```

enhancements/1845-nfd-image-compatibility/README.md
```diff
@@ -0,0 +1,182 @@
...
+This proposal introduces an idea of NFD image compatibility metadata.
+NFD labels can successfully be added to image to describe the requirements against a host or OS.
```
ArangoGutierrez, line 8:
```suggestion
NFD labels can successfully be added to the image to describe the requirements against a host or OS.
```

ArangoGutierrez (changes requested):

mfranczy:
@marquiz @ArangoGutierrez I have updated the proposal based on our discussion and consider it complete from my side.

enhancements/1845-nfd-image-compatibility/README.md
```diff
@@ -0,0 +1,184 @@
...
+
+  - **tag** - *string*  
+  This OPTIONAL property allows grouping or dividing of compatibility sets.
+
+  - **description** - *string*  
+  This OPTIONAL property is intended for a brief description of a compatibility set.
+
+Example
+
+```json
```
ChaoyiHuang, line 108: considering the node feature rules are usually expressed in yaml format, how about use yaml as an example. support both yaml and json is good.

enhancements/1845-nfd-image-compatibility/README.md
```diff
@@ -0,0 +1,184 @@
...
+  "mediaType": "application/vnd.oci.image.manifest.v1+json",
+  "artifactType": "application/vnd.k8s.nfd.image-compatibility.v1",
+  "config": {
+    "mediaType": "application/vnd.oci.empty.v1+json",
+    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
+    "size": 2
+  },
+  "layers": [
+    {
+      "mediaType": "application/vnd.k8s.nfd.image-compatibility.spec.v1+json",
```
ChaoyiHuang, line 68: it could be "application/vnd.k8s.nfd.image-compatibility.spec.v1+json" or "application/vnd.k8s.nfd.image-compatibility.spec.v1+yaml" ?  in kubernetes, usually yaml is used.

enhancements/1845-nfd-image-compatibility/README.md
mfranczy: Fixed.

marquiz:
@mfranczy this would be good to get in, does it need to be updated wrt to the latest code PR`?

mfranczy:
It doesn't have to be updated. The proposed API is still the same and the proposal doesn't include any deep implementation suggestions. It can go as it is.

marquiz (approved):
Thank you @mfranczy for writing the proposal. Let's get this in before the impl.
/ok-to-test
/lgtm

Issue #1841, opened 2024-08-18
NFD Master memory leak


subnet-dev:
**What happened**:
We use node feature discovery 0.16.4 on our GPU cluster in addition to NVIDIA's gpu-operator. The GPU operator is installed with NFD disabled so that these tools can be managed independently. Node Feature Discovery Master continuously utilises more memory and never release it. Memory usage increases by around 1.2 Gi/day in a very linear way for the master, which is the leader. I've set a memory limit of 4Gi to prevent the pod from using more memory. The configuration file `nfd-master.conf` is empty (nfd-master.conf: ‘null’).

This screenshot was taken before rolling out the deployment:
<img width="1173" alt="Screenshot 2024-08-18 at 15 50 00" src="https://github.com/user-attachments/assets/eabaa211-14eb-45d9-acdd-cc6e268fb353">

This screenshot corresponds to the last pod start-up:
<img width="1175" alt="Screenshot 2024-08-18 at 15 55 07" src="https://github.com/user-attachments/assets/951f5fb1-90f8-4222-9d54-81aca369b7e5">

Here are the master logs since start-up:
```
I0817 19:08:35.561209       1 main.go:66] "-crd-controller is deprecated, will be removed in a future release along with the deprecated gRPC API"
I0817 19:08:35.562048       1 nfd-master.go:283] "Node Feature Discovery Master" version="v0.16.4" nodeName="admin007.xxx" namespace="node-feature-discovery"
I0817 19:08:35.562145       1 nfd-master.go:1381] "configuration file parsed" path="/etc/kubernetes/node-feature-discovery/nfd-master.conf"
I0817 19:08:35.562459       1 nfd-master.go:1429] "configuration successfully updated" configuration=<
	AutoDefaultNs: true
	DenyLabelNs: {}
	EnableTaints: false
	ExtraLabelNs: {}
	Klog: {}
	LabelWhiteList: null
	LeaderElection:
	  LeaseDuration:
	    Duration: 15000000000
	  RenewDeadline:
	    Duration: 10000000000
	  RetryPeriod:
	    Duration: 2000000000
	NfdApiParallelism: 10
	NoPublish: false
	ResourceLabels: {}
	ResyncPeriod:
	  Duration: 3600000000000
 >
I0817 19:08:35.562565       1 nfd-master.go:1513] "starting the nfd api controller"
I0817 19:08:36.363181       1 updater-pool.go:142] "starting the NFD master updater pool" parallelism=10
[...]

PR #1848, opened 2024-08-20
nfd-master: proper shutdown of nfd api informers

marquiz:
Stop blocking on event channels when the api controller is stopped.
Ensures that the nfd API informer factory is properly shut down and all
resources released when stop() is called. This eliminates a memory leak
on re-configure events when leader election is enabled.

Fixes: #1841


marquiz:
Still some investigation going on
/hold

marquiz:
Came up with a neater solution
/unhold

PR #1858, opened 2024-08-30
Bump Go to v1.23

marquiz:
NOTE: api/nfd/go.mod is not bumped. There should be no need to force the importers of the API to v1.23 (yet).


marquiz:
We need https://github.com/kubernetes/test-infra/pull/33392

ArangoGutierrez (approved):
/lgtm

PR #1871, opened 2024-09-10
source/system: Add reading product name information

superbrothers:
This PR adds Add reading product name information from `/sys/devices/virtual/dmi/id/product_name`.

I want to add this feature because I want to label based on the product name of the compute machine.


marquiz (approved):
Thank you @superbrothers for the improvement. Looks fine to me
/ok-to-test

Issue #1855, opened 2024-08-28
Docs: failed to discover kernel config and builtin modules in the development environment started using tilt


ChaoyiHuang:
**What happened**:
In the [developer guide](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/docs/developer-guide/index.md#running-with-tilt) , when running with tilt, nfd worker will not be able to discover kernel config and builtin modules

the error message is:
nfd-worker │ E0826 07:07:12.558316       1 kernel.go:134] "failed to read kconfig" err="failed to read kernel config from [ /proc/config.gz /host-usr/src/linux-6.5.0-1025-gcp/.config /host-usr/src/linux/.config /host-usr/lib/modules/6.5.0-1025-gcp/config /host-usr/lib/ostree-boot/config-6.5.0-1025-gcp /host-usr/lib/kernel/config-6.5.0-1025-gcp /host-usr/src/linux-headers-6.5.0-1025-gcp/.config /lib/modules/6.5.0-1025-gcp/build/.config /host-boot/config-6.5.0-1025-gcp]"
   
nfd-worker │ E0826 07:07:12.559161       1 kernel.go:149] "failed to get builtin kernel modules" err="failed to read file /host-lib/modules/6.5.0-1025-gcp/modules.builtin: open /host-lib/modules/6.5.0-1025-gcp/modules.builtin: no such file or directory"

**What you expected to happen**:
no matter what OS is, kernel config and builtin modules should be discovered correctly

**How to reproduce it (as minimally and precisely as possible)**:

1. start a virtual machine with OS unbuntu 22.04
2. following the [developer guide](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/docs/developer-guide/index.md#running-with-tilt) 

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`): ubuntu 22.04
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

ChaoyiHuang:
the issue is due to the k3d/kind cluster created by ctlptl will run inside a container(it will serve as the virtual host). 

[...]

PR #1889, opened 2024-09-29
Doc: Fix tilt up issue in feature discovering in developer guide

ChaoyiHuang:
The issue is due to the k3d/kind cluster created by ctlptl will run inside containers(it will serve as the virtual hosts).

Host folders which will be scaned by the nfd feature discovery should be mounted into the container ( the virtual host). otherwise the nfd-worker container which run inside the virtual host will just see the default base image rootfs /boot, /lib folders, which are usually empty, leads to the discovey failure.

### DIFF
--- a/docs/developer-guide/index.md
+++ b/docs/developer-guide/index.md
@@ -256,19 +256,27 @@ the steps below.
 1. Install [kustomize](https://github.com/kubernetes-sigs/kustomize)
 1. Install [tilt](https://docs.tilt.dev/install.html)
 1. Create a local Kubernetes cluster
+    - Create image registry first
+    - Create a Kubernetes cluster. Please note that docker containers will be
+      served as controller node and worker nodes, and NFD-worker will run as a
+      DaemonSet in nested container. Therefore, to make sure the NFD-worker can
+      discover the host features, the host folders "/boot" and "/lib" should be
+      mounted into worker node docker containers when creating the Kubernetes
+      cluster.
+1. Start up node feature discovery development environment
+    To start up your Tilt development environment, run at the root of your
+    local NFD codebase.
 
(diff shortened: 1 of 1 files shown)


fmuyassarov:
Fixes:  https://github.com/kubernetes-sigs/node-feature-discovery/issues/1855

docs/developer-guide/index.md
```diff
@@ -256,19 +256,44 @@ the steps below.
...
+            - volume: /boot:/boot
+              nodeFilters:
+                - server:0
+                - agent:*
+            - volume: /lib:/lib
+              nodeFilters:
+                - server:0
+                - agent:*
+      EOF
+      ```
```
fmuyassarov, line 290: Thanks for the example! I believe we should avoid including a specific cluster installation method since it ties us to particular installers, which aren't standardized. For instance, `v1alpha5Simple` embeds k3d-specific configuration. Could we instead add a note highlighting the directories that need to be mounted?

ChaoyiHuang: sounds good

marquiz:
Thank you @ChaoyiHuang for the patch. One nit: could you wrap the lines (like the rest of the file)? Other than that looks good to me.

@fmuyassarov any further comments?

/ok-to-test

marquiz:
> One nit: could you wrap the lines

Oh the CI (mdlint) was nitting about the same thing, and a few others

ChaoyiHuang:
> > One nit: could you wrap the lines
> 
> Oh the CI (mdlint) was nitting about the same thing, and a few others

yeah, fixed.

marquiz (approved):

fmuyassarov (commented):
Thank you!
/lgtm

PR #1897, opened 2024-10-03
topology-updater: Track new resources not present during startup

### LINKED ISSUE #1424: nfd-topology-updater only tracks resources it discovered during startup
**What happened**:
Resources that will be managed by NFD topology-updater are only discovered during startup.
If a device plugin registers itself with kubelet after NFD topology-updater is already running the NRT will never be updated with new resources.

Resource discovery [link](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/pkg/resourcemonitor/noderesourcesaggregator.go#L94)
Ignoring resources not present during startup [link](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/pkg/resourcemonitor/noderesourcesaggregator.go#L246).

It may cause issues when nfd is restarted alongside device plugins, in this case NRT will stop reporting some resources.

When device plugin is created after NFD start new resources should be reported without the need of restarting NFD. 

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

ozhuraki:
If a device plugin registers itself with kubelet after NFD topology-updater is already running the NRT will never be updated with new resources.

Closes: #1424

### DIFF
--- a/pkg/resourcemonitor/noderesourcesaggregator.go
+++ b/pkg/resourcemonitor/noderesourcesaggregator.go
@@ -239,8 +239,30 @@ func getContainerDevicesFromAllocatableResources(availRes *podresourcesapi.Alloc
 // updateAvailable computes the actually available resources.
 // This function assumes the available resources are initialized to be equal to the allocatable.
 func (noderesourceData *nodeResources) updateAvailable(numaData map[int]map[corev1.ResourceName]*resourceData, ri ResourceInfo) {
+	resName := string(ri.Name)
+	resMap, ok := noderesourceData.resourceID2NUMAID[resName]
+	if !ok {
+		resMap = make(map[string]int)
+		for _, numaNodeID := range ri.NumaNodeIds {
+			if _, ok := numaData[numaNodeID]; !ok {
+				klog.InfoS("failed to find NUMA node ID under the node topology", "numaID", numaNodeID)
+				continue
+			}
+
+			if _, ok := numaData[numaNodeID][ri.Name]; !ok {
+				klog.InfoS("failed to find resource under the node topology", "resourceName", ri.Name)
+				continue
+			}
+
+			for _, resID := range ri.Data {
+				resMap[resID] = numaNodeID
+			}
+		}
+
+		noderesourceData.resourceID2NUMAID[resName] = resMap
+	}
+
 	for _, resID := range ri.Data {
-		resName := string(ri.Name)
 		resMap, ok := noderesourceData.resourceID2NUMAID[resName]
 		if !ok {
 			klog.InfoS("unknown resource", "resourceName", ri.Name)


PiotrProkop:
I don't think that's the right place to fix it as later call to `updateAvailable` will ignore the resource not present during startup here -> https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/pkg/resourcemonitor/noderesourcesaggregator.go#L246

ozhuraki:
@PiotrProkop

Thanks for the help. I checked more, would something like this account for updateAvailable()?

pkg/resourcemonitor/noderesourcesaggregator.go
```diff
@@ -152,6 +152,29 @@ func (noderesourceData *nodeResources) Aggregate(podResData []PodResources) topo
 				available:   int64(0),
 				capacity:    int64(len(noderesourceData.reservedCPUIDPerNUMA[nodeID])),
 			}
+
+			for resName, allocatable := range nodeRes {
```
PiotrProkop, line 156: this `else` statement means that given NUMA node doesn't exist, not that a resource was absent during NFD updater startup. And the missing resources are already in `perNuma` map, the problem is that they are being rejected in `updateAvailable` func

ozhuraki:
@PiotrProkop

Apologies for a delay. Thanks, updated, could you please take another look?

marquiz:
@PiotrProkop could you take a look?

/ok-to-test
/remove-lifecycle rotten

marquiz:
@ozhuraki please fix the linter error(s)

PiotrProkop:
@ozhuraki please add unit test for this functionality. I'll try to review more in depth today/tomorrow.

marquiz:
@ozhuraki any update on this one?

/remove-lifecycle rotten

pkg/resourcemonitor/noderesourcesaggregator.go
```diff
@@ -239,8 +239,30 @@ func getContainerDevicesFromAllocatableResources(availRes *podresourcesapi.Alloc
...
+				klog.InfoS("failed to find resource under the node topology", "resourceName", ri.Name)
+				continue
+			}
+
+			for _, resID := range ri.Data {
+				resMap[resID] = numaNodeID
+			}
+		}
+
+		noderesourceData.resourceID2NUMAID[resName] = resMap
```
ArangoGutierrez, line 262: This write makes the miss permanent: a resource absent from the startup snapshot is also absent from `numaData` (both are derived from the same `GetAllocatableResources` response), so the guards above always `continue` and an empty map gets cached here — after which this `!ok` branch never runs again for that resource. And even when the guards pass, the map is built from this one container's `ri.Data`, so devices of the same resource allocated to other pods still hit 'unknown resource ID'. The mapping has to come from a re-queried `GetAllocatableResources`, not from per-container allocations.

ArangoGutierrez (commented):
This targets a real gap: the aggregator snapshots `GetAllocatableResources` once at startup, so device plugins that register later never make it into the NRT. The patch can't close that gap from inside `updateAvailable`, though — the NRT zones are seeded exclusively from the startup `perNUMAAllocatable`, so even with the existing inline comments addressed, a late resource would at best get `available` decremented but still never appear with allocatable/capacity in the output. The fix needs to refresh the allocatable snapshot instead (re-query `GetAllocatableResources` in `Aggregate`, or rebuild the aggregator when an unknown resource shows up), which would also cover the symmetric case of a plugin unregistering.

Piotr's and the Copilot review comments still stand, the unit test requested in May 2025 is still missing, and both required CI jobs (verify, image-cross-build) are red. The touched code is unchanged on current master, so a rebase should be conflict-free. I'd hold this until the approach moves to refreshing the allocatable data — happy to re-review a respin.

ArangoGutierrez:
/retest

The CI failures here are environmental, not this change: `verify-master` was the dead-Keybase Codecov bug (fixed in master via #2527) and `build-image-cross-generic` hit a transient Docker Hub token 404. Re-running.

@ozhuraki — could you add a unit test for the fix? Ideally a case in `noderesourcesaggregator_test.go` where a resource appears *after* initialization (absent from the initial `allDevs`/`makeResourceMap`) and asserts that `updateAvailable` now tracks it (decrements availability) instead of logging "unknown resource". That would lock in the #1424 fix and the new lazy-mapping block.

pkg/resourcemonitor/noderesourcesaggregator.go
```diff
@@ -239,8 +239,30 @@ func getContainerDevicesFromAllocatableResources(availRes *podresourcesapi.Alloc
 // updateAvailable computes the actually available resources.
 // This function assumes the available resources are initialized to be equal to the allocatable.
 func (noderesourceData *nodeResources) updateAvailable(numaData map[int]map[corev1.ResourceName]*resourceData, ri ResourceInfo) {
+	resName := string(ri.Name)
+	resMap, ok := noderesourceData.resourceID2NUMAID[resName]
+	if !ok {
+		resMap = make(map[string]int)
+		for _, numaNodeID := range ri.NumaNodeIds {
+			if _, ok := numaData[numaNodeID]; !ok {
+				klog.InfoS("failed to find NUMA node ID under the node topology", "numaID", numaNodeID)
```
PiotrProkop, line 248: what if a new resource was reported that's on the numa node not seen before?

pkg/resourcemonitor/noderesourcesaggregator.go
```diff
@@ -239,8 +239,30 @@ func getContainerDevicesFromAllocatableResources(availRes *podresourcesapi.Alloc
...
+				continue
+			}
+
+			if _, ok := numaData[numaNodeID][ri.Name]; !ok {
+				klog.InfoS("failed to find resource under the node topology", "resourceName", ri.Name)
+				continue
+			}
```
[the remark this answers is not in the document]
PiotrProkop: I agree.

ArangoGutierrez:
Following up on my earlier review — I need to correct it. Tracing the data flow more carefully: `Aggregate` builds `perNuma` (the `numaData` passed to `updateAvailable`, and the source of the output Zones) solely from `perNUMAAllocatable`, which is the startup `GetAllocatableResources` snapshot — and `NewResourcesAggregator` runs once, outside the scan loop, so that snapshot is never refreshed.

A device plugin that registers *after* startup isn't in that snapshot, so in the new block the guard `if _, ok := numaData[numaNodeID][ri.Name]; !ok { continue }` always fires for it, `resMap` stays empty and gets cached (`resourceID2NUMAID[resName] = resMap`), and on later scans the `!ok` rebuild branch is skipped. The late resource also never appears in the output, since the Zones are built from `perNuma`. So this can't resolve #1424 as written — the missing test isn't the only gap.

To actually fix it, the allocatable snapshot needs refreshing when an unknown resource shows up — e.g. re-query `GetAllocatableResources` / rebuild the aggregator in `Aggregate` — which also keeps `allocatable`/`capacity` correct in the NRT, not just `available`. (There's also a nil-deref risk at `numaData[nodeID][ri.Name].available--` for a resource absent from `numaData`.)

Apologies for the misdirection in my earlier pass. Happy to help scope the snapshot-refresh approach.

Issue #1072, opened 2023-03-03
Status for NodeFeatureRule CRD


marquiz:
**What would you like to be added**:

Add status field to the NodeFeatureRule CRD, indicating e.g. when the NodeFeatureRule was last processed, which nodes (and when) were affected, any errors etc.

**Why is this needed**:

Better visibility to the current state of matters

marquiz:
Revived by the discussion in #1487, we could add at least the list of nodes (that matched) in the status. And that should probably be per Rule

```yaml
status:
  rules:
  - name: "rule-1"
    matchedNodes:
    - node-1
    - node-3
  - name: "rule-2"
    matchedNodes:
    - node-22
```

ArangoGutierrez:
https://github.com/kubernetes-sigs/node-feature-discovery/pull/1664#issuecomment-2129609302

marquiz:
> @marquiz is this not needed?

I think this would be cool. Just the design is totally open. At least I haven't deeply thought what (in which format) should be present in the status field.

PR #1901, opened 2024-10-06
nfd-master: Add status for NodeFeatureRule CRD

ozhuraki:
Closes: #1072


ozhuraki:
@ArangoGutierrez

Thanks for the input. Take a look, please if this would address extra updates?

api/nfd/v1alpha1/types.go
```diff
@@ -131,6 +131,8 @@ type NodeFeatureRule struct {
 
 	// Spec defines the rules to be evaluated.
 	Spec NodeFeatureRuleSpec `json:"spec"`
+	// +optional
+	Status NodeFeatureRuleStatus `json:"status"`
```
fmuyassarov, line 135:
```suggestion
	Status NodeFeatureRuleStatus `json:"status,omitempty"`
```

api/nfd/v1alpha1/types.go
```diff
@@ -139,6 +141,36 @@ type NodeFeatureRuleSpec struct {
...
 
+// NodeFeatureRuleStatus represents the status of a NodeFeatureRule
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=nfs
+// +kubebuilder:subresource:status
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +genclient
+type NodeFeatureRuleStatus struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
fmuyassarov, line 152: We don’t need to include metadata in the Status, as it doesn't add any additional value there.

api/nfd/v1alpha1/types.go
```diff
@@ -139,6 +141,36 @@ type NodeFeatureRuleSpec struct {
...
+
+	// +optional
+	Rules []RuleStatus `json:"rules"`
```
fmuyassarov, line 155:
```suggestion
	Rules []RuleStatus `json:"rules,omitempty"`
```

api/nfd/v1alpha1/types.go
```diff
@@ -139,6 +141,36 @@ type NodeFeatureRuleSpec struct {
 	Rules []Rule `json:"rules"`
 }
 
+// NodeFeatureRuleStatus represents the status of a NodeFeatureRule
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=nfs
+// +kubebuilder:subresource:status
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +genclient
```
fmuyassarov, line 149: These markers were replaced by `kubebuilder:object:root=true`. So, we could drop this off too. Reference: https://book.kubebuilder.io/migration/legacy/migration_guide_v1tov2.html?highlight=genclient#migrate-the-apis

api/nfd/v1alpha1/types.go
```diff
@@ -139,6 +141,36 @@ type NodeFeatureRuleSpec struct {
 	Rules []Rule `json:"rules"`
 }
 
+// NodeFeatureRuleStatus represents the status of a NodeFeatureRule
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:scope=Namespaced,shortName=nfs
+// +kubebuilder:subresource:status
```
fmuyassarov, line 147: These should be organized under the parent object NodeFeatureRule. Since they already exist, you could just drop them. You can find the relevant section here: https://github.com/kubernetes-sigs/node-feature-discovery/blob/955095c7eb2e7afe2276cd0b6a2dab44042c5ded/api/nfd/v1alpha1/types.go#L123. .

ozhuraki:
@fmuyassarov

Thanks, updated, please take a look

PR #1906, opened 2024-10-14
github: specify workflow permissions

marquiz:
Update gh-pages has started failing with default permissions. It needs write permissions to be able to push to the gh-pages branch.


tobiasgiese:
seems reasonable to me
/lgtm

PR #1922, opened 2024-10-22
feat: install CRDs from a subchart

### LINKED ISSUE #1637: Install CRDs from a subchart instead of using Helm 3 crds directory
Depart away from Helm 3 way to install CRDs and use a subchart for it.

**Why is this needed**:

Moving CRDs to a chart solves several issues: clean uninstall, possibility for upgrades, templating.

## Motivation and Context

My main motivation is to have a better experience when managing this operator with Argo CD. That said, moving away from Helm 3 approach to install CRDs might be helpful in other scenarios too. I'm posting the full rationale below.

Helm 3 does not manage CRDs (see https://helm.sh/docs/chart_best_practices/custom_resource_definitions/). `helm uninstall` won't remove CRDs, and `helm updgrade` won't upgrade them. Manual intervention is required with the current setup.
- CRDs installed from `crds` folder **are not included into the helm release**.
- Additionally, it is not possible to template CRDs in Helm 3 in the `crds` folder.

The following comes from helm best practices around CRDs:

> There is no support at this time for upgrading or deleting CRDs using Helm. This was an explicit decision after much community discussion due to the danger for unintentional data loss.

ref: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-1-let-helm-do-it-for-you

Another alternatively, which is what I propose and comes from Helm page:

> Another way to do this is to put the CRD definition in one chart, and then put any resources that use that CRD in another chart.
>
> In this method, each chart must be installed separately. However, this workflow may be more useful for cluster operators who have admin access to a cluster

ref: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts

**I'm completely fine to open a PR with the required refactoring, without impacting current installations**.

cmontemuino:
ROBLEM: Helm 3 only install CRDs during the chart installation. Both deletes and upgrades are not handled by Helm 3. It requires manual intervention. Additionally, it is not possible to template CRDs. This is especially important when installing the chart with an Argo CD application.

SOLUTION: add the possibility to install the CRDs from a dependency subchart. This allows clean uninstall, possibility for automated upgrades, and templating. This feature comes disabled by default.

closes #1637


TessaIO:
Thanks for the patch @cmontemuino. I saw, that you added the new CRD installation without removing the old CRDs manifests. Wouldn't that be an issue if we installed both Helm charts (the NFD and CRD) as it would try to create the same CRD twice and fail?

cmontemuino:
> Thanks for the patch @cmontemuino. I saw, that you added the new CRD installation without removing the old CRDs manifests. Wouldn't that be an issue if we installed both Helm charts (the NFD and CRD) as it would try to create the same CRD twice and fail?

Thanks for raising this point @TessaIO, as it requires more clarification.

My initial take was to make this feature completely optional. The assumption here is that there could be a use case for using Helm 3 CRDs approach.

This is what I've added in the `values.yaml`:

```yaml
# -- Toggle to install and upgrade CRDs from a subchart. Make sure to use it with `--skip-crds` to avoid conflicts.
```

And also reflected it in the `helm.md` file -- general parameters section.

I'd say a section in `helm.md` detailing this new addition makes sense. That's where I'm struggling about how the documentation is generated. For example, should I add a new section in the same lines as in https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/docs/deployment/helm.md#from-v014 ?

docs/deployment/helm.md
```diff
@@ -149,6 +149,8 @@ Chart parameters are available.
 
 | Name                                                | Type   | Default                                             | Description                                                                                                                                                                                                  |
 | --------------------------------------------------- | ------ | --------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `crds.install`                                      | bool   | false                                               | Toggle to install and upgrade CRDs from a subchart. Make sure to use it with `--skip-crds` to avoid conflicts. [More info about limitations on CRDs in Helm 3](https://helm.sh/docs/topics/charts/#limitations-on-crds) |
```
TessaIO, line 152: If i'm not mistaken, I think the `--skip-crds` flag is used for the `crds` folder and not the separate helm chart. so I don't think that they're related. Please correct if I'm wrong.

TessaIO (commented):
Thanks @cmontemuino for the clarifications. I put some comments, I'll review it one more time once this is ready.

deployment/helm/node-feature-discovery/Chart.yaml
```diff
@@ -13,3 +13,7 @@ keywords:
   - node-labels
 type: application
 version: 0.2.1
+dependencies:
+  - name: crds
+    version: 0.0.0
```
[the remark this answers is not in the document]
cmontemuino: There's a drawback to use a version here: you need bump the version and `helm dependency update` with every release.

The versioning usually makes sense when adding charts from repos, with different lifecycles.

In our case, CRDs are always generated and copied under `crds` folder: https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/hack/update_codegen.sh#L44-L46

There's a 1-1 match in the lifecycle for this chart. Thus, the `version: 0.0.0` makes the trick to avoid extra steps.

docs/deployment/helm.md
cmontemuino: You're right about the flag 👍 

The problem happens only when the chart is installed by the first time. In that scenario: `helm install xxxx --skip-crds --set crds.install=true`

(*apologies for the oversimplification above*)

Otherwise, Helm will install CRDs from both places. Again, this only happens the first time. Helm won't care about the CRDs folder during upgrades (nor uninstall)

deployment/helm/node-feature-discovery/charts/crds/values.yaml
```diff
@@ -0,0 +1,3 @@
+# -- Define annotations for CRDs
+annotations: {}
+#   argocd.argoproj.io/sync-options: Replace=false,ServerSideApply=true
```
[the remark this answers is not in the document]
cmontemuino: Got it now. Thanks!!

docs/deployment/helm.md
fmuyassarov: Are you keeping an additional copy of the CRD in the /crds directory only for the initial installation of the chart?
I'm wondering if it might be too strict to always require using `--skip-crds --set crds.install=true`.

cmontemuino: My take was make the new approach optional. That is, if you do nothing, then the CRDs get installed as they are now.
By the other hand, if you want to taste the new approach, then `--skip-crds --set crds.install=true`.

I wonder if this is a good compromise, or if we should go the sub-chart with CRDs approach in one single shot. WDYT @fmuyassarov 

Regarding additional copies, all of them are automatically generated.

cmontemuino:
It's a pity this project lacks enough contributors. I think it's a great project.

cmontemuino:
@TessaIO -- are you still waiting for changes to be done in this PR? I've provided more info to your initial comments already.

PR #1929, opened 2024-10-23
nfd-worker: use single http port for metrics and healthz

marquiz:
This PR drops the separate http servers for metrics and the health endpoint. In addition, ditch the grpc-health and replace it with plain http.

The PR is split into three commits:

1. Drop the gRPC health port
2. Convert `--metrics` to `--port`
    Use a single port for serving http. In addition to metrics we will have the healthz endpoint.
3. Add healthz endpoint


marquiz:
This is kinda RFC to get feedback. The http endpoints are "read-only" so I think we might as well serve them from a single port. Also, a single port consumed in host network scenarios. One motivation is to get rid of all direct gRPC dependencies, too.

One caveat/note is that in it's current form this doesn't deprecate the existing command line flags or Helm values. We could do that, e.g. with some sort of extra logic and override/fallback behavior around all args (`-port`, `-metrics`, `-grpc-health`) but I'm not sure it's worth the trouble.

If/when we're happy with this we can make the other daemons (`nfd-master`, `nfd-topology-updater`, `nfd-gc`) to follow this.
/hold

/cc @ArangoGutierrez @tobiasgiese @adrianchiris

tobiasgiese:
> serve them from a single port. Also, a single port consumed in host network scenarios.

IMO we should not care if we expose 1 or 3 ports, there should be enough ports free on a system. Also the ports are configurable, thus it is totally fine to me if we'd keep multiple ports. Nevertheless I like the idea having a single port :)

> One caveat/note is that in it's current form this doesn't deprecate the existing command line flags or Helm values.

I think it is okay to consider this as a breaking change and highlight this in the release notes. If we deprecate the ports we also have to ensure that the ports are listening and serve the correct content (i.e., both ports will handle the same).

IMO the change is a good improvement to our code and the UX for the users 👍🏻 
/lgtm

PiotrProkop:
> @PiotrProkop any comments?

Always :) I prefer single port implementation as it removes some code and make overall architecture simpler.

ArangoGutierrez (approved):
/lgtm

ArangoGutierrez:
unhold @marquiz ??

marquiz:
Rebased. ping @tobiasgiese @ArangoGutierrez

PR #1932, opened 2024-10-28
Introduce nfd client for image compatibilty

mfranczy:
The PR provides an initial implementation of https://github.com/kubernetes-sigs/node-feature-discovery/pull/1845. Image compatibility is defined by Node Feature Rules, so a change was also necessary there. A run strategy pattern has been implemented to control rule execution, allowing either all rules to be executed or stopping at the first failure.

~~The work is still in progress but is ready for the first review round.~~
~~Remaining tasks:~~
~~- [x] write unit tests~~
~~- [x] write documentation and add appropriate code comments~~
~~- [x] refine the commands~~


cmd/client-nfd/__debug_bin3215279485
ArangoGutierrez, line 1: Please remove this

cmd/client-nfd/main.go
```diff
@@ -0,0 +1,27 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"sigs.k8s.io/node-feature-discovery/cmd/client-nfd/subcmd"
```
ArangoGutierrez, line 20: Let's try to handle as we do for `nfd-master` and `nfd-worker` to maintain the project structure. (see how we hande flags and sub commands for those 2

ArangoGutierrez (changes requested):

api/image-compatibility/v1alpha1/spec.go
```diff
@@ -0,0 +1,35 @@
...
+package v1alpha1
+
+import (
+	nfdv1alpha1 "sigs.k8s.io/node-feature-discovery/api/nfd/v1alpha1"
+)
+
+const ArtifactType = "application/vnd.k8s.nfd.image-compatibility.v1"
+
+type Spec struct {
+	Version        string          `json:"string"`
```
ChaoyiHuang, line 26: the json field name should be "version", instead "string"

cmd/client-nfd/main.go
mfranczy: Right, I was too fast. Will do the same as you do for `nfd-master` and `nfd-worker`.

cmd/client-nfd/__debug_bin3215279485
mfranczy: Removed.

cmd/client-nfd/subcmd/compat/validate-node.go
```diff
@@ -0,0 +1,125 @@
...
+				for feature, elems := range features {
+					t.AppendSeparator()
+					for i, elem := range elems {
+
+						status := text.FgRed.Sprintf("%s", elem[pkgcompat.MatchKeyStatus])
+						if elem[pkgcompat.MatchKeyStatus] == pkgcompat.MatchStatusOK {
+							status = text.FgGreen.Sprintf("%s", elem[pkgcompat.MatchKeyStatus])
+						}
+
+						t.AppendRow(table.Row{strconv.Itoa(i + 1), fmt.Sprintf(
```
ChaoyiHuang, line 101: lots of indent layers in a function, to make it easier to read, it's better to move some logic to another function. usually > 3 indent layers in one function will lead to difficult in reading the code.

cmd/client-nfd/main.go
mfranczy: @ArangoGutierrez one question: aside from the `platform` option, I followed the style used in the [kubectl cmd](https://github.com/kubernetes-sigs/node-feature-discovery/tree/master/cmd/kubectl-nfd). Should the kubectl cmd also be refactored to align with the `nfd-master` and `nfd-worker` styles? I could open a separate PR for that.

cmd/client-nfd/subcmd/compat/validate-node.go
mfranczy: Fixed.

pkg/apis/nfd/nodefeaturerule/rule.go
```diff
@@ -31,32 +31,56 @@ import (
...
 // RuleOutput contains the output out rule execution.
 // +k8s:deepcopy-gen=false
 type RuleOutput struct {
+	Name              string
 	ExtendedResources map[string]string
 	Labels            map[string]string
 	Annotations       map[string]string
 	Vars              map[string]string
 	Taints            []corev1.Taint
+	Matches           MatchedFeatures
```
mfranczy, line 59: @marquiz, what do you think about transforming the `MatchedFeatures` type into a type that reports not only features that match with the node but also those that don’t, perhaps called `CheckedFeatures`? This way, I wouldn’t need to introduce [complex comparison functions](https://github.com/kubernetes-sigs/node-feature-discovery/blob/10972517a430ca4dee4fc06aa31e3a1a7d599332/pkg/client-nfd/compat/validate-node.go#L61-L64) to identify which expressions passed or didn’t. For image compatibility, it’s also important to return information about expressions that don’t match the node.

In several places, there are log messages with matched features, which we can retain by adding a filter for the `CheckedFeatures` type. What do you think?

mfranczy: After trying it out, it turned out to be too invasive and computation time to be longer for each rule processing. I've decided to simplify implementation on the cli side.

cmd/client-nfd/main.go
ArangoGutierrez: Let's create an issue for that and set as a TODO, but let's focus on getting this one in

ArangoGutierrez:
@marquiz PTAL

api/nfd/v1alpha1/types.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	fmt "fmt"
```
ChaoyiHuang, line 20: redundant alias

ArangoGutierrez: ++

cmd/client-nfd/subcmd/root.go
```diff
@@ -0,0 +1,46 @@
+/*
```
ArangoGutierrez, line 1: let's move all `subcmd` folder  to `pkg/client-nfd` same as `nfd-master` or `nfd-worker`

ArangoGutierrez (changes requested):

pkg/client-nfd/compat/node-validator/node-validator.go
```diff
@@ -0,0 +1,113 @@
...
+
+type Args struct {
+	RegReference *registry.Reference
+	Tags         []string
+	Platform     *ocispec.Platform
+	PlainHTTP    bool
+}
+
+type nodeValidator struct {
+	Args Args
```
ChaoyiHuang, line 50: member name is overlapped with type name

pkg/client-nfd/compat/node-validator/node-validator.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+type Args struct {
+	RegReference *registry.Reference
+	Tags         []string
+	Platform     *ocispec.Platform
+	PlainHTTP    bool
+}
+
+type nodeValidator struct {
```
ChaoyiHuang, line 49: a type can be used in other package should be started with Caption letter

examples/compatibility-6-wind-router-rules.yaml
```diff
@@ -0,0 +1,31 @@
+version: v1alpha1
```
ChaoyiHuang, line 1: it's better to remove the company name in the file name, for example compatibility-nfd-rules.yaml

examples/compatibility-6-wind-router-rules.yaml
```diff
@@ -0,0 +1,31 @@
...
+        ip6_tables: {op: Exists}
+        ip_tables: {op: Exists}
+        mpls_iptunnel: {op: Exists}
+        mpls_router: {op: Exists}
+        nf_conntrack: {op: Exists}
+        overlay: {op: Exists}
+        ppp_generic: {op: Exists}
+        vfio-pci: {op: Exists}
+        vhost-net: {op: Exists}
+        vrf: {op: Exists}
```
ChaoyiHuang, line 31: how about add another example, for example:
```
version: v1alpha1
compatibilities:
- description: "intel cpu and ubuntu OS"
  rules:
    - name: "ubuntu"
      matchFeatures:
        - feature: cpu.model
          matchExpressions:
            vendor_id: { op: In, value: [ "Intel" ] }
        - feature: kernel.config
          matchExpressions:
            CGROUP_BPF: { op: In, value: [ "y" ] }
            XFRM_INTERFACE: { op: In, value: [ "y" ] }
            NF_TABLES: { op: In, value: [ "y" ] }
            NETFILTER_XT_TARGET_NOTRACK: { op: In, value: [ "y" ] }
          matchExpressions:
            br_netfilter: { op: Exists }
            ebtables: { op: Exists }
            ifb: { op: Exists }
            ip_tables: { op: Exists }
            nf_conntrack: { op: Exists }
            overlay: { op: Exists }
            ppp_generic: { op: Exists }
            vfio-pci: { op: Exists }
            vhost-net: { op: Exists }
            vrf: { op: Exists }
        - feature: system.osrelease
          matchExpressions:
            ID: { op: In, value: [ "ubuntu" ] }
            VERSION_ID: { op: In, value: [ "20.04", "22.04" ] }
    - name: "nic"
      matchAny:
        - matchFeatures:
          - feature: pci.device
            matchExpressions:
              vendor: {op: In, value: ["0eee"]}
              class: {op: In, value: ["0200"]}
        - matchFeatures:
          - feature: pci.device
            matchExpressions:
              vendor: {op: In, value: ["0fff"]}
              class: {op: In, value: ["0200"]}
- description: "Amd cpu and RHEL"
  rules:
    - name: "RHEL"
      matchFeatures:
        - feature: cpu.model
          matchExpressions:
            vendor_id: { op: In, value: [ "Amd" ] }
        - feature: kernel.config
          matchExpressions:
            CGROUP_BPF: { op: In, value: [ "y" ] }
            MPLS_ROUTING: { op: In, value: [ "y" ] }
            MPLS_IPTUNNEL: { op: In, value: [ "y" ] }
            PPPOE: { op: In, value: [ "y" ] }
          matchExpressions:
            ip6_tables: { op: Exists }
            ip_tables: { op: Exists }
            mpls_iptunnel: { op: Exists }
            mpls_router: { op: Exists }
            vfio-pci: { op: Exists }
            vhost-net: { op: Exists }
            vrf: { op: Exists }
        - feature: system.osrelease
          matchExpressions:
            ID: { op: In, value: [ "rhel" ] }
            VERSION_ID: { op: In, value: [ "8" ] }
```
in which illustrates other way to organize the compatibilities, and show how matchAny could be used in the rule

cmd/client-nfd/subcmd/compat/validate-node.go
```diff
@@ -0,0 +1,141 @@
...
+	image     string
+	tags      []string
+	platform  options.PlatformOption
+	plainHTTP bool
+)
+
+var validateNodeCmd = &cobra.Command{
+	Use:   "validate-node",
+	Short: "Validate node based on image compatibility metadata",
+	Long:  "Validate node based on image compatibility metadata from the NFD artifact",
```
ChaoyiHuang, line 45: how about update "from the NFD artifact" to "from the image associated comaptibility artifact"

cmd/client-nfd/subcmd/compat/validate-node.go
```diff
@@ -0,0 +1,141 @@
...
+
+		for _, r := range cs.Rules {
+			printTable(r)
+		}
+		fmt.Println()
+	}
+}
+
+func printTable(r nodevalidator.RuleStatus) {
+	t := table.NewWriter()
```
ChaoyiHuang, line 94: for matchAny, (refer to the example commented in the file compatibilit-6-wind-router-rule.yaml), 

```
    - name: "nic"
      matchAny:
        - matchFeatures:
          - feature: pci.device
            matchExpressions:
              vendor: {op: In, value: ["0eee"]}
              class: {op: In, value: ["0200"]}
        - matchFeatures:
          - feature: pci.device
            matchExpressions:
              vendor: {op: In, value: ["0fff"]}
              class: {op: In, value: ["0200"]}
```


the result is printed like this:

```
┌──────────────────────────────────────────────────────────────┐
│ RULE: NIC -  FAIL                                            │
├───┬───────────────────┬─────────────────────────────┬────────┤
│ # │ FEATURE           │ EXPRESSION                  │ STATUS │
├───┼───────────────────┼─────────────────────────────┼────────┤
│ 1 │ pci.device.class  │ {op: "In", value: ["0200"]} │ FAIL   │
│ 2 │ pci.device.vendor │ {op: "In", value: ["0eee"]} │ FAIL   │
│ 3 │ pci.device.class  │ {op: "In", value: ["0200"]} │ FAIL   │
│ 4 │ pci.device.vendor │ {op: "In", value: ["0fff"]} │ FAIL   │
└───┴───────────────────┴─────────────────────────────┴────────┘
```
it's little difficult to distinguish "one of the matchAny" failed, it just listed all field result in flat

pkg/client-nfd/compat/node-validator/node-validator.go
```diff
@@ -0,0 +1,113 @@
...
+	Args Args
+}
+
+func New(opts ...NodeValidatorOpts) nodeValidator {
```
ChaoyiHuang, line 53: it's same effect to initialize a new NodeValidator without introduce extra nodeNolidatorOpts

```
func New2(args *Args) NodeValidator {
	return NodeValidator{
		Args: *args,
	}
}
```

usually a pointer is preferred to return in initialization, like 

```
func New3(args *Args) *NodeValidator {
	return &NodeValidator{
		Args: *args,
	}
}
```

api/image-compatibility/v1alpha1/spec.go
```diff
@@ -0,0 +1,35 @@
...
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	nfdv1alpha1 "sigs.k8s.io/node-feature-discovery/api/nfd/v1alpha1"
+)
+
+const ArtifactType = "application/vnd.k8s.nfd.image-compatibility.v1"
```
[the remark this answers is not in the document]
mfranczy: Right. Fixed.

pkg/client-nfd/compat/node-validator/node-validator.go
mfranczy: That's not a big deal since it's a different namespace. But it should not be exported as the type was not. Fixed.

cmd/client-nfd/subcmd/compat/validate-node.go
mfranczy: Yes, I will update the whole doc in the PR.

pkg/client-nfd/compat/node-validator/node-validator.go
mfranczy: It could work like this, although, the other commands (like `nfd-master` and `nfd-worker`) implement the options pattern. As I was asked, I am trying to stay as close as possible to other commands initialization.

examples/compatibility-6-wind-router-rules.yaml
mfranczy: What about placing something more simple (as introduced in the proposal)?
```yaml
version: v1alpha1
compatibilities:
- tag: "prefered"
  weight: 10
  description: "Prefered node configuration"
  rules:
  - name: "kernel and cpu"
    matchFeatures:
      matchExpressions:
        vfio-pci: {op: Exists}
    - feature: cpu.model
      matchExpressions:
        vendor_id: {op: In, value: ["Intel", "Amd"]}
- tag: "fallback"
  weight: 1
  description: "Minimal required configuration"
  rules:
  - name: "cpu"
    matchFeatures:
    - feature: cpu.model
      matchExpressions:
        vendor_id: {op: In, value: ["Intel", "Amd"]}
```

The example is easy to understand and run.

ChaoyiHuang: o, it's not about simplification, it's to include "matchAny" in the example

mfranczy: I will incldue it.

cmd/client-nfd/subcmd/compat/options/platform.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func (opt *PlatformOption) Parse(*cobra.Command) error {
+	var pStr string
+
+	if opt.PlatformStr == "" {
+		return nil
+	}
+
+	p := &ocispec.Platform{}
```
ArangoGutierrez, line 40:
```suggestion
	platform := &ocispec.Platform{}
```
for readability

cmd/client-nfd/subcmd/root.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/node-feature-discovery/cmd/client-nfd/subcmd/compat"
+)
+
+// RootCmd represents the base command when called without any subcommands
+var RootCmd = &cobra.Command{
+	Use:   "nfd",
+	Short: "NFD client",
+	Long:  `NFD client TBD`,
```
ArangoGutierrez, line 32: If TBD let's remove it, I think use and short are enough

ArangoGutierrez (changes requested):

cmd/client-nfd/subcmd/compat/validate-node.go
mfranczy: True, it's not easy to do this. Let me figure out how to represent matchAny.

cmd/client-nfd/subcmd/compat/options/platform.go
```diff
@@ -0,0 +1,65 @@
...
+	pStr, p.OSVersion, _ = strings.Cut(opt.PlatformStr, ":")
+	parts := strings.Split(pStr, "/")
```
ChaoyiHuang, line 42: if pStr is empty, the length of parts will not be 0, but 1, i.e, the parts will include one element which is "". 

if pStr is empty, it should go to the default branch in the next switch clause, but because the parts length is not 0, unfortunately it goes to case 1. that means the default clause will never be reached.

cmd/client-nfd/subcmd/compat/validate-node.go
mfranczy: I added `matchAny` representation by adding a separtor between the terms, like (look at the `one of available nics`):
image

pkg/client-nfd/compat/node-validator/node-validator.go
mfranczy: This structure should not be exported as it holds the internal fields for the `validate-node` cmd.

cmd/client-nfd/subcmd/compat/validate-node.go
mfranczy: I've also added the backreference and matchAny features to the example.

cmd/client-nfd/subcmd/compat/validate-node.go
mfranczy: Done.

examples/compatibility-6-wind-router-rules.yaml
mfranczy: I've also added the backreference and matchAny features to the example.

examples/image-compatibility.yaml
```diff
@@ -0,0 +1,49 @@
...
+          vendor: {op: In, value: ["0fff"]}
+          class: {op: In, value: ["0200"]}
+  - name: "prefered requirements"
+    labels:
+      my-prefered-img-requirements: "true"
+    matchFeatures:
+    - feature: rule.matched
+      matchExpressions:
+        my-prefered-kernel-and-cpu: {op: IsTrue}
+        my-prefered-img-requirements: {op: IsTrue}
```
ChaoyiHuang, line 38: reference to itself ? seems it's a self-loop reference. little complex to figure out the real use case.

maybe put the advanced back reference in another example

mfranczy: Ok I will put it to another example. The backreference feature has been described here https://kubernetes-sigs.github.io/node-feature-discovery/v0.16/usage/customization-guide.html#backreferences

cmd/client-nfd/subcmd/root.go
mfranczy: As mentioned, the client will include more subcommands so it cannot be adjusted to the `nfd-master` or `nfd-worker` style. Resolving the comment, feel free to open if you want to re-discuss.

mfranczy:
@marquiz @ArangoGutierrez I consider the PR finished and fully reviewable, please do another round once you're free.

I had to introduce a run strategy for rules execution. The validate-node command is expected to execute all rules and return all features that match the node. Initially, I wanted to return not only the matching features but also those that fail; however, implementing this required extensive changes to the Node Feature Rules code, which made the PR too large. To simplify, I decided to implement the return of failed features in upcoming PRs, as a discussion is also needed to find the most optimized solution for this.

You can view all the node feature rules API changes in this commit: [link](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1932/commits/3dd5db5879748a4ab03f9b42a2b7eb49d3880d40)

Additionally, I’m including a sample output from the  `nfd compat validate-node..` cmd
image

I belive this first implementation is good enough to be included in v0.17.0 release.

ChaoyiHuang:
> Additionally, I’m including a sample output from the `nfd compat validate-node..` cmd image
> 
> I belive this first implementation is good enough to be included in v0.17.0 release.

if matched features can be returned and output correctly, then all other unmatched feature should be failure in feature matching. 

so maybe no need to introduce runstrategy, just in the output, for the features not in the matched feature list, tag it as failed.

then we may get same report as that with runstrategy

mfranczy:
> > Additionally, I’m including a sample output from the `nfd compat validate-node..` cmd image
> > I belive this first implementation is good enough to be included in v0.17.0 release.
> 
> if matched features can be returned and output correctly, then all other unmatched feature should be failure in feature matching.

The current rules implementation doesn’t easily match executed features (whether failed or succeeded) with expressions in the spec, as it was not designed for that. Without changes, `validate-node` would need overly complex algorithms. Starting a full refactor now risks delaying inclusion in v0.17, so it’s better to start small.

> so maybe no need to introduce runstrategy, just in the output, for the features not in the matched feature list(but in the compatibility yaml), tag it as failed.
> then we may get same report as that with runstrategy

A run strategy implementation is required because, by default, rule processing stops at the first failure. `validate-node` must process all the rules.

cmd/client-nfd/subcmd/compat/validate-node.go
```diff
@@ -0,0 +1,161 @@
...
+var (
+	image     string
+	tags      []string
+	platform  options.PlatformOption
+	plainHTTP bool
+)
+
+var validateNodeCmd = &cobra.Command{
+	Use:   "validate-node",
+	Short: "Perform node validation based on its associated image compatibility artifact.",
```
ChaoyiHuang, line 48: nit, "." is not necessary

docs/reference/node-feature-client-reference.md
```diff
@@ -0,0 +1,48 @@
...
+
+---
+
+To quickly view available command line flags execute `nfd --help`.
+
+### -h, --help
+
+Print usage and exit.
+
+## Compat
```
ChaoyiHuang, line 24: when input the parameter, it's "compat", it's better to to use the term same as that in command line. the next page kubectl plugin style is not very good

docs/usage/image-compatibility.md
```diff
@@ -0,0 +1,149 @@
...
+```
+
+**Note**: The attach command is planned to be integrated into the `nfd` client
+tool. This will streamline the process, allowing you to perform the operation
+directly within the tool rather than using a separate command.
+
+### Validate the node against the image compatibility specification
+
+1. Build `nfd` client: `make build`
+1. Run `./bin/client-nfd compat validate-node --image <image-url>`
```
ChaoyiHuang, line 143: the output binanry name is "nfd" or "client-nfd". other places it uses "nfd".

ChaoyiHuang, line 143: not directly run the tool in the build node. but ot run the nfd tool in the host to be validated.

cmd/client-nfd/subcmd/compat/compat.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var CompatCmd = &cobra.Command{
+	Use:   "compat",
+	Short: "Image compatibility commands.",
```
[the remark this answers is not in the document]
mfranczy: Fixed.

docs/reference/node-feature-client-reference.md
mfranczy: I fixed the name issue. 

> the next page kubectl plugin style is not very good

Could you elaborate on this? I didn't quite understand the comment.

docs/reference/node-feature-client-reference.md
```diff
@@ -0,0 +1,48 @@
+---
+title: "Node Feature client cmdline reference"
+layout: default
+sort: 8
```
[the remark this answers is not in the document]
mfranczy: Fixed.

docs/usage/image-compatibility.md
mfranczy: Fixed.

mfranczy: Fixed.

mfranczy: The run instruction refers to the `./bin` dir to show the binary location after the build. Users will understand that they have to run the binary on the hosts they want to validate, so further explanation isn't necessary, I think.

cmd/client-nfd/subcmd/compat/validate-node.go
mfranczy: Fixed.

docs/reference/node-feature-client-reference.md
mfranczy: Discussed internally.

mfranczy:
> The current rules implementation doesn’t easily match executed features (whether failed or succeeded) with expressions in the spec, as it was not designed for that. Without changes, validate-node would need overly complex algorithms. Starting a full refactor now risks delaying inclusion in v0.17, so it’s better to start small. Additionally, I would like to discuss the planned changes with @marquiz first.

After discussions with end users, it became clear that the minimal functionality wasn't sufficient, so I revisited the problem from a new perspective.

Previously, I attempted to handle expression execution status similar to [MatchedFeatures](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/pkg/apis/nfd/nodefeaturerule/rule.go#L186), which introduced significant complexity and required extensive changes to the NFR code. I found a better approach by extending the [expression struct with an isMatch field](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1932/files#diff-2993b0b6843cd27fbf3b802d7dd4032f8be105b385fa1bc32c2cb009a1cc3d5cR302-R323), allowing the expression status to update directly during processing.

@ArangoGutierrez @marquiz apologies for the multiple updates, but this is the final version. The PR is ready for review, and I consider it complete. Please review whenever you have a moment.

api/nfd/v1alpha1/types.go
```diff
@@ -296,6 +299,34 @@ type MatchExpression struct {
...
+	}
+
+	return json.Marshal(*m)
+}
+
+func (m *MatchExpression) String() string {
+	if len(m.Value) < 1 {
+		return fmt.Sprintf("{op: %s}", m.Op)
+	}
+	return fmt.Sprintf("{op: %s, value: %s}", m.Op, m.Value)
```
ChaoyiHuang, line 329: at least for value, it may be some quoted string, "%q" (which was used in the master) may be better option

mfranczy: Fixed.

ChaoyiHuang:
for the first nfd client alpha release, it looks good to me.
For the modification in nfd rule match and it's impact on existing functions, it need more expertise review comments.

docs/usage/image-compatibility.md
```diff
@@ -0,0 +1,146 @@
...
+    This OPTIONAL property specifies the [node affinity weight](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#node-affinity-weight).
+
+  - **tag** - *string*  
+    This OPTIONAL property allows for the grouping or separation of
+    compatibility sets.
+
+  - **description** - *string*  
+    This OPTIONAL property provides a brief description of a compatibility set.
+
+### Example
```
ArangoGutierrez, line 57:
```suggestion
#### Example
```

api/nfd/v1alpha1/types.go
```diff
@@ -296,6 +299,34 @@ type MatchExpression struct {
...
 	// +optional
 	Value MatchValue `json:"value,omitempty"`
+
+	// IsMatch indicates whether the processed expression matches the host.
+	// This field should not be exposed in the CRD or used by users.
+	// It is for informational purposes only.
+	IsMatch bool `json:"-"`
+}
+
+// Override MarshalJSON to hide the IsMatch field,
```
ArangoGutierrez, line 309: I think we want to start the doc-comment with the func/method name `MarshalJSON`

ArangoGutierrez (changes requested):
2 minor comments, looks sane for me, @marquiz ?

docs/usage/image-compatibility.md
mfranczy: Fixed.

api/nfd/v1alpha1/types.go
mfranczy: Fixed.

ArangoGutierrez (approved):
I'll leave the LGTM to Markus

ChaoyiHuang (approved):
/approve

api/nfd/v1alpha1/types.go
marquiz, line 1: I think these modifications to the nfd API are the biggest pain point. Having fields that are really only for implementation is not good and takes us further from getting the api reviewed and marked stable.

I think one better way to do this (instead of polluting the api) would be for the `nodefeaturerule.Execute()` func to return a match status instead. Also in `node-validator` `RuleStatus` you'd have this match status instead of the complete rule.

Like (doesn't need to be named like this):

```golang
type MatchStatus {
    IsMatch bool
    MatchFeatures  []FeatureMatcherStatus
    MatchAny []MatchAnyStatus
}

type MatchAnyStatus {
    MatchFeatures  []FeatureMatcherStatus
}
...
```

One hack that we might be able to live with (temporarily) would be to unexport `isMatch` and have setter/getter methods for that. This way it wouldn't be seen by anybody importing the API.

Thoughts on this?

Also, the `nfdv1alpha1.Rule` contains fields that are not valid for the image compat stuff but we can deal with that later...

marquiz (commented):
One again thank you @mfranczy for the PR! And apologies for being so slow on the review. I just got constantly carried away and you need good uninterrupted to review these kind of PRs...

A few comments below. The biggest issue (which I think needs to be addressed somehow) I see are the changes to api/nfd/v1alpha1. If we can get away without those LGTM

**EDIT:** oh, and I think we should state in the docs that the nfd cli stuff is experimental/alpha whatever(?)

docs/reference/node-feature-client-reference.md
```diff
@@ -0,0 +1,64 @@
+---
+title: "Node Feature client cmdline reference"
```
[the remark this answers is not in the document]
mfranczy: Yes. It can be NFD client instead.

api/image-compatibility/v1alpha1/spec.go
```diff
@@ -0,0 +1,48 @@
...
+
+// Compatibility represents image compatibility metadata
+// that describe the image requirements for the host and OS.
+type Compatibility struct {
+	// Rules represents a list of Node Feature Rules.
+	Rules []nfdv1alpha1.Rule `json:"rules"`
+	// Weight indicates the priority of the compatibility set.
+	Weight int `json:"weight,omitempty"`
+	// Tag enables grouping or distinguishing between compatibility sets.
+	Tag string `json:"tag,omitempty"`
```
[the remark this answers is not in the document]
mfranczy: There is only allowed one tag per compatibility set.

Makefile
```diff
@@ -90,7 +90,7 @@ IMAGE_BUILD_ARGS_MINIMAL = --target minimal \
 
 all: image
 
-BUILD_BINARIES := nfd-master nfd-worker nfd-topology-updater nfd-gc kubectl-nfd
+BUILD_BINARIES := nfd-master nfd-worker nfd-topology-updater nfd-gc kubectl-nfd nfd
```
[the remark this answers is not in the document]
mfranczy: Yes, we want to include nfd cli in the container image. I prepared a k8s job which verifies if an image is compatible with the host. I am going to include it in the documentation.

pkg/apis/nfd/nodefeaturerule/rule.go
```diff
@@ -31,6 +31,36 @@ import (
 	"sigs.k8s.io/node-feature-discovery/pkg/utils"
 )
 
+// RunStrategy controls the processing of Node Feature Rules.
+type RunStrategy interface {
```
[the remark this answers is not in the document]
mfranczy: Right, it doesn't have to be an interface. A simple `failFast` flag is good enough. I will fix this in this PR.
At the beginning, I thought there may be more functions to manipulate the code execution, or gather information about executed rules.

api/nfd/v1alpha1/types.go
mfranczy: > I think these modifications to the nfd API are the biggest pain point. Having fields that are really only for implementation is not good and takes us further from getting the api reviewed and marked stable.

I understand your concern and agree that polluting the API is not ideal. Initially, I tried to avoid adding fields solely for implementation purposes, but this resulted in what I felt was an overly complicated implementation.

> I think one better way to do this (instead of polluting the api) would be for the nodefeaturerule.Execute() func to return a match status instead. Also in node-validator RuleStatus you'd have this match status instead of the complete rule.

Let me clarify the requirements to ensure we're aligned. Users expect a detailed report that shows which rules matched or failed on the host. For example, given the following rule:

```yaml
rules:
  - name: "kernel modules"
    matchFeatures:
      matchExpressions:
        vfio-pci: {op: Exists}
        ip_tables: {op: Exists}
    - feature: cpu.model
      matchExpressions:
        vendor_id: {op: In, value: ["Intel", "AMD"]}
```

Users want precise information about which expressions passed or failed. This level of detail is more valuable than just knowing which features matched. They expect the level of details as on the report [I pasted above](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1932#issuecomment-2476529739).

Initially, I used the `matchedFeatures` type to evaluate the status of each expression (passed or failed). This worked fine for `flag` and `attribute` feature types with `matchExpressions`, but for others, it became difficult to determine which matched features corresponded to a specific expression, especially since I aimed to minimize changes to the NFR code.
To address this, I introduced the `IsMatch` field and updated it directly during processing. This seemed like the simplest way to indicate whether an expression passed or failed without significant changes to the NFR code. However, I agree with you, it introduces API pollution.

> One hack that we might be able to live with (temporarily) would be to unexport isMatch and have setter/getter methods for that. This way it wouldn't be seen by anybody importing the API.

> Thoughts on this?

I will try to use something similar to status field you proposed. I will return the full information about which expression passed to be able to compare it later, it will result in more complex implementation but without API pollution.
For the comparison with already proposed solution, I will submit a commit to this PR, if we agree that's better then after I will clean the commits history to get rid of the old implementation.

> Also, the nfdv1alpha1.Rule contains fields that are not valid for the image compat stuff but we can deal with that later...

Could you point the specific ones you have in your mind? To me, it looks like that in the future we could use everything that NFR provides.

mfranczy: @marquiz @ArangoGutierrez  I was able to maintain the same functionality without changing the API.

The NFR code changes are visible under this [link](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1932/commits/b236ab35bb8bf87c218d484a31bb233547cc3764).
As suggested, I introduced the `MatchStatus` type to provide detailed information about processed rules. This type includes information about expressions that succeeded on the host, in addition to the previously existing `matchedFeatures`. Implementing this required changes to the NFR code, which I hope are acceptable.

For anyone else reviewing the changes and seeking to understand the core differences introduced:

Consider the following expression:
```yaml
feature: cpu.cpuid
matchName: {op: InRegexp, value: ["^AVX"]}
```
Previously, rule execution would only return the matched features, such as AVX, AVX2, or AVX512, without associating them with the executed expression. This PR changes that behavior. Now, in addition to returning matched features, the rule execution also provides information about the successfully executed expression.

In simpler terms, the result will now include:

    Matched features: AVX, AVX2, AVX512
    Matched expressions: cpu.cpuid: matchName: {op: InRegexp, value: ["^AVX"]}

This enhancement immediately clarifies which expressions succeeded and which failed, along with the matched features.

Makefile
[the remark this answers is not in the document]
mfranczy: Example provided: https://deploy-preview-1932--kubernetes-sigs-nfd.netlify.app/usage/image-compatibility#validate-the-k8s-cluster-node-with-the-validate-image-job

pkg/apis/nfd/nodefeaturerule/rule.go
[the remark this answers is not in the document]
mfranczy: Fixed.

api/nfd/v1alpha1/types.go
```diff
@@ -298,6 +300,13 @@ type MatchExpression struct {
 	Value MatchValue `json:"value,omitempty"`
 }
 
+func (m MatchExpression) String() string {
```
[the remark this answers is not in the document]
mfranczy: Fixed. Moved to `utils.go`.

cmd/nfd/subcmd/compat/validate-node.go
```diff
@@ -0,0 +1,224 @@
...
+	Short: "Perform node validation based on its associated image compatibility artifact",
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		var err error
+
+		if err = platform.Parse(cmd); err != nil {
+			return err
+		}
+
+		if readAccessToken && readPassword {
+			return fmt.Errorf("cannot use --read-access-token and --read-password at the same time")
```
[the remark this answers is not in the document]
mfranczy: Good catch! Fixed.

mfranczy:
> @mfranczy very good job on this one. A few minor comments/questions below.
> 
> Also, WDYT about announcing this as experimental (see [#1932 (review)](https://github.com/kubernetes-sigs/node-feature-discovery/pull/1932#pullrequestreview-2496430081))?

Thanks. I've added the experimental note in the nfd client and image compatibility docs.

docs/reference/node-feature-client-reference.md
```diff
@@ -0,0 +1,65 @@
...
+#### --tags
+
+The `--tags` flag specifies a list of tags that must match the tags
+set on the compatibility objects.
+
+#### --output-json
+
+The `--output-json` flag prints the output as a JSON object.
+
+#### --reg-username
```
marquiz, line 55: Ach, now that the flag names were updated we need to update the documentation, too. Sorry about that :see_no_evil:

mfranczy: Of course. No worries, my bad. Fixed.

api/nfd/v1alpha1/types.go
ArangoGutierrez: ++

marquiz (approved):
Wow @mfranczy, I'm ready to merge.

Let's hope something greater good comes out of this work :rocket:

marquiz:
If @ArangoGutierrez is still around I'll let him do the last check :smile: 
/assign @ArangoGutierrez

PR #1941, opened 2024-11-04
Drop support for hooks

ArangoGutierrez:


marquiz (commented):
Thanks for working on this @ArangoGutierrez 

One nit about the commit message. The hooks have already been marked as deprecated so I think a better description would be "Drop support for hooks" or something.

Another comment: we should also drop the `source.d` mounts from kustomize and Helm deployments.

Otherwise looks good

ArangoGutierrez:
> Thanks for working on this @ArangoGutierrez
> 
> One nit about the commit message. The hooks have already been marked as deprecated so I think a better description would be "Drop support for hooks" or something.
> 
> Another comment: we should also drop the `source.d` mounts from kustomize and Helm deployments.
> 
> Otherwise looks good

Done

ArangoGutierrez:
Fixes #856

marquiz (commented):
Ach, there's still some references to `source.d` in the customization guide in docs. If you could remove those, too.

ArangoGutierrez:
> source.d

Done

PR #1950, opened 2024-11-07
nfd-master: drop resourceLabels

marquiz:
Drop the resourceLabels config file option and the corresponding -resource-labels command line flag. They were deprecated in NFD v0.13 so it's time to let them go. NodeFeatureRule(s) should be used to manage ERs, instead.


marquiz:
Fine to postpone this to v0.18. OTOH I don't see this very risky to have as a late pull-in for v0.17 either.

/assign @ArangoGutierrez

ArangoGutierrez (approved):
Glad to see ExtendedResources stable now
/lgtm

PR #1962, opened 2024-11-26
go.mod: bump kubernetes patch version

marquiz:


api/nfd/go.mod
```diff
@@ -5,8 +5,8 @@ go 1.22.2
 require (
 	github.com/gogo/protobuf v1.3.2
 	github.com/stretchr/testify v1.8.4
-	k8s.io/api v0.30.3
-	k8s.io/apimachinery v0.30.3
+	k8s.io/api v0.30.7
+	k8s.io/apimachinery v0.30.7
```
fmuyassarov, line 9: maybe `v0.31.3` as anything else?

marquiz: The thinking behind sticking with v1.30.x is that the `api/nfd` module is intended to be a "library", imported by other projects. In contrast to the main nfd go module which is not. For the library, we don't need/want to force dependent projects to use the latest k8s as it's really not necessary. We'd bump when 1.30 is EOL'd (I think). Makes sense?

Issue #1998, opened 2025-01-02
Pagination size cmd line arg


ivelichkovich:
**What would you like to be added**: 

command line arg for pagination size

related to this: https://github.com/kubernetes-sigs/node-feature-discovery/pull/1815 

**Why is this needed**:

better control over load on apiserver

marquiz:
Thanks @ivelichkovich for the request. Makes sense to me. Simplest solution would be to add a config file option
/help

ivelichkovich:
I'm happy to work on this :)

ivelichkovich:
Will raise an MR soon

ivelichkovich:
so I was actually thinking the MR I linked was adding pagination to nfd-master for its informer cache so I raised an MR adding configurable pagination to the NFD master: https://github.com/kubernetes-sigs/node-feature-discovery/pull/2000 

and then here's the MR for configurable pagination to NFD GC:  https://github.com/kubernetes-sigs/node-feature-discovery/pull/2001

ArangoGutierrez:
Status: the nfd-master half merged via #2000; the nfd-gc half (#2001) was closed unmerged after going stale. @ivelichkovich do you still need configurable pagination on the gc list path? If yes it needs a new owner, if not we can close this as done.

PR #2000, opened 2025-01-04
add configurable pagination to nfd-master

ivelichkovich:
addresses scalability and api-server load concerns for large clusters by adding configurable pagination to the informer cache of nfd-master

related to: https://github.com/kubernetes-sigs/node-feature-discovery/issues/1998

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -140,6 +140,8 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 	flagset.Var(overrides.ResyncPeriod, "resync-period", "Specify the NFD API controller resync period.")
 	overrides.NfdApiParallelism = flagset.Int("nfd-api-parallelism", 10, "Defines the maximum number of goroutines responsible of updating nodes. "+
 		"Can be used for the throttling mechanism.")
+	overrides.InformerPageSize = flagset.Int64("informer-page-size", 200,
+		"The list size to use when listing NodeFeature objects to sync informer cache.")
 
 	return args, overrides
 }
--- a/docs/reference/master-commandline-reference.md
+++ b/docs/reference/master-commandline-reference.md
@@ -173,6 +173,21 @@ Example:
 nfd-master -deny-label-ns=*.vendor.com,vendor-2.io
 ```
 
+### -informer-page-size
+
+The `-informer-page-size` flag is used to control pagination
+during informer cache sync on nfd-master startup.
+This is useful to control load on api-server/etcd as listing
+NodeFeature objects can be expensive, especially in large clusters.
+
+Default: 200
+
+Example:
+
+```bash
+nfd-master -informer-page-size=20
+```
+
 ### -config
 
 The `-config` flag specifies the path of the nfd-master configuration file to
--- a/pkg/nfd-master/nfd-api-controller.go
+++ b/pkg/nfd-master/nfd-api-controller.go
@@ -56,6 +56,7 @@ type nfdApiControllerOptions struct {
 	ResyncPeriod                 time.Duration
 	K8sClient                    k8sclient.Interface
 	NodeFeatureNamespaceSelector *metav1.LabelSelector
+	ListSize                     int64
 }
 
 func init() {
@@ -95,11 +96,19 @@ func newNfdController(config *restclient.Config, nfdApiControllerOptions nfdApiC
 		// Tweak list opts on initial sync to avoid timeouts on the apiserver.
 		// NodeFeature objects are huge and the Kubernetes apiserver
 		// (v1.30) experiences http handler timeouts when the resource
-		// version is set to some non-empty value (TODO: find out why).
+		// version is set to some non-empty value
+		// https://github.com/kubernetes/kubernetes/blob/ace55542575fb098b3e413692bbe2bc20d2348ba/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L600-L616 if you set resource version to 0
+		// it serves the request from apiservers cache and doesn't use pagination otherwise pagination will default to 500
+		// so that's why this is required on large clusters
+		// So by setting this we're making it go to ETCD instead of from api-server cache, there's some WIP in k/k
+		// that seems to imply they're working on improving this behavior where you'll be able to paginate from apiserver cache
+		// it's not supported yet (2/2025), would be good to track this though kubernetes/kubernetes#108003
 		if opts.ResourceVersion == "0" {
 			opts.ResourceVersion = ""
 		}
+		opts.Limit = nfdApiControllerOptions.ListSize // value of 0 disables pagination
 	}
+
 	featureInformer := nfdinformersv1alpha1.New(informerFactory, "", tweakListOpts).NodeFeatures()
 	if _, err := featureInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
--- a/deployment/components/master-config/nfd-master.conf.example
+++ b/deployment/components/master-config/nfd-master.conf.example
@@ -2,6 +2,7 @@
 # extraLabelNs: ["added.ns.io","added.kubernets.io"]
 # denyLabelNs: ["denied.ns.io","denied.kubernetes.io"]
 # enableTaints: false
+# informerPageSize: 200
 # labelWhiteList: "foo"
 # resyncPeriod: "2h"
 # restrictions:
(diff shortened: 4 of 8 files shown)


cmd/nfd-master/main.go
```diff
@@ -122,6 +122,8 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 			"in the same format as in the config file (i.e. json or yaml). These options")
 	flagset.BoolVar(&args.EnableLeaderElection, "enable-leader-election", false,
 		"Enables a leader election. Enable this when running more than one replica on nfd master.")
+	flagset.Int64Var(&args.ListSize, "node-feature-informer-list-size", 0,
```
TessaIO, line 125: I suggest a shorter way like this.
```suggestion
	flagset.Int64Var(&args.ListSize, "informer-list-size", 0,
```

TessaIO (commented):
Thanks for the work, small comment out there. Can you please add the documentation for this new option?

cmd/nfd-master/main.go
ivelichkovich: Sure thing

ivelichkovich:
I wasn't sure if we'd want to default to zero to keep new default behavior consistent with old behavior or set it to default to 200 to match the gc pagination default size. I'm open to either

cmd/nfd-master/main.go
ivelichkovich: Changed the flag and updated docs, let me know if you'd prefer the doc update somewhere else

ivelichkovich:
This likely actually doesn't work since it's not using continue at all, there's some nuance to when/how limit is respected with regards to if api-server is serving from it's cache or from etcd and if we're using ListWatch or not that I need to research more. Any inputs appreciated though.

api/generated/informers/externalversions/nfd/v1alpha1/nodefeature.go
```diff
@@ -62,7 +62,23 @@ func NewFilteredNodeFeatureInformer(client versioned.Interface, namespace string
 				if tweakListOptions != nil {
 					tweakListOptions(&options)
 				}
-				return client.NfdV1alpha1().NodeFeatures(namespace).List(context.TODO(), options)
+				if options.Limit == 0 {
```
ivelichkovich, line 65: this works but is editing a generated file, anyone know a good clean way to have custom informer factory logic here? I could make a wrapper outside of the generated directory but not sure what the cleanest mechanism here is

ivelichkovich: ah just tested by only setting limit without changing the file, based on kubernetes audit logs it looks like the informer handles the continue stuff for you :)

pkg/nfd-master/nfd-api-controller.go
```diff
@@ -101,6 +102,7 @@ func newNfdController(config *restclient.Config, nfdApiControllerOptions nfdApiC
 			if opts.ResourceVersion == "0" {
```
ivelichkovich, line 102: btw while researching the pagination stuff I think this is likely due to this snippet of code: https://github.com/kubernetes/kubernetes/blob/ace55542575fb098b3e413692bbe2bc20d2348ba/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L600-L616 if you set resource version to 0 it serves the request from apiservers cache and doesn't use pagination otherwise pagination will default to 500 so that may explain why it blows up on large clusters

ivelichkovich, line 102: btw while researching the pagination stuff I think this is likely due to this snippet of code: https://github.com/kubernetes/kubernetes/blob/ace55542575fb098b3e413692bbe2bc20d2348ba/staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go#L600-L616 if you set resource version to 0 it serves the request from apiservers cache and doesn't use pagination otherwise pagination will default to 500 so that may explain why it blows up on large clusters

ivelichkovich: So by setting this we're making it go to ETCD instead of from api-server cache, I found some WIP in k/k that seems to imply they're working on improving this behavior where you'll be able to paginate from apiserver cache but AFAICT it's not supported yet, would be good to track this though https://github.com/kubernetes/kubernetes/issues/108003

cmd/nfd-master/main.go
```diff
@@ -122,6 +122,8 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 			"in the same format as in the config file (i.e. json or yaml). These options")
 	flagset.BoolVar(&args.EnableLeaderElection, "enable-leader-election", false,
 		"Enables a leader election. Enable this when running more than one replica on nfd master.")
+	flagset.Int64Var(&args.ListSize, "informer-list-size", 200,
```
marquiz, line 125: Just pondering (bike-shedding) on the naming. Would `-informer-paginate` or `-informer-page-size` be more descriptive, thoughts?

cmd/nfd-master/main.go
```diff
@@ -122,6 +122,8 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
...
+		"The list size to use when listing node features to sync informer cache.")
```
marquiz, line 126: I'd suggest to use the name of the CRD.

```suggestion
		"The list size to use when listing NodeFeature objects to sync informer cache.")
```

docs/reference/master-commandline-reference.md
```diff
@@ -173,6 +173,19 @@ Example:
 nfd-master -deny-label-ns=*.vendor.com,vendor-2.io
 ```
 
+### -informer-list-size
+
+The `-informer-list-size` flag is used to control pagination during informer cache sync on nfd-master startup.
+This is useful to control load on api-server/etcd as listing `nodefeatures` can be expensive, especially in large clusters.
```
marquiz, line 179:
```suggestion
This is useful to control load on api-server/etcd as listing NodeFeature objects can be expensive, especially in large clusters.
```

marquiz (commented):
Thank you @ivelichkovich for the enhancement. A few small comments below
/ok-to-test
/cc @ArangoGutierrez @adrianchiris

cmd/nfd-master/main.go
ivelichkovich: Yes I like that, I'll update

cmd/nfd-master/main.go
ivelichkovich: will do!

marquiz:
Thank you @ivelichkovich for the updates. To me, this looks pretty good, now.

One thing that would be nice would be to have the same setting as a config file option. I think that would also make this more usable with the helm chart (with like `--set master.config.informerPageSize=100`). We could track this separately (and as a separate PR) if you think that'd be better.

Also, thank you for investigating deeply the "find out why" part. Would it be possible to replace my TODO comment in the code with an explanation along the lines of these comments
https://github.com/kubernetes-sigs/node-feature-discovery/pull/2000#discussion_r1903192489
https://github.com/kubernetes-sigs/node-feature-discovery/pull/2000#discussion_r1903192746

WDYT?

ivelichkovich:
> Thank you @ivelichkovich for the updates. To me, this looks pretty good, now.
> 
> One thing that would be nice would be to have the same setting as a config file option. I think that would also make this more usable with the helm chart (with like `--set master.config.informerPageSize=100`). We could track this separately (and as a separate PR) if you think that'd be better.
> 
> Also, thank you for investigating deeply the "find out why" part. Would it be possible to replace my TODO comment in the code with an explanation along the lines of these comments [#2000 (comment)](https://github.com/kubernetes-sigs/node-feature-discovery/pull/2000#discussion_r1903192489) [#2000 (comment)](https://github.com/kubernetes-sigs/node-feature-discovery/pull/2000#discussion_r1903192746)
> 
> WDYT?

Yeah that sounds good, I can update that

marquiz:
> Yeah that sounds good, I can update that

Thanks for working on this :+1:

ivelichkovich:
Okay updated, PTAL 😄

ivelichkovich:
it looks like code coverage may be complaining about the comment lines

marquiz (commented):
@ivelichkovich please rebase the PR.

Also, you need to update `deployment/components/master-config/nfd-master.conf.example` (this is the "origin" for the conf) and after that run `make templates` to update the chart. This gets rid of the CI failure.

marquiz:
@ivelichkovich rebase and squash, other than that, lgtm

ozhuraki (commented):
LGTM, just rebase and squash

marquiz:
@ivelichkovich another rebase needed :face_with_diagonal_mouth: When that's done let's merge this

ArangoGutierrez:
@ivelichkovich PING, a rebase is needed

ivelichkovich:
rebased

marquiz (approved):
/lgtm
/retest

PR #2006, opened 2025-01-08
Add support running with OwnerReferencesPermissionEnforcement

adrianchiris:
when OwnerReferencesPermissionEnforcement validating webhook is
enabled additional permissions are required to set/update owner ref
field. NFD worker sets/updates NodeFeature owner ref field to
the worker pod and owning daemonset.

owner reference can only be updated if the worker has delete permissions
for NodeFeatures.

if owner reference has blockOwnerDeletion (as the case for the daemonset
owner reference) then it requires update permissions to the finalizers
of the owner, to avoid this, we set blockOwnerDeleteion to false for all
owners referenced from NFD worker pod when setting/updating NodeFeature
owner ref.

more info on [OwnerReferencesPermissionEnforcement](https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers/#ownerreferencespermissionenforcement)

### DIFF
--- a/deployment/base/rbac/worker-role.yaml
+++ b/deployment/base/rbac/worker-role.yaml
@@ -11,6 +11,7 @@ rules:
   - create
   - get
   - update
+  - delete
 - apiGroups:
   - ""
   resources:
--- a/deployment/helm/node-feature-discovery/templates/role.yaml
+++ b/deployment/helm/node-feature-discovery/templates/role.yaml
@@ -15,6 +15,7 @@ rules:
   - create
   - get
   - update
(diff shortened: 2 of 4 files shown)


adrianchiris:
@marquiz PTAL

marquiz (approved):
Thank you @adrianchiris. This looks good to me

marquiz:
@ArangoGutierrez wanna take a peek?

/assign @ArangoGutierrez

ArangoGutierrez (approved):
/lgtm

PR #2010, opened 2025-01-08
Bugfixes for image compatibility feature

mfranczy:
Bugfixes:
* Always pull the latest compatibility artifact
* Handle missing compatibility matches to avoid panic
* Continue procesing the rule if the feature is missing

### DIFF
--- a/pkg/apis/nfd/nodefeaturerule/rule.go
+++ b/pkg/apis/nfd/nodefeaturerule/rule.go
@@ -268,8 +268,11 @@ func evaluateFeatureMatcher(m *nfdv1alpha1.FeatureMatcher, features *nfdv1alpha1
 		fA, okA := features.Attributes[featureName]
 		fI, okI := features.Instances[featureName]
 		if !okF && !okA && !okI {
-			klog.V(2).InfoS("feature not available", "featureName", featureName)
-			return false, nil, nil
+			if failFast {
+				klog.V(2).InfoS("feature not available", "featureName", featureName)
+				return false, nil, nil
+			}
+			continue
 		}
 
 		if term.MatchExpressions != nil {
--- a/pkg/apis/nfd/nodefeaturerule/rule.go
+++ b/pkg/apis/nfd/nodefeaturerule/rule.go
@@ -268,10 +268,11 @@ func evaluateFeatureMatcher(m *nfdv1alpha1.FeatureMatcher, features *nfdv1alpha1
 		fA, okA := features.Attributes[featureName]
 		fI, okI := features.Instances[featureName]
 		if !okF && !okA && !okI {
+			klog.V(2).InfoS("feature not available", "featureName", featureName)
 			if failFast {
-				klog.V(2).InfoS("feature not available", "featureName", featureName)
 				return false, nil, nil
 			}
+			isMatch = false
 			continue
 		}
 
--- a/pkg/client-nfd/compat/node-validator/node-validator.go
+++ b/pkg/client-nfd/compat/node-validator/node-validator.go
@@ -97,6 +97,7 @@ func (nv *nodeValidator) Execute(ctx context.Context) ([]*CompatibilityStatus, e
 }
 
 func evaluateRuleStatus(rule *nfdv1alpha1.Rule, matchStatus *nodefeaturerule.MatchStatus) ProcessedRuleStatus {
+	var matchedFeatureTerms nfdv1alpha1.FeatureMatcher
 	out := ProcessedRuleStatus{Name: rule.Name, IsMatch: matchStatus.IsMatch}
 
 	evaluateFeatureMatcher := func(featureMatcher, matchedFeatureTerms nfdv1alpha1.FeatureMatcher) []MatchedExpression {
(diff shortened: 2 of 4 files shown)


pkg/apis/nfd/nodefeaturerule/rule.go
```diff
@@ -268,8 +268,11 @@ func evaluateFeatureMatcher(m *nfdv1alpha1.FeatureMatcher, features *nfdv1alpha1
 		fA, okA := features.Attributes[featureName]
 		fI, okI := features.Instances[featureName]
 		if !okF && !okA && !okI {
-			klog.V(2).InfoS("feature not available", "featureName", featureName)
-			return false, nil, nil
+			if failFast {
+				klog.V(2).InfoS("feature not available", "featureName", featureName)
```
marquiz, line 272: Should we still log this (for debugging purposes)?

mfranczy: I think yes. It's still worth to keep the reason of failing in case `failFast == true`.

marquiz: My question was merely should we also log this in case `failFast == false`?

ChaoyiHuang: in production, default logging level is warning or error? if default logging level in production is greater than info, it's fine to log that for debugging purpose no matter failFast is True or False

pkg/client-nfd/compat/node-validator/node-validator.go
```diff
@@ -97,6 +97,7 @@ func (nv *nodeValidator) Execute(ctx context.Context) ([]*CompatibilityStatus, e
 }
 
 func evaluateRuleStatus(rule *nfdv1alpha1.Rule, matchStatus *nodefeaturerule.MatchStatus) ProcessedRuleStatus {
```
ChaoyiHuang, line 99: could this function be a function of (nv *nodeValidator). and the function refactory will enhance the readbility

pkg/client-nfd/compat/node-validator/node-validator.go
```diff
@@ -97,6 +97,7 @@ func (nv *nodeValidator) Execute(ctx context.Context) ([]*CompatibilityStatus, e
...
+	var matchedFeatureTerms nfdv1alpha1.FeatureMatcher
 	out := ProcessedRuleStatus{Name: rule.Name, IsMatch: matchStatus.IsMatch}
 
 	evaluateFeatureMatcher := func(featureMatcher, matchedFeatureTerms nfdv1alpha1.FeatureMatcher) []MatchedExpression {
```
ChaoyiHuang, line 103: nested function make it become a little long for evaluateRuleStatus

pkg/client-nfd/compat/node-validator/node-validator.go
```diff
@@ -163,11 +164,17 @@ func evaluateRuleStatus(rule *nfdv1alpha1.Rule, matchStatus *nodefeaturerule.Mat
 	}
 
 	if matchFeatures := rule.MatchFeatures; matchFeatures != nil {
-		out.MatchedExpressions = evaluateFeatureMatcher(matchFeatures, matchStatus.MatchedFeaturesTerms)
+		if matchStatus.MatchFeatureStatus != nil {
```
ChaoyiHuang, line 167: non supported features in nfd could be added as some test cases. the tool should deal with unrecognized feature name or expression as well. in the future may check whether non-supported features are in image-compatibility artifact

pkg/apis/nfd/nodefeaturerule/rule.go
marquiz: Default is `V(0)` so this is not logged by default settings.

pkg/client-nfd/compat/node-validator/node-validator.go
marquiz: We could add the refactoring stuff as TODO notes and/or create a GitHub issue for tracking and concentrate solely on bugfixes in this PR. WDYT?

mfranczy: I agree. I plan to do refactoring but in a different PR, in this specific one I would prefer only to concentrate on the bugfixes I provided.

pkg/client-nfd/compat/node-validator/node-validator.go
mfranczy: @ChaoyiHuang I will provide the refactored code but in a different PR. I would prefer that it doesn't grow too much that we can the fixes fast.

pkg/apis/nfd/nodefeaturerule/rule.go
mfranczy: > My question was merely should we also log this in case failFast == false?

Ah, yes, I will adjust the code.

mfranczy: The code is adjusted.

pkg/client-nfd/compat/node-validator/node-validator.go
mfranczy: I have added a unit test to check the behaviour. Currently, it's allowed to add any feature to the spec, if it doesn't exist the rule will fail. The user will get a log about it (in debug mode). In the future, we can think of additional behaviours (like stopping the execution if the feature doesn't exist etc.).

marquiz (approved):
Looks good to go from my pov.

@ChaoyiHuang WDYT?

pkg/client-nfd/compat/node-validator/node-validator.go
marquiz: Also, doing refactoring in a separate PR makes backporting to release branch cleaner and less risky

pkg/client-nfd/compat/node-validator/node-validator.go
ChaoyiHuang: TODO notes sounds good

ChaoyiHuang:
> Looks good to go from my pov.
> 
> @ChaoyiHuang WDYT?

LGTM

PR #2013, opened 2025-01-09
[release-0.16] Add support running with OwnerReferencesPermissionEnforcement

marquiz:
when OwnerReferencesPermissionEnforcement validating webhook is enabled additional permissions are required to set/update owner ref field. NFD worker sets/updates NodeFeature owner ref field to the worker pod and owning daemonset.

owner reference can only be updated if the worker has delete permissions for NodeFeatures.

if owner reference has blockOwnerDeletion (as the case for the daemonset owner reference) then it requires update permissions to the finalizers of the owner, to avoid this, we set blockOwnerDeleteion to false for all owners referenced from NFD worker pod when setting/updating NodeFeature owner ref.

Signed-off-by: adrianc <adrianc@nvidia.com>
(cherry picked from commit 3f012c2d5a140bd80abf128f9d7d45ee988c0794)

Backports: #2006

### DIFF
--- a/deployment/base/rbac/worker-role.yaml
+++ b/deployment/base/rbac/worker-role.yaml
@@ -11,6 +11,7 @@ rules:
   - create
   - get
   - update
+  - delete
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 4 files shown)


fmuyassarov (commented):
/lgtm

PR #2031, opened 2025-01-28
source/pci: Group identical PCI entries

ozhuraki:


marquiz (commented):
Thanks @ozhuraki for working on this.

We should only group devices where all attributes are identical. So we need to contain all attributes in the lookup key (e.g. with json.Marshal or something).

Also, we'd probably need a config option and/or feature gate for this so that it can be turned off. I've seen usages where people create per-device labels (like `dev-0`, `dev-1`, ...) by looping over the devices so we don't want to completely break those.

marquiz:
ping @ozhuraki, any updates on this? I think this would be an important optimization for the next release

ozhuraki:
@marquiz 
Apologies for a delay. Updated, please take a look.

source/pci/utils.go
```diff
@@ -80,13 +87,35 @@ func detectPci() ([]nfdv1alpha1.InstanceFeature, error) {
...
+			if result == 0 {
+				entry.Count += 1
+				devGrouped[info.Attributes["vendor"]][info.Attributes["device"]] = entry
+			}
+		}
+	}
+
+	for _, dev := range devInfo {
+		entry := devGrouped[dev.Attributes["vendor"]][dev.Attributes["device"]]
+		dev.Attributes["count"] = string(entry.Count)
```
marquiz, line 118: Need to use strconv for this

marquiz:
ping @ozhuraki any updates on this?

marquiz:
@ozhuraki any update on this enhancement? We should probably put this behind a feature gate to make it possible to disable the behavior (for users dependent on the old behavior, e.g. iterating over each device)

source/pci/utils.go
```diff
@@ -80,13 +87,35 @@ func detectPci() ([]nfdv1alpha1.InstanceFeature, error) {
...
 		}
-		devInfo = append(devInfo, *info)
+
+		b, err := json.Marshal(info.Attributes)
+		if err != nil {
+			return nil, err
+		}
+
+		if entry, ok := devGrouped[info.Attributes["vendor"]][info.Attributes["device"]]; !ok {
+			devGrouped[info.Attributes["vendor"]] = make(map[string]DevGroupedEntry)
```
ArangoGutierrez, line 104: This `make` runs whenever the (vendor, device) lookup misses, so every new device key of an already-seen vendor wipes that vendor's earlier entries: wiped keys re-enter the `!ok` branch (duplicate `devInfo` appends), and the fixup loop below then reads zero-value entries, so with the rootfs-1 mock sysfs (9 of 10 devices are vendor 8086) all but the last-seen 8086 device get count 0. Minimal guard if you keep the nested map:

```suggestion
			if devGrouped[info.Attributes["vendor"]] == nil {
				devGrouped[info.Attributes["vendor"]] = make(map[string]DevGroupedEntry)
			}
```

source/pci/utils.go
```diff
@@ -80,13 +87,35 @@ func detectPci() ([]nfdv1alpha1.InstanceFeature, error) {
...
+			devGrouped[info.Attributes["vendor"]][info.Attributes["device"]] = DevGroupedEntry{Bytes: b, Count: 1}
+			devInfo = append(devInfo, *info)
+		} else {
+			result := bytes.Compare(b, devGrouped[info.Attributes["vendor"]][info.Attributes["device"]].Bytes)
+			if result == 0 {
```
ArangoGutierrez, line 109: When the same (vendor, device) pair appears with different attributes -- e.g. two identical NICs where only one exposes `sriov_totalvfs`, or differing `iommu_group/type` -- this branch drops the device entirely: it is neither appended to `devInfo` nor counted, so it vanishes from the instance list and e.g. a `.sriov.capable` label can be lost. Non-identical devices need to stay as separate instances; keying the map by the marshaled attribute bytes (as @marquiz suggested) fixes this and the vendor-map clobbering in one go.

source/pci/utils.go
```diff
@@ -80,13 +87,35 @@ func detectPci() ([]nfdv1alpha1.InstanceFeature, error) {
...
+				entry.Count += 1
+				devGrouped[info.Attributes["vendor"]][info.Attributes["device"]] = entry
+			}
+		}
+	}
+
+	for _, dev := range devInfo {
```
ArangoGutierrez, line 116: `TestPciSource` asserts exact attribute maps, so it will fail once every instance gains `count` (CI has not run here due to needs-ok-to-test). Beyond updating the expected features, testdata rootfs-1 contains no two identical devices, so the grouping path is never exercised -- a duplicate-device case would have caught both the zero-count problem and the `string(int)` conversion.

ArangoGutierrez (commented):
This collapses PCI devices with identical attributes into a single instance with a `count` attribute, a useful size optimization for nodes with many identical devices. The current grouping has two correctness bugs (inline): the per-vendor map is re-created on every new device key, yielding zero counts and duplicate instances, and devices sharing vendor/device but differing in other attributes are silently dropped from the instance list -- both disappear if you restructure around a single map keyed by the marshaled attributes as @marquiz suggested in March. The existing `TestPciSource` will fail as-is and nothing exercises the grouping path, and `count` needs adding to the pci.device attribute table in `docs/usage/customization-guide.md`. Earlier feedback is also still open: strconv for the count conversion and the feature gate/config option for users relying on per-device instances (the documented `{{ .pci.device | len }}` template example changes meaning under this PR, which underlines that ask). Needs that rework plus updated tests/docs and a CI run (PR is needs-ok-to-test and lifecycle/rotten) before it's ready for another review pass.

source/pci/utils.go
ArangoGutierrez, line 104: **Bug:** this `make(...)` runs on *every* `!ok`, clobbering the vendor's inner map. The second *distinct* device of a vendor wipes the first's entry; a later *identical* device then sees `!ok` and gets appended again (duplicate) with `Count` reset to 1 — so counts are wrong and grouping breaks for any vendor with >1 device. Guard it:
```go
if _, ok := devGrouped[info.Attributes["vendor"]]; !ok {
    devGrouped[info.Attributes["vendor"]] = make(map[string]DevGroupedEntry)
}
```

source/pci/utils.go
```diff
@@ -80,13 +87,35 @@ func detectPci() ([]nfdv1alpha1.InstanceFeature, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		if entry, ok := devGrouped[info.Attributes["vendor"]][info.Attributes["device"]]; !ok {
+			devGrouped[info.Attributes["vendor"]] = make(map[string]DevGroupedEntry)
+			devGrouped[info.Attributes["vendor"]][info.Attributes["device"]] = DevGroupedEntry{Bytes: b, Count: 1}
+			devInfo = append(devInfo, *info)
+		} else {
+			result := bytes.Compare(b, devGrouped[info.Attributes["vendor"]][info.Attributes["device"]].Bytes)
```
ArangoGutierrez, line 108: **Bug:** when `bytes.Compare != 0` (same `(vendor, device)`, different attributes — e.g. distinct PCI address) the device is neither counted nor appended to `devInfo`, so it's **silently dropped**. Two cards of the same model would lose one. Distinct entries must still be reported — reconsider the grouping key/identity.

source/pci/utils.go
ArangoGutierrez, line 118: **Bug:** `string(entry.Count)` converts the int to a Unicode rune, not its decimal text — `string(5)` is `"\x05"`, not `"5"`. Use `strconv.Itoa(entry.Count)` (add the `strconv` import). `go vet`'s `stringintconv` flags this, so `verify-master` would fail on it too.

source/pci/utils.go
```diff
@@ -68,6 +70,11 @@ func readPciDevInfo(devPath string) (*nfdv1alpha1.InstanceFeature, error) {
 	return nfdv1alpha1.NewInstanceFeature(attrs), nil
 }
 
+type DevGroupedEntry struct {
```
ArangoGutierrez, line 73: Nit: `DevGroupedEntry` is only used internally — prefer unexported `devGroupedEntry`.

ArangoGutierrez (changes requested):
Thanks for this — and apologies it went stale. I took a careful pass: the grouping idea is reasonable, but there are a few correctness issues to resolve before it can go in, plus a broader question.

**Correctness (details inline):**
1. The `make(...)` in the not-found branch is unconditional and clobbers the vendor's inner map → wrong counts and duplicated identical devices.
2. `string(entry.Count)` yields a rune, not the decimal count — use `strconv.Itoa`.
3. Same-`(vendor, device)` devices with differing attributes are dropped instead of reported.

**Tests:** please add a table test over synthetic device sets (identical; same-model/different-address; multiple vendors) — it would have caught all three above.

**Semantics:** this changes the PCI source's emitted instances (fewer instances + a new `count` attribute), which can affect existing `NodeFeatureRule`/label matches. Worth a docs note and a quick confirmation that grouping is the intended behavior.

Marking request-changes for now; happy to re-review once addressed. Thanks!

PR #2038, opened 2025-01-29
Use Sprig template functions and add asLabelValue function

### LINKED ISSUE #2033: Use Sprig template functions
https://github.com/Masterminds/sprig

This library provides ease-of-use helpers, like `split` and `replace`.

**Why is this needed**:

Currently, I cannot use `system.dmiid.product_name` because the value has a space in it, and spaces are disallowed in labels. If I could do `{{.Value | replace " " "-" }}` it would fix it for me. As a result, I have to match the product_names I expect and manually write labels for them.

This will be true of any data that's not controlled by the cluster operator to be put into the content of a label.

fahhem:
Fixes #2033

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module sigs.k8s.io/node-feature-discovery
 go 1.24
 
 require (
+	github.com/Masterminds/sprig/v3 v3.3.0
 	github.com/fsnotify/fsnotify v1.8.0
 	github.com/google/go-cmp v0.6.0
 	github.com/google/uuid v1.6.0
@@ -42,7 +43,10 @@ require (
 
 require (
 	cel.dev/expr v0.19.0 // indirect
+	dario.cat/mergo v1.0.1 // indirect
 	github.com/JeffAshton/win_pdh v0.0.0-20161109143554-76bb4ee9f0ab // indirect
+	github.com/Masterminds/goutils v1.1.1 // indirect
+	github.com/Masterminds/semver/v3 v3.3.0 // indirect
 	github.com/Microsoft/go-winio v0.6.2 // indirect
 	github.com/NYTimes/gziphandler v1.1.1 // indirect
 	github.com/OneOfOne/xxhash v1.2.8 // indirect
@@ -87,6 +91,7 @@ require (
 	github.com/gorilla/websocket v1.5.0 // indirect
 	github.com/grpc-ecosystem/go-grpc-prometheus v1.2.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.20.0 // indirect
+	github.com/huandu/xstrings v1.5.0 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/jaypipes/pcidb v1.0.1 // indirect
 	github.com/josharian/intern v1.0.0 // indirect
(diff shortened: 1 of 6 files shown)


pkg/utils/template.go
```diff
@@ -0,0 +1,77 @@
...
+package utils
+
+import (
+	"bytes"
+    "fmt"
+	"strings"
+    "text/template"
+
+	"github.com/Masterminds/sprig/v3"
+)
```
TessaIO, line 26: nit: formatting issues here

pkg/utils/template.go
```diff
@@ -0,0 +1,77 @@
...
+type TemplateHelper struct {
+	template *template.Template
+}
+
+func NewTemplateHelper(name string) (*TemplateHelper, error) {
+	tmpl := template.New("").Funcs(sprig.FuncMap()).Option("missingkey=error")
+    tmpl = tmpl.Funcs(template.FuncMap{
+        "asLabelValue": func(s string) string {
+            // Replace spaces and periods with dashes
+            return strings.ReplaceAll(strings.ReplaceAll(s, " ", "-"), ".", "-")
```
marquiz, line 37: I'd suggest to add this in a separate PR so we can polish it there. The func needs to be more complex  as
>  label must be an empty string or consist of alphanumeric characters, `-`, `_` or `.`, and must start and end with an alphanumeric character
> A label key and value must begin with a letter or number, and may contain letters, numbers, hyphens, dots, and underscores, up to 63 characters each.

So we'd need something like this:
```golang
	re := regexp.MustCompile(`[^a-zA-Z0-9._-]+`)
	return re.ReplaceAllString(input, "_")
```

And probably even more complex than that so that it would guarantee that the end result is accepted as a label value. Do length checking (trimming) and somehow handle cases where the value would start with non-alphanumeric.

WDYT?

pkg/utils/template.go
marquiz, line 1: Is there a reason for moving the file? If not, I'd suggest to keep it where it was to make the review easier.

marquiz (commented):
Thank you @fahhem for the contribution. That was quick :smile: A few comments below.

Also, could you update the documentation (https://kubernetes-sigs.github.io/node-feature-discovery/v0.16/usage/customization-guide.html#templating) to mention that the sprig funcs are supported?

pkg/utils/template.go
fahhem: I'm not familiar enough with Go to write this correctly in the time that I have. I'm happy to remove this and let you add a better one later

fahhem:
Looks like all the tests pass now!

pkg/utils/template.go
marquiz: > It's split out from rule.go so that it's shared with validate.go, otherwise I'd have to duplicate this to validate templates correctly

Ach, of course :man_facepalming: Small suggestion/nit, though: I think a better location would be to put this in new package under pkg/apis/nfd, say `pkg/apis/nfd/utils/` or `pkg/apis/nfd/template/` as the functionality is very much tied to the API and NodeFeatureRule. WDYT?

go.mod
marquiz, line 1: nit: could you run `go mod tidy` to clean up `go.{mod,sum}`?

marquiz (commented):
@fahhem thanks for the update, looks very good now, just two small nits/suggestions below. I think you could also squash the commits into one.

pkg/utils/template.go
fahhem: I tried this, but my Go-fu is weak and it wouldn't let me reference the new location, it kept telling me to `go get ...` but I'm pretty sure that's not what I'm supposed to do for same-repo "dependencies"

fahhem:
I don't understand why that test is failing, there's no tracebacks or anything, it just dies after 2 hours

marquiz:
> I don't understand why that test is failing, there's no tracebacks or anything, it just dies after 2 hours

Ach, that is some flake in the cross-build (arm64) builds that I'm not sure why it gets stuck. Just need to re-test until it succeeds
/retest

marquiz:
Thanks @fahhem for the update. I put my review comment in a form of a PR
https://github.com/SonicInfra/node-feature-discovery/pull/1

If you think that is ok you can just squash it into your commit.

fahhem:
Thanks! I'll squash them now

marquiz:
@fahhem could you do a (clean) rebase the patch on top of `master` to get rid of the merge commit?

fahhem:
@marquiz okay, done!

ArangoGutierrez (approved):

ArangoGutierrez (approved):
/lgtm

PR #2040, opened 2025-02-04
Allow DS_PROMETHEUS substitution for grafana dashboard

szinn:
This PR allows the substitution to be used since the datasource was not being picked up correctly


marquiz (approved):
Thank you @szinn for the improvement.

Other ideas/enhancements for the NFD example dashboard are very much welcome, too. What we have here was just a very quick sketch by me :blush: 

/lgtm

PR #2041, opened 2025-02-05
nfd-worker: Add support to configure the ownerReference to Node

### LINKED ISSUE #2039: Default ownerReference for NodeFeature should be Node

I think the default/out-of-box behavior when NFD is installed should be that the `NodeFeature` CRs should have their owner reference set to `v1.Node` object.

**Why is this needed**:

The rationale is basically summarized at https://ahmet.im/blog/nfd-incident/. Basically, any other alternative is worse:

* **[Owner is DaemonSet Pod (current default)](https://github.com/kubernetes-sigs/node-feature-discovery/blob/4f24a38ad48e364b4164baf01d81a2967e59e2c7/pkg/nfd-worker/nfd-worker.go#L256-L278)**: Means your node labels are gonna get cleared during a rolling update of daemonset. No-go for a lot of installations that want to guarantee labels will always be there.
* **[No owner (configured via a CLI flag)](https://github.com/kubernetes-sigs/node-feature-discovery/blob/4f24a38ad48e364b4164baf01d81a2967e59e2c7/cmd/nfd-worker/main.go#L126-L127):** Means you'll leak NodeFeature CRs (though the controller can totally clean these up during periodic resyncs if it has that logic in nfd-gc). 

**Parenting to v1.Node** has the following advantages:
- NodeFeature resource doesn't get randomly deleted by the controller (and cause incidents like the one linked above) and its lifespan is now tied to Node itself.
- Eliminates the need for nfd-gc as Kubernetes garbage collector in kube-controller-manager would now handle the removal.

I can't think of any downsides to having a single ownerReference set to the Node object.

ozhuraki:
Closes: #2039

### DIFF
--- a/docs/deployment/uninstallation.md
+++ b/docs/deployment/uninstallation.md
@@ -31,3 +31,12 @@ kubectl -n node-feature-discovery wait job.batch/nfd-master --for=condition=comp
 
 > **NOTE:** You must run prune before removing the RBAC rules (serviceaccount,
 > clusterrole and clusterrolebinding).
+
+## Removing stale NodeFeature objects
+
+In case NFD-Worker is configured to set the owner reference to Node or Pod,
+manually delete stale NodeFeature objects.
+
+```bash
+kubectl delete -n node-feature-discovery NodeFeature $(kubectl get -o json -n node-feature-discovery NodeFeature | jq ".items[].metadata.name")
+```
--- a/docs/reference/worker-commandline-reference.md
+++ b/docs/reference/worker-commandline-reference.md
@@ -179,6 +179,23 @@ Example:
 nfd-worker -no-owner-refs
 ```
 
+### -owner-ref-pod
+
+The `-owner-ref-pod` flag sets the owner references
+of the NodeFeature object to Pod.
+
+> **NOTE:** This flag takes precedence over the
+> [`core.OwnerRefNode`](worker-configuration-reference.md#coreownerrefpod)
+> configuration file option.
+
+Default: *false*
+
+Example:
+
+```bash
+nfd-worker -owner-ref-pod
+```
+
 ### -oneshot
 
 The `-oneshot` flag causes nfd-worker to exit after one pass of feature
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -81,6 +81,7 @@ type coreConfig struct {
 	LabelWhiteList utils.RegexpVal
 	NoPublish      bool
 	NoOwnerRefs    bool
+	OwnerRefPod    bool
 	FeatureSources []string
 	Sources        *[]string
 	LabelSources   []string
@@ -101,6 +102,7 @@ type Args struct {
 	Options     string
 	Port        int
 	NoOwnerRefs bool
+	OwnerRefPod bool
 
 	Overrides ConfigOverrideArgs
 }
@@ -109,6 +111,7 @@ type Args struct {
 type ConfigOverrideArgs struct {
 	NoPublish      *bool
 	NoOwnerRefs    *bool
+	OwnerRefPod    *bool
 	FeatureSources *utils.StringSliceVal
 	LabelSources   *utils.StringSliceVal
 }
@@ -253,33 +256,53 @@ func (w *nfdWorker) setOwnerReference() error {
 	ownerReference := []metav1.OwnerReference{}
 
 	if !w.config.Core.NoOwnerRefs {
-		// Get pod owner reference
-		podName := os.Getenv("POD_NAME")
-		// Add pod owner reference if it exists
-		if podName != "" {
-			if selfPod, err := w.k8sClient.CoreV1().Pods(w.kubernetesNamespace).Get(context.TODO(), podName, metav1.GetOptions{}); err != nil {
-				klog.ErrorS(err, "failed to get self pod, cannot inherit ownerReference for NodeFeature")
-				return err
-			} else {
-				for _, owner := range selfPod.OwnerReferences {
-					owner.BlockOwnerDeletion = ptr.To(false)
-					ownerReference = append(ownerReference, owner)
+		if w.config.Core.OwnerRefPod {
+			// Get pod owner reference
+			podName := os.Getenv("POD_NAME")
+			// Add pod owner reference if it exists
+			if podName != "" {
+				if selfPod, err := w.k8sClient.CoreV1().Pods(w.kubernetesNamespace).Get(context.TODO(), podName, metav1.GetOptions{}); err != nil {
+					klog.ErrorS(err, "failed to get self pod, cannot inherit ownerReference for NodeFeature")
+					return err
+				} else {
+					for _, owner := range selfPod.OwnerReferences {
+						owner.BlockOwnerDeletion = ptr.To(false)
+						ownerReference = append(ownerReference, owner)
+					}
 				}
-			}
 
-			podUID := os.Getenv("POD_UID")
-			if podUID != "" {
-				ownerReference = append(ownerReference, metav1.OwnerReference{
-					APIVersion: "v1",
-					Kind:       "Pod",
-					Name:       podName,
-					UID:        types.UID(podUID),
-				})
+				podUID := os.Getenv("POD_UID")
+				if podUID != "" {
+					ownerReference = append(ownerReference, metav1.OwnerReference{
+						APIVersion: "v1",
+						Kind:       "Pod",
+						Name:       podName,
+						UID:        types.UID(podUID),
+					})
+				} else {
+					klog.InfoS("Cannot append POD ownerReference to NodeFeature, POD_UID not specified")
+				}
 			} else {
-				klog.InfoS("Cannot append POD ownerReference to NodeFeature, POD_UID not specified")
+				klog.InfoS("Cannot set NodeFeature owner references, POD_NAME not specified")
 			}
 		} else {
-			klog.InfoS("Cannot set NodeFeature owner references, POD_NAME not specified")
+			// Get node owner reference
+			nodeName := os.Getenv("NODE_NAME")
+			if nodeName != "" {
+				if selfNode, err := w.k8sClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{}); err != nil {
+					klog.ErrorS(err, "failed to get self node, cannot inherit ownerReference for NodeFeature")
+					return err
+				} else {
+					ownerReference = append(ownerReference, metav1.OwnerReference{
+						APIVersion: "v1",
+						Kind:       "Node",
+						Name:       nodeName,
+						UID:        selfNode.UID,
+					})
+				}
+			} else {
+				klog.InfoS("Cannot set NodeFeature owner reference to Node, NODE_NAME not specified")
+			}
 		}
 	}
 
@@ -504,6 +527,9 @@ func (w *nfdWorker) configure(filepath string, overrides string) error {
 	}
 	if w.args.Overrides.NoOwnerRefs != nil {
 		c.Core.NoOwnerRefs = *w.args.Overrides.NoOwnerRefs
+	}
+	if w.args.Overrides.OwnerRefPod != nil {
+		c.Core.OwnerRefPod = *w.args.Overrides.OwnerRefPod
 	}
 	if w.args.Overrides.FeatureSources != nil {
 		c.Core.FeatureSources = *w.args.Overrides.FeatureSources
--- a/pkg/nfd-worker/nfd-worker_test.go
+++ b/pkg/nfd-worker/nfd-worker_test.go
@@ -48,9 +48,12 @@ func TestRun(t *testing.T) {
 		Convey("When publishing features from fake source", func() {
 			os.Setenv("NODE_NAME", "fake-node")
 			os.Setenv("KUBERNETES_NAMESPACE", "fake-ns")
+			b := true
 			args := &worker.Args{
-				Oneshot: true,
+				Oneshot:     true,
+				NoOwnerRefs: true,
 				Overrides: worker.ConfigOverrideArgs{
+					NoOwnerRefs:    &b,
 					FeatureSources: &utils.StringSliceVal{"fake"},
 					LabelSources:   &utils.StringSliceVal{"fake"},
 				},
--- a/test/e2e/utils/rbac.go
+++ b/test/e2e/utils/rbac.go
@@ -63,6 +63,11 @@ func ConfigureRBAC(ctx context.Context, cs clientset.Interface, ns string) error
 		return err
 	}
 
+	_, err = createClusterRoleWorker(ctx, cs)
+	if err != nil {
+		return err
+	}
+
 	_, err = createClusterRoleGC(ctx, cs)
 	if err != nil {
 		return err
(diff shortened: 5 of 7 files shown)


marquiz (commented):
Thanks @ozhuraki for this.

I started to wonder should we make this default behavior, instead, name the option `OwnerRefPod` (defaulting to false). In this case we'd also need to update the `docs/deployment/uninstallation.md` with instructions how to remove stale nodefeature objects. Probably update other pages as well (like `docs/deployment/uninstallation.md#uninstallation`) and note/warn about stale NodeFeature files after uninstallation.

/cc @ArangoGutierrez @adrianchiris @ahmetb 
/ok-to-test

ahmetb:
/lifecycle frozen

Still very much interested in this. 🙏

marquiz:
/reopen
ping @ozhuraki

marquiz:
ping @ozhuraki, any update on this? I think this would still be a valuable enhancement (my [comment](https://github.com/kubernetes-sigs/node-feature-discovery/pull/2041#pullrequestreview-2725984403) still holds)

pkg/nfd-worker/nfd-worker.go
```diff
@@ -253,33 +256,53 @@ func (w *nfdWorker) setOwnerReference() error {
...
-		if podName != "" {
-			if selfPod, err := w.k8sClient.CoreV1().Pods(w.kubernetesNamespace).Get(context.TODO(), podName, metav1.GetOptions{}); err != nil {
-				klog.ErrorS(err, "failed to get self pod, cannot inherit ownerReference for NodeFeature")
-				return err
-			} else {
-				for _, owner := range selfPod.OwnerReferences {
-					owner.BlockOwnerDeletion = ptr.To(false)
-					ownerReference = append(ownerReference, owner)
+		if w.config.Core.OwnerRefPod {
+			// Get pod owner reference
```
ArangoGutierrez, line 260: Main concern with the current shape: wrapping the pre-existing Pod path in `if OwnerRefPod` and making Node the `else` flips the default owner from Pod to Node. Existing deployments upgrading would see their `NodeFeature` objects stop being owned by the worker Pod/DaemonSet.

Could we keep Pod as the default and make Node the opt-in? And, larger question: would you be open to a single enum flag `-owner-ref=none|pod|node` (deprecating `-no-owner-refs`) so we don't have two booleans encoding three states?

pkg/nfd-worker/nfd-worker.go
```diff
@@ -253,33 +256,53 @@ func (w *nfdWorker) setOwnerReference() error {
...
-			klog.InfoS("Cannot set NodeFeature owner references, POD_NAME not specified")
+			// Get node owner reference
+			nodeName := os.Getenv("NODE_NAME")
+			if nodeName != "" {
+				if selfNode, err := w.k8sClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{}); err != nil {
+					klog.ErrorS(err, "failed to get self node, cannot inherit ownerReference for NodeFeature")
+					return err
+				} else {
+					ownerReference = append(ownerReference, metav1.OwnerReference{
+						APIVersion: "v1",
```
ArangoGutierrez, line 297: For the Node case, worth setting `Controller` and `BlockOwnerDeletion` explicitly — there's no other controller claiming `NodeFeature`, and being explicit matches the Pod branch and K8s conventions.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -253,33 +256,53 @@ func (w *nfdWorker) setOwnerReference() error {
...
-				klog.InfoS("Cannot append POD ownerReference to NodeFeature, POD_UID not specified")
+				klog.InfoS("Cannot set NodeFeature owner references, POD_NAME not specified")
 			}
 		} else {
-			klog.InfoS("Cannot set NodeFeature owner references, POD_NAME not specified")
+			// Get node owner reference
+			nodeName := os.Getenv("NODE_NAME")
+			if nodeName != "" {
+				if selfNode, err := w.k8sClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{}); err != nil {
+					klog.ErrorS(err, "failed to get self node, cannot inherit ownerReference for NodeFeature")
```
ArangoGutierrez, line 293: Small thing while we're here: could we thread a `context.Context` into this function instead of `context.TODO()`? Pre-existing on the Pod branch, nice to fix on the way through.

pkg/nfd-worker/nfd-worker_test.go
```diff
@@ -48,9 +48,12 @@ func TestRun(t *testing.T) {
 		Convey("When publishing features from fake source", func() {
 			os.Setenv("NODE_NAME", "fake-node")
 			os.Setenv("KUBERNETES_NAMESPACE", "fake-ns")
+			b := true
 			args := &worker.Args{
-				Oneshot: true,
+				Oneshot:     true,
```
ArangoGutierrez, line 53: Setting `NoOwnerRefs: true` here skips the new owner-ref logic entirely — the new code isn't exercised by any test. Could we add a table-driven test over the `{none, pod, node}` configurations, pre-creating a fake Node with a known UID and asserting the resulting `OwnerReferences`? That would also cover the Node-not-found error path.

docs/deployment/uninstallation.md
```diff
@@ -31,3 +31,12 @@ kubectl -n node-feature-discovery wait job.batch/nfd-master --for=condition=comp
...
 > **NOTE:** You must run prune before removing the RBAC rules (serviceaccount,
 > clusterrole and clusterrolebinding).
+
+## Removing stale NodeFeature objects
+
+In case NFD-Worker is configured to set the owner reference to Node or Pod,
+manually delete stale NodeFeature objects.
+
+```bash
+kubectl delete -n node-feature-discovery NodeFeature $(kubectl get -o json -n node-feature-discovery NodeFeature | jq ".items[].metadata.name")
```
ArangoGutierrez, line 41: Small fix: `jq ".items[].metadata.name"` emits quoted strings, which `kubectl delete` will reject. Two options — add `-r` to jq, or simpler: `kubectl delete -n node-feature-discovery nodefeatures.nfd.k8s-sigs.io --all`.

ArangoGutierrez (changes requested):
Sorry this sat so long — thanks for your patience. The feature itself makes sense and I'd like to land a version of it. A few things to work through:

- The branch is behind master and needs a rebase; CI has gone red since April.
- Main design question — see the inline note on `setOwnerReference`. This currently changes the default owner from Pod to Node, which is a behavior change on upgrade for existing users. I'd prefer keeping Pod as the default and making Node opt-in, and ideally collapsing the two booleans into a single `-owner-ref=none|pod|node` flag (deprecating `-no-owner-refs`). Happy to discuss alternatives.
- The test update in `nfd-worker_test.go` sets `NoOwnerRefs: true`, which skips the new owner-ref logic entirely — we'd need tests that actually exercise the new Pod and Node paths.
- Could you also add `ownerRefPod` to `deployment/helm/node-feature-discovery/values.yaml`? Helm users won't otherwise discover the new option.

If any of this is a non-starter for you, let me know — happy to talk it through.

test/e2e/utils/rbac.go
```diff
@@ -239,6 +257,24 @@ func createRoleWorker(ctx context.Context, cs clientset.Interface, ns string) (*
...
+func createClusterRoleWorker(ctx context.Context, cs clientset.Interface) (*rbacv1.ClusterRole, error) {
+	cr := &rbacv1.ClusterRole{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "nfd-worker-e2e",
+		},
+		Rules: []rbacv1.PolicyRule{
+			{
+				APIGroups: []string{""},
+				Resources: []string{"nodes"},
+				Verbs:     []string{"get", "list"},
```
ArangoGutierrez, line 270: This grants `nodes get` to the e2e harness only — the shipped RBAC is unchanged: `deployment/base/rbac/worker-role.yaml` and helm `templates/role.yaml` still give the worker a namespaced Role with just `nodefeatures` + `pods get`. With Node as the default owner path, both deployment methods need a new worker ClusterRole + ClusterRoleBinding granting `nodes get`, otherwise every stock install fails the lookup. The code only ever does a Get, so `list` can be dropped here too.

ArangoGutierrez (commented):
Took another pass. The threads from the previous review are still open (Pod-vs-Node default and the single `-owner-ref=none|pod|node` idea, tests that actually exercise the new paths, `ownerRefPod` in helm values), so I'm not repeating them inline. New finding this round: only the e2e harness gains `nodes get` — the shipped kustomize `worker-role.yaml` and helm `role.yaml` don't — and since Node becomes the default owner path and a failed lookup propagates from `setOwnerReference()` into `Run()`, a stock deployment exits at startup after upgrading. Two smaller items inline: a dead `Args.OwnerRefPod` field and a wrong option name in the CLI reference. The branch still conflicts with master (`configure()` has since grown a ctx parameter) and CI has been red since April — needs the deployment RBAC, the rebase, and the earlier review threads resolved before this can merge.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -253,33 +256,53 @@ func (w *nfdWorker) setOwnerReference() error {
...
+					return err
```
ArangoGutierrez, line 294: When this Get fails, the error propagates from `setOwnerReference()` through `configure()` into `Run()` and the worker exits without publishing anything — and with the shipped RBAC lacking `nodes get` for the worker SA, that is the default outcome on upgrade. Owner refs are best-effort metadata: the NODE_NAME-unset case below already degrades to no owner ref, so an apiserver error here should log and continue rather than take feature discovery down.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -101,6 +102,7 @@ type Args struct {
 	Options     string
 	Port        int
 	NoOwnerRefs bool
+	OwnerRefPod bool
```
ArangoGutierrez, line 105: `Args.OwnerRefPod` is never read — `configure()` only consumes `Args.Overrides.OwnerRefPod`. Same dead-field shape as the pre-existing `Args.NoOwnerRefs`; better to drop this one than extend the pattern.

docs/reference/worker-commandline-reference.md
```diff
@@ -179,6 +179,23 @@ Example:
 nfd-worker -no-owner-refs
 ```
 
+### -owner-ref-pod
+
+The `-owner-ref-pod` flag sets the owner references
+of the NodeFeature object to Pod.
+
+> **NOTE:** This flag takes precedence over the
+> [`core.OwnerRefNode`](worker-configuration-reference.md#coreownerrefpod)
```
ArangoGutierrez, line 188: Copy-paste slip: the config option this references is `core.ownerRefPod`, not `core.OwnerRefNode`.
```suggestion
> [`core.ownerRefPod`](worker-configuration-reference.md#coreownerrefpod)
```

pkg/nfd-worker/nfd-worker.go
```diff
@@ -253,33 +256,53 @@ func (w *nfdWorker) setOwnerReference() error {
...
 			} else {
-				klog.InfoS("Cannot append POD ownerReference to NodeFeature, POD_UID not specified")
+				klog.InfoS("Cannot set NodeFeature owner references, POD_NAME not specified")
 			}
 		} else {
-			klog.InfoS("Cannot set NodeFeature owner references, POD_NAME not specified")
+			// Get node owner reference
+			nodeName := os.Getenv("NODE_NAME")
+			if nodeName != "" {
+				if selfNode, err := w.k8sClient.CoreV1().Nodes().Get(context.TODO(), nodeName, metav1.GetOptions{}); err != nil {
```
ArangoGutierrez, line 292: **Blocker:** this needs cluster-scoped `get nodes`, but the production worker RBAC (`deployment/base/rbac/worker-role.yaml`) is a namespaced `Role` granting only `nodefeatures` + `pods get` — and a `Role` can't grant `nodes` at all. So this `Get` returns Forbidden on a normal deploy and the function then `return err`s, breaking NodeFeature creation by default. Please add a worker `ClusterRole` + `ClusterRoleBinding` (`nodes: [get]`) to the deployment manifests (helm template + `deployment/base/rbac`), not just the e2e RBAC.

pkg/nfd-worker/nfd-worker_test.go
```diff
@@ -48,9 +48,12 @@ func TestRun(t *testing.T) {
...
+				NoOwnerRefs: true,
```
ArangoGutierrez, line 54: Setting `NoOwnerRefs: true` here disables the owner-ref path, so the new Node-owner-reference logic — the core of this PR — ends up untested. Please add a case that injects a fake Node and asserts the resulting `OwnerReference{Kind: "Node", Name, UID}`.

ArangoGutierrez (changes requested):
Thanks — the Node-ownership goal (fix #2039: NodeFeature surviving pod restarts) makes sense. A few things before it can merge:

**Blocker — production RBAC (see inline):** the new default fetches the Node via the API, but the deployed worker RBAC (`deployment/base/rbac/worker-role.yaml`) is a namespaced `Role` granting only `nodefeatures` + `pods get` — no `nodes`, and a `Role` can't authorize the cluster-scoped `nodes` resource at all. Only `test/e2e/utils/rbac.go` is updated. So on a standard deploy the `Nodes().Get` returns Forbidden and `setOwnerReference` `return err`s → NodeFeature creation breaks by default. Needs a worker `ClusterRole` + `ClusterRoleBinding` with `nodes: [get]` in the helm template and `deployment/base/rbac`.

**Test:** the new Node-owner-ref logic is untested — the only test change *disables* owner refs.

**Rebase:** `needs-rebase` (and it overlaps `pkg/nfd-worker/nfd-worker.go` with #2464 — worth sequencing the two).

**Behavior change:** default owner ref Pod→Node is user-visible (GC semantics) — worth a release note + a maintainer nod. Docs look good.

Marking request-changes; happy to re-review once addressed. Thanks!

ArangoGutierrez:
Two more findings from a fresh pass, on top of the open items:

- `test/e2e/utils/rbac.go`: `createClusterRoleWorker` / `createClusterRoleBindingWorker` call `Update()` on freshly-built objects with no `ResourceVersion`. On a clean cluster that returns `NotFound` and `ConfigureRBAC` aborts — should be `Create` (or Create, then Update on `AlreadyExists`).
- `pkg/nfd-worker/nfd-worker.go`: the Node `OwnerReference` is constructed without `Controller` or `BlockOwnerDeletion` (the Pod branch sets `BlockOwnerDeletion: false`). Set them explicitly (`Controller: ptr.To(true)`, `BlockOwnerDeletion: ptr.To(false)`) to match conventions and the Pod branch.

The earlier blockers still stand: the production worker RBAC is a namespaced `Role` and can't grant cluster-scoped `nodes get`, so the new default path returns Forbidden on a stock deploy; the Node-owner-ref path has no test; and the branch needs a rebase.

ArangoGutierrez:
Thanks for the work here, and apologies it sat this long.

This is superseded by #2545, which merged on 8 July (7d333448) and shipped in v0.19.0. That
PR made the NodeFeature owner reference configurable via `-owner-refs` (node, pod, ds) and
covers the same two files this one touches. You went through three rounds of review here, and
the outcome you were pushing for did land, just through the other PR.

One thing your issue asked for is still open: the default is still `pod,ds` rather than
`node` (cmd/nfd-worker/main.go:128), so #2039 remains a live decision about whether to change
that default. Input there would be welcome.

Closing this one as superseded.

PR #2055, opened 2025-02-17
feat: add support to differentiate specific hypervisors on s390x

### LINKED ISSUE #2037: Add support to differentiate specific hypervisors on s390x

Add support to differentiate hypervisors on s390x nodes. While the existing `HYPERVISOR` label works for `KVM`, it does not identify `z/VM` as virtual machine. Using tools like `lscpu`, this feature would add finer-grained labels for hypervisors such as `z/VM` and `PR/SM (LPAR).`

lscpu ouput for z/VM: 
```
# lscpu
[...]
Virtualization features:
  Hypervisor:             z/VM 7.2.0
  Hypervisor vendor:      IBM
  Virtualization type:    full
[...]

# lscpu --json | \
  jq -r '.lscpu[] | select(.field == "Hypervisor:") | .data'
z/VM 7.2.0
```

**Why is this needed**:

The current `HYPERVISOR` label is insufficient for s390x nodes, as it only identifies `KVM`. Differentiating `z/VM `and `LPAR` is essential for accurate workload scheduling and operational requirements in s390x environments.

routerhan:
Resolves #2037

### DIFF
--- a/source/cpu/cpu.go
+++ b/source/cpu/cpu.go
@@ -19,6 +19,7 @@ package cpu
 import (
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 
@@ -263,7 +264,43 @@ func getCPUModel() map[string]string {
 	cpuModelInfo["family"] = strconv.Itoa(cpuid.CPU.Family)
 	cpuModelInfo["id"] = strconv.Itoa(cpuid.CPU.Model)
 
+	hypervisor, err := getHypervisor()
+	if err != nil {
+		klog.ErrorS(err, "failed to detect hypervisor")
+	} else if hypervisor != "" {
+		cpuModelInfo["hypervisor"] = hypervisor
+	}
+
 	return cpuModelInfo
+}
+
+// getHypervisor detects the hypervisor on s390x by reading /proc/sysinfo.
+// If the file does not exist, it returns an empty string with no error.
+func getHypervisor() (string, error) {
+	if _, err := os.Stat("/proc/sysinfo"); os.IsNotExist(err) {
+		return "", nil
+	}
+
+	data, err := os.ReadFile("/proc/sysinfo")
+	if err != nil {
+		return "", err
+	}
+
+	hypervisor := "PR/SM"
+	for _, line := range strings.Split(string(data), "\n") {
+		if strings.Contains(line, "Control Program:") {
+			parts := strings.SplitN(line, ":", 2)
+			if len(parts) == 2 {
+				hypervisor = strings.TrimSpace(parts[1])
+			}
+			break
+		}
+	}
+	// Replace forbidden symbols
+	fullRegex := regexp.MustCompile("[^-A-Za-z0-9_.]+")
+	hypervisor = fullRegex.ReplaceAllString(hypervisor, "_")
+
+	return hypervisor, nil
 }
 
 func discoverTopology() map[string]string {
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -921,6 +921,7 @@ The following features are available for matching:
 |                  |              | **`family`** | int    | CPU family |
 |                  |              | **`vendor_id`** | string | CPU vendor ID |
 |                  |              | **`id`** | int        | CPU model ID |
+|                  |              | **`hypervisor`** | string | Hypervisor type information from `/proc/sysinfo` (s390x-only feature) |
 | **`cpu.pstate`** | attribute    |          |            | State of the Intel pstate driver. Does not exist if the driver is not enabled. |
 |                  |              | **`status`** | string | Status of the driver, possible values are 'active' and 'passive' |
 |                  |              | **`turbo`**  | bool   | 'true' if turbo frequencies are enabled, otherwise 'false' |
--- a/go.mod
+++ b/go.mod
@@ -13,7 +13,7 @@ require (
 	github.com/k8stopologyawareschedwg/podfingerprint v0.2.2
 	github.com/klauspost/cpuid/v2 v2.2.10
 	github.com/onsi/ginkgo/v2 v2.23.3
-	github.com/onsi/gomega v1.36.3
+	github.com/onsi/gomega v1.36.2
 	github.com/opencontainers/image-spec v1.1.1
 	github.com/opencontainers/runc v1.2.6
 	github.com/prometheus/client_golang v1.21.1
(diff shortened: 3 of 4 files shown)


source/cpu/cpuid_linux_s390x.go
```diff
@@ -96,4 +102,29 @@ func getCpuidFlags() []string {
...
+	data, err := os.ReadFile("/proc/sysinfo")
+	if err != nil {
+		klog.ErrorS(err, "failed to read /proc/sysinfo")
+		return ""
+	}
+
+	for _, line := range strings.Split(string(data), "\n") {
+		if strings.Contains(line, "Control Program:") {
+			parts := strings.SplitN(line, ":", 2)
+			if len(parts) == 2 {
```
mfranczy, line 122: If this condition fails, should not you break the loop? I suppose there is only one `Control Program` line in `/proc/sysinfo`?

source/cpu/cpuid_linux_s390x.go
```diff
@@ -24,6 +24,12 @@ unsigned long gethwcap() {
 }
 */
 import "C"
+import (
```
mfranczy, line 27: Could you add a new line between your imports and `import "C"`?

routerhan: Done.

source/cpu/cpuid_linux_s390x.go
routerhan: I rework on it. Hope it make more sense. 
Thank you for reviewing.

source/cpu/cpuid_linux_s390x.go
```diff
@@ -96,4 +102,29 @@ func getCpuidFlags() []string {
...
+	ret["hypervisor"] = getHypervisor()
+
+	return ret
+}
+
+func getHypervisor() string {
+	data, err := os.ReadFile("/proc/sysinfo")
+	if err != nil {
+		klog.ErrorS(err, "failed to read /proc/sysinfo")
+		return ""
```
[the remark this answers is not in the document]
routerhan: Agreed. `getHypervisor()` now returns err and be handled in `getCpuidAttributes()`

ozhuraki (approved):

mfranczy (approved):
lgtm 

/cc @marquiz

marquiz (commented):
Thank you @routerhan for the enhancement.

Sorry, I'm a bit late in the game and rampaging here :roll_eyes:

First observation, I'd consider naming the feature `hypervisor_name`. The reason being that on x86_64 side we already have `HYPERVISOR` (flag telling if we're virtualized). We might want to add this new feature on the x86 side, and using the same name would be a bonus (`HYPERVISOR` and `hypervisor` are not truly clashing but definitely confusing if we have both).

Second thought/question. We have the cpu model feature (`model` and `getCPUModel()`) so we could consider putting this there. Then also renaming would not be important. Writing this I'm not really certain would it be more logical under `cpuid` or `model` feature :blush: 

Thoughts @routerhan @mfranczy @ArangoGutierrez ?

/ok-to-test

mfranczy:
> Writing this I'm not really certain would it be more logical under cpuid or model feature 😊

Good point! I believe cpuid should strictly refer to cases where it directly invokes the CPUID instruction for clarity.
CPU model is a better option, but I think `system` category is the best choice here :smile: 

My reasoning is that this check primarily determines which virtualization software is running on the node, rather than verifying CPU virtualization capabilities or whether the node itself is a virtual machine.

marquiz:
> whether the node itself is a virtual machine.

Isn't this exactly what this PR is doing(?)

mfranczy:
> Isn't this exactly what this PR is doing(?)

My understanding of this PR is that it detects which hypervisor is running on the node that hosts VMs, not if it the node is a guest itself. But maybe I am wrong @routerhan?

mfranczy:
Ah that's my bad, sorry.

> Add support to differentiate hypervisors on s390x nodes. While the existing HYPERVISOR label works for KVM, it does not identify z/VM as virtual machine. Using tools like lscpu, this feature would add finer-grained labels for hypervisors such as z/VM and PR/SM (LPAR).

I should read the issue before reviewing the PR.

routerhan:
Thank you @marquiz for the feedbacks.

> Second thought/question. We have the cpu model feature (`model` and `getCPUModel()`) so we could consider putting this there. Then also renaming would not be important. Writing this I'm not really certain would it be more logical under `cpuid` or `model` feature 😊
> Thoughts @routerhan @mfranczy @ArangoGutierrez ?

It also makes sense to me to put it under `model` feature.
So here is what I plan to do: 
I'll put `getHypervisor()` into `cpu.go` file, and calling `getHypervisor()` from `getCPUModel()`

What do you think @marquiz @mfranczy ?

marquiz:
> It also makes sense to me to put it under `model` feature. So here is what I plan to do: I'll put `getHypervisor()` into `cpu.go` file, and calling `getHypervisor()` from `getCPUModel()`
> 
> One more thing, this PR is s390x-specific solution, other architecture like x86 would throw an error, since they do not have `/proc/sysinfo` for detecting hypervisor information.

AFAIU, `/proc/sysinfo` is very specific to s390x so from that point-of-view it (`getHypervisor`) could reside in an arch specific source file. But I think would be good enough to check the existence of `/proc/sysinfo` and only try to read it if it exists (not spitting out errors on other archs). And, only set the feature if hypervisor != "". So, I'll leave it up to you, as long as we don't see unnecessary misleading errors on other archs.

source/cpu/cpu.go
```diff
@@ -263,9 +263,42 @@ func getCPUModel() map[string]string {
...
+	if _, err := os.Stat("/proc/sysinfo"); os.IsNotExist(err) {
+		return "", nil
+	}
+
+	data, err := os.ReadFile("/proc/sysinfo")
+	if err != nil {
+		return "", err
+	}
+
+	hypervisor := "PR/SM"
```
marquiz, line 288: By default we return "PR/SM" even if the machine is not a VM(?)

routerhan: The PR/SM is always the underlying hypervisor, and this is what we expect to see as the default.
This is also consistent to other tooling like `lscpu`.

marquiz: Sorry for my ignorance, but trying to educate myself. On Z you're always having a hypervisor, never running os on "bare metal"?

routerhan: Hi @marquiz, 

The underlying hypervisor that partitions the mainframe resources into logical partitions (LPARs) is always PR/SM. Depending on the use case, in an LPAR either
- a Linux guest is running directly (i.e. Hypervisor is PR/SM),
- or a second hypervisor (KVM or z/VM) is running in the LPAR and partitions the LPAR resources further. The nodes then run as guests in KVM or z/VM.

Thanks to @werled for s390x detailed knowledge.

marquiz: Got that. Thank you @routerhan for educating me.

marquiz (approved):
Thanks @routerhan. Looks good to me
/assign @ArangoGutierrez 
ping @mfranczy

mfranczy (approved):
lgtm

routerhan:
> lgtm

Hi @mfranczy, thanks for reviewing, but I think the lgtm tag was not set properly. (?)

PR #2056, opened 2025-02-17
Introduce memory.hugepages feature

mfranczy:
I added `memory.hugepages`  feature with the following attributes:
* `enabled`  - true` if total number of huge pages (of any page size) have been configured, otherwise `false`
* `hugepages-<page-size>` - total number of huge pages (e.g., `hugepages-1Gi=16`)


mfranczy:
/cc @ArangoGutierrez @marquiz 

@ChaoyiHuang please take a look as well.

ChaoyiHuang:
> What is the use case for this? Hugepages should be available as available resources/capacity in the Node object

some product requires both 2MB and 1G size hugepages, it is good enough to check the initial capacity (boot parameter). how to allocate and recycle the hugepages should be the responsibility of the applications.

dynamic capacity detection during runtime is unpredicatble, and lead to contention if multiple applications demand hugepages. the issue should be addressed in resources allocation in scheduling.

mfranczy:
@marquiz looks like I misunderstood the requirement from our environments. Sorry for that. I have to change the implementation, but still the `memory.hugepages` is something we need.

We require a pre-flight check on nodes to verify if huge pages are enabled, and if they are, we need information about configured page sizes. This helps determine whether the image is compatibile with the node. This should be done ahead of the scheduling phase. However, we don't need to check capacity or available huge page resources, as you pointed out, we can leave it to scheduler.

mfranczy:
I updated the PR, as mentioned before, it's only to check if required page sizes have been configured on node, capacity / available resources is left to scheduler. 

We need this for image compatibility.

docs/usage/customization-guide.md
```diff
@@ -966,6 +966,9 @@ The following features are available for matching:
 |                  |              | **`node_count`** | int | Number of NUMA nodes |
 | **`memory.swap`**  | attribute  |          |            | Swap enabled on node |
 |                  |              | **`enabled`** | bool  | `true` if swap partition detected, `false` otherwise |
+| **`memory.hugepages`**  | attribute  |          |       | Discovery of huge pages size on node |
+|                  |              | **`enabled`** | bool  | `true` if huge pages size has been configured, otherwise `false` |
+|                  |              | **`hugepages-<page-size>`** | bool   | `true` if the corresponding huge page size is enabled (e.g., `hugepages-2Mi`), the attribute is absent if not enabled |
```
ChaoyiHuang, line 971: hugepages size maybe in kB, not Mi

source/memory/memory.go
```diff
@@ -179,6 +192,46 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+
+	reg := regexp.MustCompile(`hugepages-(\d+)`)
+	for _, entry := range subdirs {
+		matches := reg.FindStringSubmatch(entry.Name())
+		if len(matches) != 2 {
+			klog.ErrorS(fmt.Errorf("unexpected directory format"), "unexpectedDir", entry.Name())
+			continue
+		}
+
+		sizeFormat := matches[1] + "Ki"
```
ChaoyiHuang, line 222: just iterate the name of subdir is not enough

```
root@aiagent:/sys/kernel/mm/hugepages# ls
hugepages-1048576kB  hugepages-2048kB
root@aiagent:/sys/kernel/mm/hugepages# cat hugepages-1048576kB/nr_hugepages
0
root@aiagent:/sys/kernel/mm/hugepages# cat hugepages-2048kB/nr_hugepages
0
root@aiagent:/sys/kernel/mm/hugepages# cat /proc/cmdline 
BOOT_IMAGE=/boot/vmlinuz-6.8.0-1020-gcp root=PARTUUID=2c3e7707-8110-4bed-8029-add7d3035646 ro console=ttyS0,115200 panic=-1
```

even no hugepage configured, there are still subdir hugepages-1048576kB  hugepages-2048kB

source/memory/memory.go
```diff
@@ -179,6 +192,46 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+		quantity, err := resource.ParseQuantity(sizeFormat)
+		if err != nil {
+			klog.ErrorS(err, "Skipping invalid size format", "sizeFormat", sizeFormat)
+			continue
+		}
+
+		hugePages["hugepages-"+quantity.String()] = "true"
```
ChaoyiHuang, line 229: test cases are good to have

source/memory/memory.go
```diff
@@ -179,6 +192,46 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+			klog.ErrorS(fmt.Errorf("unexpected directory format"), "unexpectedDir", entry.Name())
+			continue
+		}
+
+		sizeFormat := matches[1] + "Ki"
+		quantity, err := resource.ParseQuantity(sizeFormat)
+		if err != nil {
+			klog.ErrorS(err, "Skipping invalid size format", "sizeFormat", sizeFormat)
+			continue
+		}
```
ChaoyiHuang, line 227: kernel parameters : https://www.kernel.org/doc/Documentation/admin-guide/kernel-parameters.txt

source/memory/memory.go
mfranczy: Yes, you're right the allocation must be checked.

docs/usage/customization-guide.md
mfranczy: Huge page sizes are in kB, `resource.Quantity` converts the size to the most readable unit.
For example 2048kB will be converted to 2Mi, 1048576kB to 1Gi etc.

source/memory/memory.go
mfranczy: Fixed.

docs/usage/customization-guide.md
mfranczy: Additionally, `/sys/kernel/mm/hugepages` always report page sizes in KB. From the linux kernel doc:

> The root huge page control directory in sysfs is:
>
>	/sys/kernel/mm/hugepages
>
>For each huge page size supported by the running kernel, a subdirectory
>will exist, of the form:
>
>	hugepages-${size}kB

source/memory/memory.go
mfranczy: Yes, but before writing tests I would like to agree on the final solution with everybody.

source/memory/memory.go
ChaoyiHuang: > However checking huge pages configuration over kernel args eleminates discovery of configuration made at runtime.

without hugepage size configured in the kernel boot parameters, it's not able to enable hugepages at runtime. for example, if you didnot configure hugepagesz=1G, then it's not able to mount 1G hugepages filesystem or change the amount of hugepages in OS

source/memory/memory.go
```diff
@@ -179,6 +192,73 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+
+		pageSize, err := getHugePageSize(basePath, entry.Name(), reg)
+		if err != nil {
+			klog.ErrorS(err, "Skipping huge page entry", "entry", entry.Name())
+			continue
+		}
+		if pageSize != "" {
+			if hugePages["enabled"] == "false" {
```
ChaoyiHuang, line 223: no need

source/memory/memory.go
mfranczy: @ChaoyiHuang does the reply satisfy your concern? Or you had something else in your mind that I missed, if so please elaborate.

source/memory/memory.go
```diff
@@ -179,6 +192,73 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+
+	reg := regexp.MustCompile(`hugepages-(\d+)`)
+	for _, entry := range subdirs {
+		if !entry.IsDir() {
+			continue
+		}
+
+		pageSize, err := getHugePageSize(basePath, entry.Name(), reg)
+		if err != nil {
+			klog.ErrorS(err, "Skipping huge page entry", "entry", entry.Name())
```
ChaoyiHuang, line 219: double entry here

source/memory/memory.go
```diff
@@ -179,6 +192,73 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+	return hugePages, nil
+}
+
+func getHugePageSize(basePath, dirName string, reg *regexp.Regexp) (string, error) {
+	matches := reg.FindStringSubmatch(dirName)
+	if len(matches) != 2 {
+		return "", fmt.Errorf("unexpected directory format: %s", dirName)
+	}
+
+	allocatedPagesFile := filepath.Join(basePath, dirName, "nr_hugepages")
```
ChaoyiHuang, line 239: allocatedPagesFile -> numberHugePages

source/memory/memory.go
```diff
@@ -179,6 +192,73 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+	allocatedPagesRaw, err := os.ReadFile(allocatedPagesFile)
+	if err != nil {
+		return "", fmt.Errorf("unable to read allocated pages from %s", allocatedPagesFile)
```
ChaoyiHuang, line 242: unable to read allocated pages from

->

unable to read total hugepages from

the size of hugepages indicated in the fileName

source/memory/memory.go
```diff
@@ -179,6 +192,73 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+	}
+
+	allocatedPages, err := strconv.Atoi(strings.TrimSpace(string(allocatedPagesRaw)))
```
ChaoyiHuang, line 245: allocatedPages -> totalPages

allocation in hugepages mostly mean mmap or fallocate etc allocation.
use "allocate" as prefix will lead to missunderstanding

source/memory/memory.go
ChaoyiHuang: read nr_hugepages also work here

source/memory/memory.go
mfranczy: Fixed.

source/memory/memory.go
mfranczy: Right. Fixed.

source/memory/memory.go
mfranczy: Fixed with a bit different words, but the context stays the same.

source/memory/memory.go
mfranczy: Right. Fixed.

source/memory/memory.go
mfranczy: The function accepts the following args:
`klog.ErrorS(err error, msg string, keysAndValues ...interface{})`

I would not say it's a double entry, personally I don't think it's bad to use the same noun in the message and as a key.

source/memory/memory.go
mfranczy: I added the tests.

mfranczy:
The PR is ready for another round of review.

ChaoyiHuang:
LGTM. thanks for the update.

ArangoGutierrez:
Any thoughts @marquiz ??

source/memory/memory.go
```diff
@@ -179,6 +192,71 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+	}
+
+	return hugePages, nil
+}
+
+func getHugePageSize(basePath, dirName string, reg *regexp.Regexp) (string, error) {
+	matches := reg.FindStringSubmatch(dirName)
+	if len(matches) != 2 {
+		return "", fmt.Errorf("unexpected directory format: %s", dirName)
+	}
```
marquiz, line 235: nit: I'd probably simplify by not using regexp for this simple thing

```suggestion
func getHugePageSize(basePath, dirName string) (string, error) {
	if !strings.HasPrefix(dirname, "hugepages-") {
		return "", fmt.Errorf("unexpected hugepages directory name: %s", dirName)
	}
	sizeFormat := strings.strings.TrimPrefix(dirName, "hugepages-") + "Ki"
```

Optionally we could do this checking/trimming also outside this function and just pass the hugepagesize (as an arg

docs/usage/customization-guide.md
marquiz, line 971: Hmm, I started to think that it would possibly be useful to have this as an int, after all.

For example, I was thinking about some usage scenarios where you'd have `hugepages-2Mi=0`. In the future, when nodes support dynamic resiziing, you know that those hugepages are supported on the system even if not created, yet. Then you could have some management system deployed for adjusting the number of hugepages.

WDYT

source/memory/memory.go
```diff
@@ -179,6 +192,71 @@ func detectNv() ([]nfdv1alpha1.InstanceFeature, error) {
...
+		return "", nil
+	}
+
+	sizeFormat := matches[1] + "Ki"
+	quantity, err := resource.ParseQuantity(sizeFormat)
+	if err != nil {
+		return "", fmt.Errorf("invalid size format: %s", sizeFormat)
+	}
+
+	return "hugepages-" + quantity.String(), nil
```
marquiz, line 257: If we want to indicate that the naming matches with Kubernetes resource names we could possibly use `corev1.ResourceHugePagesPrefix` here.

docs/usage/customization-guide.md
ChaoyiHuang: > Hmm, I started to think that it would possibly be useful to have this as an int, after all.
> 
> For example, I was thinking about some usage scenarios where you'd have `hugepages-2Mi=0`. In the future, when nodes support dynamic resiziing, you know that those hugepages are supported on the system even if not created, yet. Then you could have some management system deployed for adjusting the number of hugepages.
> 
> WDYT

good idea

source/memory/memory.go
mfranczy: I removed the regexp and change the `getHugePageSize` function to `getHugePageTotalCount`.
The logic is a little bit different than suggested although it's simplified.

docs/usage/customization-guide.md
mfranczy: Adjusted to the comment.

source/memory/memory.go
mfranczy: Done.

marquiz (approved):
Thank you @mfranczy for this enhancement. Looks good to me

mfranczy:
Thanks @marquiz, I forgot to add a new line in huge pages test file. Fixed now.

PR #2065, opened 2025-02-28
Change ExecuteGroupRule to return detailed result

marquiz:
Modify the ExecuteGroupRule() which is used for NodeFeatureGroups to return detailed match status, similar to the Execute (used for NodeFeatureRules). This prepares for using the simpler GroupRule type in the image compatibility API.

### DIFF
--- a/pkg/apis/nfd/nodefeaturerule/rule.go
+++ b/pkg/apis/nfd/nodefeaturerule/rule.go
@@ -149,37 +149,48 @@ func Execute(r *nfdv1alpha1.Rule, features *nfdv1alpha1.Features, failFast bool)
 
 // ExecuteGroupRule executes the GroupRule against a set of input features, and return true if the
 // rule matches.
-func ExecuteGroupRule(r *nfdv1alpha1.GroupRule, features *nfdv1alpha1.Features, failFast bool) (bool, error) {
-	matched := false
-	if len(r.MatchAny) > 0 {
+func ExecuteGroupRule(r *nfdv1alpha1.GroupRule, features *nfdv1alpha1.Features, failFast bool) (MatchStatus, error) {
+	var (
+		matchStatus MatchStatus
+		isMatch     bool
+	)
+	if n := len(r.MatchAny); n > 0 {
+		matchStatus.MatchAny = make([]*MatchFeatureStatus, 0, n)
 		// Logical OR over the matchAny matchers
 		for _, matcher := range r.MatchAny {
-			if isMatch, matches, err := evaluateMatchAnyElem(&matcher, features, failFast); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/apis/nfd/nodefeaturerule/rule.go
```diff
@@ -149,37 +149,48 @@ func Execute(r *nfdv1alpha1.Rule, features *nfdv1alpha1.Features, failFast bool)
...
+					// there's no need to evaluate other matchers in MatchAny
+					break
+				}
 			}
+			matchStatus.MatchAny = append(matchStatus.MatchAny, featureStatus)
 		}
-		if !matched {
-			return false, nil
+		if !isMatch {
+			return matchStatus, nil
```
mfranczy, line 176: If `matchAny` fails then you stop processing [matchFeatures](https://github.com/kubernetes-sigs/node-feature-discovery/pull/2065/files#diff-69d932ac4c96e47966ef5df8d26739c2d4fab714f6e2acb675c83230a8004dfdR180). Currently, the spec allows combination of both. I think it would be better to process `matchFeatures` as well to get a full report.

mfranczy: I just realised that `Execute` has the same issue. In case of `matchAny` fails then `matchFeatures` is not processed. I belive we should change this behaviour and process the whole spec.

marquiz: Good point, and well thought :+1: Yes, I agree we should fix this. You could submit a fix to `Execute` and I can then adjust this PR. Note: I'll be OoO for a week starting from tomorrow

pkg/apis/nfd/nodefeaturerule/rule.go
```diff
@@ -149,37 +149,48 @@ func Execute(r *nfdv1alpha1.Rule, features *nfdv1alpha1.Features, failFast bool)
...
+				if failFast {
+					// there's no need to evaluate other matchers in MatchAny
+					break
+				}
 			}
+			matchStatus.MatchAny = append(matchStatus.MatchAny, featureStatus)
 		}
-		if !matched {
-			return false, nil
+		if !isMatch {
```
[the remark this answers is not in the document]
marquiz: Done

mfranczy (approved):
Looks good to me.

marquiz:
Adjusted to correspond to changes in #2068

ArangoGutierrez (approved):
/lgtm

marquiz:
Updated: removed one incorrect error log message

ArangoGutierrez (approved):
/lgtm

PR #2070, opened 2025-03-03
Remove deprecated toleration/affinity value

ArangoGutierrez:
Fix this warning during helm install

```bash
W0303 17:28:58.017492   44880 warnings.go:70] spec.template.spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].preference.matchExpressions[0].key: node-role.kubernetes.io/master is use "node-role.kubernetes.io/control-plane" instead
```

Cause of the Warning
	•	node-role.kubernetes.io/master is deprecated.
	•	Since Kubernetes v1.20, the preferred label is node-role.kubernetes.io/control-plane.
	•	The warning is triggered because node-role.kubernetes.io/master is still used in your chart’s nodeAffinity rules.

### DIFF
--- a/deployment/components/common/master-affinity.yaml
+++ b/deployment/components/common/master-affinity.yaml
@@ -6,23 +6,9 @@
       - weight: 1
         preference:
           matchExpressions:
-          - key: "node-role.kubernetes.io/master"
-            operator: In
-            values: [""]
-      - weight: 1
-        preference:
-          matchExpressions:
           - key: node-role.kubernetes.io/control-plane
             operator: In
             values: [""]
-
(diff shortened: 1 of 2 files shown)


fmuyassarov (commented):
If we do this, I think we should drop https://github.com/kubernetes-sigs/node-feature-discovery/blob/19f6b4f04f5abdc1c7daedaad3981423a6f46c8c/deployment/components/common/master-affinity.yaml#L9 too, WDYT?

fmuyassarov (commented):
/lgtm

PR #2071, opened 2025-03-04
Add new comparison operators

mfranczy:
This PR extends the operators list with the following:
* Le - lower than or equal
* Ge - greater than or equal
* LeGe - range that includes boundary values


marquiz (approved):
Looks good to me
/ok-to-test

ArangoGutierrez (approved):
/lgtm

PR #2072, opened 2025-03-04
Drop gRPC startup probe

fmuyassarov:
Remove gRPC startup probe from master following gRPC [logging removal](https://github.com/kubernetes-sigs/node-feature-discovery/pull/2004).


ArangoGutierrez (approved):
/lgtm

PR #2076, opened 2025-03-07
Control post-delete hook

alekseysen:
There are situations where the labels generated by the NFD need to be preserved even after its removal.

One such use case arises when NFD is installed as part of another project. By design, uninstalling the parent project does not remove all its components. However, the remaining components still rely on the labels created by NFD. If NFD is removed, these labels will be deleted, causing dependent components to fail.


marquiz:
@alekseysen I didn't fully understand the description. How/why does NFD get uninstalled? Is the problem that uninstalling the parent project also removes NFD or is NFD removed manually in your case?

alekseysen:
hi @marquiz,

we've got following use-case: NFD chart is deployed as a sub-chart from our project. Obviously, we've got some workloads (e.g. deployments. daemonsets) which require NFD labels to be scheduled on nodes. These workloads aren't deployed as a part of helm chart.
In a case if our helm chart is uninstalled helm deletes NFD too. This procedure can be done as a part of upgrade/maintenance procedure. 
 
To maintain system integrity and ensure continued functionality, there must be an option that allows any modifications made by NFD to be preserved upon its removal. These changes are essential for system stability, compatibility, and optimized performance, preventing potential disruptions or regressions. Removing these modifications could lead to broken dependencies, security vulnerabilities, or degraded system performance. By retaining these enhancements, organizations can safeguard their infrastructure while ensuring seamless transitions and future adaptability.

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -15,6 +15,8 @@ featureGates:
 
 priorityClassName: ""
 
+cleanupLabels: true
```
marquiz, line 18: nitpick about the naming. It cleans up other stuff too (i.e. annotations, extended resources and taints), so I'd suggest something like `postDeleteCleanup` or smth. WDYT @alekseysen

marquiz (commented):
Thank you @alekseysen for the detailed description. I think this change is good, especially as it's behind an option.

We'd also need to document the new value in `docs/deployment/helm.md`

/ok-to-test

marquiz (approved):
Thanks @alekseysen looks good to me.
/assign @ArangoGutierrez

deployment/helm/node-feature-discovery/templates/post-delete-job.yaml
```diff
@@ -96,3 +97,4 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
+{{- end }}
```
ArangoGutierrez, line 100: New line needed

ArangoGutierrez (changes requested):

marquiz:
@ArangoGutierrez PTAL

fmuyassarov (commented):
/lgtm
/hold 
to give time for @ArangoGutierrez to have a look as well.

marquiz:
@ArangoGutierrez any comments?

alekseysen:
@ArangoGutierrez could you please let me know if there is anything else is missing in this request to be approved?

PR #2083, opened 2025-03-11
Replace golang.org/x/exp/maps with the standard library's maps package.

fmuyassarov:
Replace [golang.org/x/exp/maps](https://pkg.go.dev/golang.org/x/exp/maps) with the standard library's [maps](https://pkg.go.dev/maps) package.


marquiz (approved):
Nice, finally we get this improvement :smile:  Thank you @fmuyassarov

PR #2096, opened 2025-03-20
Add VersionRange operator for NFR/NFG API

mfranczy:
The PR adds a new `VersionRange` operator.
The input must be a semantic version and falls within a range that includes the boundary values. Boundaries can be either a valid semantic version or an empty string (""), which represents an infinite limit.

Example of usage
```
- feature: kernel.version
  matchExpressions:
    full: {op: VersionRange, value: ["5.15", "7.0.0"]}
- feature: kernel.version
  matchExpressions:
    full: {op: VersionRange, value: ["", "7.0.0"]}
- feature: kernel.version
  matchExpressions:
    full: {op: VersionRange, value: ["5.15", "5.25"]}
- feature: kernel.version
  matchExpressions:
    full: {op: VersionRange, value: ["5.15", ""]}
```

### DIFF
--- a/api/nfd/go.mod
+++ b/api/nfd/go.mod
@@ -1,7 +1,6 @@
 module sigs.k8s.io/node-feature-discovery/api/nfd
 
-go 1.22.2
-toolchain go1.23.7
+go 1.24
 
 require (
 	github.com/stretchr/testify v1.8.4
--- a/api/nfd/v1alpha1/types.go
+++ b/api/nfd/v1alpha1/types.go
@@ -360,6 +360,9 @@ const (
 	// MatchIsFalse returns true if the input holds the value "false". The
 	// expression must not have any values.
 	MatchIsFalse MatchOp = "IsFalse"
+	// MatchVersionRange returns true if the input is version that falls into the
+	// specified version range. Both the input and value must be semantic versions.
+	MatchVersionRange MatchOp = "VersionRange"
 )
 
 const (
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -816,6 +816,7 @@ below.
 |  `GeLe`         | 2            | Input falls within a range that includes the boundary values. Both the input and value must be integer numbers. |
 |  `IsTrue`       | 0            | Input is equal to "true" |
 |  `IsFalse`      | 0            | Input is equal "false" |
+|  `VersionRange` | 2            | Input falls within a range that includes the boundary values. Both the input and value must contain semantic versions. |
 
 The `value` field of MatchExpression is a list of string arguments to the
 operator.
--- a/docs/usage/customization-guide.md
+++ b/docs/usage/customization-guide.md
@@ -816,7 +816,7 @@ below.
 |  `GeLe`         | 2            | Input falls within a range that includes the boundary values. Both the input and value must be integer numbers. |
 |  `IsTrue`       | 0            | Input is equal to "true" |
 |  `IsFalse`      | 0            | Input is equal "false" |
-|  `VersionRange` | 2            | Input falls within a range that includes the boundary values. Both the input and value must contain semantic versions. |
+|  `VersionRange` | 2            | The input must be a semantic version and falls within a range that includes the boundary values. Boundaries can be either a valid semantic version or an empty string (""), which represents an infinite limit. |
 
 The `value` field of MatchExpression is a list of string arguments to the
 operator.
--- a/pkg/apis/nfd/nodefeaturerule/expression.go
+++ b/pkg/apis/nfd/nodefeaturerule/expression.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/klog/v2"
 
 	nfdv1alpha1 "sigs.k8s.io/node-feature-discovery/api/nfd/v1alpha1"
+	"sigs.k8s.io/node-feature-discovery/pkg/utils"
 )
 
 const (
@@ -53,6 +54,7 @@ var matchOps = map[nfdv1alpha1.MatchOp]struct{}{
 	nfdv1alpha1.MatchGeLe:         {},
 	nfdv1alpha1.MatchIsTrue:       {},
 	nfdv1alpha1.MatchIsFalse:      {},
+	nfdv1alpha1.MatchVersionRange: {},
 }
 
 // evaluateMatchExpression evaluates the MatchExpression against a single input value.
@@ -166,6 +168,42 @@ func evaluateMatchExpression(m *nfdv1alpha1.MatchExpression, valid bool, value i
 				return false, fmt.Errorf("invalid expression, 'value' field must be empty for Op %q (have %v)", m.Op, m.Value)
 			}
 			return value == "false", nil
+		case nfdv1alpha1.MatchVersionRange:
+			if len(m.Value) != 2 {
+				return false, fmt.Errorf("invalid expression, 'value' field must contain exactly two elements for Op %q (have %v)", m.Op, m.Value)
+			}
+			versions := map[string]string{
+				"min": m.Value[0],
+				"max": m.Value[1],
+			}
+
+			for k, v := range versions {
+				if v == "" {
+					continue
+				}
+				ver, err := utils.ExtractSemVer(v)
+				if err != nil {
+					return false, fmt.Errorf("invalid expression, 'value' field must contain semantic versions or empty string for Op %q (have: %v)", m.Op, m.Value)
+				}
+				versions[k] = ver
+			}
+
+			ver, err := utils.ExtractSemVer(value)
+			if err != nil {
+				return false, fmt.Errorf("invalid expression, the flag/attribute/instance value is not a semantic version %q", value)
+			}
+			versions["current"] = ver
+
+			if versions["min"] != "" && versions["max"] != "" && utils.CompareVersions(versions["min"], versions["max"]) == 1 {
+				return false, fmt.Errorf("invalid expressions, minVersion value %q is greater than maxVersion value %q", versions["min"], versions["max"])
+			}
+			if versions["min"] != "" && utils.CompareVersions(versions["current"], versions["min"]) < 0 {
+				return false, nil
+			}
+			if versions["max"] != "" && utils.CompareVersions(versions["current"], versions["max"]) > 0 {
+				return false, nil
+			}
+			return true, nil
 		default:
 			return false, fmt.Errorf("unsupported Op %q", m.Op)
 		}
--- a/pkg/apis/nfd/nodefeaturerule/expression_test.go
+++ b/pkg/apis/nfd/nodefeaturerule/expression_test.go
@@ -265,6 +265,28 @@ func TestEvaluateMatchExpressionValues(t *testing.T) {
 
 		{name: "44", op: nfdv1alpha1.MatchIsFalse, key: "foo", input: I{"foo": "true"}, result: assert.False, err: assert.Nil},
 		{name: "45", op: nfdv1alpha1.MatchIsFalse, key: "foo", input: I{"foo": "false"}, result: assert.True, err: assert.Nil},
+
+		{name: "46", op: nfdv1alpha1.MatchVersionRange, values: V{"", ""}, key: "foo", input: I{"foo": "3.5.0-flavor"}, result: assert.True, err: assert.Nil},
+		{name: "47", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5.0", "4.0.1"}, key: "foo", input: I{"foo": "3.5.0-flavor"}, result: assert.True, err: assert.Nil},
+		{name: "48", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5.0", "4.0.1"}, key: "foo", input: I{"foo": "3.5-flavor"}, result: assert.True, err: assert.Nil},
+		{name: "49", op: nfdv1alpha1.MatchVersionRange, values: V{"3.0.0", "4.0.1"}, key: "foo", input: I{"foo": "3-flavor"}, result: assert.True, err: assert.Nil},
+		{name: "50", op: nfdv1alpha1.MatchVersionRange, values: V{"3", "4"}, key: "foo", input: I{"foo": "3.5.0"}, result: assert.True, err: assert.Nil},
+		{name: "51", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5.0", "4.0.1"}, key: "foo", input: I{"foo": "3.5.0"}, result: assert.True, err: assert.Nil},
+		{name: "52", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5.0", "4.0.1"}, key: "foo", input: I{"foo": "3.5"}, result: assert.True, err: assert.Nil},
+		{name: "53", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5", "4.0"}, key: "foo", input: I{"foo": "3.5"}, result: assert.True, err: assert.Nil},
+		{name: "54", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5", "4.0"}, key: "foo", input: I{"foo": "3.5.2"}, result: assert.True, err: assert.Nil},
+		{name: "55", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5", ""}, key: "foo", input: I{"foo": "3.6.2"}, result: assert.True, err: assert.Nil},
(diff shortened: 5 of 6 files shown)


mfranczy:
PTAL @ChaoyiHuang 

/cc @marquiz @ArangoGutierrez

api/nfd/go.mod
```diff
@@ -1,7 +1,6 @@
 module sigs.k8s.io/node-feature-discovery/api/nfd
 
-go 1.22.2
-toolchain go1.23.7
+go 1.24
```
marquiz, line 3: Is this necessary bump?

pkg/apis/nfd/nodefeaturerule/expression_test.go
```diff
@@ -265,6 +265,28 @@ func TestEvaluateMatchExpressionValues(t *testing.T) {
 
 		{name: "44", op: nfdv1alpha1.MatchIsFalse, key: "foo", input: I{"foo": "true"}, result: assert.False, err: assert.Nil},
 		{name: "45", op: nfdv1alpha1.MatchIsFalse, key: "foo", input: I{"foo": "false"}, result: assert.True, err: assert.Nil},
+
+		{name: "46", op: nfdv1alpha1.MatchVersionRange, values: V{"", ""}, key: "foo", input: I{"foo": "3.5.0-flavor"}, result: assert.True, err: assert.Nil},
```
marquiz, line 269: Should we test and what happens if both min and max are empty strings (`""`) and the input is not valid semver?

docs/usage/customization-guide.md
```diff
@@ -816,6 +816,7 @@ below.
 |  `GeLe`         | 2            | Input falls within a range that includes the boundary values. Both the input and value must be integer numbers. |
 |  `IsTrue`       | 0            | Input is equal to "true" |
 |  `IsFalse`      | 0            | Input is equal "false" |
+|  `VersionRange` | 2            | The input must be a semantic version and falls within a range that includes the boundary values. Boundaries can be either a valid semantic version or an empty string (""), which represents an infinite limit. |
```
ChaoyiHuang, line 819: if the boundary values must be included, then it's not possible to express range like >= 5.10, < 6.6

pkg/apis/nfd/nodefeaturerule/expression_test.go
```diff
@@ -265,6 +265,28 @@ func TestEvaluateMatchExpressionValues(t *testing.T) {
...
+		{name: "54", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5", "4.0"}, key: "foo", input: I{"foo": "3.5.2"}, result: assert.True, err: assert.Nil},
+		{name: "55", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5", ""}, key: "foo", input: I{"foo": "3.6.2"}, result: assert.True, err: assert.Nil},
+		{name: "56", op: nfdv1alpha1.MatchVersionRange, values: V{"3.6", ""}, key: "foo", input: I{"foo": "3.6.2"}, result: assert.True, err: assert.Nil},
+		{name: "57", op: nfdv1alpha1.MatchVersionRange, values: V{"3.6.2", ""}, key: "foo", input: I{"foo": "3.6.2"}, result: assert.True, err: assert.Nil},
+		{name: "58", op: nfdv1alpha1.MatchVersionRange, values: V{"", "4.5"}, key: "foo", input: I{"foo": "4.4.9"}, result: assert.True, err: assert.Nil},
+		{name: "59", op: nfdv1alpha1.MatchVersionRange, values: V{"", "4.4.9"}, key: "foo", input: I{"foo": "4.4.9"}, result: assert.True, err: assert.Nil},
+		{name: "60", op: nfdv1alpha1.MatchVersionRange, values: V{"invalid", "invalid"}, key: "foo", input: I{"foo": "3.5.0"}, result: assert.False, err: assert.NotNil},
```
ChaoyiHuang, line 283: test cases for min, max semver with diffirent length, and value with variable length too. for example, [5.5, 6.6.3], value 5.6.4, 5.4, 5.5.2, 5.5, 6, 6.6 etc

pkg/apis/nfd/nodefeaturerule/expression_test.go
```diff
@@ -265,6 +265,28 @@ func TestEvaluateMatchExpressionValues(t *testing.T) {
 
 		{name: "44", op: nfdv1alpha1.MatchIsFalse, key: "foo", input: I{"foo": "true"}, result: assert.False, err: assert.Nil},
 		{name: "45", op: nfdv1alpha1.MatchIsFalse, key: "foo", input: I{"foo": "false"}, result: assert.True, err: assert.Nil},
+
+		{name: "46", op: nfdv1alpha1.MatchVersionRange, values: V{"", ""}, key: "foo", input: I{"foo": "3.5.0-flavor"}, result: assert.True, err: assert.Nil},
+		{name: "47", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5.0", "4.0.1"}, key: "foo", input: I{"foo": "3.5.0-flavor"}, result: assert.True, err: assert.Nil},
+		{name: "48", op: nfdv1alpha1.MatchVersionRange, values: V{"3.5.0", "4.0.1"}, key: "foo", input: I{"foo": "3.5-flavor"}, result: assert.True, err: assert.Nil},
```
ChaoyiHuang, line 271: what about the input {"foo": "flavor-3.5"}

docs/usage/customization-guide.md
marquiz: Hmm, yes, good point. I started to wonder if we should extend Gt/Ge/Lt/Le to support semvers, too (in practice it would the cover floating point numbers as well) :thinking:

mfranczy: I like the idea of extending Gt/Ge/Lt/Le, I would add GeLe, GtLt to the list too. 
The fewer operators we introduce, the better.

api/nfd/go.mod
mfranczy: Globally on my machine I have go1.24 installed. That's also the required version by the main go.mod of the project. I could not generate changes I made to NFD API because it complained about outdated toolchain and go version. After running `go mod tidy` the go.mod from NFD API changed to:

```
go 1.23
toolchain go1.24
```

Probably, that's the better idea just to bump to go 1.23 to enforce smaller change to people.
Any thoughts?

mfranczy: Makes perfect sense.

docs/usage/customization-guide.md
mfranczy: I've been prototyping and thinking about this for a while. I’ve come to the conclusion that mixing versions, floating-point, and integer numbers within the current `MatchExpression` structure is problematic.

For example, when using the same comparison operators for both versions and floating-point numbers, we get conflicting results:
* 2.10 < 2.2 → false (as a version)
* 2.10 < 2.2 → true (as a float, since 2.10 is interpreted as 2.1)

I see two possible solutions:

The first one, extend MatchExpression with a `Type` field

```go
    type MatchExpression struct {
        Op MatchOp `json:"op"`
        Value MatchValue `json:"value,omitempty"`
        // Type defines the value type for specific operators.
        // For example, Ge/Gt/Le/Lt could support "number" (including floats) or "version".
        // Some operators, like InRegexp, would always require a string.
        // +optional
        Type string
    }
```
The `Type` field would be tightly coupled with the operator—some would not allow the usage of the field (e.g., InRegexp always requires strings), while others (e.g., Ge) would support multiple types like number or version.

The second approach is to introduce a separate set of operators for version comparisons.

I believe the first approach is more flexible and avoids unnecessary duplication.

@marquiz @ChaoyiHuang, what are your thoughts?

marquiz: > For example, when using the same comparison operators for both versions and floating-point numbers, we get conflicting results:
> 
>     * 2.10 < 2.2 → false (as a version)
>     * 2.10 < 2.2 → true (as a float, since 2.10 is interpreted as 2.1)

Argh, you're right, stupid me :man_facepalming: 

> I see two possible solutions:
> 
> The first one, extend MatchExpression with a `Type` field

Quickly thought, I like this idea. Having it as optional shouldn't break any existing users. For now, it would be only supported for the comparison operators (similar to `value` field is not accepted for IsTrue and folks).

I wonder should we make a separation between integers and floating point numbers :thinking: Currently require integer numbers and extending to floats is not exactly the same but otoh I have hard time imagining that this extension would actually break anyone. What I'm after here is just the naming of the types, `Number` vs. `Integer` vs. `Float`, not that this PR would be required to add float support.

Probably need to sleep over this and try to think about the possible problems and caveats.

> The second approach is to introduce a separate set of operators for version comparisons.
> 
> I believe the first approach is more flexible and avoids unnecessary duplication.

Yes, I'm inclined in that direction that the first option looks better.

pkg/apis/nfd/nodefeaturerule/expression_test.go
mfranczy: I will add tests for that in separate test suite for `ExtractVersion` func.

pkg/apis/nfd/nodefeaturerule/expression_test.go
mfranczy: I will add tests for that in separate test suite for `CompareVersions` func.

docs/usage/customization-guide.md
mfranczy: > I wonder should we make a separation between integers and floating point numbers 🤔 Currently require integer numbers and extending to floats is not exactly the same but otoh I have hard time imagining that this extension would actually break anyone. What I'm after here is just the naming of the types, Number vs. Integer vs. Float, not that this PR would be required to add float support.

From a user perspective, I think we should not, it would be more convinient to support the type `Number` which includes integer and float. The implementation would have to take care of type resolution. I belive it is better just to put the type `Number`, rather than explicitly `Integer` or `Float` in the spec.

marquiz: Having slept over this, I agree

pkg/apis/nfd/nodefeaturerule/expression_test.go
mfranczy: Done.

pkg/apis/nfd/nodefeaturerule/expression_test.go
mfranczy: Done.

mfranczy:
@marquiz @ChaoyiHuang I updated the PR, it's good enough for another review round.

btw @marquiz, i don't see prow has configured a job for https://github.com/kubernetes-sigs/node-feature-discovery/tree/master/api/nfd, it looks like over time one test started reporting errors:

```
--- FAIL: TestInstanceFeatureSet (0.00s)                                                                                                                                                                                                      
    feature_test.go:90:                                                                                                
                Error Trace:    /home/nmm/Workspace/node-feature-discovery/api/nfd/v1alpha1/feature_test.go:90                                                                                                                                
                Error:          Not equal:                                                                             
                                expected: []v1alpha1.InstanceFeature{}                                                                                                                                                                        
                                actual  : []v1alpha1.InstanceFeature(nil)                                              
                                                                                                                                                                                                                                              
                                Diff:                                                                                  
                                --- Expected                                                                                                                                                                                                  
                                +++ Actual                                                                             
                                @@ -1,3 +1,2 @@                                                                                                                                                                                               
                                -([]v1alpha1.InstanceFeature) {       
                                -}                                                                                                                                                                                                            
                                +([]v1alpha1.InstanceFeature) <nil>   
                                                                                                                                                                                                                                              
                Test:           TestInstanceFeatureSet                                                                 
FAIL                                                                                                                                                                                                                                          
FAIL    sigs.k8s.io/node-feature-discovery/api/nfd/v1alpha1     0.005s
FAIL
```

marquiz:
> it looks like over time one test started reporting errors

:man_facepalming: Thanks for the heads-up @mfranczy, see #2107

api/nfd/v1alpha1/utils.go
marquiz, line 1: Hmm, started to think could we use some existing semver library, e.g.?

- https://github.com/blang/semver
- https://github.com/Masterminds/semver

api/nfd/v1alpha1/types.go
```diff
@@ -296,16 +296,44 @@ type MatchExpression struct {
 	// In other cases Value should contain at least one element.
 	// +optional
 	Value MatchValue `json:"value,omitempty"`
+
+	// Type defines the value type for specific operators.
+	// The currently supported type is 'version' for Gt,Ge,Lt,Le,GtLt,GeLe operators.
```
ChaoyiHuang, line 301: version or semver

api/nfd/v1alpha1/types.go
```diff
@@ -360,6 +388,9 @@ const (
 	// MatchIsFalse returns true if the input holds the value "false". The
 	// expression must not have any values.
 	MatchIsFalse MatchOp = "IsFalse"
+	// MatchVersionRange returns true if the input is version that falls into the
+	// specified version range. Both the input and value must be semantic versions.
```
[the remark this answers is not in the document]
mfranczy: That's leftover from the previous commit. I am aware that kernel doesn't follow semver. I forgot to remove this.

api/nfd/v1alpha1/utils.go
mfranczy: As @ChaoyiHuang already mentioned, not everyone follows semver (e.g., kernel versioning upstream/downstream). I implemented the common aspects that align with semver.

marquiz: Hmm, but isn't what we support here a subset of semver?

I.e. 1, 1.2 and 1.2.3 are all valid semvers -> we could use semver library

mfranczy: > K. Should we have semver as a separate type? Possible future extension, could be just a TODO note. WDYT?

That's a good question, probably we could add this as a separate type, or give some more thoughts how to do it. For now I will place a TODO note.

marquiz, line 1: I think this stuff belongs to `pkg/apis/nfd/nodefeaturerule/` instead (everything else except for addition of the `Type` field)

pkg/apis/nfd/nodefeaturerule/expression.go
```diff
@@ -122,40 +122,94 @@ func evaluateMatchExpression(m *nfdv1alpha1.MatchExpression, valid bool, value i
...
-
-			l, err := strconv.Atoi(value)
-			if err != nil {
-				return false, fmt.Errorf("not a number %q", value)
-			}
-			r, err := strconv.Atoi(m.Value[0])
-			if err != nil {
-				return false, fmt.Errorf("not a number %q in %v", m.Value[0], m)
+			if m.Type != "" && m.Type != nfdv1alpha1.TypeVersion {
+				return false, fmt.Errorf("invalid expression, 'type' field only accepts %q value for Op %q (have %q)", nfdv1alpha1.TypeVersion, m.Op, m.Type)
```
marquiz, line 126: I'd suggest to add this one as the default: case below

pkg/apis/nfd/nodefeaturerule/expression.go
```diff
@@ -122,40 +122,94 @@ func evaluateMatchExpression(m *nfdv1alpha1.MatchExpression, valid bool, value i
...
+				if (m.Op == nfdv1alpha1.MatchGtLt && lRange == nfdv1alpha1.CmpGt && rRange == nfdv1alpha1.CmpLt) ||
+					(m.Op == nfdv1alpha1.MatchGeLe &&
+						(lRange == nfdv1alpha1.CmpGt || lRange == nfdv1alpha1.CmpEq) &&
+						(rRange == nfdv1alpha1.CmpLt || rRange == nfdv1alpha1.CmpEq)) {
+					return true, nil
+				}
+
+				return false, nil
+
+			default:
```
marquiz, line 195:
```suggestion
			"":
```

and return a invalid type error as the default: case

api/nfd/v1alpha1/utils.go
mfranczy: Fixed.

pkg/apis/nfd/nodefeaturerule/expression.go
mfranczy: Fixed.

pkg/apis/nfd/nodefeaturerule/expression.go
mfranczy: Fixed.

marquiz (approved):
Nice job @mfranczy. I would be ready to merge this one

marquiz:
/assign @ArangoGutierrez 
@ChaoyiHuang PTAL

PR #2097, opened 2025-03-20
Revert "nfd-master: use only unbuffered chans in the nfd api-controller"

marquiz:
Fixes a bug where updates to NodeFeatureGroup objects were not happening after the initial sync.

We need to use buffered channels for the "update all" channels as there is the default: case in select that is sending the update all signal (to the channel). An alternative would be to remove the default: case from the select block but it is more efficient this way - just skip if the reader side hasn't yet got the "update all" signal.

This PR also adds new e2e-tests for NodeFeatureGroup to cover this case (plus test the recently added templating and backrefs).


ArangoGutierrez (approved):
/lgtm

PR #2104, opened 2025-03-25
Allow post-delete job to be disabled

alekseysen:
There are situations where the modifications performed by the NFD need to be preserved even after its removal.

One such use case arises when NFD is installed as part of another project. By design, uninstalling the parent project does not remove all its components. However, the remaining components still rely on the labels created by NFD. If NFD is removed, these labels will be deleted, causing dependent components to fail.

To maintain system integrity and ensure continued functionality, adding the 'postDeleteCleanup' option that allows any modifications made by NFD to be preserved upon its removal.


alekseysen:
Hi @marquiz ,
I did something wrong with original PR #2076 . Could you review a new one, please?

Changes:
renamed the variable to postDeleteCleanup
Added a documentation

ozhuraki (approved):

PR #2121, opened 2025-04-09
Add detection of libc implementation and its version

mfranczy:
The PR is still in progress, I am still trying to find more robust approach for this.


mfranczy:
Hm, maybe better approach will be to use CGO and execute `gnu_get_libc_version` and something for musl...

mfranczy:
After intensive research, I didn't find a better way than just parsing the libc.so binaries to find out about the version and type. I will finish implementation with this design.

mfranczy:
/cc @marquiz @ArangoGutierrez 

PTAL @ChaoyiHuang 

This design represents the best solution we can achieve by directly reading ELF/binary files. While it's not perfect, I believe it's sufficient as a temporary measure. To accurately determine the version, I think we should execute the ldd from host to retrieve its version output.

The challenge lies in how host directories are mounted into containers. By default, each directory is prefixed with 'host'. In a separate discussion, I'd like to explore the idea of using a common parent directory for all host directories. This would allow us to chroot and execute host binaries within the container (like ldd).

Anyway, if anybody has a better idea, please share it.

source/system/testdata/libc/glibc
mfranczy, line 1: The glibc, musl, and unknown libc binaries are sourced from https://github.com/mfranczy/libc-fake. I created these for unit tests, and by embedding the libc name and version into the .rodata section, I can precisely validate the solution. The builds are neither stripped nor optimized, making them easy to disassemble and inspect the contents, if required.

source/system/libc.go
```diff
@@ -0,0 +1,112 @@
...
+		// External solution has to be applied, like executing ldd directly.
+		return muslType, "", nil
+	default:
+		return unknownType, "", nil
+	}
+}
+
+func extractGlibcVersion(data []byte) string {
+	lines := bytes.SplitSeq(data, []byte("\n"))
+	for line := range lines {
```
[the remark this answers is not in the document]
mfranczy: Wrong. It's not an index. It's the byte type.

source/system/system.go
```diff
@@ -118,6 +119,12 @@ func (s *systemSource) Discover() error {
 		s.features.Attributes[DmiIdFeature] = nfdv1alpha1.NewAttributeFeatures(dmiAttrs)
 	}
 
+	libcAttrs, err := getLibcAttributes()
+	if err != nil {
+		klog.Error(err, "failed to detect libc version")
```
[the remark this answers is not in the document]
mfranczy: Fixed.

source/system/libc_linux_amd64.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"sigs.k8s.io/node-feature-discovery/pkg/utils/hostpath"
+)
+
+func getCommonLibcPaths() []string {
+	return []string{
+		// glibc
+		hostpath.UsrDir.Path("lib64/libc.so.6"),
+		hostpath.LibDir.Path("x86_64-linux-gnu/libc.so.6"),
```
ChaoyiHuang, line 27: should ARM, 32 bit CPU also be included in detection

source/system/libc_linux_amd64.go
```diff
@@ -0,0 +1,33 @@
...
+limitations under the License.
+*/
+
+package system
+
+import (
+	"sigs.k8s.io/node-feature-discovery/pkg/utils/hostpath"
+)
+
+func getCommonLibcPaths() []string {
```
ChaoyiHuang, line 23: this function will be built in any cpu and os, so it'll produce conflict with libc_stub.go getCommonLibcPaths for !linux && !amd64

mfranczy: The name of the file contains os and cpu tags `.._linux_amd64`, it won't be duplicated with the stub file.

source/system/libc_linux_amd64.go
mfranczy: Before expanding to different architectures, I think it's better to discuss the idea I already shared, which is: change how NFD volumes are mounted into containers and execute the `ldd` binary from the host.

marquiz: Escaping the container this way starts to sound very dubious to me :monocle_face: How about having (an optional) initcontainer/sidecar for doing the scary stuff? Or (for glibc at least) mount and grep the `ldd` script for the version (possible problems: ldd does not exists, ldd/binutils is different version than libc)?

Thoughts?

source/system/libc_linux_amd64.go
ChaoyiHuang: oh, I just forget to check the file name

source/system/libc_linux_amd64.go
mfranczy: > Escaping the container this way starts to sound very dubious to me 🧐

You're right, it's definitely risky. Executing chroot is far from ideal and would require the container to run with the SYS_CHROOT capability. That opens up a range of issues, from security concerns to violating the principle of container isolation.

> Or (for glibc at least) mount and grep the ldd script for the version (possible problems: ldd does not exist, or ldd/binutils is a different version than libc)?

Do you mean mounting and then executing `ldd --version` to grep the output? Or literally grepping the ldd binary/script to extract the libc version?

* If you meant the first option:
That won't work reliably. Running ldd without isolating the environment (e.g., via chroot) risks linking against the container's own shared libraries instead of the host;s, especially if paths overlap. To get accurate results, I'd need to copy the host's shared libraries and use chroot or a similar mechanism to recreate the host's runtime environment. Without this, ldd might resolve the wrong libraries and produce misleading output.

* If you meant the second option:
That also wouldn't help, ldd itself doesn't contain the libc version string. It calls the `gnu_get_libc_version` function at runtime, which is resolved through dynamic linking. Btw. this PR is using a similar approach to what you suggest: scanning .rodata sections of shared libraries to locate the version string, since it's embedded in the binaries (but only for glibc - not for musl).

@marquiz The sidecar approach could work, but it won't be a native or guaranteed discovery. On Kubernetes, we could handle this using an init container. But if someone runs the CLI directly on the host, they’d need to provide the glibc version manually, perhaps via an environment variable.

@ChaoyiHuang would the last approach work for you?

ChaoyiHuang: providing the glibc version manually doesn't work well in many cases.

marquiz: > * If you meant the second option:

Yeah, this was what I was thinking about. Looking at ldd on my Fedora 42 I see

```
$ cat /usr/bin/ldd
...
while test $# -gt 0; do
  case "$1" in
  --vers | --versi | --versio | --version)
    echo 'ldd (GNU libc) 2.41'
...
```

So it's hardcoded in the script and doesn't look at the libc library at all :thinking:

mfranczy: @marquiz damn, you're right. I am stupid, I was debugging the invoked linker executed by ldd, not the ldd script itself. Then we have a winner, for musl it also shows the version in the script. For the possible problems, if they exist then we will just report uknown for libc type and its version, does it sound good?

ChaoyiHuang: it seems using "strings /usr/lib/x86_64-linux-gnu/libc.so.6 | grep GLIBC" is more easier, and no need to mount /user/bin/ldd

marquiz: > musl it also shows the version in the script.

Just made another quick check with alpine docker image with no luck :/

```
$ docker run --rm -it alpine cat /usr/bin/ldd
#!/bin/sh
exec /lib/ld-musl-x86_64.so.1 --list "$@"
$
```

marquiz: > it seems using "strings /usr/lib/x86_64-linux-gnu/libc.so.6 | grep GLIBC" is more easier, and no need to mount /user/bin/ldd

In the NFD image we cannot run any scripts or executables, we're essentially a scratch image with just the (statically linked) nfd binaries

mfranczy: > > it seems using "strings /usr/lib/x86_64-linux-gnu/libc.so.6 | grep GLIBC" is more easier, and no need to mount /user/bin/ldd
> 
> In the NFD image we cannot run any scripts or executables, we're essentially a scratch image with just the (statically linked) nfd binaries

Currently this PR does something similar as @ChaoyiHuang suggested, although it doesn't execute anything. It just reads .rodata of ELF.

jjacobelli:
Quick question unrelated to the current discussions, but what's the usage of knowing the host libc version?

marquiz:
> Quick question unrelated to the current discussions, but what's the usage of knowing the host libc version?

A good question. So @mfranczy are you running some host binaries inside the container or planting some binaries from the container into the host? Or possibly both? Or something else?

ChaoyiHuang:
> A good question. So @mfranczy are you running some host binaries inside the container or planting some binaries from the container into the host? Or possibly both? Or something else?

considering the case, a new libc interface or feature in the newer libc was used by the app in the container, but this interface or feature has not been supported for the older version libc in the host. no matter the newer glibc was built into the (base) image or not, if the host libc does not provide this interface or feature, then failure happen. sometimes there is part overlap for kernel compatibility, sometimes separately.

that's the links discussed at the begining of the image compatibility working group initiation

marquiz:
> considering the case, a new libc interface or feature in the newer libc was used by the app in the container, but this interface or feature has not been supported for the older version libc in the host.

I get this. But in normal situations you're not dependent on the host libc. The workload is containerized, it uses whatever libc is in the container (base) image, right(?) I was trying to understand the usage scenarios where the workload is dependent on the host libc.

ChaoyiHuang:
> > considering the case, a new libc interface or feature in the newer libc was used by the app in the container, but this interface or feature has not been supported for the older version libc in the host.
> 
> I get this. But in normal situations you're not dependent on the host libc. The workload is containerized, it uses whatever libc is in the container (base) image, right(?) I was trying to understand the usage scenarios where the workload is dependent on the host libc.

HPC community mentioned "some host pieces like drivers and some libraries must be injected at runtime.  We have encountered this in HPC use cases with GPU and interconnect support." also MPI libaries may be injected from host. It would help that HPC community can describe in further detail for real cases.

marquiz:
> Legacy stuff, like early device plugins calling host binaries, after talking this through internally. I am closing the PR.

Good resolution for now, if you ask me :sweat_smile:

PR #2129, opened 2025-04-22
nfd-master: clean up api controller

marquiz:
Drop effectively unused parameter. The NodeFeatureAPI feature gate has been locked to true.


fmuyassarov (commented):
/lgtm

PR #2130, opened 2025-04-22
nfd-master: cleanup nil checks

marquiz:
The nfdController must never be nil now that the gRPC API has been removed.


fmuyassarov (commented):
/lgtm

Issue #2063, opened 2025-02-26
Serious memory leak in non-leader nfd-master pod


age9990:
**What happened**:
After upgrading from v0.14.4 to v0.16.6, memory usage of non-leader nfd-master pods gradually increased, eventually OOM

**What you expected to happen**:
Like in v0.14.4, memory usage of all nfd-master pods are stable

**How to reproduce it (as minimally and precisely as possible)**:
Install gpu-operator v24.9.2 with default configuration, the memory leak occurred

**Anything else we need to know?**:
We set replicas=3 for nfd-master, leader nfd-master pod did not suffer from memory leak, the other two non-leader both suffer from memory leak. If I delete the leader pod to force a leader re-election, the memory usage of newly selected leader would gradually decreased to normal level and remain stable.

**Environment**:
- Kubernetes version (use `kubectl version`): K8S 1.24
- Cloud provider or hardware configuration: Bare Metal
- OS (e.g: `cat /etc/os-release`): Ubuntu 22.04
- Kernel (e.g. `uname -a`): 6.2.0-39-generic
- Install tools: Helm
- Network plugin and version (if this is a network-related bug):
- Others:

cbrherms:
Seeing the same issue here with a leak in non-leader until oom.

marquiz:
Thank you for reporting the issue @age9990. It hasn't got the attention it deserved. I believe #2131 fixes the issue. Will be backported to v0.16 and v0.17.

marquiz:
@age9990 @cbrherms v0.16.8 and v0.17.3 are now out, hopefully fixing the issue:
https://github.com/kubernetes-sigs/node-feature-discovery/releases/tag/v0.16.8
https://github.com/kubernetes-sigs/node-feature-discovery/releases/tag/v0.17.3

PR #2131, opened 2025-04-22
nfd-master: fix memory leak when leader election is enabled

marquiz:
Fix a serious memory leak of non-leaders that was caused by bad chan usage. With leader election enabled, the nfdAPIUpdateHandler() was not started for non-leader instances, and thus, there was no reader for the chans that the nfdController uses to communicate what objects to update. This, in turn caused blocking in the kubernetes API informer context (trying to queue data into the chan), piling up requests on each NodeFeature update, consuming more and more memory (that would not be released unless we became the leader).

Fixes: #2063

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -149,6 +149,9 @@ type nfdMaster struct {
 	updaterPool    *updaterPool
 	deniedNs
 	config *NFDConfig
+
+	// isLeader indicates if this instance is the leader, changing dynamically
+	isLeader bool
 }
 
 // NewNfdMaster creates a new NfdMaster server instance.
@@ -305,10 +308,11 @@ func (m *nfdMaster) Run() error {
 
 	// Run updater that handles events from the nfd CRD API.
 	if m.args.EnableLeaderElection {
-		go m.nfdAPIUpdateHandlerWithLeaderElection()
+		go m.startLeaderElectionHandler()
 	} else {
-		go m.nfdAPIUpdateHandler()
+		m.isLeader = true
 	}
+	go m.nfdAPIUpdateHandler()
 
 	// Register health probe (at this point we're "ready and live")
 	httpMux.HandleFunc("/healthz", m.Healthz)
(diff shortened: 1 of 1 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -1310,11 +1319,15 @@ func (m *nfdMaster) nfdAPIUpdateHandlerWithLeaderElection() {
 		RenewDeadline: m.config.LeaderElection.RenewDeadline.Duration,
 		Callbacks: leaderelection.LeaderCallbacks{
 			OnStartedLeading: func(_ context.Context) {
-				m.nfdAPIUpdateHandler()
+				m.isLeader = true
 			},
 			OnStoppedLeading: func() {
 				// We lost the lock.
 				klog.InfoS("leaderelection lock was lost")
+				// Exit, makes sure that in-flight requests/re-tries will be
```
ArangoGutierrez, line 1327: Shouldn't we `isLeader = false` when stopping being the leader?

marquiz:
Still testing
/hold

pkg/nfd-master/nfd-master.go
marquiz: I thought about adding it, but put it in the TODO comment instead. We do stop() and thus exit the process so isLeader=false would be kinda nop. I can add it if you think it's better that way.

marquiz:
This patch is probably the simplest thing to do. Alternatives could be e.g.:

- only start the api controller when we get the lead
  - pros: significantly less memory usage and load on apiserver when non-leader
  - cons: slow hand-over as it will not start syncing informer caches until being the leader
- do `!isLead` filtering on the api controller side (don't try to push stuff to chans when no leader).
  - pros: not many, maybe slightly nicer in the sense that no data is passed through the chans
  - cons: more error prone (subject to similar errors in the future), multiple points to inject the filtering

pkg/nfd-master/nfd-master.go
ArangoGutierrez: I think is a good first approach, I am good to lgtm

marquiz:
This should fix the most pressing issue. Follow-up PRs can improve if people are willing to work on this.
/unhold

PR #2135, opened 2025-04-23
[release-0.17] nfd-master: fix memory leak when leader election is enabled

marquiz:
Fix a serious memory leak of non-leaders that was caused by bad chan usage. With leader election enabled, the nfdAPIUpdateHandler() was not started for non-leader instances, and thus, there was no reader for the chans that the nfdController uses to communicate what objects to update. This, in turn caused blocking in the kubernetes API informer context (trying to queue data into the chan), piling up requests on each NodeFeature update, consuming more and more memory (that would not be released unless we became the leader).

(cherry picked from commit a595439063c97fd01f7c57e9e49f2781904c8af8)

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -159,6 +159,9 @@ type nfdMaster struct {
 	updaterPool    *updaterPool
 	deniedNs
 	config *NFDConfig
+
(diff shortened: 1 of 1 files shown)


ArangoGutierrez (approved):
/lgtm

PR #2146, opened 2025-04-30
helm: Inherit resource limits for prune job

jcpunk:
This adds in some resource limits so the prune job can't run away with the system.


marquiz (approved):
Thanks @jcpunk 

/ok-to-test
/assign @fmuyassarov

PR #2148, opened 2025-05-02
helm: Add optional PodDisruptionBudget

jcpunk:
This permits setting the podDisruptionBudget if desired.


deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -126,6 +126,11 @@ master:
     value: ""
     effect: "NoSchedule"
 
+  podDisruptionBudget:
+    enabled: false
```
marquiz, line 130: For consistency, I'd suggest s/enabled/enable/ as all the other similar Helm values are named `enable`, like `gc.enable`.

```suggestion
    enable: false
```

marquiz (commented):
Thank you @jcpunk for the enhancement. I had one comment regarding the naming (below). Also, we'd need to document the new values in the tables in `docs/deployment/helm.md`

deployment/helm/node-feature-discovery/values.yaml
jcpunk: Fixed.

marquiz (approved):
Looks good to me
/ok-to-test
/assign @ArangoGutierrez @fmuyassarov

PR #2151, opened 2025-05-08
scripts: Add helm-push to push-image.sh script

ozhuraki:
@marquiz

I am not quite sure if this is enough, would it be possible to smoke test on your side?

### DIFF
--- a/scripts/test-infra/push-image.sh
+++ b/scripts/test-infra/push-image.sh
@@ -10,3 +10,7 @@ gcloud auth configure-docker
 
 # Build and push images
 IMAGE_ALL_PLATFORMS=linux/amd64,linux/arm64 make push-all $VERSION_OVERRIDE
+
+go install helm.sh/helm/v3/cmd/helm@v3.17.3
+
+make helm-push $VERSION_OVERRIDE


marquiz (commented):
@ozhuraki thanks. Please try to run this with the cloudbuild image we are using (see the `cloudbuild.yaml`). One gap I suspect is that Helm is not present in the image -> will fail.


/ok-to-test

ozhuraki:
@marquiz

> One gap I suspect is that Helm is not present in the image

Thanks, checked. Yes, helm is missing in that image.

What is the way to add it there?

marquiz:
> What is the way to add it there?

I think we could add smth like

```bash
go install helm.sh/helm/v3/cmd/helm@${HELM_VERSION}
```

into the script. We could default HELM_VERSION to the latest release now available. Or alternatively just not use a variable but hard code the version in the script.

marquiz (approved):
Only one way to find out ;)
/lgtm

ozhuraki:
@marquiz

Thanks, added helm into image. Sorry for the late response, was still checking locally.

Thanks a lot for your help!

Issue #2075, opened 2025-03-07
Watch features.d changes


tkatila:
**What would you like to be added**:

Changes in the label files in the `features.d` directory are now propagated every `sleep-interval`. I'd like to have it so that NFD worker watches the `features.d` directory and runs the labeling procedure on every change.

**Why is this needed**:

We have a use case where a workload requires additional node configuration on top of an extended resource (device plugin). The workload is structured so that a container requests an extended resource and is `nodeSelector` limited to nodes with a special "configuration complete" label. With the default 60sec sleep duration, there is an additional 1-59 sec delay before the workload is scheduled to a node. For the labels to appear quicker to nodes would make them more convenient to control scheduling.

marquiz:
Thanks @tkatila for the proposal. I think this makes sense. It could also make it sligthly less racy if we rely on fsnotify events instead of a fixed interval there.

PR #2156, opened 2025-05-12
nfd-worker: Watch features.d changes

ozhuraki:
Closes: #2075

### DIFF
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -30,6 +30,7 @@ import (
 
 	"maps"
 
+	"github.com/fsnotify/fsnotify"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"golang.org/x/net/context"
@@ -121,6 +122,8 @@ type nfdWorker struct {
 	k8sClient           k8sclient.Interface
 	nfdClient           nfdclient.Interface
 	stop                chan struct{} // channel for signaling stop
+	fsEvent             chan fsnotify.Event
+	fsWatcher           *fsnotify.Watcher
 	featureSources      []source.FeatureSource
 	labelSources        []source.LabelSource
 	ownerReference      []metav1.OwnerReference
@@ -304,6 +307,29 @@ func (w *nfdWorker) Run() error {
 	labelTrigger.Reset(w.config.Core.SleepInterval.Duration)
 	defer labelTrigger.Stop()
 
+	featureFilesDir := "/etc/kubernetes/node-feature-discovery/features.d/"
+
+	info, err := os.Stat(featureFilesDir)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
+	}
+
+	if info != nil && info.IsDir() {
+		watcher, err := fsnotify.NewWatcher()
+		if err != nil {
+			return err
+		}
+		defer watcher.Close()
+
+		err = watcher.Add(featureFilesDir)
+		if err != nil {
+			return fmt.Errorf("unable to access %v: %w", featureFilesDir, err)
+		}
+		w.fsWatcher = watcher
+	}
+
 	httpMux := http.NewServeMux()
 
 	// Register to metrics server
@@ -341,6 +367,13 @@ func (w *nfdWorker) Run() error {
 				return err
 			}
 
+		case event := <-w.fsWatcher.Events:
+			if event.Op&fsnotify.Create == fsnotify.Create || event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Remove == fsnotify.Remove || event.Op&fsnotify.Rename == fsnotify.Rename || event.Op&fsnotify.Chmod == fsnotify.Chmod {
+				err = w.runFeatureDiscovery()
+				if err != nil {
+					return err
+				}
+			}
 		case <-w.stop:
 			klog.InfoS("shutting down nfd-worker")
 			return nil


marquiz (commented):
Thanks @ozhuraki for taking a stab at this.

I think we should refactor the code and re-architecture this more to make the code more maintainable. There might be other sources we'd also make react to events in a similar way. Basically, it should be the source (`source/local` in this case) which should be able to notify the main event loop that features have been updated. Also, no need to run re-discovery of all features.

ozhuraki:
@marquiz

Thanks, makes sense. I will move this into source/local.

ozhuraki:
@ArangoGutierrez

Thanks, updated, please take a look.

source/local/local.go
```diff
@@ -318,6 +321,47 @@ func getFileContent(fileName string) ([][]byte, error) {
 	return lines, nil
 }
 
+func (s *localSource) runNotifier(ch chan struct{}) {
+	for {
+		select {
+		case event := <-s.fsWatcher.Events:
+			if event.Op&fsnotify.Create == fsnotify.Create || event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Remove == fsnotify.Remove || event.Op&fsnotify.Rename == fsnotify.Rename || event.Op&fsnotify.Chmod == fsnotify.Chmod {
```
marquiz: Small suggestion, could be more readable with something like:

```golang
			opAny := fsnotify.Create | fsnotify.Write | fsnotify.Remove | fsnotify.Rename | fsnotify.Chmod
			
			if event.Op&opAny != 0 {
```

WDYT?

source/local/local.go
```diff
@@ -318,6 +321,47 @@ func getFileContent(fileName string) ([][]byte, error) {
...
+				klog.InfoS("fsnotify event", event)
```
marquiz: I'd suggest not to log these as "debug", (to avoid potentially flooding the logs). Also the usage of klog.InfoS is not correct.


```suggestion
				klog.V(2).Infos("fsnotify event", "eventName", event.Name, "eventOp", event.Op)
```

source/local/local.go
```diff
@@ -318,6 +321,47 @@ func getFileContent(fileName string) ([][]byte, error) {
...
+	}
+}
+
+// SetChannel method of the EventSource Interface
+func (s *localSource) SetChannel(ch chan struct{}) error {
+	info, err := os.Stat(featureFilesDir)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
```
marquiz: If it does not exist we probably want to exit, too (return nil)?

source/local/local.go
```diff
@@ -318,6 +321,47 @@ func getFileContent(fileName string) ([][]byte, error) {
 	return lines, nil
 }
 
+func (s *localSource) runNotifier(ch chan struct{}) {
+	for {
+		select {
+		case event := <-s.fsWatcher.Events:
+			if event.Op&fsnotify.Create == fsnotify.Create || event.Op&fsnotify.Write == fsnotify.Write || event.Op&fsnotify.Remove == fsnotify.Remove || event.Op&fsnotify.Rename == fsnotify.Rename || event.Op&fsnotify.Chmod == fsnotify.Chmod {
+				klog.InfoS("fsnotify event", event)
+				ch <- struct{}{}
```
marquiz: Looking at this I think we need some filtering of the events. For example, using temporary files in the dir (as recommended by our documentation, to avoid races) we will get a ton of events, each of them causing rediscovery. E.g. wait for one second (to get the slew of fsnotify events) and then send the feature event.

source/local/local.go
```diff
@@ -318,6 +321,47 @@ func getFileContent(fileName string) ([][]byte, error) {
...
+// SetChannel method of the EventSource Interface
+func (s *localSource) SetChannel(ch chan struct{}) error {
+	info, err := os.Stat(featureFilesDir)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
+	}
+
+	if info != nil && info.IsDir() {
```
marquiz: nit: maybe we should also check if `s.fsWatcher` is not nil? Kinda theoretical but thinking if this would be called multiple times

source/source.go
```diff
@@ -77,6 +77,14 @@ type SupplementalSource interface {
 	DisableByDefault() bool
 }
 
+// EventSource is an interface for a source that can send events
+type EventSource interface {
+	Source
+
+	// SetChannel sets the channel
+	SetChannel(chan struct{}) error
```
marquiz: nit/suggestion: maybe indicate in the name what channel we're setting. E.g. `SetNotifyChannel` or smth, maybe you can come up with a better name. And, maybe it should be **Add**SmthChannel calling the method multiple times will not unwire the old notifier channels.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -341,6 +348,12 @@ func (w *nfdWorker) Run() error {
 				return err
 			}
 
+		case <-w.sourceEvent:
+			err = w.runFeatureDiscovery()
```
marquiz: We don't want to run full feature discovery of all sources. Just the one that evented and then update the NodeFeature object. Prolly requires a bit of refactoring on the nfd-worker.go side.

Either do s.Discover() here or then in the source (and then just notify nfd-worker that it needs to re-advertise features).

marquiz:
ping @ozhuraki any update on this?

ozhuraki:
@marquiz @ArangoGutierrez

Thanks for the review. Updated, please take a look.

source/local/local.go
```diff
@@ -318,6 +321,49 @@ func getFileContent(fileName string) ([][]byte, error) {
...
+		case event := <-s.fsWatcher.Events:
+			opAny := fsnotify.Create | fsnotify.Write | fsnotify.Remove | fsnotify.Rename | fsnotify.Chmod
+			if event.Op&opAny != 0 {
+				klog.V(2).InfoS("fsnotify event", "eventName", event.Name, "eventOp", event.Op)
+				ch <- s.Name()
+			}
+		case err := <-s.fsWatcher.Errors:
+			klog.ErrorS(err, "failed to to watch features.d changes")
+		}
+		time.Sleep(1 * time.Second)
```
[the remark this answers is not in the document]
marquiz: @ozhuraki I think copilot is up to something here :) If we get a burst of say 10 events, we'd read them one-per-second, causing a streak of 10 1-per-second updates in nfd-worker. We'd only want a 1 second total delay, i.e. read events as fast as we can but "group" them into one event that gets sent to the ch.

Elsewhere we've used e.g. a patter of `rateLimit := time.After(time.Second)...`, you could come up with something better, thoug...

pkg/nfd-worker/nfd-worker.go
```diff
@@ -341,6 +381,12 @@ func (w *nfdWorker) Run() error {
 				return err
 			}
 
+		case sourceName := <-w.sourceEvent:
+			err = w.runFeatureDiscoveryBySourceName(sourceName)
```
marquiz: I think we could simplify this quite bit by having sourceEvent of type `chan FeatureSource`. We could just do `s.Discover()` here and ditch runFeatureDiscoveryBySourceName altogether. Need to split `runFeatureDiscovery()` into two parts to avoid copy-pasting code: feature-discovery and feature-advertisement (create labels etc). Then call the feature-advertisement part here.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -248,6 +249,36 @@ func (w *nfdWorker) runFeatureDiscovery() error {
...
+			if err := s.Discover(); err != nil {
+				klog.ErrorS(err, "feature discovery failed", "source", s.Name())
+			}
+			klog.V(3).InfoS("feature discovery completed", "featureSource", s.Name(), "duration", time.Since(currentSourceStart))
+		}
+	}
+
+	discoveryDuration := time.Since(discoveryStart)
+	klog.V(2).InfoS("feature discovery of all sources completed", "duration", discoveryDuration)
+	featureDiscoveryDuration.WithLabelValues(utils.NodeName()).Observe(discoveryDuration.Seconds())
```
marquiz, line 251: We shouldn't update the metrics here as the metric is for "all enabled feature sources" and we're only doing one here. We can think about per-source metrics in a separate PR

pkg/nfd-worker/nfd-worker.go
```diff
@@ -304,6 +335,15 @@ func (w *nfdWorker) Run() error {
 	labelTrigger.Reset(w.config.Core.SleepInterval.Duration)
 	defer labelTrigger.Stop()
 
+	w.sourceEvent = make(chan string)
+	eventSources := source.GetAllEventSources()
+	for _, s := range eventSources {
+		if err := s.SetNotifyChannel(w.sourceEvent); err != nil {
+			klog.ErrorS(err, "failed to set notify channel for event source", "source", s.Name())
+			return fmt.Errorf("failed to set notify channel for event source %s: %w", s.Name(), err)
```
marquiz: Let's not error out in this case (just log the error). The design pattern of nfd-worker has been that log errors if we cannot access some data but don't crash. Let's follow that

pkg/nfd-worker/nfd-worker.go
```diff
@@ -304,6 +335,15 @@ func (w *nfdWorker) Run() error {
 	labelTrigger.Reset(w.config.Core.SleepInterval.Duration)
 	defer labelTrigger.Stop()
 
+	w.sourceEvent = make(chan string)
+	eventSources := source.GetAllEventSources()
+	for _, s := range eventSources {
```
marquiz: I think this stuff should be done in the `configure()` function. The other similar stuff is there, too.

marquiz (commented):
Thanks @ozhuraki for the update. Comments below

ozhuraki:
@marquiz

Thanks for the review. Updated, please take a look.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -216,10 +217,23 @@ func (i *infiniteTicker) Reset(d time.Duration) {
 	default:
 		// If the sleep interval is not a positive number the ticker will act
 		// as if it was set to an infinite duration by not ticking.
-		i.Stop()
+		i.Ticker.Stop()
 	}
 }
 
+// Publish labels.
+func (w *nfdWorker) publishLabels() error {
```
marquiz: nit: naming, this is not publishing only labels but both labels and "raw" features, via the NodeFeatureObject. I'd suggest to rename to `publishFeaturesAndLabels` or `publishNodeFeatureObject` or something similar

source/local/local.go
```diff
@@ -318,6 +322,63 @@ func getFileContent(fileName string) ([][]byte, error) {
...
+		if err != nil {
+			errWatcher := watcher.Close()
+			if errWatcher != nil {
+				klog.ErrorS(errWatcher, "failed to close fsnotify watcher")
+			}
+			return fmt.Errorf("unable to access %v: %w", featureFilesDir, err)
+		}
+		s.fsWatcher = watcher
+
+		go s.runNotifier(ctx, ch)
```
marquiz, line 374: Creating a new wacher and goroutine on each call (without clean-up) is risky business. If we ever call SetNotifyChannel again that will leave orphaned and stale goroutines and watchers running.

We should only need one watcher/notifier instance for the lifetime of the application. For example, only create those if s.fsWatcher is not nil. If it exists, SetNotifyChannel() only sets the channel.

Possible further refactoring is splitting the watcher/notifier initialization into a separate method to keep SetNotifyChannel() cleaner/terser.

source/local/local.go
```diff
@@ -318,6 +322,63 @@ func getFileContent(fileName string) ([][]byte, error) {
 	return lines, nil
 }
 
+func (s *localSource) runNotifier(ctx context.Context, ch chan *source.FeatureSource) {
+	rateLimit := time.After(time.Second)
```
marquiz: suggestion: use time.Ticker so there's no need to re-initialize later

```suggestion
	rateLimit := time.NewTicker(time.Second)
```

ozhuraki:
@marquiz

Thanks for the review. Updated, please take a look.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -525,6 +548,14 @@ func (w *nfdWorker) configure(filepath string, overrides string) error {
 		s.SetConfig(c.Sources[s.Name()])
 	}
 
+	w.sourceEvent = make(chan *source.FeatureSource)
```
marquiz: nit: future-proofness, close the old channel if it exists

pkg/nfd-worker/nfd-worker.go
```diff
@@ -525,6 +548,14 @@ func (w *nfdWorker) configure(filepath string, overrides string) error {
...
+	eventSources := source.GetAllEventSources()
+	for _, s := range eventSources {
```
marquiz: I think we should move these to `configureCore()`.

Also, we should probably honor the feature-sources flag/config so that if somebody does e.g. `-feature-sources=all,-local` it doesn't get activated.

pkg/nfd-worker/nfd-worker.go
marquiz: Hmm, I take my words back a bit. This might actually be not so good idea - if we close the channel here the other end will panic if it tries to send to the closed channel 🙄

Please think a way to handle this in "golang way". Maybe the safest way would be to set up the context in the configure() (and call cancel() before setting a new one if cancel is non-nil).

source/local/local.go
```diff
@@ -318,6 +322,64 @@ func getFileContent(fileName string) ([][]byte, error) {
...
+func (s *localSource) SetNotifyChannel(ctx context.Context, ch chan *source.FeatureSource) error {
+	info, err := os.Stat(featureFilesDir)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
+	}
+
+	if info != nil && info.IsDir() {
+		watcher, err := fsnotify.NewWatcher()
```
marquiz: Move this under inside the if s.fsWatcher == nil, right(?)

pkg/nfd-worker/nfd-worker.go
marquiz: > Also, we should probably honor the feature-sources flag/config so that if somebody does e.g. `-feature-sources=all,-local` it doesn't get activated.

This is still unresolved, if I'm not mistaken

pkg/nfd-worker/nfd-worker.go
```diff
@@ -237,12 +252,9 @@ func (w *nfdWorker) runFeatureDiscovery() error {
 	if w.config.Core.SleepInterval.Duration > 0 && discoveryDuration > w.config.Core.SleepInterval.Duration/2 {
 		klog.InfoS("feature discovery sources took over half of sleep interval ", "duration", discoveryDuration, "sleepInterval", w.config.Core.SleepInterval.Duration)
 	}
-	// Get the set of feature labels.
-	labels := createFeatureLabels(w.labelSources, w.config.Core.LabelWhiteList.Regexp)
 
-	// Update the node with the feature labels.
-	if !w.config.Core.NoPublish {
-		return w.advertiseFeatures(labels)
+	if err := w.publishNodeFeatureObject(); err != nil {
```
marquiz: nit: simplify 

```suggestion
	return w.publishNodeFeatureObject()
```

source/local/local.go
```diff
@@ -318,6 +322,63 @@ func getFileContent(fileName string) ([][]byte, error) {
...
+// SetNotifyChannel method of the EventSource Interface
+func (s *localSource) SetNotifyChannel(ctx context.Context, ch chan *source.FeatureSource) error {
+	info, err := os.Stat(featureFilesDir)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return err
+		}
+	}
+
+	if info != nil && info.IsDir() {
```
marquiz: nit: the error handling is somewhat unorthodox. Return right away if os.Stat fails.

```suggestion
		}
		return nil
	}

	if info.IsDir() {
```

marquiz (commented):
Thanks for the update @ozhuraki. Small nits/suggestions below.

I'm still a bit concnerned about the handling of the notifier goroutine (leaking of goroutines if SetNotifyChannel gets called more than once. But I guess we can live with that for now as this is more of a concern for future

**EDIT:** Please rebase to get rid of the netlify failures

ozhuraki:
@marquiz

Thanks, updated! Please take a look.

ArangoGutierrez (approved):
/lgtm

Issue #2127, opened 2025-04-15
Allow templating of updateStrategy for worker DaemonSet


jon-rei:
**What would you like to be added**:

Currently, the default updateStrategy for the DaemonSet is used, which is the following by default

```yaml
    updateStrategy:
      rollingUpdate:
        maxSurge: 0
        maxUnavailable: 1
```

The whole updateStrategy should be templatable with Helm to adjust the maxUnavailable setting or even other settings.

**Why is this needed**:

In our environment, we have hundreds of nodes, so a rollout takes a long time. Also, if a node is unhealthy, the rollout will stop. To overcome this, we always set maxUnavailable to 10% to speed up the rollout.

fmuyassarov:
This issue was addresses by https://github.com/kubernetes-sigs/node-feature-discovery/pull/2157. 
Closing as done.
/close

PR #2157, opened 2025-05-15
add configurable  worker daemonset updateStrategy in Helm chart

subnet-dev:
Related to: #2127

I remain open to any suggestions or proposals for improvement.

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/worker.yaml
+++ b/deployment/helm/node-feature-discovery/templates/worker.yaml
@@ -13,6 +13,10 @@ metadata:
   {{- end }}
 spec:
   revisionHistoryLimit: {{ .Values.worker.revisionHistoryLimit }}
+  {{- with .Values.worker.updateStrategy }}
+  updateStrategy:
+    {{- toYaml . | nindent 4 }}
+  {{- end}}
   selector:
     matchLabels:
       {{- include "node-feature-discovery.selectorLabels" . | nindent 6 }}
--- a/deployment/helm/node-feature-discovery/values.yaml
+++ b/deployment/helm/node-feature-discovery/values.yaml
@@ -469,6 +469,12 @@ worker:
 
   priorityClassName: ""
 
+  updateStrategy: {}
+    # type: RollingUpdate
+    # rollingUpdate:
+    #   maxSurge: 0
+    #   maxUnavailable: "10%"
+
 topologyUpdater:
   config: ### <NFD-TOPOLOGY-UPDATER-CONF-START-DO-NOT-REMOVE>
     ## key = node name, value = list of resources to be excluded.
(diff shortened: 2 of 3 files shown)


deployment/helm/node-feature-discovery/templates/worker.yaml
```diff
@@ -13,6 +13,13 @@ metadata:
   {{- end }}
 spec:
   revisionHistoryLimit: {{ .Values.worker.revisionHistoryLimit }}
+  updateStrategy:
+    type: {{ ((.Values.worker).updateStrategy).type | default "RollingUpdate" }}
+    {{- if or (not ((.Values.worker).updateStrategy).type) (eq .Values.worker.updateStrategy.type "RollingUpdate") }}
+    rollingUpdate:
+      maxSurge: {{ (((.Values.worker).updateStrategy).rollingUpdate).maxSurge | default 1 }}
```
marquiz, line 20: I think this doesn't work as intended, e.f. if you want to set maxSurge to 0. 

I suggest using something like this for the values:

```suggestion
      {{-with Values.worker.updateStrategy.rollingUpdate.maxSurge }}
      maxSurge: {{ . }}
      {{- end }}
```

Alternatively we could have a simple

```yaml
        {{- with .Values.worker.updateStrategy }}
          {{- toYaml . | nindent 2 }}
        {{- end}}
```

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -469,6 +469,12 @@ worker:
 
   priorityClassName: ""
 
+  # updateStrategy:
+  #   type: RollingUpdate
+  #   rollingUpdate:
+  #     maxSurge: 10%
+  #     maxUnavailable: 10%
```
marquiz, line 476: Depending on how the template is written, we might need

```suggestion
  updateStrategy: {}
    # type: RollingUpdate
    # rollingUpdate:
    #  maxSurge: 10%
    #  maxUnavailable: 10%
```
ALTERNATIVELY, we code put defaults, here, too

```suggestion
  updateStrategy: {}
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 1
      maxUnavailable: 1
```

marquiz (commented):
Thanks @subnet-dev for the patch.

In addition to the chart, we also need to document the new values in `docs/deployment/helm.md`

marquiz (approved):
@subnet-dev thanks for the update. I'm good with this approach (we don't seem to have a consistent way how to handle these, each field separately vs. dict and toYaml).

@ArangoGutierrez @fmuyassarov wanna have a say?

/ok-to-test

fmuyassarov (commented):
Would it be okay to squash c86d142a237f5140bbf73da4486e884222110031 and b7ddbe5f11a13ed8e73756932c61b0d07f85212b commits before we merge it? Otherwise, LGTM.

fmuyassarov (commented):
/lgtm

PR #2165, opened 2025-05-27
docs: re-align helm values tables

marquiz:
Use reference-style links to shorten the lines. Make maintenance a bit easier. Also, drop unneeded/unsupported backslashes from the one template value.


marquiz:
Rebased. @fmuyassarov PTAL

PR #2168, opened 2025-05-28
nfd-gc: rename -metrics to -port

marquiz:
Align with other daemons, -metrics has been deprecated. This patch also refactors the code to make it easier to add healthz endpoint.


fmuyassarov (approved):
/lgtm

PR #2170, opened 2025-06-07
feat: add ability to export feature labels for static node

vsoch:
Problem: We want to use NFD (labels and other functionality) for HPC, but a Kubernetes environment is expected.
Solution: Add an `nfd-worker --export` mode that allows for generation and export of static labels.

For the HPC use case, we want to be able to export static features without Kubernetes, either to the screen (for inspection) or a file (for saving and use with a scheduler, or other compatibility checking algorithm).  Adding support for this is fairly easy - we need to just expose the flag, disable the check for Kubernetes assets, and then generate the labels once and exit (without starting or sending to a server). The usage is simple and intuitive:

```bash
# Export labels to the terminal for inspection
nfd-worker --export

# Save to a file path
nfd-worker --export=labels.json
```
The flag is also useful for cases beyond HPC when a quick inspection is desired. Note that instead of adding two flags (a boolean for export, and string for a filepath) I opted for the design above - one flag that supports being used as a boolean or providing a path. A new flag type is added to support that, and it is generic so could be used for other future cases. For my implementation, for us to check how this new flag was set, we need to pass it to the initFlags function:

```go
args, overrides := initFlags(flags, &exportFlag)
```
And I think this is OK for now, but if more flags were defined in this manner, we would want to pass a structure with all of them. I decided that was overkill for the time being. For another design choice, I think the generation of the labels should be in the same function to `Run()`, and to support this I check `w.args.Export` before setting up any servers / Prometheus, and run the generation (that will handle the export to terminal or file and exit early, if applicable).  I have also added tests and documentation, and I tested on an HPC node (in user space) to make sure the labels generated without issue.

I realize I didn't open an issue for discussion - so am happy to have it here! To better explain the use case, this is a feature intended to better support compatibility checking, but for HPC. We won't be able to run compatibility checks assuming labels are available via Kubernetes, but we will be able to export them to files that represent nodes, and then be able to use associated tools to check compatibility artifacts against the known features. This is what I'm working on today. Thanks!

### DIFF
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -102,7 +102,10 @@ type Args struct {
 	Port        int
 	NoOwnerRefs bool
 
-	Overrides ConfigOverrideArgs
+	// Options for export
+	Export     bool
+	ExportPath string
+	Overrides  ConfigOverrideArgs
 }
 
 // ConfigOverrideArgs are args that override config file options
@@ -178,7 +181,8 @@ func NewNfdWorker(opts ...NfdWorkerOption) (NfdWorker, error) {
 	}
 
 	// k8sClient might've been set via opts by tests
-	if nfd.k8sClient == nil {
+	// Export mode does not assume in Kubernetes environment
+	if nfd.k8sClient == nil && !nfd.args.Export {
 		kubeconfig, err := utils.GetKubeconfig(nfd.args.Kubeconfig)
 		if err != nil {
 			return nfd, err
@@ -221,6 +225,7 @@ func (i *infiniteTicker) Reset(d time.Duration) {
 }
 
 // Run feature discovery.
+// singleExport will derive labels and save or print
 func (w *nfdWorker) runFeatureDiscovery() error {
 	discoveryStart := time.Now()
 	for _, s := range w.featureSources {
@@ -239,6 +244,29 @@ func (w *nfdWorker) runFeatureDiscovery() error {
 	}
 	// Get the set of feature labels.
 	labels := createFeatureLabels(w.labelSources, w.config.Core.LabelWhiteList.Regexp)
+
+	// Show exported labels and exit
+	if w.args.Export {
+		exportedLabels, err := json.MarshalIndent(labels, "", "    ")
+		if err != nil {
+			return err
+		}
+		// No export path provided
+		if w.args.ExportPath == "" {
+			fmt.Println(string(exportedLabels))
+		} else {
+
+			// Export labels to provided filename
+			fd, err := os.Create(w.args.ExportPath)
+			if err != nil {
+				return err
+			}
+			defer fd.Close()
+			_, err = fmt.Fprint(fd, string(exportedLabels))
+			return err
+		}
+		return err
+	}
 
 	// Update the node with the feature labels.
 	if !w.config.Core.NoPublish {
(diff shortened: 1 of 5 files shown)


ChaoyiHuang:
```
# Export labels to the terminal for inspection
nfd-worker --export

# Save to a file path
nfd-worker --export=labels.json
```

how about using stdout to unify the commandline, for example

```
# Export labels to the terminal for inspection
nfd-worker --export=stdout

# Save to a file path
nfd-worker --export=labels.json
```

vsoch:
@ChaoyiHuang thank you for taking a look! How would you distinguish these two?

```bash
# Export labels to the terminal for inspection
nfd-worker --export=stdout
nfd-worker --export
```
They are both going to dump to the output stream. Did you have in mind using different streams, or minimizing other output for the rest of the run for one case? Or are you saying to keep the command the same, but have `stdout`? I think the issue with that would be not knowing if a user wants to write to a file called `stdout` vs standard output. With just `--export` it is clear because the variable isn't defined.

ChaoyiHuang:
> th just `--export` it is clear because the variable isn't defined.

keep the command the same format (stdout is standard output). 

--export is also OK, just have to code a little more.

ChaoyiHuang:
it may just export to standard output, and no need to support export to path. 

if want to save the content to file, it can be done via redirecting the output to file, for example:

```
nfd-worker --export > labels.json
```

so the code can be simplified to just support exporting to standard output

```
nfd-worker --export
```

vsoch:
The nfd worker generates a lot of logs (and other unexpected output) so I wasn't sure how to cleanly do that. Whereas with writing the file explicitly, I can control exactly the contents.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -178,7 +181,8 @@ func NewNfdWorker(opts ...NfdWorkerOption) (NfdWorker, error) {
 	}
 
 	// k8sClient might've been set via opts by tests
-	if nfd.k8sClient == nil {
+	// Export mode does not assume in Kubernetes environment
+	if nfd.k8sClient == nil && !nfd.args.Export {
```
mfranczy, line 185: I believe it's better to keep the worker k8s centric (as it implements the logic for updating k8s resources), operate on raw features rather than labels, and add a new subcommand to the [nfd client](https://github.com/kubernetes-sigs/node-feature-discovery/tree/master/cmd/nfd) for dumping features.

Raw features can easily be transformed into labels if needed. However, since labels are derived from raw features and the [worker configuration](https://kubernetes-sigs.github.io/node-feature-discovery/v0.16/reference/worker-configuration-reference#corelabelsources), relying solely on labels may cause certain node capabilities to be missed due to default or custom configurations.

The same principle applies to the image compatibility implementation in NFD, where comparisons are performed by [checking the features](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/pkg/client-nfd/compat/node-validator/node-validator.go#L74) (not the labels).

Thoughts?

vsoch: I agree - I'll update the PR this week. Thanks @mfranczy!

marquiz: Yes, I tend to agree as well. Mostly because much of the nfd-worker code is currently heavily based on the assumption that we're running in a kubernetes environment. Starting from the assumption that a kubeconfig must be found... Wiring the export functionality __nicely__ into nfd-worker would require heavy refactoring, I believe, and it's much easier to put the stuff into the nfd client (I believe) as @mfranczy suggested. Especially if/when you're not interested in all the config options of nfd-worker.

vsoch:
Superseded by https://github.com/kubernetes-sigs/node-feature-discovery/pull/2183.

Thanks for the discussion here - closing for follow up in #2183.

PR #2183, opened 2025-06-25
feat: add ability to export features for static node

vsoch:
Problem: We want to statically export NFD features for HPC.
Solution: Add an `nfd export features` mode that allows for generation and export of static feature labels.

For the HPC use case, we want to be able to export static features, either to the screen (for inspection) or a file (for saving and use with a scheduler, or other compatibility checking algorithm).  Adding support for this is fairly easy - we need to just add the subcommand. The usage is simple and intuitive:

```bash
# Export raw features to the terminal for inspection
nfd export features

# Save to a file path
nfd export features --path features.json

# or the same for labels
nfd export labels
nfd export labels --path labels.json
```

The flag is also useful for cases beyond HPC when a quick inspection is desired. There are ways this could be extended (requesting specific groups or filters) but I think it's better to start simple.

This is a follow up of https://github.com/kubernetes-sigs/node-feature-discovery/pull/2170. If you'd like tests added, I did an attempt and it looks like the features aren't exposed in the test environment - we would need a mock. Please point me to an example for how to do that. Thanks!

### DIFF
--- a/cmd/nfd/subcmd/export/export.go
+++ b/cmd/nfd/subcmd/export/export.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 The Kubernetes Authors.
+Copyright 2025 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -14,33 +14,22 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package subcmd
+package export
 
 import (
 	"fmt"
 	"os"
 
 	"github.com/spf13/cobra"
-
-	"sigs.k8s.io/node-feature-discovery/cmd/nfd/subcmd/compat"
-	"sigs.k8s.io/node-feature-discovery/cmd/nfd/subcmd/export"
 )
 
-// RootCmd represents the base command when called without any subcommands
-var RootCmd = &cobra.Command{
-	Use:   "nfd",
-	Short: "Node Feature Discovery client",
+var ExportCmd = &cobra.Command{
+	Use:   "export",
+	Short: "Export commands",
 }
 
-func init() {
-	RootCmd.AddCommand(compat.CompatCmd)
-	RootCmd.AddCommand(export.ExportCmd)
-}
-
-// Execute adds all child commands to the root command and sets flags appropriately.
-// This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	if err := RootCmd.Execute(); err != nil {
+	if err := ExportCmd.Execute(); err != nil {
 		fmt.Println(err)
 		os.Exit(1)
 	}
--- a/cmd/nfd/subcmd/export/features.go
+++ b/cmd/nfd/subcmd/export/features.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package export
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+func NewExportCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "features",
+		Short: "Export features for given node",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			sources := map[string]source.FeatureSource{}
+			for k, v := range source.GetAllFeatureSources() {
+				if ts, ok := v.(source.SupplementalSource); ok && ts.DisableByDefault() {
+					continue
+				}
+				sources[k] = v
+			}
+
+			// Discover all feature sources
+			for _, s := range sources {
+				if err := s.Discover(); err != nil {
+					return fmt.Errorf("error during discovery of source %s: %w", s.Name(), err)
+				}
+			}
+
+			features := source.GetAllFeatures()
+			exportedLabels, err := json.MarshalIndent(features, "", "    ")
+			if err != nil {
+				return err
+			}
+
+			if outputPath != "" {
+				err = writeToFile(outputPath, string(exportedLabels))
+			} else {
+				fmt.Println(string(exportedLabels))
+			}
+			return err
+		},
+	}
+	cmd.Flags().StringVar(&outputPath, "path", "", "export to this JSON path")
+	return cmd
+}
+
+func init() {
+	ExportCmd.AddCommand(NewExportCmd())
+}
--- a/cmd/nfd/subcmd/export/labels.go
+++ b/cmd/nfd/subcmd/export/labels.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package export
+
+import (
+	"encoding/json"
+	"fmt"
+	"maps"
+	"regexp"
+	"slices"
+	"sort"
+
+	"github.com/spf13/cobra"
+	"k8s.io/klog/v2"
+
+	worker "sigs.k8s.io/node-feature-discovery/pkg/nfd-worker"
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+func NewLabelsCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "labels",
+		Short: "Export feature labels for given node",
+		RunE: func(cmd *cobra.Command, args []string) error {
+
+			// Determine enabled feature sources
+			featureSources := make(map[string]source.FeatureSource)
+			for n, s := range source.GetAllFeatureSources() {
+				if ts, ok := s.(source.SupplementalSource); !ok || !ts.DisableByDefault() {
+					err := s.Discover()
+					if err != nil {
+						return err
+					}
+					featureSources[n] = s
+				}
+			}
+			featureSourceList := slices.Collect(maps.Values(featureSources))
+			sort.Slice(featureSourceList, func(i, j int) bool { return featureSourceList[i].Name() < featureSourceList[j].Name() })
+
+			// Determine enabled label sources
+			labelSources := make(map[string]source.LabelSource)
+			for n, s := range source.GetAllLabelSources() {
+				if ts, ok := s.(source.SupplementalSource); !ok || !ts.DisableByDefault() {
+					labelSources[n] = s
+				}
+			}
+			labelSourcesList := slices.Collect(maps.Values(labelSources))
+			sort.Slice(labelSourcesList, func(i, j int) bool {
+				iP, jP := labelSourcesList[i].Priority(), labelSourcesList[j].Priority()
+				if iP != jP {
+					return iP < jP
+				}
+				return labelSourcesList[i].Name() < labelSourcesList[j].Name()
+			})
+
+			labels := worker.Labels{}
+			labelWhiteList := *regexp.MustCompile("")
+
+			// Get labels from all enabled label sources
+			for _, source := range labelSourcesList {
+				labelsFromSource, err := worker.GetFeatureLabels(source, labelWhiteList)
+				if err != nil {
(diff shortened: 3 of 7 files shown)


cmd/nfd/subcmd/export/features.go
```diff
@@ -0,0 +1,103 @@
...
+					}
+				}
+			}
+			exportedLabels, err := json.MarshalIndent(featureListing, "", "    ")
+			if err != nil {
+				return err
+			}
+
+			if outputPath != "" {
+				fd, err := os.Create(outputPath)
```
ChaoyiHuang: It's possible that the user want to export the content into file located not in the current directory. so if the file path does not exist, os.Create will report error "The directory containing the file must already exist. If there is an error, it will be of type *PathError ."

cmd/nfd/subcmd/export/features.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+var (
+	outputJSON bool
```
ChaoyiHuang: outputJSON not used in the file

vsoch: Good catch! Fixed.

cmd/nfd/subcmd/export/features.go
```diff
@@ -0,0 +1,102 @@
...
+				if ts, ok := v.(source.SupplementalSource); ok && ts.DisableByDefault() {
+					continue
+				}
+				sources[k] = v
+			}
+
+			// Discover all feature sources
+			for _, s := range sources {
+				if err := s.Discover(); err != nil {
+					fmt.Errorf("error during discovery of source %s: %w", s.Name(), err)
```
mfranczy: `fmt.Errorf` output is not used, I assume you wanted to write:

`return fmt.Errorf("error during discovery of source %s: %w", s.Name(), err)`

instead of 

`return err`

cmd/nfd/subcmd/export/features.go
```diff
@@ -0,0 +1,102 @@
...
+			}
+
+			for featureName, featureSet := range features.Instances {
+				for _, element := range featureSet.Elements {
+					for attrName, attrValue := range element.Attributes {
+						featureListing[fmt.Sprintf("%s.%s", featureName, attrName)] = attrValue
+					}
+				}
+			}
+			exportedLabels, err := json.MarshalIndent(featureListing, "", "    ")
```
mfranczy: Currently, key collisions occur during processing. When multiple PCI devices are present on the host, only the last processed device’s attributes are retained, as all devices share the same key paths (e.g., pci.device.class). Earlier values are overwritten.

I would suggest simply executing `json.MarshalIndent(features, "", "  ")` as this:
1. eliminates key collisions
2. produces output that is already compatible with the `nfd compat` command
3. preserves the flag, attribute, and instance types which are crucial for the NFD match expression engine

However, if returning domain-style paths is required, it may be better to introduce an additional flag, such as `--output-domain` and add a counter to the path, for example: `pci.devices.1...`

mfranczy:
That's a nice functionality, not only for non-k8s use cases, but also for taking a snapshot of the node's state from the perspective of enabled features.

cmd/nfd/subcmd/export/features.go
vsoch: I see what you mean (40 devices [here](https://gist.github.com/vsoch/9f8fb46020d245b6dda1691d6350d795)) are reduced to 6. The same issue exists for others that are kept as arrays.

The problem as I see it is the user might eventually want to map features to key value pairs (matching to Kubernetes) and if we don't flatten in some way, we place the burden on them to do it right. Is the problem just avoided for node features because pci devices are not included? If we currently have:

```
    "pci.device.class": "0c03",
    "pci.device.device": "1669",
    "pci.device.iommu_group/type": "DMA-FQ",
    "pci.device.subsystem_device": "50d9",
    "pci.device.subsystem_vendor": "17aa",
    "pci.device.vendor": "1022",
```
What if we added the index to the key?

```
    "pci.device.0.class": "0c03",
    "pci.device.0.device": "1669",
    "pci.device.0.iommu_group/type": "DMA-FQ",
    "pci.device.0.subsystem_device": "50d9",
    "pci.device.0.subsystem_vendor": "17aa",
    "pci.device.0.vendor": "1022",
...
```
I put that example here: https://gist.github.com/vsoch/9f8fb46020d245b6dda1691d6350d795#file-features-in-list-json. It includes all features, but could be problematic if the order is somehow changed. We would need the tool to not care about the index of a set, as long as it is present.

And for use cases - I like that the output works with nfd compat, but (at least for the hpc use case) it's unlikely nfd compat will be used if we are going to read a cache of nodes from elsewhere. Did you have a use case in mind?

And what would it look like if someone enabled the export for the kubernetes labels?

cmd/nfd/subcmd/export/features.go
vsoch: Does checking the error is not nil and returning if it isn't not cover that? Why do we need special processing for a path error? 

If you are suggesting we create the directory, my 0.02 is that we shouldn't generally be making directories to export a json file to - that's on the user to choose an existing path.

cmd/nfd/subcmd/export/features.go
mfranczy: > The problem as I see it is the user might eventually want to map features to key value pairs (matching to Kubernetes) and if we don't flatten in some way, we place the burden on them to do it right.

Right, in that case I would propose to implement two separate commands:
* `nfd export features` - to export features in their raw format
* `nfd export labels` - to export labels

> We would need the tool to not care about the index of a set, as long as it is present.

I think instead of adding the index to the domain path, we should follow the default NFD labels format as described [here](https://kubernetes-sigs.github.io/node-feature-discovery/v0.17/usage/features.html#pci) for PCI. In this case it's `pci-<class>_<vendor>.present: true`. Other sources should follow their own default format too.

There are methods for getting the labels already. I think it's a matter of executing [GetAllLabelSources](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/source/source.go#L129). Hopefully, you should instantly get all labels.

> And for use cases - I like that the output works with nfd compat, but (at least for the hpc use case) it's unlikely nfd compat will be used if we are going to read a cache of nodes from elsewhere. Did you have a use case in mind?

I like the idea of having two distinct commands as it keeps the functionality aligned with NFD's conventions while still allowing usage outside of k8s.

I personally find the following use cases valid:
* `nfd export features` - for tools that require knowledge of features type (like nfd compat). This command will allow for snapshotting the state of the node and simplify testing.
* `nfd export labels` - for HPC and similar use cases, as well for testing the worker configuration in the future.

What do you think?

cmd/nfd/subcmd/export/export.go
```diff
@@ -0,0 +1,36 @@
+/*
+Copyright 2024 The Kubernetes Authors.
```
mfranczy: Should be 2025 :)

cmd/nfd/subcmd/export/features.go
vsoch: I like that interface! I will work on updating the PR for it.

cmd/nfd/subcmd/export/export.go
vsoch: But... but... but! https://youtu.be/vhfsbHnM7dI :laughing:

cmd/nfd/subcmd/export/features.go
vsoch: @mfranczy all set!

cmd/nfd/subcmd/export/features.go
ChaoyiHuang: ok to only work for existing path, let user to create non existing path before export

cmd/nfd/subcmd/export/export.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+var ExportCmd = &cobra.Command{
+	Use:   "export",
+	Short: "Export features commands",
```
ChaoyiHuang: Export features or labels commands

vsoch: Good catch! What do you think of just "Export commands" to be generic for the export group? :thinking:

ChaoyiHuang: of course much more friendly for extention :)

mfranczy (approved):
lgtm

/cc @marquiz @ArangoGutierrez

cmd/nfd/subcmd/export/labels.go
```diff
@@ -0,0 +1,106 @@
...
+				}
+				defer fd.Close()
+				_, err = fmt.Fprint(fd, string(exportedLabels))
+				return err
+			} else {
+				fmt.Println(string(exportedLabels))
+			}
+			return nil
+		},
+	}
```
ChaoyiHuang, line 95: line 86~line 99 is to export content to file, same code for features and lables, how about wrape these lines in a function for deduplication.

vsoch: Good thinking! All set.

cmd/nfd/subcmd/export/features.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"sigs.k8s.io/node-feature-discovery/source"
+)
+
+var (
```
ChaoyiHuang: this variable is used both in features and lables export, it could be defined in NewExportCmd(or renamed to NewFeaturesCmd) and NewLabelsCmd, in which the nested function can call the local variable in the parent function.

or move this variable to util file.

overall LGTM

vsoch: Moved to utils. Thanks!

vsoch:
Ping @adrianchiris @zvonkok @ArangoGutierrez @marquiz we've gone through several rounds of review and have LGTM from the compatibility working group. Can any of you review this?

vsoch:
Ping @adrianchiris @zvonkok @ArangoGutierrez @marquiz we've gone through several rounds of review and have LGTM from the compatibility working group. Can any of you review this?

cmd/nfd/subcmd/export/export.go
ArangoGutierrez: Although I enjoyed the video, let's go for 2025.

cmd/nfd/subcmd/export/utils.go
```diff
@@ -0,0 +1,21 @@
+package export
```
ArangoGutierrez, line 17: Missing LICENSE header

cmd/nfd/subcmd/export/utils.go
```diff
@@ -0,0 +1,21 @@
...
+
+import (
+	"fmt"
+	"os"
+)
+
+var (
+	outputPath string
```
ArangoGutierrez, line 25: If Var is global for the pkg `export`, why have it as an entry value for `writeToFile`?

cmd/nfd/subcmd/export/utils.go
```diff
@@ -0,0 +1,21 @@
...
+)
+
+// writeToFile saves string content to file
```
ArangoGutierrez:
```suggestion
// writeToFile saves string content to a file at the path set by outputPath
```

cmd/nfd/subcmd/export/export.go
vsoch: This was fixed (an old comment) - I'll resolve this time!

cmd/nfd/subcmd/export/utils.go
vsoch: It is global (and private) because it's a shared flag, and I usually see those defined in this way. For the latter, I think it's better practice to not have package functions rely on global variables.

cmd/nfd/subcmd/export/utils.go
vsoch: This was also added.

cmd/nfd/subcmd/export/utils.go
```diff
@@ -0,0 +1,37 @@
...
+	"fmt"
+	"os"
+)
+
+var (
+	outputPath string
+)
+
+// writeToFile saves string content to a file at the path set by outputPath
+func writeToFile(outputPath, content string) error {
```
ChaoyiHuang: it will be good to use another name for the function parameter "outputPath" here, to avoid the confusion of global variable name and local varible name

vsoch: Sure, done.

docs/usage/nfd-export.md
```diff
@@ -0,0 +1,58 @@
...
+
+## Export
+
+If you are interested in exporting features or labels in a generic
+context, the nfd client supports an export mode, where both can be
+derived on the command line.
+
+### Features
+
+**Feature export is in the experimental `v1alpha1` version.**
```
fmuyassarov: Is it OK to refer to this as `v1alpha1` even if isn't an API but a command line option?

vsoch: We could just say it is an experimental version. Would that be better?

fmuyassarov: probably would be better, not to give a wrong impression to the readers.

vsoch: Will do! I’m running to the bakery to be there when it opens but will change this when I get back. Thanks for the review!

vsoch: Done!

fmuyassarov (commented):
Thank you. LGTM

ArangoGutierrez (approved):
PING @marquiz  for lgtm

ArangoGutierrez:
@vsoch please check the failing test, PR is now approved

ArangoGutierrez (approved):
/lgtm

vsoch:
Thanks to all! :tada:

PR #2191, opened 2025-07-07
update: add support for 'imagePullSecrets' propagation by root helm chart

heyvister1:
nfd serves as sub-chart in our project, and we would like to propagate `imagePullSecrets` content in root `values.yaml` to nfd sub-chart. Therefore I have added `.Values.global.imagePullSecrets` and created a helper helm func to update accordingly.

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/_helpers.tpl
+++ b/deployment/helm/node-feature-discovery/templates/_helpers.tpl
@@ -105,3 +105,22 @@ Create the name of the service account which nfd-gc will use
     {{ default "default" .Values.gc.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
+
+{{/*
+imagePullSecrets helper - uses local values or falls back to global values
+*/}}
+{{- define "node-feature-discovery.imagePullSecrets" -}}
+{{- $imagePullSecrets := list -}}
+{{- if .Values.imagePullSecrets -}}
+  {{- range .Values.imagePullSecrets -}}
+    {{- $imagePullSecrets = append $imagePullSecrets . -}}
+  {{- end -}}
+{{- else if and .Values.global .Values.global.imagePullSecrets -}}
+  {{- range .Values.global.imagePullSecrets -}}
+    {{- $imagePullSecrets = append $imagePullSecrets . -}}
+  {{- end -}}
+{{- end -}}
+{{- if $imagePullSecrets -}}
+{{- $imagePullSecrets | toJson }}
+{{- end -}}
+{{- end -}}
--- a/docs/deployment/helm.md
+++ b/docs/deployment/helm.md
@@ -170,6 +170,12 @@ Metrics are configured to be exposed using prometheus operator API's by
 default. If you want to expose metrics using the prometheus operator
 API's you need to install the prometheus operator in your cluster.
 
+### Global parameters
+
+| Name                                        | Type    | Default                          | Description                                                                                                                |
+|---------------------------------------------|---------|----------------------------------|----------------------------------------------------------------------------------------------------------------------------|
+| `global.imagePullSecrets`                   | array   | []                               | An optional list of references to secrets with the same meaning as `imagePullSecrets`. If `imagePullSecrets` is specified, it takes precedence over `global.imagePullSecrets`. |
+
 ### Master pod parameters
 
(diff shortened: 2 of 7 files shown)


heyvister1:
cc @adrianchiris

deployment/helm/node-feature-discovery/templates/_helpers.tpl
```diff
@@ -105,3 +105,22 @@ Create the name of the service account which nfd-gc will use
...
 {{- end -}}
+
+{{/*
+imagePullSecrets helper - uses local values or falls back to global values
+*/}}
+{{- define "node-feature-discovery.imagePullSecrets" -}}
+{{- $imagePullSecrets := list -}}
+{{- if .Values.imagePullSecrets -}}
+  {{- range .Values.imagePullSecrets -}}
+    {{- $imagePullSecrets = append $imagePullSecrets (dict "name" .) -}}
```
adrianchiris: until today, IIUC the format of Values.imagePullSecrets was a list of dict with the name key (thats the only way it could have worked)

so before this change if someone wanted to specify image pull secret in values.yaml he/she would:
```
values.yaml:
------------
imagePullSecrets:
  - name: my-image-secret
  - name: my other secret
```

this change breaks that as now it would be:
```
imagePullSecrets:
  - my-pull-secret
  - my-other-pull-secret
```

so i would do the following:
1. keep the original format to not break user API with the chart
2. modify this helper to take that into account
3. explicitly define a global section in the values.yaml with imagePullSecrets so that its clear we are using it. the image pull secret format should be the same as we have today IMO else it would be confusing.

this is a double edged sword though since global section in helm means all charts get whatever is defined there.
this means that its expected (in this example) for all charts to accept imagePullSecrets in this format.

from your implementation i think this might not be the case.

heyvister1: Thanks @adrianchiris I have removed addition of "name" key in helper func to stick with original format.

adrianchiris:
/lgtm

can you please update `docs/deplyment/helm.md` as well ?

marquiz (approved):
Thank you @heyvister1 for the contribution. Looks good to me
/ok-to-test

@adrianchiris PTAL

docs/deployment/helm.md
```diff
@@ -170,6 +170,9 @@ Metrics are configured to be exposed using prometheus operator API's by
 default. If you want to expose metrics using the prometheus operator
 API's you need to install the prometheus operator in your cluster.
 
+> `imagePullSecrets` assignment is done through `node-feature-discovery.imagePullSecrets` in [deployment/helm/node-feature-discovery/templates/_helpers.tpl](https://github.com/kubernetes-sigs/node-feature-discovery/blob/{{ site.release }}/deployment/helm/node-feature-discovery/templates/_helpers.tpl)
```
adrianchiris: i was thinking to just extend  general parameters above. or better yet add a new section: "Global Parameters"

heyvister1: Done

marquiz (approved):
New table is better :+1:

marquiz:
@adrianchiris PTAL

docs/deployment/helm.md
```diff
@@ -170,6 +170,12 @@ Metrics are configured to be exposed using prometheus operator API's by
 default. If you want to expose metrics using the prometheus operator
 API's you need to install the prometheus operator in your cluster.
 
+### Global parameters
+
+| Name                                        | Type    | Default                          | Description                                                                                                                |
+|---------------------------------------------|---------|----------------------------------|----------------------------------------------------------------------------------------------------------------------------|
+| `global.imagePullSecrets`                   | array   | []                               | In case global `imagePullSecrets` are used, their values will be appended, when local `.Values.imagePullSecrets` are not specified. Local values take precedence over global values. |
```
adrianchiris: i find the description hard to understand.

suggested reword:

An optional list of references to secrets with the same meaning as `imagePullSecrets`. If `imagePullSecrets` is specified, it takes precedence over `global.imagePullSecrets`

heyvister1: Done

adrianchiris (approved):
/lgtm

marquiz:
Thanks @adrianchiris. Why prow ignored your lgtm :disappointed: 
/lgtm

PR #2192, opened 2025-07-07
export swap behavior via label

### LINKED ISSUE #2178: add swap detailed feature label

> a. NFD labels reflect hardware state (swap device presence), not Kubernetes configuration state. For instance, it cannot distinguish between NoSwap, LimitedSwap, or future swap modes.

This was disscuessed in https://github.com/kubernetes/kubernetes/issues/132416#issuecomment-2992329382.

**Why is this needed**:

> Add swap labels to swap-enabled nodes to allow admin to target swap demanding pods selectively for scheduling on specific nodes.

`For instance, it cannot distinguish between NoSwap, LimitedSwap, or future swap modes.`

fmuyassarov:
This patch exports a label indicating the kubelet's configured swap behavior. By default, it reads from `/var/lib/kubelet/config.yaml`. If a custom kubelet config path is used, it can be set via the `-kubelet-config-path` flag.
If the swap behavior is not specified, `NoSwap` is assumed (matching kubelet's default).

**Note**: `feature.node.kubernetes.io/memory-swap.behavior` (or in general kubelet's `memorySwap.swapBehavior`) reflects whether k8s workloads are allowed to use swap. A node may have swap enabled, but if kubelet is set to `NoSwap`, pods cannot use it. As such, the behavior label will be exported only if node level swap is enabled.
```YAML
      feature.node.kubernetes.io/memory-swap: "true"
      feature.node.kubernetes.io/memory-swap.behavior: LimitedSwap
```
Fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/2178


deployment/components/common/worker-mounts.yaml
```diff
@@ -22,6 +22,9 @@
   - name: features-d
     hostPath:
       path: "/etc/kubernetes/node-feature-discovery/features.d/"
+  - name: kubelet-dir
+    hostPath:
+      path: "/var/lib/kubelet/config.yaml"
```
fmuyassarov: @marquiz Do you have any suggestions for handling this part?
When a custom kubelet config path is used, I want to ensure the user-specified path (via the `kubelet-config-path` flag) is correctly propagated. There are two places where this path needs to be dynamic. 
1. when mounting the path so the NFD worker can read the file
2. when passing the path to the worker (this part is already handled).
The issue is that it's easy to parameterize this in Helm, but I haven't found a clean way to do it with Kustomize.

Also, for the Helm specifically, do you think it makes sense to place the flag under `worker.config`, or would another location be more appropriate?

marquiz (commented):
@fmuyassarov thank you for the patch and working on this. Let's get this finalized.

My main comment is that in topology-updater we already read the kubelet config (read from the kubelet configz endpoint by default), helpers are in `pkg/utils/kubeconf/`. Could we do the same on nfd-worker?

fmuyassarov:
> @fmuyassarov thank you for the patch and working on this. Let's get this finalized.
> 
> My main comment is that in topology-updater we already read the kubelet config (read from the kubelet configz endpoint by default), helpers are in `pkg/utils/kubeconf/`. Could we do the same on nfd-worker?

I will check that.

fmuyassarov:
Hi @marquiz . I've reworked the patch to utilize the configz. The current state is missing the documentation update which I thought of adding once I get a green light that this the right approach. Please take a look.

deployment/base/rbac/worker-role.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: rbac.authorization.k8s.io/v1
-kind: Role
+kind: ClusterRole
```
marquiz: With the principle of least privileges in mind, I think we should create a clusterrole for "nodes/proxy" (only) and keep the existing role as is. WDYT?

pkg/nfd-worker/nfd-worker.go
```diff
@@ -255,9 +277,10 @@ func (w *nfdWorker) setOwnerReference() error {
 	if !w.config.Core.NoOwnerRefs {
 		// Get pod owner reference
 		podName := os.Getenv("POD_NAME")
+		podNamespace := os.Getenv("POD_NAMESPACE")
 		// Add pod owner reference if it exists
 		if podName != "" {
-			if selfPod, err := w.k8sClient.CoreV1().Pods(w.kubernetesNamespace).Get(context.TODO(), podName, metav1.GetOptions{}); err != nil {
+			if selfPod, err := w.k8sClient.CoreV1().Pods(podNamespace).Get(context.TODO(), podName, metav1.GetOptions{}); err != nil {
```
marquiz: What's wrong with `w.kubernetesNamespace`?

pkg/nfd-worker/nfd-worker.go
```diff
@@ -720,3 +749,38 @@ func (c *sourcesConfig) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+func getKubeletConfigFunc(uri, apiAuthTokenFile string) (func() (*kubeletconfigv1beta1.KubeletConfiguration, error), error) {
```
marquiz: Would it be possible to share code with nfd-topology-updater?

marquiz (commented):
Thanks for the update @fmuyassarov. A few quick comments below.

The bigger question to me is how should we hierarchically place the new feature. The feature is related to swap, but specifically kubernetes/kubelet config setting. I could think e.g. a new `system.kubelet` or `system.kubernetes` feature. Then kubeletconfig config opts would not be command line flags for nfd-worker but config options for the source. And we would get robustness, nfd-worker would not refuse to start if it was not able to get the kubelet config

All in all, I'm not questioning the need for the feature. Just need to think what would be the most logical way to expose and implement it. Need to sleep over this.

deployment/base/rbac/worker-role.yaml
fmuyassarov: Sure.

pkg/nfd-worker/nfd-worker.go
fmuyassarov: Yup, I can move the code around to avoid duplication.

pkg/nfd-worker/nfd-worker.go
fmuyassarov: If I recall correctly from my test, if `kubernetesNamespace` is empty it will refuse because we are not using the clusterRole. But if I follow your comment earlier, keeping the Role as is and creating the ClusterRole only for the nodes/proxy, then we can keep `w.kubernetesNamespace`.

ArangoGutierrez: So we can expect this to be reverted?

deployment/base/rbac/worker-role.yaml
ArangoGutierrez: ++

ArangoGutierrez (commented):
2 comments

cmd/nfd-worker/main.go
```diff
@@ -134,3 +155,8 @@ func initFlags(flagset *flag.FlagSet) (*worker.Args, *worker.ConfigOverrideArgs)
 
 	return args, overrides
 }
+
+func isIPv6(addr string) bool {
+	ip := net.ParseIP(addr)
+	return ip != nil && strings.Count(ip.String(), ":") >= 2
```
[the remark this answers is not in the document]
fmuyassarov: this just a copy paste of the same helper function from topology updater. But I can fix it in both places.

pkg/nfd-worker/nfd-worker.go
fmuyassarov: > So we can expect this to be reverted?

yes

marquiz:
@fmuyassarov independent of where we logically put the new feature, I think the detection should be outside nfd-worker core package. If we keep it in memory.swap, then probably the feature name should somehow reflect that it's **kubelet** behavior.

fmuyassarov:
> @fmuyassarov independent of where we logically put the new feature, I think the detection should be outside nfd-worker core package. If we keep it in memory.swap, then probably the feature name should somehow reflect that it's **kubelet** behavior.

The reason for splitting the detection between memory and worker pkgs is that, memory pkg doesn't have access to the kubeconfig & kubelet. Technically speaking, there is no need to do anything within the memory pkg. But, the reason I used memory pkg is because we are detecting the memory related feature. Sure, it is via kubelet, but the main thing is not kubelet, but the memory. Since we already had a logic of detecting the swap memory within the memory pkg, I thought of  extending it for swap memory type too.

pkg/nfd-worker/nfd-worker.go
fmuyassarov: @marquiz It looks like we still need this. Following your earlier comment about using a ClusterRole only for `nodes/proxy`, I updated the code to create a ClusterRole and ClusterRoleBinding in addition to the existing Role and RoleBinding. However, during my test runs, since I’m not providing the `KUBERNETES_NAMESPACE` variable, it defaults to an empty value for `w.kubernetesNamespace`, which causes the following issue:

```
E0918 10:19:43.598034       1 nfd-worker.go:282] "failed to get self pod, cannot inherit ownerReference for NodeFeature" err="the server could not find the requested resource"
```
```
Tracking new pod rollout (nfd-worker-5m584):
     ┊ Scheduled       - <1s
     ┊ Initialized     - <1s
     ┃ Not Ready       - (ContainersNotReady): containers with unready status: [nfd-worker]
[event: pod node-feature-discovery/nfd-worker-5m584] Container image "ttl.sh/gcr.io_k8s-staging-nfd_node-feature-discovery:tilt-bc71b0b17aea3fb5" already present on machine
I0918 10:19:41.895350       1 nfd-worker.go:315] "Node Feature Discovery Worker" version="" nodeName="eseldb12u01" namespace=""
E0918 10:19:41.904694       1 nfd-worker.go:282] "failed to get self pod, cannot inherit ownerReference for NodeFeature" err="the server could not find the requested resource"
E0918 10:19:41.904738       1 main.go:73] "error while running" err="the server could not find the requested resource"
[event: pod node-feature-discovery/nfd-worker-5m584] Container image "ttl.sh/gcr.io_k8s-staging-nfd_node-feature-discovery:tilt-bc71b0b17aea3fb5" already present on machine
[event: pod node-feature-discovery/nfd-worker-5m584] Back-off restarting failed container nfd-worker in pod nfd-worker-5m584_node-feature-discovery(784862c3-3fbe-445a-8cc2-e7bee0675246)
Detected container restart. Pod: nfd-worker-5m584. Container: nfd-worker.
I0918 10:19:43.506399       1 nfd-worker.go:315] "Node Feature Discovery Worker" version="" nodeName="eseldb12u01" namespace=""
E0918 10:19:43.598034       1 nfd-worker.go:282] "failed to get self pod, cannot inherit ownerReference for NodeFeature" err="the server could not find the requested resource"
E0918 10:19:43.598101       1 main.go:73] "error while running" err="the server could not find the requested resource"
[event: pod node-feature-discovery/nfd-worker-5m584] Back-off restarting failed container nfd-worker in pod nfd-worker-5m584_node-feature-discovery(784862c3-3fbe-445a-8cc2-e7bee0675246)
[event: pod node-feature-discovery/nfd-worker-5m584] Back-off restarting failed container nfd-worker in pod nfd-worker-5m584_node-feature-discovery(784862c3-3fbe-445a-8cc2-e7bee0675246)
[event: pod node-feature-discovery/nfd-worker-5m584] Back-off restarting failed container nfd-worker in pod nfd-worker-5m584_node-feature-discovery(784862c3-3fbe-445a-8cc2-e7bee0675246)
[event: pod node-feature-discovery/nfd-worker-5m584] Container image "ttl.sh/gcr.io_k8s-staging-nfd_node-feature-discovery:tilt-bc71b0b17aea3fb5" already present on machine
[event: pod node-feature-discovery/nfd-worker-5m584] Back-off restarting failed container nfd-worker in pod nfd-worker-5m584_node-feature-discovery(784862c3-3fbe-445a-8cc2-e7bee0675246)
Detected container restart. Pod: nfd-worker-5m584. Container: nfd-worker.
```

But if we rely on the downward API to fetch its own namespace then we can avoid leaving `w.kubernetesNamespace` empty. Perhaps there are some issues of this approach that I'm unaware of, but I'm willing to discuss it.

marquiz: @fmuyassarov `w.kubernetesNamespace` is initialized with `utils.GetKubernetesNamespace()`. That, in turn, takes the value from `/var/run/secrets/kubernetes.io/serviceaccount/namespace` or if not found, fallbacks to `KUBERNETES_NAMESPACE` env variable. That should work. In normal deployments there should be no need to set the env variable but in tilt you might need to.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -169,12 +175,25 @@ func NewNfdWorker(opts ...NfdWorkerOption) (NfdWorker, error) {
 		stop:                make(chan struct{}),
 	}
 
+	if nfd.args.ConfigFile != "" {
```
ArangoGutierrez, line 178: This check now runs before the opts loop, but `args` is only populated by `WithArgs()` inside that loop — so `configFilePath` is always empty and `-config` is silently dropped. It needs to stay after the opts are applied (as on master), independent of the struct reassignment below that Copilot flagged.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -312,6 +331,12 @@ func (w *nfdWorker) Run() error {
 	httpMux.Handle("/metrics", promhttp.HandlerFor(promRegistry, promhttp.HandlerOpts{}))
 	registerVersion(version.Get())
 
+	klConfig, err := w.kubeletConfigFunc()
+	if err != nil {
+		return err
+	}
+	memory.SetSwapMode(klConfig.MemorySwap.SwapBehavior)
```
ArangoGutierrez, line 338: This resolves the kubelet config once for the lifetime of the worker, whereas topology-updater re-runs `kubeletConfigFunc` on interval-based updates exactly so configz changes get picked up. Since feature discovery re-runs every sleep-interval anyway, consider re-resolving per pass — otherwise a kubelet reconfig leaves a stale `behavior` label until the pod restarts.

deployment/components/common/env.yaml
```diff
@@ -13,3 +13,11 @@
       valueFrom:
         fieldRef:
           fieldPath: metadata.uid
+    - name: NODE_ADDRESS
```
ArangoGutierrez, line 16: Only the kustomize deployment gets this — the Helm `worker.yaml` template sets no `NODE_ADDRESS` (only `topologyupdater.yaml` does) and grants no `nodes/proxy`, so a Helm-deployed worker running this image exits at `parseArgs` (`os.Exit(1)`) and crashloops. The Helm chart needs the same env and RBAC updates.

ArangoGutierrez (commented):
This moves `getKubeletConfigFunc` into `pkg/utils/kubeconf` and wires configz into nfd-worker to publish kubelet swap behavior as `memory-swap.behavior`, following the earlier review direction. The constructor rework is the main blocker: the `nfd` struct reassignment Copilot flagged is still in place, and there is a second ordering bug noted inline that drops `-config` entirely. The open threads from the last round also still stand in the current diff: `POD_NAMESPACE` vs `w.kubernetesNamespace`, not failing the whole worker when the kubelet config is unreachable, and the placement/naming question for the feature. `verify-master` and `e2e-test-master` are red and the branch is `lifecycle/rotten`, so this needs a rebase; Helm parity and docs are missing, and `memory_test.go` could take a table-driven case for the behavior values (default `NoSwap`, `LimitedSwap`, unset). The feature is wanted per #2178, but this needs the constructor fixes, the open threads resolved, Helm/docs coverage, and green CI before it is mergeable.

deployment/base/rbac/worker-clusterrole.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: nfd-worker
+rules:
+- apiGroups:
+  - ""
+  resources: ["pods"]
+  verbs: ["get"]
+- apiGroups: [""]
+  resources: ["nodes/proxy"]
+  verbs: ["get"]
```
ArangoGutierrez, line 12: From the earlier thread this was meant to be `nodes/proxy` only — `pods get` is already granted by the namespaced Role, so duplicating it here widens the grant cluster-wide for no gain.
```suggestion
- apiGroups: [""]
  resources: ["nodes/proxy"]
  verbs: ["get"]
```

source/memory/memory.go
```diff
@@ -106,12 +113,16 @@ func (s *memorySource) Discover() error {
...
-	// Detect Swap
+	// Detect Swap and Swap Behavior
 	if swap, err := detectSwap(); err != nil {
 		klog.ErrorS(err, "failed to detect Swap nodes")
 	} else {
 		s.features.Attributes[SwapFeature] = nfdv1alpha1.AttributeFeatureSet{Elements: swap}
+		swap["behavior"] = defaultSwapBehavior
+		if swapBehavior != "" {
+			swap["behavior"] = swapBehavior
+		}
```
ArangoGutierrez, line 124: This runs whenever `detectSwap()` succeeds, so the `memory.swap.behavior` attribute gets published (possibly `LimitedSwap`) even on nodes where swap is disabled — a NodeFeatureRule matching on it would hit swapless nodes. Gating it on `enabled == "true"` would match the label semantics described in the PR. Also note configz reports the configured value, not the effective one (kubelet ignores `LimitedSwap` on cgroup v1) — worth a caveat when the docs land.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -169,12 +175,25 @@ func NewNfdWorker(opts ...NfdWorkerOption) (NfdWorker, error) {
...
 	}
 
-	if nfd.args.ConfigFile != "" {
-		nfd.configFilePath = filepath.Clean(nfd.args.ConfigFile)
+	kubeletConfigFunc, err := kubeconf.GetKubeletConfigFunc(nfd.args.KubeletConfigURI, nfd.args.APIAuthTokenFile)
+	if err != nil {
+		return nil, err
+	}
+
+	nfd = &nfdWorker{
```
ArangoGutierrez, line 191: **Blocker:** this reassigns `nfd` to a brand-new struct with only `kubeletConfigFunc` set, discarding the `config: &NFDConfig{}`, `kubernetesNamespace`, `stop` channel (and `configFilePath`) initialized above. After this, `w.config` is nil, `w.stop` is a **nil channel** (a receive on it blocks forever), and `w.kubernetesNamespace` is empty — so the worker nil-derefs / hangs at runtime. This should be `nfd.kubeletConfigFunc = kubeletConfigFunc` (and remove the second `opts` loop).

pkg/nfd-worker/nfd-worker.go
```diff
@@ -169,12 +175,25 @@ func NewNfdWorker(opts ...NfdWorkerOption) (NfdWorker, error) {
...
+		nfd.configFilePath = filepath.Clean(nfd.args.ConfigFile)
+	}
+
 	for _, o := range opts {
 		o.apply(nfd)
 	}
 
-	if nfd.args.ConfigFile != "" {
-		nfd.configFilePath = filepath.Clean(nfd.args.ConfigFile)
+	kubeletConfigFunc, err := kubeconf.GetKubeletConfigFunc(nfd.args.KubeletConfigURI, nfd.args.APIAuthTokenFile)
```
ArangoGutierrez, line 186: This calls `GetKubeletConfigFunc` unconditionally, so `NewNfdWorker` hard-fails whenever `KubeletConfigURI` is empty (`url.ParseRequestURI("")` errors) — e.g. unit tests building the worker via opts without setting it (likely why CI is red). It also makes the kubelet `configz` endpoint a mandatory startup dependency for *every* worker (`Run()` returns err if `kubeletConfigFunc()` fails), not just when swap-behavior is desired. Consider making it lazy/optional.

deployment/base/rbac/worker-clusterrole.yaml
```diff
@@ -0,0 +1,12 @@
...
+kind: ClusterRole
+metadata:
+  name: nfd-worker
+rules:
+- apiGroups:
+  - ""
+  resources: ["pods"]
+  verbs: ["get"]
+- apiGroups: [""]
+  resources: ["nodes/proxy"]
```
ArangoGutierrez, line 11: Security: `nodes/proxy` is a powerful grant — it permits proxying arbitrary requests to the kubelet. Adding it to the worker (a DaemonSet on every node) is a notable privilege increase / blast-radius widening. Please call this out explicitly, and consider reading kubelet config from the local file via hostPath (as topology-updater supports) to avoid needing it.

deployment/base/rbac/worker-clusterrolebinding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: nfd-worker
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: nfd-worker
+subjects:
+- kind: ServiceAccount
+  name: nfd-worker
+  namespace: default
```
ArangoGutierrez, line 12: This hardcodes `namespace: default`, but NFD deploys to `node-feature-discovery` (or a configurable ns). Unless a kustomize namespace transformer rewrites subject namespaces, this binding won't target the real `nfd-worker` SA. Should track the deployment namespace.

ArangoGutierrez (changes requested):
Thanks for this (fixes #2178) — the feature is useful, but there are a couple of blocking issues:

**Blocker — worker construction (inline):** `NewNfdWorker` reassigns `nfd` to a fresh `&nfdWorker{kubeletConfigFunc: …}`, dropping the `config`, `kubernetesNamespace`, and `stop` set just above. That leaves a nil `config` and a **nil `stop` channel** (receives block forever) → the worker nil-derefs / hangs at runtime. Almost certainly a refactor slip — should be `nfd.kubeletConfigFunc = kubeletConfigFunc` (and drop the duplicate opts loop).

**Blocker — mandatory kubelet dependency (inline):** `GetKubeletConfigFunc` is called unconditionally at construction, so `NewNfdWorker` hard-fails on empty `KubeletConfigURI` (tests / misconfig), and every worker now requires reaching the kubelet `configz` endpoint at startup. This likely explains the red `verify`/`e2e`.

**Security (inline):** granting the worker DaemonSet `nodes/proxy` is a significant privilege increase — please call it out, and consider the local-file path (hostPath) the topology-updater uses to avoid it.

**RBAC (inline):** the ClusterRoleBinding hardcodes `namespace: default`.

**Tests:** no coverage for the new `swap.behavior` label / `SetSwapMode`. **Nits:** trailing newline dropped from `worker-role.yaml` + `kustomization.yaml`; `defaultSwapBehavior` appears unused.

Marking request-changes (also `lifecycle/rotten` → needs a rebase). Happy to re-review once addressed — thanks!

PR #2193, opened 2025-07-08
tilt: use anonymous, ephemeral image registry

fmuyassarov:
Use anonymous, ephemeral image registry [ttl.sh](https://ttl.sh/) for local development.


marquiz (approved):
/lgtm

PR #2199, opened 2025-07-22
Bugfix: NFR and NFG rules processing

mfranczy:
There is an error that mostly influenced processing of matchAny. The status of the rule was not transfered correctly.


mfranczy:
fyi @ChaoyiHuang

PR #2200, opened 2025-07-23
github: set JEKYLL_GITHUB_TOKEN env

marquiz:
This hopefully rate limiting warnings and errors in building the documentation.

PR #2201, opened 2025-07-23
Makefile: pass JEKYLL_GITHUB_TOKEN env to doc builder container

marquiz:
Missed this one in #2200

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ JEKYLL_ENV ?= development
 SITE_BUILD_CMD := $(CONTAINER_RUN_CMD) --rm -i -u "`id -u`:`id -g`" \
 	$(shell [ -t 0 ] && echo '-t') \
 	-e JEKYLL_ENV=$(JEKYLL_ENV) \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -21,6 +21,7 @@ JEKYLL_ENV ?= development
 SITE_BUILD_CMD := $(CONTAINER_RUN_CMD) --rm -i -u "`id -u`:`id -g`" \
 	$(shell [ -t 0 ] && echo '-t') \
 	-e JEKYLL_ENV=$(JEKYLL_ENV) \
+	-e JEKYLL_GITHUB_TOKEN="$$JEKYLL_GITHUB_TOKEN" \
```
fmuyassarov, line 24:
```suggestion
	-e JEKYLL_GITHUB_TOKEN="$JEKYLL_GITHUB_TOKEN" \
```

marquiz: This is passed to the make target (to be executed). It takes it from the env (not from a makefile variable) so we need the double  symbol

fmuyassarov: Do'h, makes sense.

PR #2255, opened 2025-08-05
nfd-worker: Add FeatureAllowList, FeatureDenyList

### LINKED ISSUE #2026: Ability to white list feature attributes on the NodeFeature object

Add logic to filter the "features" that gets added to the nodefeature object.
https://github.com/kubernetes-sigs/node-feature-discovery/blob/d7257a9bc95238c52e74121a1feb95331d52ca53/pkg/nfd-worker/nfd-worker.go#L668

**Why is this needed**:

The filter logic currently available, filters the labels that gets added to the node. But, the nodefeature object is still big, with all the feature attributes added to the NodeFeature.
:
For example, we added labelSource and label white lists to get the feature flags for these PCI devices as below:
 
  labels:
    feature.node.kubernetes.io/pci-0300_1a03.present: "true"
 
But, the pci devices and attributes in the NodeFetaure is quite large and lists hundreds of devices we don't care about. We wanted to support this feature.

Another example:

Obviously the ability to filter would be different for each feature. As another example, the kernel feature is also quite large (several thousand lines), when we may only care about one or two flags for labels - kernel version in our case.
 
This means we need to enable kernel feature, which is very large as well.

ozhuraki:
Closes: #2026

### DIFF
--- a/cmd/nfd-master/main.go
+++ b/cmd/nfd-master/main.go
@@ -65,6 +65,10 @@ func main() {
 			args.Overrides.DenyLabelNs = overrides.DenyLabelNs
 		case "label-whitelist":
 			args.Overrides.LabelWhiteList = overrides.LabelWhiteList
+		case "feature-allowlist":
+			args.Overrides.FeatureAllowList = overrides.FeatureAllowList
+		case "feature-denylist":
+			args.Overrides.FeatureDenyList = overrides.FeatureDenyList
 		case "enable-taints":
 			args.Overrides.EnableTaints = overrides.EnableTaints
 		case "no-publish":
@@ -121,16 +125,22 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
 	args.Klog = klogutils.InitKlogFlags(flagset)
 
 	overrides := &master.ConfigOverrideArgs{
-		LabelWhiteList: &utils.RegexpVal{},
-		DenyLabelNs:    &utils.StringSetVal{},
-		ExtraLabelNs:   &utils.StringSetVal{},
-		ResyncPeriod:   &utils.DurationVal{Duration: time.Duration(1) * time.Hour},
+		LabelWhiteList:  &utils.RegexpVal{},
+		FeatureAlowList: &utils.RegexpVal{},
+		FeatureDenyList: &utils.RegexpVal{},
+		DenyLabelNs:     &utils.StringSetVal{},
+		ExtraLabelNs:    &utils.StringSetVal{},
+		ResyncPeriod:    &utils.DurationVal{Duration: time.Duration(1) * time.Hour},
 	}
 	flagset.Var(overrides.ExtraLabelNs, "extra-label-ns",
 		"Comma separated list of allowed extra label namespaces")
 	flagset.Var(overrides.LabelWhiteList, "label-whitelist",
 		"Regular expression to filter label names to publish to the Kubernetes API server. "+
 			"NB: the label namespace is omitted i.e. the filter is only applied to the name part after '/'.")
+	flagset.Var(overrides.FeatureAllowList, "feature-allowlist",
+		"Regular expression to filter feature names to publish to the Kubernetes API server")
+	flagset.Var(overrides.FeatureDenyList, "feature-denylist",
+		"Regular expression to filter out feature names")
 	overrides.EnableTaints = flagset.Bool("enable-taints", false,
 		"Enable node tainting feature")
 	overrides.NoPublish = flagset.Bool("no-publish", false,
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -77,14 +77,16 @@ type NFDConfig struct {
 }
 
 type coreConfig struct {
-	Klog           klogutils.KlogConfigOpts
-	LabelWhiteList utils.RegexpVal
-	NoPublish      bool
-	NoOwnerRefs    bool
-	FeatureSources []string
-	Sources        *[]string
-	LabelSources   []string
-	SleepInterval  utils.DurationVal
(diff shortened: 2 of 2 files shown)


marquiz (commented):
Thanks @ozhuraki for starting to work on this.

I have two main comments:

1. The main motivation for this feature is decreasing the size of the NodeFeature object(s). We need filtering for the "raw" features that get into those CRs. For the labels we have adequate controls (I think).
2. We need some more expressive mechanism than one-regexp-to-rule-them-all. Like *Only show cpu.cpuid.AVX512\* and cpu.cpuid.AMX but nothing else* or *Filter out all PCI devices with class 08*

cmd/nfd-master/main.go
```diff
@@ -121,16 +125,22 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
...
 	}
 	flagset.Var(overrides.ExtraLabelNs, "extra-label-ns",
 		"Comma separated list of allowed extra label namespaces")
 	flagset.Var(overrides.LabelWhiteList, "label-whitelist",
 		"Regular expression to filter label names to publish to the Kubernetes API server. "+
 			"NB: the label namespace is omitted i.e. the filter is only applied to the name part after '/'.")
+	flagset.Var(overrides.FeatureAllowList, "feature-allowlist",
+		"Regular expression to filter feature names to publish to the Kubernetes API server")
+	flagset.Var(overrides.FeatureDenyList, "feature-denylist",
+		"Regular expression to filter out feature names")
```
marquiz, line 143: We need something more structured than plain regexp. I'd suggest to drop the command line flags and make this a config-file only option.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -77,14 +77,16 @@ type NFDConfig struct {
...
-	NoPublish      bool
-	NoOwnerRefs    bool
-	FeatureSources []string
-	Sources        *[]string
-	LabelSources   []string
-	SleepInterval  utils.DurationVal
+	Klog             klogutils.KlogConfigOpts
+	LabelWhiteList   utils.RegexpVal
+	FeatureAllowList utils.RegexpVal
+	FeatureDenyList  utils.RegexpVal
```
marquiz, line 83: We need something more structured. One idea could be to follow something we have in NodeFeatureRules. We have stuff organized in hierarchy like `<source>.<feature>.<element>`. So following the NodeFeatureRule model we could have

```yaml
featureDenyList:
  - feature: "kernel.config"
```

and then more refined configuration

```yaml
featureDenyList:
  - feature: "pci.device"
    class: {op: In, value: ["0880"]}
```

pkg/nfd-worker/nfd-worker.go
```diff
@@ -564,6 +568,21 @@ func GetFeatureLabels(source source.LabelSource, labelWhiteList regexp.Regexp) (
 
 	for k, v := range features {
 		name := k
+
+		allowed := featureAllowList.MatchString(name)
+		if !allowed {
+			klog.InfoS("feature does not match the allowlist", "feature", name, "regexp", featureAllowList.String())
+		}
+
+		denied := featureDenyList.MatchString(name)
```
ArangoGutierrez, line 577: An empty pattern matches every string, and `newDefaultConfig()` seeds `FeatureDenyList` with `regexp.MustCompile("")` — so with default config every feature is denied and the worker publishes no labels at all. Empty-matches-all is exactly why the allowlist default works, but it inverts for a denylist; gate the deny check on a non-empty pattern:
```suggestion
		denied := featureDenyList.String() != "" && featureDenyList.MatchString(name)
```

cmd/nfd-master/main.go
```diff
@@ -121,16 +125,22 @@ func initFlags(flagset *flag.FlagSet) (*master.Args, *master.ConfigOverrideArgs)
...
+		DenyLabelNs:     &utils.StringSetVal{},
+		ExtraLabelNs:    &utils.StringSetVal{},
+		ResyncPeriod:    &utils.DurationVal{Duration: time.Duration(1) * time.Hour},
 	}
 	flagset.Var(overrides.ExtraLabelNs, "extra-label-ns",
 		"Comma separated list of allowed extra label namespaces")
 	flagset.Var(overrides.LabelWhiteList, "label-whitelist",
 		"Regular expression to filter label names to publish to the Kubernetes API server. "+
 			"NB: the label namespace is omitted i.e. the filter is only applied to the name part after '/'.")
+	flagset.Var(overrides.FeatureAllowList, "feature-allowlist",
```
ArangoGutierrez, line 140: These are nfd-worker core-config options, but the flags are registered in the nfd-master binary and assigned to `master.ConfigOverrideArgs` fields that don't exist (pkg/nfd-master isn't changed in this PR), so this doesn't compile. Dropping the flags per the thread below resolves it; if a flag survives, it belongs in cmd/nfd-worker against `worker.ConfigOverrideArgs`.

ArangoGutierrez (commented):
This adds regexp allow/deny filtering for the worker's feature labels. Two new issues on top of the existing threads: it doesn't compile (the flags in cmd/nfd-master assign `master.ConfigOverrideArgs` fields that don't exist — pkg/nfd-master isn't touched), and the default config publishes zero labels because the empty deny pattern matches every feature name (see inline). @marquiz's design questions from September are still unanswered — filter the raw features that land in the NodeFeature object rather than labels, structured matchers instead of a single regexp, config-file only — and those decide the shape here, so I'd settle them before polishing further. Whatever the final shape, it still needs unit tests (deny-over-allow precedence, both-empty default), docs/example-config updates, and a decision on which name form the patterns match (`labelWhiteList` matches the final label name; these match the raw source key). With the merge conflict, red tide, and lifecycle/rotten on top: needs a rebase and the open design feedback addressed before this can move toward merge.

PR #2263, opened 2025-08-19
scripts/test-infra: bump golangci-lint to v2.4.0

marquiz:
Also fix or silence errors from golangci-lint v2.4.0:

- unchecked return values (errcheck)
- nolint for Close() for read-only files and http servers
- unnecessary naming of embedded fields of structs (staticcheck)
- multiple imports of the same package (staticcheck)
- use switch-case instead of if/else (staticcheck)
- remove type from var declaration (staticcheck)
- remove "Yoda" condition from if (staticcheck)
- don't use dot imports

### DIFF
--- a/cmd/nfd-gc/main.go
+++ b/cmd/nfd-gc/main.go
@@ -68,7 +68,7 @@ func parseArgs(flags *flag.FlagSet, osArgs ...string) *nfdgarbagecollector.Args 
 
 	_ = flags.Parse(osArgs)
 	if len(flags.Args()) > 0 {
-		fmt.Fprintf(flags.Output(), "unknown command line argument: %s\n", flags.Args()[0])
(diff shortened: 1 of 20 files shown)


cmd/nfd-gc/main.go
```diff
@@ -68,7 +68,7 @@ func parseArgs(flags *flag.FlagSet, osArgs ...string) *nfdgarbagecollector.Args
 
 	_ = flags.Parse(osArgs)
 	if len(flags.Args()) > 0 {
-		fmt.Fprintf(flags.Output(), "unknown command line argument: %s\n", flags.Args()[0])
+		_, _ = fmt.Fprintf(flags.Output(), "unknown command line argument: %s\n", flags.Args()[0])
```
ArangoGutierrez, line 71: Sometimes I wonder, is this better than the comment `//noerrcheck` comment?

marquiz: Yeah, maybe the nolint is cleaner

PR #2280, opened 2025-09-10
deps: bump k8s.io to v1.34

marquiz:


marquiz:
v1.34.1 has been tagged but not all the sub-tags (all the subpackages) are not there yet. Let's wait for those and then bump to v1.34.1 directly.
/hold

PR #2297, opened 2025-09-25
helm: add artifacthub metadata file to OCI registry

marquiz:
See:
https://artifacthub.io/docs/topics/repositories/helm-charts/#oci-support

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,6 +184,12 @@ helm-lint:
 helm-push:
 	helm package deployment/helm/node-feature-discovery --version $(CHART_VERSION) --app-version $(IMAGE_TAG_NAME)
 	helm push node-feature-discovery-$(CHART_VERSION).tgz oci://${IMAGE_REGISTRY}/charts
+	# Push artifacthub.io metadata
+	# Ref: https://artifacthub.io/docs/topics/repositories/helm-charts/#oci-support
+	oras push \
+	    ${IMAGE_REGISTRY}/charts/node-feature-discovery:artifacthub.io \
+	    --config /dev/null:application/vnd.cncf.artifacthub.config.v1+yaml \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -184,6 +184,12 @@ helm-lint:
 helm-push:
 	helm package deployment/helm/node-feature-discovery --version $(CHART_VERSION) --app-version $(IMAGE_TAG_NAME)
 	helm push node-feature-discovery-$(CHART_VERSION).tgz oci://${IMAGE_REGISTRY}/charts
+	# Push artifacthub.io metadata
+	# Ref: https://artifacthub.io/docs/topics/repositories/helm-charts/#oci-support
+	oras push \
+	    ${IMAGE_REGISTRY}/charts/node-feature-discovery:artifacthub.io \
```
fmuyassarov, line 190: nit
```suggestion
	    ${IMAGE_REGISTRY}/charts/${IMAGE-NAME}/:artifacthub.io \
```

marquiz: I actually had that :sweat_smile:  But then realized that chart-name != image-name. We could specify a CHART_NAME variable but that wouldn't make much sense imo because there's no other charts than `deployment/helm/node-feature-discovery`. WDYT?

fmuyassarov: I think [IMAGE_NAME](https://github.com/kubernetes-sigs/node-feature-discovery/blob/86e7497dd058c9ace829a164b14750c19fa03364/Makefile#L42) is a constant and is equal to `node-feature-discovery`. It is just yet another var as IMAGE_REGISTRY. But this is just a nit. Let's keep it as it is.
/lgtm

marquiz: :+1: But as it's a variable one MAY override it and image building still works as expected. However, if we'd use that for the helm chart the helm-push would fail.

PR #2300, opened 2025-09-25
cloudbuild: fix build

marquiz:
Bump cloudbuild image to the latest version and match oras version with the golang version within.


fmuyassarov (commented):
/lgtm

PR #2302, opened 2025-09-26
Move the default values for images to Dockerfile to fix InvalidDefaultArgInFrom warnings

fmuyassarov:
Add the default values for images in Dockerfile to fix build checks warnings which might come from an empty variable.

```
 => WARN: InvalidDefaultArgInFrom: Default value for ARG ${BUILDER_IMAGE} results in empty or inval  0.0s
 => WARN: InvalidDefaultArgInFrom: Default value for ARG ${BASE_IMAGE_FULL} results in empty or inv  0.0s
 => WARN: InvalidDefaultArgInFrom: Default value for ARG ${BASE_IMAGE_MINIMAL} results in empty or   0.0s
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ ARG BASE_IMAGE_FULL
 ARG BASE_IMAGE_MINIMAL
 
 # Build node feature discovery
-FROM ${BUILDER_IMAGE} AS builder
+FROM ${BUILDER_IMAGE:-golang} AS builder
 
 # Get (cache) deps in a separate layer
 COPY go.mod go.sum /go/node-feature-discovery/
@@ -23,7 +23,7 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
     make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
 
 # Create full variant of the production image
-FROM ${BASE_IMAGE_FULL} AS full
+FROM ${BASE_IMAGE_FULL:-debian:stable-slim} AS full
 
 # Run as unprivileged user
 USER 65534:65534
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -8,9 +8,9 @@ IMAGE_BUILD_CMD ?= docker build
 IMAGE_BUILD_EXTRA_OPTS ?=
 IMAGE_PUSH_CMD ?= docker push
 CONTAINER_RUN_CMD ?= docker run
-BUILDER_IMAGE ?= golang:1.25-trixie
-BASE_IMAGE_FULL ?= debian:bookworm-slim
-BASE_IMAGE_MINIMAL ?= scratch
+BUILDER_IMAGE ?=
```
marquiz: This we'd need to leave as is because `hack/prepare-release.sh` breaks otherwise

Dockerfile
```diff
@@ -3,7 +3,7 @@ ARG BASE_IMAGE_FULL
 ARG BASE_IMAGE_MINIMAL
 
 # Build node feature discovery
-FROM ${BUILDER_IMAGE} AS builder
+FROM ${BUILDER_IMAGE:-golang:1.25-trixie} AS builder
```
marquiz:
```suggestion
FROM ${BUILDER_IMAGE:-golang} AS builder
```

Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
     make install VERSION=$VERSION HOSTMOUNT_PREFIX=$HOSTMOUNT_PREFIX
 
 # Create full variant of the production image
-FROM ${BASE_IMAGE_FULL} AS full
+FROM ${BASE_IMAGE_FULL:-debian:bookworm-slim} AS full
```
marquiz:
```suggestion
FROM ${BASE_IMAGE_FULL:-debian:stable-slim} AS full
```

marquiz (commented):
@fmuyassarov thanks for the enhancement. Good to get rid of those annoying warnings.

However, I suggest to keep the Makefile unmodified. Also, put generic defaults in the Dockerfile so that we never need to modify the dockerfile just to bump the base images. WDYT?

fmuyassarov:
> @fmuyassarov thanks for the enhancement. Good to get rid of those annoying warnings.
> 
> However, I suggest to keep the Makefile unmodified. Also, put generic defaults in the Dockerfile so that we never need to modify the dockerfile just to bump the base images. WDYT?

Thanks for the review @marquiz.
Sounds good. Updated accordingly.

PR #2303, opened 2025-09-26
scripts/test-infra/push-image: fix env variable parsing

marquiz:
The _GIT_TAG variable is always specified (set by the prow job). We can determine whether we're building a tag or branch from the _PULL_BASE_REF.


fmuyassarov (commented):
/lgtm

PR #2304, opened 2025-09-27
cloudbuild: export _PULL_BASE_REF as an env variable

marquiz:
Needed by the latest version of push-image.sh script.


fmuyassarov (commented):
/lgtm

PR #2305, opened 2025-09-29
docs: document export sub-command in the nfd reference

marquiz:


vsoch (approved):
LGTM. I don't see where it shows up in the preview (I searched for "export" and no result), so if someone wants to throw up a link it would be helpful.

marquiz:
> I don't see where it shows up in the preview (I searched for "export" and no result)

Indeed, the netlify previews are broken :man_facepalming: Try to fix that with #2306

fmuyassarov:
@marquiz can we re-trigger the netlify to rebuild the preview after your last fix?
Also, maybe it is worth of mentioning this changes in the release notes for v0.18? WDYT?

marquiz:
> can we re-trigger the netlify to rebuild the preview after your last fix?

Rebased so fresh netlify preview

> Also, maybe it is worth of mentioning this changes in the release notes for v0.18? WDYT?

Yes, iirc I mentioned that in the image compatibility section but maybe we could be more verbose there

docs/reference/node-feature-client-reference.md
```diff
@@ -63,3 +63,23 @@ The `--registry-password-stdin` flag enables reading of registry password from s
 #### --registry-token-stdin
 
 The `--registry-token-stdin` flag enables reading of registry token from stdin.
+
+## export
+
+Commands for exporting node features locally.
```
fmuyassarov: nit:
```suggestion
Commands for exporting node features to the terminal.
```

docs/reference/node-feature-client-reference.md
```diff
@@ -63,3 +63,23 @@ The `--registry-password-stdin` flag enables reading of registry password from s
...
+
+### features
+
+Export features of the system in JSON format.
```
fmuyassarov:
```suggestion
Export features of the system in JSON format to the terminal.
```

docs/reference/node-feature-client-reference.md
marquiz: That's slightly imprecise imo, I used "stdout or a file" instead. WDYT?

fmuyassarov: I think that's better

mfranczy:
Nice!

PR #2309, opened 2025-10-02
helm: fixes

marquiz:
A bunch of fixes to the Helm chart:

1. Drop unused master.featureApi from values.yaml
2. Drop featureRulesController parameter
    The nfd-master does not have such commandline flag as `-featurerules-controller` anymore.
3. Fix master.extraEnvs
4. Fix PodDisruptionBudget


fmuyassarov:
@marquiz seems we need a manual cherry-picking :) 
/lgtm

PR #2311, opened 2025-10-02
helm: add validation schema

marquiz:
Add auto-generated values schema for the Helm chart. Utilizes the https://github.com/losisin/helm-values-schema-json.git Helm plugin and annotations in the values file to generate the schema. The PR also adds CI step to check that the schema is in sync with the values file.


Makefile
```diff
@@ -179,6 +179,12 @@ mdlint:
 	ruby:slim \
 	/workdir/scripts/test-infra/mdlint.sh
 
+.PHONY: helm-schema
+helm-schema:
+	# Requires helm plugin install https://github.com/losisin/helm-values-schema-json.git
+	cd deployment/helm/node-feature-discovery/ && \
+	helm schema
+
```
fmuyassarov, line 191: Do you think it is a bad idea to actually install the plugin here directly so that we allow local validations as well?

fmuyassarov (commented):
Looks good, just one question

Makefile
marquiz: I was thinking about this but then thought I didn't want to unconditionally mess up with people's system.

Alternatives:
- check if the plugin is installed and print instructions
- wonder if there's a possibility to install/run the plugin locally

WDYT?

fmuyassarov: `check if the plugin is installed and print instructions`... sounds good

marquiz: > `check if the plugin is installed and print instructions`... sounds good

Thanks, done

marquiz:
Update: rebased and added a check that the helm schema plugin exists

marquiz:
Some fixes still on way
/hold

marquiz:
Corrected the startup/liveness/readiness probe config and validation

fmuyassarov (commented):
Good improvement @marquiz 
/lgtm

Issue #2258, opened 2025-08-12
Drop obsolete lableSelector for `node-role.kubernetes.io/master`


jcpunk:
**What would you like to be added**:

The control plane label `node-role.kubernetes.io/master` has been deprecated for a few years. Folks running clusters without the modern label `node-role.kubernetes.io/control-plane` are hopefully zero due to the number of security fixes.

The existing deployments have both the old and the new label.

**Why is this needed**:
Various linters and other tools flag the obsolete labelSelector on application of the mainfests.  For example:
```
node-feature-discovery/Deployment/node-feature-discovery-master: spec.template.spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution[0].preference.matchExpressions[0].key: node-role.kubernetes.io/master is use "node-role.kubernetes.io/control-plane" instead
```

Since the deployment already has the modern label, this warning doesn't result in an action item.

This is present in helm chart version `0.17.3`

marquiz:
Fixed by #2070, released in [v0.18](https://github.com/kubernetes-sigs/node-feature-discovery/releases/tag/v0.18.0). Closing the issue

/close

PR #2321, opened 2025-10-03
helm: Drop obsolete lableSelector for node-role.kubernetes.io/master

ozhuraki:
Closes: #2258


marquiz:
Fixed in #2070. Let's not change this for the v0.17 to avoid breaking any users of that release branch
/close

PR #2324, opened 2025-10-06
helm: improve chart readme

marquiz:
1. Auto-generate chart readme
    Use github.com/norwoodj/helm-docs to auto-generate chart readme from the values.yaml file.
1. Makefile: add helm-docs target
1. scripts/test-infra: add step to verify Helm README
1. hack/prepare-release.sh: update Helm README

### DIFF
--- a/hack/prepare-release.sh
+++ b/hack/prepare-release.sh
@@ -138,9 +138,15 @@ if [ -z "$no_patching" ]; then
         -i deployment/helm/node-feature-discovery/Chart.yaml
     sed -e s"/pullPolicy:.*/pullPolicy: IfNotPresent/" \
         -e s"!gcr.io/k8s-staging-nfd/node-feature-discovery!registry.k8s.io/nfd/node-feature-discovery!" \
+        -e s"!kubernetes-sigs.github.io/node-feature-discovery/master!kubernetes-sigs.github.io/node-feature-discovery/$docs_version!" \
         -i deployment/helm/node-feature-discovery/values.yaml
     sed -e s"!kubernetes-sigs.github.io/node-feature-discovery/master!kubernetes-sigs.github.io/node-feature-discovery/$docs_version!" \
         -i deployment/helm/node-feature-discovery/README.md
+    sed -e s"!.*metadata\.oci_repo.*!{{- define \"metadata.oci_repo\" -}}oci://registry.k8s.io/nfd/charts/node-feature-discovery!" \
+        -e s"/.*metadata\.oci_tag.*/{{- define \"metadata.oci_tag\" -}}$semver/" \
+        -e s"/.*metadata\.docs_version.*/{{- define \"metadata.docs_version\" -}}$docs_version/" \
+        -i deployment/helm/node-feature-discovery/_metadata.gotmpl
+    make helm-docs
 
     # Patch e2e test
     echo Patching test/e2e/node_feature_discovery.go flag defaults to registry.k8s.io/nfd/node-feature-discovery and $release


Makefile
```diff
@@ -189,6 +189,17 @@ helm-schema:
 	cd deployment/helm/node-feature-discovery/ && \
 	helm schema
 
+.PHONY: helm-docs
+helm-docs:
+	@if ! command -v helm-docs >/dev/null; then \
+	    echo "ERROR: helm-docs command not found! Please install the plugin with:"; \
+	    echo "       go install github.com/norwoodj/helm-docs/cmd/helm-docs"; \
```
fmuyassarov:
```suggestion
	    echo "       go install github.com/norwoodj/helm-docs/cmd/helm-docs@1.14.2"; \
```
otherwise this will fail...

hack/prepare-release.sh
```diff
@@ -138,9 +138,15 @@ if [ -z "$no_patching" ]; then
...
         -e s"!gcr.io/k8s-staging-nfd/node-feature-discovery!registry.k8s.io/nfd/node-feature-discovery!" \
+        -e s"!kubernetes-sigs.github.io/node-feature-discovery/master!kubernetes-sigs.github.io/node-feature-discovery/$docs_version!" \
         -i deployment/helm/node-feature-discovery/values.yaml
     sed -e s"!kubernetes-sigs.github.io/node-feature-discovery/master!kubernetes-sigs.github.io/node-feature-discovery/$docs_version!" \
         -i deployment/helm/node-feature-discovery/README.md
+    sed -e s"!.*metadata\.oci_repo.*!{{- define \"metadata.oci_repo\" -}}oci://registry.k8s.io/nfd/charts/node-feature-discovery!" \
+        -e s"/.*metadata\.oci_tag.*/{{- define \"metadata.oci_tag\" -}}$semver/" \
+        -e s"/.*metadata\.docs_version.*/{{- define \"metadata.docs_version\" -}}$docs_version/" \
+        -i deployment/helm/node-feature-discovery/_metadata.gotmpl
+    make helm-docs
```
fmuyassarov, line 149: If the docs isn't in snyc with the values this will fail the release right? Do we want that?

fmuyassarov (commented):
Not directly related to this but part of this as well. Now that we have three targets related to Helm:
```
helm-schema
helm-docs
helm-lint
```
I wonder if we should have a single target that wraps all of these, so there’s less risk of the PR author forgetting to run a specific target and having to go back and forth between failed CI results and local fixes.

Makefile
marquiz: Thanks. I now listed helm docs in the `tools` section of go mod so we got rid of this (and the install in the test-infra/verify script).

hack/prepare-release.sh
marquiz: The make helm-docs here ensures that the docs (I assume you refer to docs as the Helm chart README) are updated/in-sync. This script makes changes that the user then needs to commit and submit as a PR.

marquiz:
> I wonder if we should have a single target that wraps all of these, so there’s less risk of the PR author forgetting to run a specific target and having to go back and forth between failed CI results and local fixes.

Good point, makes sense. What if I address this in a separate PR?

fmuyassarov:
> > I wonder if we should have a single target that wraps all of these, so there’s less risk of the PR author forgetting to run a specific target and having to go back and forth between failed CI results and local fixes.
> 
> Good point, makes sense. What if I address this in a separate PR?

That's even better I think. +1

hack/prepare-release.sh
fmuyassarov: I guess this is okay as it is, assuming that now we have this `helm-docs` target call as part of the mandatory CI check. 
But what I was referring to was that, in case if there a mismatch (in the Helm docs) in the HEAD of the repo and we start doing a release, the release process will fail because this script is run when we try to cut a new release and at that point it is too late to ask the merged PR author for the fix. Of course, we can fix it ourselves, and more than that I don't think we will run into that kind of situation since this target is now part of the `scripts/test-infra/verify.sh` too.

marquiz: The thing is, this script is run manually before the release. It put's everything in place so that references (links, tags etc) in the docs and deployment files point to the correct release. This is not run automatically (maybe we could change this, but that's another story...)

PR #2330, opened 2025-10-07
helm: common documentation in html docs and chart readme

marquiz:
Use the helm-docs tool to genearate the docs/deployment/helm.md file. The main benefit is that the chart values tables are autogenerated and don't need any manual maintenance, anymore.

This patch also updates the chart readme, syncing the content with docs/deployment/helm.md so that they have essentially the same content.


marquiz:
Rebased

fmuyassarov (commented):
/lgtm

ArangoGutierrez (approved):
/lgtm

PR #2331, opened 2025-10-07
Makefile: add helm-sync target

marquiz:
One helm target to rule them all.


marquiz:
Now we can start bike shedding about the proper name of the makefile target :sweat_smile: I'm not super happy with the `helm-sync` but that was a quick initial thought. 

/assign @fmuyassarov @ArangoGutierrez

marquiz:
I was also thinking helm-update and helm-generate but there's also the helm lint. Maybe helm-generate would still be better :thinking:

marquiz:
> I'm fine with` helm-sync`, another option is `helm-generate`.

Changed to `helm-generate`

/retitle Makefile: add helm-generate target

PR #2339, opened 2025-10-13
source/kernel: Add module whitelist

ozhuraki:


marquiz (commented):
Hi @ozhuraki what's the context/motivation for this? I'd much rather see a generic mechanism (á la #2255) for enabling/disabling features than rather unwieldy per-feature regexps.

source/kernel/kernel.go
```diff
@@ -90,6 +94,11 @@ func (s *kernelSource) SetConfig(conf source.Config) {
 	default:
 		panic(fmt.Sprintf("invalid config type: %T", conf))
 	}
+
+	r, err := regexp.Compile(s.config.ModuleWhiteList)
+	if err == nil {
+		s.moduleWhiteList = r
+	}
```
ArangoGutierrez, line 101: If the regexp doesn't compile, the error is dropped and `s.moduleWhiteList` stays nil, so the next `Discover()` panics with a nil-pointer dereference in `getLoadedModules()` — a config typo like `nvidia(` crashes the worker. On config reload a bad value silently keeps the previous regexp instead. Never leave the field nil:

```suggestion
	r, err := regexp.Compile(s.config.ModuleWhiteList)
	if err != nil {
		klog.ErrorS(err, "invalid moduleWhiteList, falling back to matching all modules", "moduleWhiteList", s.config.ModuleWhiteList)
		r = regexp.MustCompile(".*")
	}
	s.moduleWhiteList = r
```

Alternatively a custom type with `UnmarshalJSON` (like `utils.DurationVal`) would reject the bad value at config-parse time through the existing error path.

source/kernel/modules.go
```diff
@@ -27,7 +27,7 @@ import (
 
 const kmodProcfsPath = "/proc/modules"
 
-func getLoadedModules() ([]string, error) {
+func getLoadedModules(s *kernelSource) ([]string, error) {
```
ArangoGutierrez, line 30: Passing the whole `*kernelSource` here couples the file parsers to the source struct just to reach one field. Passing the `*regexp.Regexp` itself (`getLoadedModules(allow *regexp.Regexp)`) keeps them self-contained and makes a table-driven test for the filter (hit/miss/default `.*`) easy to add — there is currently no test exercising the new behavior.

ArangoGutierrez (commented):
This adds a kernel-source config regexp that filters which modules end up in the `kernel.loadedmodule`/`kernel.enabledmodule` features, defaulting to `.*` so existing deployments see no change. The design question @marquiz raised in October is still unanswered and remains the main blocker; one more data point for it: filtering at `Discover()` time changes what NodeFeatureRules/custom rules can match, not just which labels get emitted, which strengthens the case for a generic mechanism over per-feature regexps. On the code itself there is one real bug: an invalid regexp leaves the compiled matcher nil and the worker panics on the next discovery pass (see inline). If this goes ahead it also needs docs (worker-configuration-reference.md, nfd-worker.conf.example, helm examples), a test for the filter, and a PR description stating the motivation. Not mergeable as-is: please reply to the open design question first — the PR is `lifecycle/rotten` and prow never ran (`needs-ok-to-test`), so it will auto-close soon without a `/remove-lifecycle rotten`.

source/kernel/kernel.go
```diff
@@ -40,14 +41,16 @@ const (
 
 // Configuration file options
 type Config struct {
-	KconfigFile string
-	ConfigOpts  []string `json:"configOpts,omitempty"`
+	KconfigFile     string
+	ConfigOpts      []string `json:"configOpts,omitempty"`
+	ModuleWhiteList string
```
ArangoGutierrez, line 46: `ConfigOpts` above carries an explicit json tag — add `json:"moduleWhiteList,omitempty"` for consistency with the documented camelCase config surface. Also, since this is a new option, `moduleAllowlist` would match current Kubernetes naming guidance; `labelWhiteList`/`deviceClassWhitelist` are legacy names we are stuck with, no need to add another.

source/kernel/kernel.go
```diff
@@ -90,6 +94,11 @@ func (s *kernelSource) SetConfig(conf source.Config) {
 	default:
 		panic(fmt.Sprintf("invalid config type: %T", conf))
 	}
+
+	r, err := regexp.Compile(s.config.ModuleWhiteList)
```
ArangoGutierrez, line 98: **Bug:** if `regexp.Compile` fails (an invalid user-supplied regex), the error is silently swallowed and `s.moduleWhiteList` stays nil — then `MatchString` nil-derefs and panics the worker at discovery. Please `klog.ErrorS` the error and fall back to a match-all regex (or reject the config). Same nil risk if `Discover()` ever runs before `SetConfig`.

source/kernel/modules.go
```diff
@@ -40,13 +40,20 @@ func getLoadedModules() ([]string, error) {
 		if len(line) == 0 {
 			continue
 		}
+
+		mod := strings.Fields(line)[0]
+		// Skip if module doesn't match ModuleWhiteList
+		if !s.moduleWhiteList.MatchString(mod) {
```
ArangoGutierrez, line 46: This nil-derefs if `s.moduleWhiteList` was never compiled (the invalid-regex path above, or `Discover` before `SetConfig`). Guard for nil or guarantee a compiled default.

source/kernel/kernel.go
ArangoGutierrez, line 46: Inclusive naming: please rename `ModuleWhiteList` → `ModuleAllowList` (and the flag/config key + the `moduleWhiteList` field) — k8s guidance has moved away from 'whitelist'.

ArangoGutierrez (changes requested):
Thanks — the allow-list idea is useful, and defaulting to `.*` keeps it backward-compatible. A few things before merge:

**Bug (inline):** in `SetConfig`, a failed `regexp.Compile` (invalid user regex) is silently swallowed, leaving `moduleWhiteList` nil → `MatchString` then nil-derefs and **panics the worker** at discovery. Please log the error and fall back to a match-all regex (and guard the nil).

**Inclusive naming:** rename `ModuleWhiteList` → `ModuleAllowList` (field, flag/config key, and `moduleWhiteList`) — k8s/NFD have moved off "whitelist" and maintainers will likely require it.

**Tests:** add a test for the filtering — it would also cover the invalid-regex/nil path.

**Docs:** document the new config option in `worker-configuration-reference.md`.

Marking request-changes (also `lifecycle/rotten` → will need a rebase). Thanks!

PR #2340, opened 2025-10-14
fix: allow helm chart to use object type in worker and master config

almaslennikov:
Having only "null" as type breaks deployment of the chart if config is not empty


almaslennikov:
@marquiz thanks for the clarification, added the annotations to values.yaml and re-ran the helm generation

PR #2359, opened 2025-11-04
images: Publish s390x and ppc64le images

fidencio:
Kata Containers has added NFD as a dependency, and we have folks from those two architectures willing to get NFD working as expected for their virt / TEE capabilities.

However, we cannot run tests with s390x / ppc64le right now as the images are not published by NFD (thus, not helping us to fully add NFD as a requirement).


marquiz (approved):
/lgtm

PR #2363, opened 2025-11-04
Increase image building timeout

marquiz:
Enabling new archs (ppc64le and s390x) increased the image build time considerably.


fmuyassarov (commented):
/lgtm

PR #2367, opened 2025-11-07
source/kernel: Add kvm availability

seungukshin:
Add support for testing if a node can use virtualization for use by kata-containers (see https://katacontainers.slack.com/archives/C879ACQ00/p1761731112286659). On AMD and Intel platforms this can be determined with CPU feature flags, but on arm64 the situation is more complex. There 2 two flags - `ID_AA64PFR0_EL1.EL2` ([doc](https://developer.arm.com/documentation/ddi0601/2025-09/AArch64-Registers/ID-AA64PFR0-EL1--AArch64-Processor-Feature-Register-0)) and `ID_AA64MMFR1_EL1.VH` ([doc](https://developer.arm.com/documentation/ddi0601/2025-09/AArch64-Registers/ID-AA64MMFR1-EL1--AArch64-Memory-Model-Feature-Register-1))

* `ID_AA64PFR0_EL1.EL2` & 0b0001 != 0 (kernel runs at EL1, hyp world at EL2 - nVHE)
* `ID_AA64MMFR1_EL1.VH` & 0b0001 != 0 (kernel is at EL2 using VHE), not visible from user space

Instead, check if `kvm` is initialized and has created `/sys/devices/virtual/misc/kvm`.

### DIFF
--- a/source/kernel/kernel.go
+++ b/source/kernel/kernel.go
@@ -36,6 +36,7 @@ const (
 	SelinuxFeature       = "selinux"
 	VersionFeature       = "version"
 	EnabledModuleFeature = "enabledmodule"
+	KvmFeature           = "kvm"
 )
 
 // Configuration file options
@@ -157,6 +158,13 @@ func (s *kernelSource) Discover() error {
 	} else {
 		s.features.Attributes[SelinuxFeature] = nfdv1alpha1.NewAttributeFeatures(nil)
 		s.features.Attributes[SelinuxFeature].Elements["enabled"] = strconv.FormatBool(selinux)
+	}
+
+	if kvm, err := KvmEnabled(); err != nil {
+		klog.ErrorS(err, "failed to detect kvm status")
+	} else {
+		s.features.Attributes[KvmFeature] = nfdv1alpha1.NewAttributeFeatures(nil)
+		s.features.Attributes[KvmFeature].Elements["enabled"] = strconv.FormatBool(kvm)
 	}
 
 	klog.V(3).InfoS("discovered features", "featureSource", s.Name(), "features", utils.DelayedDumper(s.features))
--- a/deployment/base/nfd-crds/cr-sample.yaml
+++ b/deployment/base/nfd-crds/cr-sample.yaml
@@ -47,6 +47,9 @@ spec:
         - feature: kernel.selinux
           matchExpressions:
             enabled: {op: IsFalse}
+        - feature: kernel.kvm
+          matchExpressions:
+            enabled: {op: IsFalse}
         - feature: kernel.version
           matchExpressions:
             major: {op: In, value: ["5"]}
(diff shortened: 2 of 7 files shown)


source/kernel/kernel.go
```diff
@@ -114,6 +115,10 @@ func (s *kernelSource) GetLabels() (source.FeatureLabels, error) {
 		labels["selinux.enabled"] = "true"
 	}
 
+	if enabled, ok := features.Attributes[KvmFeature].Elements["enabled"]; ok && enabled == "true" {
+		labels["kvm.enabled"] = "true"
```
marquiz: Do we need/want the default node label?

marquiz (commented):
Thank you @seungukshin for the contribution. Could you sign the CLA?

seungukshin:
> Thank you @seungukshin for the contribution. Could you sign the CLA?

Thank you @marquiz and sorry for the delay.
I'm working on CLA from my company side, and probably I could proceed this this week.

source/kernel/kernel.go
seungukshin: Thank you for looking into this @marquiz .
My thought is if there is no label, then consider it as not supported.
Do you think default value with false would better?

marquiz: No. My question was about mitigating the "node label clutter", i.e. do we want/need this to be always created (when kvm is supported) or would the NodeFeature be enough)

seungukshin: Thank you for clarification, @marquiz .
I think the NodeFeature would be enough because virtualization is not for all use cases. I think usual use cases would be using pods which does not need virtualization.
But, kata-containers / CoCo (confidential containers) need to figure out virtualization functionality on nodes because it uses virtual machine.

marquiz: Thanks. Then I'd suggest removing it from the `labels` (and corresponding documentation). Can be easily added later (if it's seen essential) but harder to remove after the fact.

marquiz:
> I'm working on CLA from my company side, and probably I could proceed this this week.

No problem

source/kernel/kernel.go
seungukshin: Thank you for reviewing this change @marquiz .
I removed the labels part so it does not create a label by default.
But I kept the documentation changes because they are only for feature explanation.

seungukshin: My bad. Actually `features.md` describes built-in labels. so I removed kvm form that doc.

marquiz (approved):
Thank you @seungukshin, looks good to me now.

One ask: could you rebase so that we get rid of the CI netlify errors?`

/assign @fmuyassarov @ArangoGutierrez

seungukshin:
Thank you @marquiz for your review.

> One ask: could you rebase so that we get rid of the CI netlify errors?`

I forgot syncing my forked repo. I just rebased it.

marquiz:
ping @fmuyassarov @ArangoGutierrez

seungukshin:
just rebase on the latest master branch.
@fmuyassarov @ArangoGutierrez , could you take a look?

PR #2388, opened 2025-12-07
feat(cpu): detect hypervisor on x86_64 and arm64 via CPUID

### LINKED ISSUE #2267: [FEATURE] list virtualization type

Labels covering what, if any, virtualization platform the nodes running on.  For example, `kvm` or `hyperv`.  It would also be handy to note hosts on bare metal.

If a host lies and says it is one thing when it is actually another, I wouldn't consider that a bug in nfd.

**Why is this needed**:

When looking at host specific details, it is helpful to know what sort of runtime platform is involved.

whatever125:
Added detection of the virtualization platform on x86/arm64 archs using klauspost/cpuid lib.

- Preserved existing s390x implementation using `/proc/sysinfo`.
- On x86 and arm64 `cpu.model.hypervisor` now reports hypervisor type, or `none` for bare-metal, and raw CPUID string for unknown hypervisors.

Resolves #2267

### DIFF
--- a/source/cpu/cpu.go
+++ b/source/cpu/cpu.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -278,17 +279,12 @@ func getCPUModel() map[string]string {
 // and on x86_64/arm64 using the cpuid library.
 // Returns normalized hypervisor string, "unknown" for unknown vendor, or "none" for bare metal.
 func getHypervisor() (string, error) {
-	// try using /proc/sysinfo for s390x
-	if _, err := os.Stat("/proc/sysinfo"); err == nil {
-		hv, discovered := getHypervisorFromProcSysinfo()
-		if discovered {
-			return hv, nil
-		}
-	} else {
-		klog.ErrorS(err, "failed to stat /proc/sysinfo")
+	// use /proc/sysinfo for s390x
+	if runtime.GOARCH == "s390x" {
+		return getHypervisorFromProcSysinfo()
 	}
 
-	// fallback to cpuid
+	// use cpuid lib for x86/arm
 	if cpuid.CPU.VM() {
 		return getHypervisorFromCPUID()
 	}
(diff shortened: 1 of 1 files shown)


source/cpu/cpu.go
```diff
@@ -274,33 +274,78 @@ func getCPUModel() map[string]string {
...
-	for _, line := range strings.Split(string(data), "\n") {
-		if strings.Contains(line, "Control Program:") {
-			parts := strings.SplitN(line, ":", 2)
-			if len(parts) == 2 {
-				hypervisor = strings.TrimSpace(parts[1])
+	// use cpuid lib for x86/arm
+	if cpuid.CPU.VM() {
+		hv := strings.ToLower(cpuid.CPU.HypervisorVendorID.String())
+	
+		switch hv {
```
marquiz:
```suggestion
		switch hv := strings.ToLower(cpuid.CPU.HypervisorVendorID.String()); hv {
```

source/cpu/cpu.go
```diff
@@ -274,33 +274,78 @@ func getCPUModel() map[string]string {
...
+		case "qemu":
+			return "qemu", nil
+		case "apple":
+			return "apple", nil
+		case "qnx":
+			return "qnx", nil
+		case "acrn":
+			return "acrn", nil
+		case "sre":
+			return "sre", nil
```
marquiz: Couldn't we put all these cases where we don't mangle the value under the default case?


```suggestion
		default:
			return hv, nil
```

source/cpu/cpu.go
```diff
@@ -274,33 +274,78 @@ func getCPUModel() map[string]string {
...
+// and on x86_64/arm64 using the cpuid library.
+// Returns normalized hypervisor string or "none" for bare metal.
+// Returns "", nil only when hypervisor information is unavailable on s390x
 func getHypervisor() (string, error) {
-	if _, err := os.Stat("/proc/sysinfo"); os.IsNotExist(err) {
-		return "", nil
-	}
+	// use /proc/sysinfo for s390x
+	if _, err := os.Stat("/proc/sysinfo"); err == nil {
+		data, err := os.ReadFile("/proc/sysinfo")
```
marquiz: I wonder if it would make the code more maintainable and readable to split it into multiple functions, something like:

```suggestion
	if _, err := os.Stat("/proc/sysinfo"); err == nil {
		return getHypervisorFromProcSysinfo
	} else {
		klog.Error(err, "failed to stat /proc/sysinfo")
	}
	if cpuid.CPU.VM() {
		return getHyperVisorFromCPUID(), nil
	}
	return "none", nil
```

Also, if the stat fails, we might still want to fall back to the cpuid method (like above). Thoughts?

whatever125:
Thank you for the feedback @marquiz. I've updated the code as you suggested:
1. Split into helper functions, updated fallback logic
2. Used default in switch case. Now only `cpuid.VendorUnknown` is treated separately to get vendor from `cpuid.CPU.HypervisorVendorString`. Otherwise, the default `cpuid.CPU.HypervisorVendorID.String()` is used
3. Got rid of string handling for simplicity and maintainability

source/cpu/cpu.go
```diff
@@ -274,33 +274,65 @@ func getCPUModel() map[string]string {
...
+// Returns normalized hypervisor string, "unknown" for unknown vendor, or "none" for bare metal.
 func getHypervisor() (string, error) {
-	if _, err := os.Stat("/proc/sysinfo"); os.IsNotExist(err) {
-		return "", nil
+	// try using /proc/sysinfo for s390x
+	if _, err := os.Stat("/proc/sysinfo"); err == nil {
+		hv, discovered := getHypervisorFromProcSysinfo()
+		if discovered {
+			return hv, nil
+		}
```
marquiz: If this does not find anything (`discovered == false`), do we want to fall back to `getHypervisorFromCPUID()`? Probably not, or WDYT?

whatever125: after digging deeper, i've reconsidered the logic:
* PR/SM virtualization is always present on s390x, so it should be the default fallback value (consistent with the original code)
* `/proc/sysinfo` exists only on s390x. its absence means another arch -> go to cpuid
* not being able to read it -> just log and return the error
* cpuid doesn't support s390x, so we shouldn't fall back to it in any case
* parsing `/proc/sysinfo` and not finding "Control Program:" means PR/SM virtualization by default -> return "PR/SM", not a blank string

so this is what i propose:
```go
func getHypervisor() (string, error) {
	// use /proc/sysinfo for s390x
	if _, err := os.Stat("/proc/sysinfo"); err == nil {
		return getHypervisorFromProcSysinfo()
	}
	// use cpuid lib for x86/arm
	if cpuid.CPU.VM() {
		return getHypervisorFromCPUID()
	}
	// no hv discovered
	return "none", nil
}
```
and then the `getHypervisorFromProcSysinfo()` is just the old `getHypervisor()`.

maybe we could use `runtime.GOARCH == "s390x"` for a cleaner arch detection instead of doing stat on `/proc/sysinfo`. what do you think?

marquiz: > maybe we could use `runtime.GOARCH == "s390x"` for a cleaner arch detection instead of doing stat on `/proc/sysinfo`. what do you think?

That works :+1: Would make sense with the fact that `/proc/sysinfo` only exists on s390x

whatever125:
@marquiz thanks! i've updated the code to use `runtime.GOARCH` and implemented the logic we've discussed

PR #2402, opened 2025-12-25
Add NFD image compatibility scheduler proposal.

Xunli-Yang:
Add KEP: NFD image compatibility scheduler proposal.
### What the proposal does?
Building upon the first phase of [KEP-1845 Proposal](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/enhancements/1845-nfd-image-compatibility/README.md), which completed node compatibility validation. This proposal introduces a compatibility scheduling plugin. The compatibility scheduler plugin automatically analyzes the compatibility requirements of container images, filters suitable nodes for scheduling, and ensures that containers run on compatible nodes.
### Special notes for reviewer:
Based on the discussions on [node-feature-discovery Slack channel](https://app.slack.com/client/T09NY5SBT/C01A82FKSQK), this proposal has presented three solutions and intends to get consensus on the implementation direction.
Co-authored-by: @mfranczy @ChaoyiHuang


mfranczy (changes requested):
Thanks for the work on this proposal.

Regarding Solution 2, after reviewing it, I want to clarify that the current sqlite  design does not fully reflect what I originally had in mind when we discussed it. Because of that and the fact that I am no longer working on image compatibility I would prefer to be removed from the list of coauthors.

In my original idea, sqlite was meant to act strictly as an internal, local sink of derived node feature data, based on my earlier work and experiments around scaling to very large clusters. The intent was not to introduce a new control plane data store, but a cached database used to reduce pressure on etcd from large or frequent feature payloads.

Sqlite was intended to be used only as a precomputation engine, never queried from the scheduler hot path (which seems you do in the diagram) and not treated as part of the scheduling model itself. The expected flow was that a controller evaluates NFG offline and publishes a result (for example a bitset or membership list). The scheduler would then consume only those results via informers or pod mutation (e.g. applying node selectors), without querying NFG directly during scheduling. This is what I meant by the scheduler plugin filtering nodes purely in memory.

The diagram does not reflect what is described in the text, nor what I had in mind.

Because of this mismatch, and since we were not working on the proposal together but rather discussing ideas based on my past work on the project, I amm not comfortable being listed as a coauthor on the proposal. Please remove me.

Xunli-Yang:
@mfranczy,

Thank you for your contributions to this work, and I am sincerely sorry for the misrepresentation of your design thought.

I would like to clarify that the diagram in Solution 2 was intended to provide a complete picture of the overall workflow. The SQLite database is solely for data pre-computation and is not part of the scheduler. I apologize for the misunderstanding caused by my representation in the diagram.

Additionally, I originally considered you as the proposer of this part of the work and the ideas for Solution 2, and I am sorry for adding you as a co-author without prior communication. I fully respect and understand your decision to be removed as a co-author.

To address this, we will:

1. Revise the diagram to eliminate ambiguity and clearly indicate that SQLite is not part of the scheduler.
2. Close the current PR and create a new one to remove you from the list of co-authors. I hope this will not cause you any inconvenience.

Once again, I apologize for my oversight. Thank you for all your contributions. 

Best wishes for your work in your new focus area and Happy New Year!

Issue #2397, opened 2025-12-22
explicitly set `hostUsers` for nfd


jcpunk:
**What would you like to be added**:

Kubernetes 1.33+ offers user namespaces which further isolates users from the system and other pods on the same host.

It would be helpful to explicitly set `hostUsers: true` where host privileges are required and allow setting `hostUsers: false` on parts that can have full isolation.

In my experience with kubernetes, explicit is always better than implicit.

**Why is this needed**:

At some point `hostUsers` default setting will become user customizable. When this happens a number of folks will set this enable user namespaces by default. Explicitly setting pods that require elevated host permissions will both clearly document the required permission set and prevent unexpected "foot guns".

Note: `hostUsers: false` is incompatible with `kind`.

marquiz:
Thanks @jcpunk for the suggestion. This will be a good improvement. Thinking aloud, we might need hostUsers for the topology-updater (to access kubelet's configz endpoint) but not for others

jcpunk:
Anything that uses `hostNetwork` should probably also set `hostUsers: true`.

PR #2411, opened 2026-01-13
feat(helm): Allow control of hostUsers

jcpunk:
By default this will render nothing, so folks without the feature gate should see no changes.

Fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/2397

### DIFF
--- a/deployment/helm/node-feature-discovery/README.md
+++ b/deployment/helm/node-feature-discovery/README.md
@@ -175,6 +175,7 @@ NFD.
 | master.extraArgs | list | `[]` | Additional [command line arguments](https://kubernetes-sigs.github.io/node-feature-discovery/master/reference/master-commandline-reference) to pass to nfd-master. |
 | master.extraEnvs | list | `[]` | Additional environment variables to set in the nfd-master container. |
 | master.hostNetwork | bool | `false` | Run the container in the host's network namespace. |
+| master.hostUsers | bool | `nil` | Run the container with host user ids. NOTE: if hostNetwork is true, hostUsers should be true. |
 | master.dnsPolicy | string | `"ClusterFirstWithHostNet"` | NFD master pod [dnsPolicy](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy). |
 | master.config | string | `nil` | NFD master [configuration](https://kubernetes-sigs.github.io/node-feature-discovery/master/reference/master-configuration-reference). |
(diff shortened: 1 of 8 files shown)


deployment/helm/node-feature-discovery/README.md
```diff
@@ -175,6 +175,7 @@ NFD.
 | master.extraArgs | list | `[]` | Additional [command line arguments](https://kubernetes-sigs.github.io/node-feature-discovery/master/reference/master-commandline-reference) to pass to nfd-master. |
 | master.extraEnvs | list | `[]` | Additional environment variables to set in the nfd-master container. |
 | master.hostNetwork | bool | `false` | Run the container in the host's network namespace. |
+| master.hostUsers | string | `nil` | Run the container with host user ids. NOTE: if hostNetwork is true, hostUsers should be true. |
```
ArangoGutierrez: Here you say string but declare the var as a bool

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -65,6 +65,10 @@ master:
   # -- Run the container in the host's network namespace.
   # @section -- NFD-Master
   hostNetwork: false
+  # @schema type: [boolean, null]
+  # -- Run the container with host user ids. NOTE: if hostNetwork is true, hostUsers should be true.
+  # @section -- NFD-Master
+  hostUsers: NULL
```
ArangoGutierrez: Shouldn't it be false instead of NULL?

jcpunk: The string `false` will opt into user namespaces.  For folks on kubernetes 1.32 and earlier, if they don't have the right feature flags enabled, having `hostUsers` set could cause scheduling issues.

deployment/helm/node-feature-discovery/README.md
jcpunk: I think that is a byproduct of me using `NULL` instead of `null`... I'll see about fixing.

jcpunk: hmmm, not sure why this is getting picked up as `string` by the helm schema but not the doc generation...

ArangoGutierrez: @marquiz any advice here?

jcpunk:
Rebased

deployment/helm/node-feature-discovery/README.md
marquiz: looks like @jcpunk managed to fix this

PR #2412, opened 2026-01-14
source/local: Fix fsnotify watcher resource leak

ArangoGutierrez:
Close the fsnotify watcher when the context is cancelled to prevent resource leaks. The watcher reference is also reset to nil to allow proper re-initialization if SetNotifyChannel is called again.

### DIFF
--- a/source/local/local.go
+++ b/source/local/local.go
@@ -73,7 +73,7 @@ type localSource struct {
 	features   *nfdv1alpha1.Features
 	config     *Config
 	cancelFunc context.CancelFunc // cancels the active notifier goroutine
-	done       chan struct{}      // closed when notifier goroutine exits
+	wg         sync.WaitGroup     // tracks the active notifier goroutine
 	mu         sync.Mutex         // serializes SetNotifyChannel and protects fields
 }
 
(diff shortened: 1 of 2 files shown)


ArangoGutierrez:
@marquiz  PTAL

ArangoGutierrez:
> LGTM

Thanks!!

/cc @marquiz

source/local/local.go
```diff
@@ -356,22 +393,32 @@ func (s *localSource) SetNotifyChannel(ctx context.Context, ch chan *source.Feat
...
+		if prevDone != nil {
+			<-prevDone
+		}
+
+		// Create a cancellable context for the notifier goroutine
+		notifyCtx, cancel := context.WithCancel(ctx)
+		s.cancelFunc = cancel
+		s.done = make(chan struct{})
+
+		go s.runNotifier(notifyCtx, ch, watcher, s.done)
```
marquiz: @ArangoGutierrez thanks for the update and apologies for my sluggishness on this. Looks pretty good, now. But... 🙈 

Would it be possible to simplify with [WaitGroup.Go()](https://pkg.go.dev/sync@go1.25.7#WaitGroup.Go)? Get rid of the done channed 


```suggestion
		s.wg.Go(runNotifier(notifyCtx, ch, watcher))
```

And then replace the prevDone stuff with simpler `s.wg.Wait()`. WDYT?

ArangoGutierrez: done!

marquiz (approved):
Good improvement @ArangoGutierrez. Ready to be merged imo

/assign @fmuyassarov

PR #2414, opened 2026-01-15
fix flag passing to config override informerpagesize

ivelichkovich:
fix flag passing to config override informerpagesize, it never gets set to the overrides and essentially ends up getting ignored


cmd/nfd-master/main.go
```diff
@@ -22,8 +22,6 @@ import (
 	"os"
 	"time"
 
-	"k8s.io/klog/v2"
```
ArangoGutierrez, line 25: unrelated change

ivelichkovich: fixed

PR #2415, opened 2026-01-15
nfd-master: prevent label removal on incomplete cache sync

### LINKED ISSUE #2408: Node labels removed and re-added on nfd-master restart
**What happened**:

When nfd-master restart, node labels are removed and then re-added during startup, even though the node features have not changed.

The node labels should stay untouched.

We have a prod cluster with 2k nodes, nearly everytime the nfd-master restart, we can see nodes' labels got removed and then added back.

nfd version: v0.18.2, very similar to this issue: https://github.com/kubernetes-sigs/node-feature-discovery/issues/1802

We only observed this behavior on several large clusters: 1k+ nodes. We can confirm the nodefeature does not change, as we found not all node will get the label added back, if we label the node with some irrelevant label, we could see nfd master pickup the changes and add back the nfd labels to the node.

not sure if it is related to list from ETCD directly.

Time line of the nfd-master:
>2026-01-07 19:17:21.765 - "starting the nfd api controller"
2026-01-07 19:20:53.166 - "informer caches synced" duration="3m31.4009872s"
2026-01-07 19:20:53.181 - "starting the NFD master updater pool" parallelism=10
2026-01-07 19:20:53.181 - "http server starting" port=":8080"
2026-01-07 19:20:53.181 - attempting to acquire leader lease kube-system/nfd-master.nfd.kubernetes.io...
2026-01-07 19:21:10.536 - successfully acquired lease kube-system/nfd-master.nfd.kubernetes.io
2026-01-07 19:21:14.186 - "will process all nodes in the cluster"

Request received from the api-server:
>1/7/2026, 7:17:21.795 PM - /nodefeatures?limit=200 → 200
1/7/2026, 7:17:31.774 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615567111)
1/7/2026, 7:17:44.843 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615567111)
1/7/2026, 7:17:53.568 PM - /nodefeatures?continue=...&limit=200 → 429 (RV 615561940)
1/7/2026, 7:18:10.124 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615567111)
1/7/2026, 7:18:36.574 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615567111)
1/7/2026, 7:18:40.567 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615561940)
1/7/2026, 7:18:57.726 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615567111)
1/7/2026, 7:19:03.211 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615561940)
1/7/2026, 7:19:21.653 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615567111)
1/7/2026, 7:19:36.705 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615561940)
1/7/2026, 7:19:46.684 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615567111)
1/7/2026, 7:20:09.740 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615567111)
1/7/2026, 7:20:13.360 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615561940)
1/7/2026, 7:20:35.855 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615561940)
1/7/2026, 7:20:40.265 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615567111)
1/7/2026, 7:20:53.123 PM - /nodefeatures?...&resourceVersion=615567111&watch=true → 200
1/7/2026, 7:20:54.261 PM - /nodefeatures?limit=200&resourceVersion=615567111 → 200
1/7/2026, 7:21:03.284 PM - /nodefeatures?continue=...&limit=200 → 410 Expired (RV 615567111)
1/7/2026, 7:21:03.526 PM - /nodefeatures?limit=200&resourceVersion=615567111 → 410 Expired
1/7/2026, 7:21:03.652 PM - /nodefeatures?limit=200 → 200 (fresh list, RV 615638081)
1/7/2026, 7:21:04.285 PM - /nodefeatures?continue=...&limit=200 → 410 Expired (RV 615561940)
1/7/2026, 7:21:04.358 PM - /nodefeatures?limit=200 → 200 (fresh list)
1/7/2026, 7:21:12.952 PM - /nodefeatures?continue=...&limit=200 → 200 (RV 615638081)
1/7/2026, 7:21:15.662 PM - /nodefeatures?...&resourceVersion=615638308&watch=true → 200
1/7/2026, 7:21:17.160 PM - /nodefeatures?...&resourceVersion=615639272&watch=true → 200
1/7/2026, 7:21:18.912 PM - /nodefeatures?...&resourceVersion=615641654&watch=true → 200
1/7/2026, 7:21:26.771 PM - /nodefeatures?continue=...&limit=200 → 500 ServerTimeout (RV 615638081)
1/7/2026, 7:28:25.931 PM - /nodefeatures?...&resourceVersion=615797355&watch=true → 200

**Environment**:
- Kubernetes version (use `kubectl version`): 1.33.2
- Cloud provider or hardware configuration: azure
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

ArangoGutierrez:
This fixes an issue where nfd-master would remove all NFD-managed labels from nodes when NodeFeature objects were missing from the informer cache. This commonly occurs in large clusters (1k+ nodes) during startup due to pagination failures (410 Expired errors) when syncing the informer cache.

The fix introduces a pending delete tracking mechanism:
- When a NodeFeature is explicitly deleted (DeleteFunc), the node is marked as having a pending delete
- When processing a node update, if no NodeFeature is found in cache:
  - If there's a pending delete marker: proceed with label removal
  - If no pending delete marker: skip label removal (cache miss)

Additionally, a 'processedOnce' mechanism ensures that label removal is only skipped for nodes that haven't been successfully processed before. This prevents regressions where labels from deleted NodeFeatureRules wouldn't be cleaned up.

Also includes:
- Fallback to object name in DeleteFunc when node-name label is missing
- Cleanup of stale pendingDeletes/processedOnce when nodes are deleted
- Comprehensive unit and E2E tests

Fixes: #2408

### DIFF
--- a/pkg/nfd-master/nfd-master.go
+++ b/pkg/nfd-master/nfd-master.go
@@ -29,6 +29,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/google/uuid"
@@ -154,14 +155,29 @@ type nfdMaster struct {
 
 	// isLeader indicates if this instance is the leader, changing dynamically
 	isLeader bool
+
+	// pendingDeletesMu protects pendingDeletes and processedOnce
+	pendingDeletesMu sync.Mutex
+	// pendingDeletes tracks nodes that have pending NodeFeature delete events.
+	// This is used to distinguish between "NodeFeature was deleted" vs
+	// "NodeFeature is missing from cache due to incomplete sync".
+	// Only nodes in this set should have their labels removed when no
+	// NodeFeature is found.
+	pendingDeletes sets.Set[string]
+	// processedOnce tracks nodes that have been processed at least once.
+	// This is used to allow label removal on subsequent processing even if
+	// no NodeFeature exists (e.g., when NodeFeatureRules are deleted).
+	processedOnce sets.Set[string]
 }
 
 // NewNfdMaster creates a new NfdMaster server instance.
 func NewNfdMaster(opts ...NfdMasterOption) (NfdMaster, error) {
 	nfd := &nfdMaster{
-		nodeName:  utils.NodeName(),
-		namespace: utils.GetKubernetesNamespace(),
-		stop:      make(chan struct{}),
+		nodeName:       utils.NodeName(),
+		namespace:      utils.GetKubernetesNamespace(),
+		stop:           make(chan struct{}),
+		pendingDeletes: sets.New[string](),
+		processedOnce:  sets.New[string](),
 	}
 
 	for _, o := range opts {
@@ -421,7 +437,7 @@ func (m *nfdMaster) prune() error {
 		klog.InfoS("pruning node...", "nodeName", node.Name)
 
 		// Prune labels and extended resources
-		err := m.updateNodeObject(m.k8sClient, &node, Labels{}, Annotations{}, ExtendedResources{}, []corev1.Taint{})
+		err := m.updateNodeObject(m.k8sClient, &node, Labels{}, Annotations{}, ExtendedResources{}, []corev1.Taint{}, false)
 		if err != nil {
 			nodeUpdateFailures.Inc()
 			return fmt.Errorf("failed to prune node %q: %v", node.Name, err)
@@ -587,7 +603,10 @@ func (m *nfdMaster) nfdAPIUpdateAllNodes() error {
 
 // getAndMergeNodeFeatures merges the NodeFeature objects of the given node into a single NodeFeatureSpec.
 // The Name field of the returned NodeFeatureSpec contains the node name.
-func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeFeature, error) {
+// The second return value indicates whether any NodeFeature objects were found in the cache
+// (before namespace filtering). This helps distinguish between "no NodeFeature exists" and
+// "NodeFeature exists but is filtered out by namespace selector".
+func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeFeature, bool, error) {
 	nodeFeatures := &nfdv1alpha1.NodeFeature{
 		ObjectMeta: metav1.ObjectMeta{
 			Name: nodeName,
@@ -597,8 +616,11 @@ func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeF
 	sel := k8slabels.SelectorFromSet(k8slabels.Set{nfdv1alpha1.NodeFeatureObjNodeNameLabel: nodeName})
 	objs, err := m.featureLister.List(sel)
 	if err != nil {
-		return &nfdv1alpha1.NodeFeature{}, fmt.Errorf("failed to get NodeFeature resources for node %q: %w", nodeName, err)
+		return &nfdv1alpha1.NodeFeature{}, false, fmt.Errorf("failed to get NodeFeature resources for node %q: %w", nodeName, err)
 	}
+
+	// Track whether we found any objects in the cache (before namespace filtering)
+	foundInCache := len(objs) > 0
 
 	filteredObjs := []*nfdv1alpha1.NodeFeature{}
 	for _, obj := range objs {
@@ -607,9 +629,9 @@ func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeF
 		}
 	}
 
-	// Node without a running NFD-Worker
+	// Node without a running NFD-Worker or all NodeFeatures filtered by namespace
 	if len(filteredObjs) == 0 {
-		return &nfdv1alpha1.NodeFeature{}, nil
+		return &nfdv1alpha1.NodeFeature{}, foundInCache, nil
 	}
 
 	// Sort our objects
@@ -665,7 +687,7 @@ func (m *nfdMaster) getAndMergeNodeFeatures(nodeName string) (*nfdv1alpha1.NodeF
 		klog.V(4).InfoS("merged nodeFeatureSpecs", "newNodeFeatureSpec", utils.DelayedDumper(features))
 	}
 
-	return nodeFeatures, nil
+	return nodeFeatures, foundInCache, nil
 }
 
 // isThirdPartyNodeFeature determines whether a node feature is a third party one or created by nfd-worker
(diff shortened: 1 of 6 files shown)


marquiz (commented):
Thanks @ArangoGutierrez for the PR. And apologies for forgetting/neglecting this. It'd be really good to get the issue fixed.

I have basically one reservation, or edge case, that would be nice to somehow resolve: what if the NodeFeature object truly is gone when the nfd-master starts? 🧐 If there is no NodeFeature object the nfd-master should wipe the nfd-originated node labels. AFAIU, the implementation here does nothing in this case: there is not pending delete marker so it ignores the update. If nothing else, would it be possible to do the cleanup in this case e.g. after each periodic cache sync?

ArangoGutierrez:
Thanks for the feedback @marquiz — you were right that the original implementation would leave stale labels indefinitely when a NodeFeature was genuinely deleted before nfd-master started.

I've pushed a set of commits that address this concern along with other improvements identified during review:

**Periodic reconciliation (your concern):** Added a ticker in `nfdAPIUpdateHandler` that fires at the `ResyncPeriod` interval (default 1h) and sets `updateAll = true`. On the first pass, nodes with no NodeFeature and no pending delete have removal skipped (defensive against cache miss). After one resync period, `processedOnce` is `true` so `skipRemoval` no longer applies and stale labels get cleaned up normally. Trade-off: up to 1 resync period of stale labels for this edge case.

**Additional improvements on top:**
- Renamed `skipLabelRemoval` → `skipRemoval` (it also protects annotations)
- Extended skip-removal protection to extended resources and taints for consistency
- Replaced three separate lock/unlock cycles in `nfdAPIUpdateOneNode` with an atomic `resolveNodeState` method

All changes have unit tests. Happy to discuss the approach if you have concerns about the resync-period trade-off.

ArangoGutierrez:
@marquiz Ready for a new review

ArangoGutierrez:
@marquiz Ready for a new review

pkg/nfd-master/nfd-master.go
```diff
@@ -345,6 +361,18 @@ func (m *nfdMaster) nfdAPIUpdateHandler() {
 	nodeFeatureGroup := make(map[string]struct{})
 	updateAllNodeFeatureGroups := false
 	rateLimit := time.After(time.Second)
+
+	// Periodic reconciliation: re-process all nodes at the resync interval
+	// to catch nodes whose NodeFeatures were genuinely absent (not cache misses).
+	// A nil channel blocks forever in select, effectively disabling the case
+	// when ResyncPeriod is zero (which would panic in time.NewTicker).
+	var resyncTickerC <-chan time.Time
+	if m.config.ResyncPeriod.Duration > 0 {
```
ArangoGutierrez, line 370: With `-resync-period=0` (valid today, it just disables informer resync) this backstop is silently off: a node that hit the first-pass skip only gets cleaned up when some other event re-queues it, which in a quiet cluster may be never. Suggest a warning log when the ticker is disabled plus a sentence in the resyncPeriod docs, since the option now has correctness implications.

pkg/nfd-master/nfd-master.go
```diff
@@ -674,22 +711,109 @@ func (m *nfdMaster) isThirdPartyNodeFeature(nodeFeature nfdv1alpha1.NodeFeature,
 }
 
 func (m *nfdMaster) nfdAPIUpdateOneNode(cli k8sclient.Interface, node *corev1.Node) error {
-	// Merge all NodeFeature objects into a single NodeFeatureSpec
-	nodeFeatures, err := m.getAndMergeNodeFeatures(node.Name)
+	// Merge all NodeFeature objects into a single NodeFeatureSpec.
+	// foundInCache indicates whether any NodeFeature objects were found in the
+	// informer cache (before namespace filtering).
+	nodeFeatures, foundInCache, err := m.getAndMergeNodeFeatures(node.Name)
 	if err != nil {
```
ArangoGutierrez, line 718: Noting for reviewers: foundInCache is whole-node granularity. A partial cache miss — the node has worker plus third-party NodeFeatures and the paginated list lost only one of them — still reports true, so the missing object's labels are removed on the first pass. Detecting that is arguably out of scope, but the comment above the skip logic should state the limitation explicitly.

pkg/nfd-master/nfd-master.go
```diff
@@ -674,22 +711,109 @@ func (m *nfdMaster) isThirdPartyNodeFeature(nodeFeature nfdv1alpha1.NodeFeature,
...
+	// - No NodeFeature found in cache (!foundInCache)
+	// - No pending delete marker (not an explicit deletion)
+	// - This node hasn't been processed before (first time during startup)
+	//
+	// After initial processing, subsequent updates proceed normally to allow
+	// cleanup when NodeFeatures are deleted or NodeFeatureRules change.
+	// Atomically resolve pending-delete and processed-once state.
+	// This consumes any pending delete marker and marks the node as processed.
+	hasPendingDelete, alreadyProcessed := m.resolveNodeState(node.Name)
+
```
ArangoGutierrez, line 738: resolveNodeState consumes the marker and sets processedOnce before the patch is attempted. If refreshNodeFeatures fails, the updater pool requeues the node and the retry runs with alreadyProcessed=true, i.e. removal enabled while the cache may still be incomplete — and the apiserver conditions that cause 410 cache misses are the same ones that make patch failures likely. One option: revert the processedOnce insert on error when the node was not previously processed.

pkg/nfd-master/nfd-master.go
```diff
@@ -674,22 +711,109 @@ func (m *nfdMaster) isThirdPartyNodeFeature(nodeFeature nfdv1alpha1.NodeFeature,
...
+	skipRemoval := false
+	if !hasFeatures {
+		if !hasPendingDelete && !foundInCache && !alreadyProcessed {
+			skipRemoval = true
+			klog.V(2).InfoS("first processing of node with no NodeFeature in cache, "+
+				"skipping removal (possible startup cache miss)",
```
ArangoGutierrez: This log line is the first thing an operator needs when debugging why stale labels survived for up to a resync period, but V(2) is hidden at default verbosity. Consider a default-level InfoS and/or a counter metric next to nodeUpdateRequests/nodeUpdateFailures.

test/e2e/node_feature_discovery_test.go
```diff
@@ -620,6 +620,37 @@ var _ = NFDDescribe(Label("nfd-master"), func() {
 				expectedLabels[targetNodeName] = k8sLabels{}
 				eventuallyNonControlPlaneNodes(ctx, f.ClientSet).Should(MatchLabels(expectedLabels, nodes))
 			})
+
+			It("labels should be properly removed when NodeFeature is deleted (pending delete tracking)", func(ctx context.Context) {
```
ArangoGutierrez, line 624: This flow passes even without this PR: after the create step the node is in processedOnce, so removal proceeds regardless of the pending-delete marker, and the case above already covers delete-triggered cleanup. The behavior unique to this change — first pass with an empty informer cache against a pre-labeled node — has no test; a unit test driving nfdAPIUpdateOneNode with an empty featureLister would pin the skip-decision block (e.g. catch an inverted alreadyProcessed check).

ArangoGutierrez (commented):
Self-review notes for the other reviewers, since this is my own PR. The current revision pairs the pending-delete marker with a one-shot processedOnce guard and uses the ResyncPeriod ticker (1h default) as the cleanup backstop for the truly-deleted-NodeFeature case raised by marquiz; skip-removal now covers labels, annotations, extended resources and taints symmetrically.

Areas where I'd like a second pair of eyes: the guard state is consumed before the node patch is attempted, so a failed first pass retries with removal re-enabled; foundInCache is whole-node granularity, so a partial cache miss is not protected; and the skip path only logs at V(2). Details inline.

The open Copilot threads (tombstone unwrap in DeleteFunc, nil node on non-NotFound getNode errors, tracking-annotation overwrite under skipRemoval) still need resolution before this is ready for lgtm. CI is green across the prow jobs; tide is only waiting on the lgtm label.

ArangoGutierrez:
/cc @PiotrProkop

Green and /approve'd — just needs your /lgtm to merge. Thanks Piotr!

pkg/nfd-master/nfd-master.go
```diff
@@ -674,22 +711,109 @@ func (m *nfdMaster) isThirdPartyNodeFeature(nodeFeature nfdv1alpha1.NodeFeature,
...
+	// Merge all NodeFeature objects into a single NodeFeatureSpec.
+	// foundInCache indicates whether any NodeFeature objects were found in the
+	// informer cache (before namespace filtering).
+	nodeFeatures, foundInCache, err := m.getAndMergeNodeFeatures(node.Name)
 	if err != nil {
 		return fmt.Errorf("failed to merge NodeFeature objects for node %q: %w", node.Name, err)
 	}
 
+	// Check if we have any features for this node
+	hasFeatures := len(nodeFeatures.Spec.Labels) > 0 ||
```
PiotrProkop: consider making `HasFeatures` a `NodeFeature` func:
```go
func (n *NodeFeature) HasFeatures() bool 
```

pkg/nfd-master/nfd-master.go
```diff
@@ -674,22 +711,109 @@ func (m *nfdMaster) isThirdPartyNodeFeature(nodeFeature nfdv1alpha1.NodeFeature,
...
+	if !hasFeatures {
+		if !hasPendingDelete && !foundInCache && !alreadyProcessed {
+			skipRemoval = true
+			klog.V(2).InfoS("first processing of node with no NodeFeature in cache, "+
+				"skipping removal (possible startup cache miss)",
+				"nodeName", node.Name)
+		} else if hasPendingDelete {
+			klog.V(2).InfoS("NodeFeature deleted, proceeding with removal",
+				"nodeName", node.Name)
+		}
```
PiotrProkop: for better readability can you:

```suggestion
        if hasPendingDelete {
        		klog.V(2).InfoS("NodeFeature deleted, proceeding with removal",
				"nodeName", node.Name)
		} else {
		    if !foundInCache && !alreadyProcessed {
			    skipRemoval = true
			    klog.V(2).InfoS("first processing of node with no NodeFeature in cache, "+
				    "skipping removal (possible startup cache miss)",
				    "nodeName", node.Name)
		}
```

ArangoGutierrez:
@PiotrProkop addressed both your review suggestions and rebased on latest master:

- `HasFeatures` is now a method on `NodeFeature` (in `api/nfd/v1alpha1/feature.go`), replacing the inline check.
- Restructured the cache-miss `if/else` per your suggestion. I kept the outer `if !nodeFeatures.HasFeatures()` guard (your snippet dropped it) so the "NodeFeature deleted" log doesn't fire when features are present — semantics are unchanged.

Verified locally: `go build ./...` + `go vet` clean, and `pkg/nfd-master` + `api/nfd` unit tests pass. PTAL — `/lgtm` when you're happy. Thanks!

PiotrProkop:
thanks!
/lgtm

PR #2422, opened 2026-01-23
Remove default CPU limits

haavard:
This was already done for Helm: https://github.com/kubernetes-sigs/node-feature-discovery/pull/1728

This brings makes Kustomization deployments in line with Helm by removing CPU limits by default, for the same reasons (<https://github.com/kubernetes-sigs/node-feature-discovery/issues/1724>, <https://home.robusta.dev/blog/stop-using-cpu-limits>)


fmuyassarov (commented):
Thanks!

/ok-to-test
/lgtm

Issue #2424, opened 2026-01-28
Use finegrained kubelet API authorization


vinayakankugoyal:
Granting nodes/proxy with even the get verb can be used by attackers to gain privileges. This risk has been highlighted in the kubernetes documentation in https://kubernetes.io/docs/concepts/security/rbac-good-practices/#access-to-proxy-subresource-of-nodes and https://kubernetes.io/docs/concepts/security/api-server-bypass-risks/#kubelet-api and more recently in - https://grahamhelton.com/blog/nodes-proxy-rce.

We found the following instances of nodes/proxy being used in this repository

https://github.com/search?q=%22nodes%2Fproxy%22+%28repo%3Akubernetes-sigs%2Fnode-feature-discovery+OR+repo%3Akubernetes-sigs%2Fnode-feature-discovery-operator%29&type=code

Please update the workloads to use fine-grained kubelet API authorization [KEP](https://github.com/kubernetes-sigs/kindnet/issues/kep.k8s.io/2862) and replace nodes/proxy with more fine-grained permissions. See [this](https://kubernetes.io/docs/reference/access-authn-authz/kubelet-authn-authz/#fine-grained-authorization) link for how to update your RBAC config.

liggitt:
cc @ffromani

liggitt:
it looks like this is only used by nfd-topology-updater to fetch `https://%s:%d/configz` directly from the kubelet API?

https://github.com/kubernetes-sigs/node-feature-discovery/blob/55dc2681e846bf10816d69b9f0fb52f7b48e33fd/cmd/nfd-topology-updater/main.go#L94

if so, `nodes/configz` would grant this permission by default in 1.33+ clusters

---

I don't see a use of the kubelet API or nodes/proxy inside nfd-gc ... is that a remnant that could be dropped from the nfd-gc role?

liggitt:
Opened https://github.com/kubernetes-sigs/node-feature-discovery/pull/2425 and https://github.com/kubernetes-sigs/node-feature-discovery-operator/pull/266 to try to address https://github.com/kubernetes-sigs/node-feature-discovery/issues/2424#issuecomment-3814053383

marquiz:
> it looks like this is only used by nfd-topology-updater to fetch `https://%s:%d/configz` directly from the kubelet API?

Yes, that should be the case.
 
> if so, `nodes/configz` would grant this permission by default in 1.33+ clusters

[...]

PR #2425, opened 2026-01-28
`nodes/proxy` permission cleanup

liggitt:
xref https://github.com/kubernetes-sigs/node-feature-discovery/issues/2424

Switches the permission in node-feature-discovery from `nodes/proxy` to `nodes/configz` (works in 1.33+ clusters by default, avoids overgranting of broad `nodes/proxy` permissions)

Removes the (apparently) unused `nodes/proxy` permissions from the nfd-gc role (I couldn't find anything in the code base that was actually making use of that permission)

cc @ffromani

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/clusterrole.yaml
+++ b/deployment/helm/node-feature-discovery/templates/clusterrole.yaml
@@ -118,12 +118,6 @@ rules:
   - list
   - watch
 - apiGroups:
-  - ""
-  resources:
-  - nodes/proxy
-  verbs:
-  - get
-- apiGroups:
   - topology.node.k8s.io
   resources:
   - noderesourcetopologies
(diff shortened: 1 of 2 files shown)


fmuyassarov:
Since we use that role for obtaining kubelet's configuration in topology-updater, it should be alright to just stick with `configz` endpoint. +1 from my side. 

@marquiz any objections on this?

deployment/helm/node-feature-discovery/templates/clusterrole.yaml
```diff
@@ -120,7 +120,7 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - nodes/proxy
+  - nodes/configz
```
dttung2905: Cmiiw, should this be deleted instead? From my understanding, for gc we dont need nodes/proxy GET perm while for topology updater , it should be changed to nodes/configz

liggitt: ah, yes... I missed that this bit was for gc

liggitt: fixed

tariq1890:
Will this fix be backported to the `release-0.18` branch?

PR #2429, opened 2026-02-03
fix: rename gomega.go to gomega_test.go

faganihajizada:
## Problem description

The file ztest/e2e/gomega.goz references functions (`getNonControlPlaneNodes`, `getNode`) defined in `_test.go` files. Since go build excludes `_test.go` files, it fails with:

```
test/e2e/gomega.go:41:10: undefined: getNonControlPlaneNodes
test/e2e/gomega.go:181:14: undefined: getNode
```

## Solution

Fix go build `./test/e2e/...` failure by renaming `gomega.go` to `gomega_test.go`.

The file contains custom Gomega matchers used exclusively by tests, so it should be a test file.

## Test

```
go build ./test/e2e/... # Now passes
go test -c ./test/e2e/... # Still works
```

### DIFF
--- a/test/e2e/helpers.go
+++ b/test/e2e/helpers.go
@@ -1,0 +1,64 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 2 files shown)


faganihajizada:
> However, I suggest a bit different approach. Instead of renaming to `_test.go` move the functions to a separate non-test file, e.g. `test/e2e/helpers.go`

Thank you @marquiz. I agree with the approach you propose and I pushed new commit to address it.

PR #2431, opened 2026-02-06
safely remove existing buildx builder before recreate

ashokpariya0:
Check if nfd-builder exists before attempting removal to avoid unnecessary cleanup error

We see below errors when running
./hack/init-buildx.sh

failed to remove nfd-builder: no builder "nfd-builder" found
**ERROR: failed to remove one or more builders**

reference: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_node-feature-discovery/2428/pull-node-feature-discovery-build-image-cross-generic/2019112349893398528

Changes in this PR:
- Check if nfd-builder exists before removing it

### DIFF
--- a/hack/init-buildx.sh
+++ b/hack/init-buildx.sh
@@ -19,7 +19,7 @@ export DOCKER_CLI_EXPERIMENTAL=enabled
 
 # We can skip setup if the current builder already has multi-arch
 # AND if it isn't the docker driver, which doesn't work
-current_builder="$(docker buildx inspect nfd-builder || true)"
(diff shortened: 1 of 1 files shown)


ashokpariya0:
> I think we also need to fix it in l22 and make it something like below, otherwise we won't get rid of that ERROR message fully.
> 
> ```
> current_builder="$(docker buildx inspect nfd-builder 2>/dev/null || true)
> ```
Yes, updated [change](https://github.com/kubernetes-sigs/node-feature-discovery/pull/2431/changes/efee42f008b33f266b808ccce38a28bd52dd3be7)

Issue #2430, opened 2026-02-06
Security Vulnerability: Go Package: golang.org/x/net <= 0.44.0 - Multiple Remote Denial of Service Vulnerabilities - 0.45.0


anjaniprayaga:
**What happened**: golang.org/x/net package used in the latest version <= 0.44.0, 
Description:
1) The parser implements the HTML specification, which contains a number of algorithms which are quadratic in complexity by design. This causes the processing time to scale non-linearly with respect to the size of the input for some HTML documents. We have imposed a depth limit of 512 for nested HTML tags, which should be high enough for the vast majority of valid HTML documents, to address this. [CVE-2025-47911] 2) The parser also misimplemented a portion of the HTML specification for table related tags. This could cause the parser to enter an infinite loop when encountering specific combinations of tags. [CVE-2025-58190] Vendor Affected Components: Go Package: golang.org/x/net: < 0.45.0

**What you expected to happen**: Update Go Package: golang.org/x/net to version 0.45.0 or later.

**How to reproduce it (as minimally and precisely as possible)**: Please generate the sbom, or please find the below snippet attached.

<img width="1575" height="647" alt="Image" src="https://github.com/user-attachments/assets/df69d8f0-e82e-4952-8e93-eb7d65a0b115" />

**Anything else we need to know?**: NA

**Environment**:
- Kubernetes version (use `kubectl version`): NA
- Cloud provider or hardware configuration: NA
- OS (e.g: `cat /etc/os-release`): NA
- Kernel (e.g. `uname -a`): NA
- Install tools: NA
- Network plugin and version (if this is a network-related bug): NA
- Others:NA

PR #2432, opened 2026-02-07
deps: update dependencies and get rid of external CVE

fmuyassarov:
fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/2430

### DIFF
--- a/api/nfd/go.mod
+++ b/api/nfd/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/node-feature-discovery/api/nfd
 
-go 1.23.0
+go 1.24.0
 
 require (
 	github.com/stretchr/testify v1.8.4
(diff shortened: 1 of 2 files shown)


api/nfd/go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/node-feature-discovery/api/nfd
 
-go 1.23.0
+go 1.25.0
```
marquiz: 1.24 should be enough?

fmuyassarov: Sure. May I ask why not 1.25?

marquiz: This is kinda "library" module. No need to force downstream users to use the latest golang version unless absolutely necessary

Issue #2389, opened 2025-12-09
Add a flag to disable leader election for NFD-master


Wennn:
**What would you like to be added**:
Add a flag to disable leader election for NFD-master in values.yaml
**Why is this needed**:
When we run NFD-master on a large-scale cluster, the leader election may time out very easily, causing the NFD-master pod to panic.
In this scenario, increasing replicas won’t help because there’s always only one instance that can process the node labels. Therefore, we would like to explicitly disable the leader election to reduce the likelihood of NFD-master panic.

marquiz:
Makes sense. I guess the logic should be if the replicacount is 1 then don't enable leader election
/help

nmn3m:
Hi @marquiz,

I see two possible approaches for this feature:

**Option 1: Automatic based on replicaCount**
```yaml
master:
  replicaCount: 1  # leader election disabled
  # replicaCount: 3  # leader election enabled
```
- Leader election only enabled when `replicaCount > 1`
- Simpler configuration

**Option 2: Explicit flag**
```yaml
master:
  replicaCount: 1
  enableLeaderElection: false
```
- Direct control to disable leader election
- Addresses the original issue's need to "explicitly disable" for large-scale clusters

**Question:** The issue mentions wanting to explicitly disable leader election to avoid timeouts. With Option 1, there's no way to force it off if needed. Do you think Option 1 is sufficient, or should we provide the explicit flag for edge cases?

Happy to pick this up if you'd like!

nmn3m:
Hi @ArangoGutierrez 👋  

We’re considering two options for NFD-master leader election:  
1) auto-disable when `replicaCount == 1`, or  
2) add an explicit `enableLeaderElection` flag in `values.yaml`.

Given the timeout issues on large clusters, do you have a recommendation?  
Thanks for your input!

marquiz:
I think option 1 is better. Leader is for cases where replica count is greater than 1 as otherwise all instances of nfd-master will try to update all nodes (races plus a lot of traffic to the api server).

repairman29:
🎯 **PING RECEIVED!**

**@repairman29** has claimed this bounty via [Echeo](https://echeo.io).

**Match Score:** 85%  
**Status:** Ready to start working!

[View conversation in Echeo](https://echeo.io/messages) to coordinate.

repairman29:
Hey there! This looks like something I could help with. I'd be happy to take a look and see what we can do.

[...]

PR #2435, opened 2026-02-10
nfd-master: Auto-disable leader election if replicaCount is 1

ozhuraki:
Closes: #2389

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/master.yaml
+++ b/deployment/helm/node-feature-discovery/templates/master.yaml
@@ -121,7 +121,9 @@ spec:
             {{- if .Values.master.instance | empty | not }}
             - "-instance={{ .Values.master.instance }}"
             {{- end }}
+            {{- if gt (int .Values.master.replicaCount) 1 }}
             - "-enable-leader-election"
+            {{- end }}
             {{- if .Values.master.extraLabelNs | empty | not }}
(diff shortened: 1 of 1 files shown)


pkg/nfd-master/nfd-master.go
```diff
@@ -310,10 +311,10 @@ func (m *nfdMaster) Run() error {
...
 	// Run updater that handles events from the nfd CRD API.
-	if m.args.EnableLeaderElection {
-		go m.startLeaderElectionHandler()
-	} else {
+	if m.config.ReplicaCount == 1 {
 		m.isLeader = true
+	} else if m.args.EnableLeaderElection {
+		go m.startLeaderElectionHandler()
 	}
 	go m.nfdAPIUpdateHandler()
```
fmuyassarov: Maybe to avoid accidental false election triggers we could be explicit here that we want to start the election only when there is more than one replica? In other words,

```go
if m.config.ReplicaCount == 1 {
	m.isLeader = true
} else if m.config.ReplicaCount > 1 && m.args.EnableLeaderElection {
	go m.startLeaderElectionHandler()
}
```

ozhuraki:
@marquiz 
Thanks for the help and sorry for a delay with this. Enforced at helm level, please take a look.

Issue #2440, opened 2026-02-16
s390x: CPU model attributes (vendor_id/family/id) reporting default zero values


ashokpariya0:
**What happened**:
On s390x (IBM zSystems) nodes, NFD reports generic or placeholder/default values for CPU model attributes. Specifically:

  **feature.node.kubernetes.io/cpu-model.family=0
   feature.node.kubernetes.io/cpu-model.id=0
   eature.node.kubernetes.io/cpu-model.vendor_id=VendorUnknown**

feature.node.kubernetes.io/cpu-model.family is reported as 0.
feature.node.kubernetes.io/cpu-model.id is reported as 0.
feature.node.kubernetes.io/cpu-model.vendor_id is reported as VendorUnknown

**What you expected to happen**:
NFD should identify the specific hardware attributes of the IBM Z system by parsing architecture-specific files. It should:

Map the machine type from /proc/cpuinfo to cpu-model.id (e.g., 8561 for z15).
Map the hex identification string to cpu-model.family.
Sanitize the vendor string (e.g., IBM_S390) to be a valid Kubernetes label.

**How to reproduce it (as minimally and precisely as possible)**:
Install NFD on an s390x node (LPAR or VM)
Inspect the node labels:

**Anything else we need to know?**:
The current cpuid library used by NFD is optimized for x86/ARM64. On s390x, /proc/cpuinfo and /proc/sysinfo are the standard sources for hardware topology. Without these accurate labels, it is impossible to use NodeFeatureRules to distinguish between different generations of IBM Z hardware (e.g., z14 vs z15 vs z16) which have different vector and cryptographic instructions.

**Environment**:
- Kubernetes version (use `kubectl version`):
- Cloud provider or hardware configuration: IBM zSystems (s390x)
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

PR #2441, opened 2026-02-16
feat(cpu): implement cpu model and vendor id detection

ashokpariya0:
- Implement specialized s390x attribute discovery from /proc/cpuinfo.
- Map s390x 'machine' type to 'cpu-model.id' and 'identification' string to 'cpu-model.family'.

This PR fixes: https://github.com/kubernetes-sigs/node-feature-discovery/issues/2440

This change ensures that IBM Z systems report accurate hardware generations instead of default zero values.

This PR fixes an issue where s390x (IBM Z) nodes report generic or incorrect CPU model attributes. Because NFD’s default discovery logic is primarily tailored for x86/ARM64, s390x systems currently fallback to "Unknown" or zeroed values. This change introduces a dedicated discovery path for s390x by parsing host-specific files (/proc/cpuinfo and /proc/sysinfo).

With this change we get correct value

 **kubectl get node control-3 -o json | jq '.metadata.labels | with_entries(select(.key |contains("feature.node.kubernetes.io")))'**
{

  
  "feature.node.kubernetes.io/cpu-model.family": "0DBBF8",
    "feature.node.kubernetes.io/cpu-model.id": "8561",
  "feature.node.kubernetes.io/cpu-model.vendor_id": "IBM_S390",
 
  
}

### DIFF
--- a/source/cpu/cpu.go
+++ b/source/cpu/cpu.go
@@ -282,6 +282,20 @@ func getCPUModel() map[string]string {
 		cpuModelInfo["hypervisor"] = hypervisor
 	}
 
+	if runtime.GOARCH == "s390x" {
+		s390xAttrs := getS390xAttributes()
+		if v := s390xAttrs["vendor_id"]; v != "" {
+			cpuModelInfo["vendor_id"] = v
+		}
+		if f := s390xAttrs["family"]; f != "" && f != "0" {
+			cpuModelInfo["family"] = f
+		}
+		if id := s390xAttrs["id"]; id != "" && id != "0" {
+			cpuModelInfo["id"] = id
+		}
+
+	}
+
 	return cpuModelInfo
 }
 
@@ -325,6 +339,38 @@ func getHypervisorFromProcSysinfo() (string, error) {
 	hypervisor = fullRegex.ReplaceAllString(hypervisor, "_")
 
 	return hypervisor, nil
+}
+
+func getS390xAttributes() map[string]string {
+	attrs := make(map[string]string)
+	attrs["vendor_id"] = ""
+	attrs["family"] = "0"
+	attrs["id"] = "0"
+
+	if data, err := os.ReadFile("/proc/cpuinfo"); err == nil {
+		lines := strings.Split(string(data), "\n")
+		for _, line := range lines {
+			parts := strings.SplitN(line, ":", 2)
+			if len(parts) < 2 {
+				continue
+			}
+			key := strings.TrimSpace(parts[0])
+			val := strings.TrimSpace(parts[1])
+
+			switch key {
+			case "vendor_id":
+				// Sanitize IBM/S390 -> IBM_S390
+				attrs["vendor_id"] = regexp.MustCompile("[^-A-Za-z0-9_.]+").ReplaceAllString(val, "_")
+			case "machine":
+				attrs["id"] = val
+			case "identification":
+				if attrs["family"] == "0" { // Only set once for the first processor found
+					attrs["family"] = val
+				}
+			}
+		}
+	}
+	return attrs
 }
 
 func getHypervisorFromCPUID() (string, error) {


ashokpariya0:
could you please look into this pr ? @marquiz @adrianchiris

source/cpu/cpu.go
```diff
@@ -281,6 +281,20 @@ func getCPUModel() map[string]string {
 	} else if hypervisor != "" {
 		cpuModelInfo["hypervisor"] = hypervisor
 	}
+	if runtime.GOARCH == "s390x" {
+		s390xAttrs := getS390xAttributes()
+		if s390xAttrs["vendor_id"] != "" {
```
marquiz: Should we skip setting the other values in case `["vendor_id"] == ""`?

source/cpu/cpu.go
```diff
@@ -281,6 +281,20 @@ func getCPUModel() map[string]string {
...
+			cpuModelInfo["vendor_id"] = s390xAttrs["vendor_id"]
+		}
+
+		cpuModelInfo["family"] = s390xAttrs["family"]
+		cpuModelInfo["id"] = s390xAttrs["id"]
+
+		hv, err := getHypervisorFromProcSysinfo()
+		if err == nil && hv != "none" && hv != "" {
```
marquiz: I'd suggest to NOT move the getHypervisorFromProcSysinfo() here, let it stay in getHypervisor. Add this fix (`hv != "none"`) in getHypervisorFromProcSysinfo().

WDYT

source/cpu/cpu.go
ashokpariya0: Done

source/cpu/cpu.go
ashokpariya0: Done, [change](https://github.com/kubernetes-sigs/node-feature-discovery/compare/55de82b4ff1af5b0903fb46e3152c31a80de2cc9..b0eb028dd6dec86fe8da23d6a0eb776f97be6231)

source/cpu/cpu.go
```diff
@@ -278,10 +278,24 @@ func getCPUModel() map[string]string {
 	hypervisor, err := getHypervisor()
 	if err != nil {
 		klog.ErrorS(err, "failed to detect hypervisor")
-	} else if hypervisor != "" {
+	} else if hypervisor != "" && hypervisor != "none" {
```
marquiz: Let's not change the behavior other than what the PR subject describes.

getHypervisor() is documented as:

```
Returns normalized hypervisor string, "unknown" for unknown vendor, or "none" for bare metal.
```

ashokpariya0: Addressed

marquiz (commented):
Thanks for the update. The only question I have (missed in my earlier comments), is changing the default hypervisor from "PR/SM" to "none" to desired. It's kinda out-of-scope of the PR title...

@routerhan any comments

ashokpariya0:
> Thanks for the update. The only question I have (missed in my earlier comments), is changing the default hypervisor from "PR/SM" to "none" to desired. It's kinda out-of-scope of the PR title...
> 
> @routerhan any comments

>
I feel changing the default hypervisor from "PR/SM" to "none" is necessary to correctly reflect environments where no hypervisor is detected (e.g., bare metal or unsupported platforms). 
The previous default value of "PR/SM" could lead to incorrect reporting ??
Anyway, I’ve reverted that change. If needed, I can open a separate PR.

routerhan:
> > Thanks for the update. The only question I have (missed in my earlier comments), is changing the default hypervisor from "PR/SM" to "none" to desired. It's kinda out-of-scope of the PR title...
> > @routerhan any comments
> 
> > 
> 
> I feel changing the default hypervisor from "PR/SM" to "none" is necessary to correctly reflect environments where no hypervisor is detected (e.g., bare metal or unsupported platforms). The previous default value of "PR/SM" could lead to incorrect reporting ?? Anyway, I’ve reverted that change. If needed, I can open a separate PR.

Thanks for pointing out, the "PR/SM" indicates the bare metal case on s390x. So I agree it can be classified as "none".

marquiz:
> Anyway, I’ve reverted that change. If needed, I can open a separate PR.

@ashokpariya0 a separate PR :+1: Also because then this fix get's nicely "documented" in the commit history

ashokpariya0:
can we please merge this pr?

PR #2449, opened 2026-03-01
build: disable CGO for multi-arch builds

fmuyassarov:
Add `CGO_ENABLED=0` to Makefile to prevent GCC invocation during cross-compilation. This is an attempt to fix the build failure caused by GCC internal compiler error (segfault in cc1) when compiling runtime/cgo.


marquiz (approved):
/lgtm

PR #2452, opened 2026-03-02
add --load flag to docker buildx build

fmuyassarov:
Export the built image to the local Docker daemon so it is available for subsequent steps, example docker images. By default, buildx builds images using BuildKit and leaves the result in the build cache without exporting it anywhere.

_Note_: building images for multi arch with `--load`parameter is now supported and fixed in docker buildx. See https://github.com/docker/buildx/issues/59.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -113,9 +113,8 @@ image: yamls
 	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
 
 image-all: ensure-buildx yamls
-# --load : not implemented yet, see: https://github.com/docker/buildx/issues/59
-	$(IMAGE_BUILDX_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
-	$(IMAGE_BUILDX_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
+	$(IMAGE_BUILDX_CMD) --load $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
+	$(IMAGE_BUILDX_CMD) --load $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
 
 # clean NFD labels on all nodes
 # devel only


fmuyassarov:
well, I guess we need the Prow infra to bump their docker before we move on :)

ozhuraki (approved):
LGTM, in case you decide to move with docker upgrade

ArangoGutierrez (commented):
Adding `--load` to `image-all` fixes a real local-dev annoyance — today the multi-arch build result stays in the buildx cache and never shows up in `docker images`. The blocker is that multi-arch `--load` requires the containerd image store in the daemon, so the required cross-build Prow job fails (details inline) and waiting for a generic docker bump may not resolve it; an opt-in flag would keep CI green without depending on infra. Separately, `docs/developer-guide/index.md` (lines 58-60) still says docker cannot load manifest-lists and links the same buildx#59 this PR drops from the Makefile — that paragraph should be updated here too. Not mergeable as-is with the required job red; good to go once the flag is opt-in (or Prow's daemon actually supports containerd-store loads) and the doc is refreshed.

Makefile
```diff
@@ -113,9 +113,8 @@ image: yamls
 	$(IMAGE_BUILD_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
 
 image-all: ensure-buildx yamls
-# --load : not implemented yet, see: https://github.com/docker/buildx/issues/59
-	$(IMAGE_BUILDX_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
-	$(IMAGE_BUILDX_CMD) $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_MINIMAL)
+	$(IMAGE_BUILDX_CMD) --load $(IMAGE_BUILD_ARGS) $(IMAGE_BUILD_ARGS_FULL)
```
ArangoGutierrez, line 116: Multi-arch `--load` only works when the daemon runs the containerd image store (recent engines default to it on fresh installs only), which is exactly what the required `pull-node-feature-discovery-build-image-cross-generic` job hits — `scripts/test-infra/build-image-cross.sh` runs this target with `IMAGE_ALL_PLATFORMS=linux/amd64,linux/arm64`. Rather than waiting on Prow infra, consider making this opt-in: an empty-default `IMAGE_BUILDX_LOAD ?=` variable referenced here, or simply document `make image-all IMAGE_BUILD_EXTRA_OPTS=--load`, which already works today against a containerd-store daemon. Opt-in also spares the default 5-platform build from exporting every arch variant of both images into the local daemon.

PR #2463, opened 2026-03-17
source/cpu: Expose CPU cores information

ozhuraki:


source/cpu/cpu.go
```diff
@@ -299,6 +308,15 @@ func getCPUModel() map[string]string {
 	return cpuModelInfo
 }
 
+func getCPUCores() map[string]string {
+	cpuCoresInfo := make(map[string]string)
+	cpuCoresInfo["physical"] = strconv.Itoa(cpuid.CPU.PhysicalCores)
```
ArangoGutierrez, line 313: These cpuid-derived counts don't hold up across platforms: on x86 they come from CPUID leaf 0xB / CPUID.1, which describes a single package, so a 2-socket node reports half its cores; on arm64 the library just returns `runtime.NumCPU()` with `ThreadsPerCore=1`; on ppc64le/s390x all three are 0/0/1 (wrong on POWER SMT). Deriving them from sysfs like `discoverTopology()` already does — unique (physical_package_id, core_id) pairs for physical, CPU entry count for logical — gives node-level, arch-independent numbers and can account for offline CPUs, which CPUID enumeration can't.

source/cpu/cpu.go
```diff
@@ -40,6 +40,7 @@ const Name = "cpu"
 
 const (
 	CpuidFeature       = "cpuid"
+	Cpucores           = "cpu_cores"
```
ArangoGutierrez, line 43: This surfaces as feature `cpu.cpu_cores` in NodeFeatureRules — the `cpu` prefix is redundant and the underscore diverges from the single-word feature names here. These elements fit the existing `topology` feature next to `socket_count` and `hardware_multithreading`, so extending `discoverTopology()` would avoid a new feature name entirely.

source/cpu/cpu.go
```diff
@@ -161,6 +162,11 @@ func (s *cpuSource) GetLabels() (source.FeatureLabels, error) {
 		labels["model."+k] = v
 	}
 
+	// CPU cores
+	for k, v := range features.Attributes[Cpucores].Elements {
+		labels["cores."+k] = v
```
ArangoGutierrez, line 167: Counts have deliberately stayed attribute-only in this source: `topology.socket_count` gets no default label, and the security counts are skipLabel'ed for the same reason. Three always-on numeric labels per node is cardinality NFD avoids by default — I'd drop this hunk and let users opt in via NodeFeatureRule.

ArangoGutierrez (commented):
This adds physical/threads_per_core/logical CPU counts from klauspost/cpuid as a new `cpu_cores` attribute feature plus three default labels. The data source is the main problem: cpuid-derived counts are per-package on multi-socket x86, fall back to `runtime.NumCPU()` on arm64, and are 0 on ppc64le/s390x, so the published values are wrong on a fair share of real nodes — the sysfs topology walk in `discoverTopology()` is the right basis. Counts have also been kept attribute-only so far (`topology.socket_count` has no default label), so I'd drop the GetLabels hunk and document the new elements in the `cpu.topology` table of customization-guide.md instead. Prow CI hasn't run yet (needs-ok-to-test); no merge conflicts. Needs the discovery moved to sysfs and the attribute/label split sorted out before this is mergeable.

docs/usage/features.md
```diff
@@ -65,6 +65,9 @@ feature.node.kubernetes.io/<feature> = <value>
 | **`cpu-model.vendor_id`**           | string | Comparable CPU vendor ID. |
 | **`cpu-model.family`**              | int    | CPU family. |
 | **`cpu-model.id`**                  | int    | CPU model number. |
+| **`cpu-cores.physical `**           | int    | Number of CPU physical cores. |
+| **`cpu-cores.threads_per_core`**    | int    | Number of threads per CPU core. |
+| **`cpu-cores.logical`**             | int    | Number of CPU physical cores. |
```
ArangoGutierrez, line 70: Two typos: trailing space inside the backticks on `cpu-cores.physical `, and the `logical` row describes physical cores. Also, if these end up attribute-only (see comment on GetLabels), this label table is the wrong place — the `cpu.topology` table in customization-guide.md is where they'd go.
```suggestion
| **`cpu-cores.physical`**            | int    | Number of CPU physical cores. |
| **`cpu-cores.threads_per_core`**    | int    | Number of threads per CPU core. |
| **`cpu-cores.logical`**             | int    | Number of CPU logical cores. |
```

ArangoGutierrez, line 70: This describes `cpu-cores.logical` as "Number of CPU physical cores" — looks copy-pasted from the physical row. Should be "Number of CPU logical cores."

docs/usage/features.md
```diff
@@ -65,6 +65,9 @@ feature.node.kubernetes.io/<feature> = <value>
 | **`cpu-model.vendor_id`**           | string | Comparable CPU vendor ID. |
 | **`cpu-model.family`**              | int    | CPU family. |
 | **`cpu-model.id`**                  | int    | CPU model number. |
+| **`cpu-cores.physical `**           | int    | Number of CPU physical cores. |
```
ArangoGutierrez, line 68: Minor: stray trailing space inside the backticks — `cpu-cores.physical ` should be `cpu-cores.physical`.

source/cpu/cpu.go
ArangoGutierrez, line 43: Nit: `Cpucores = "cpu_cores"` is inconsistent with its siblings — `Cpumodel = "model"` (no `cpu_` prefix) and the `*Feature` suffix on `CpuidFeature`/`CstateFeature`. Consider `CpucoresFeature` with value `"cores"`. Non-blocking.

ArangoGutierrez (changes requested):
Thanks — clean addition, and the implementation looks correct (nice use of `strconv.Itoa`). A couple of doc fixes and a nit before merge:

- `cpu-cores.logical` is documented as "Number of CPU physical cores" (copy-paste from the physical row) — should be logical (inline).
- stray trailing space inside the `cpu-cores.physical ` cell (inline).
- minor const-naming inconsistency (inline, non-blocking).

The code itself is good once the docs are corrected. Thanks!

PR #2464, opened 2026-03-17
nfd-worker: Don't connect to master in case --no-publish is used

ozhuraki:
This is useful together with Oneshot option for quick out of cluster tests.

### DIFF
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -179,6 +179,17 @@ func NewNfdWorker(opts ...NfdWorkerOption) (NfdWorker, error) {
 		nfd.configFilePath = filepath.Clean(nfd.args.ConfigFile)
 	}
 
+	if err := nfd.readConfig(nfd.configFilePath, nfd.config); err != nil {
+		return nil, err
+	}
+
+	if (nfd.args.Overrides.NoPublish != nil && *nfd.args.Overrides.NoPublish) || nfd.config.Core.NoPublish {
+		// Don't connect to apiserver in case --no-publish is used.
+		// This is useful together with Oneshot option for quick
+		// out of cluster tests.
+		return nfd, nil
+	}
+
 	// k8sClient might've been set via opts by tests
 	if nfd.k8sClient == nil {
 		kubeconfig, err := utils.GetKubeconfig(nfd.args.Kubeconfig)
@@ -489,16 +500,7 @@ func (w *nfdWorker) configureCore(ctx context.Context, c coreConfig) error {
 	return nil
 }
 
-// Parse configuration options
-func (w *nfdWorker) configure(ctx context.Context, filepath string, overrides string) error {
-	// Create a new default config
-	c := newDefaultConfig()
-	confSources := source.GetAllConfigurableSources()
-	c.Sources = make(map[string]source.Config, len(confSources))
-	for _, s := range confSources {
-		c.Sources[s.Name()] = s.NewConfig()
-	}
-
+func (w *nfdWorker) readConfig(filepath string, c *NFDConfig) error {
 	// Try to read and parse config file
 	if filepath != "" {
 		data, err := os.ReadFile(filepath)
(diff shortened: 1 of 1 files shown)


marquiz (commented):
Thanks @ozhuraki, this is a welcome enhancement for development/debug workflow. Two quick comments:
1. This is a bit more complicated and needs more work: we need to respect the `noPublish` setting from the config file, too
2. s/master/apiserver/ in the commit message and comments

ozhuraki:
@marquiz
Thanks, makes sense. Updated, PTAL.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -179,6 +179,17 @@ func NewNfdWorker(opts ...NfdWorkerOption) (NfdWorker, error) {
 		nfd.configFilePath = filepath.Clean(nfd.args.ConfigFile)
 	}
 
+	if err := nfd.readConfig(nfd.configFilePath, nfd.config); err != nil {
```
ArangoGutierrez, line 182: Nit: the config file now gets read and parsed twice on startup (again via `configure()` in `Run()`), duplicating the "configuration file parsed" log and the `core.sources` deprecation warning. Since this parse only peeks at `core.noPublish` — a bare `&NFDConfig{}` with no `Sources` pre-population silently drops the per-source sections — parsing into a local throwaway config instead of `nfd.config` would make that clearer.

pkg/nfd-worker/nfd-worker.go
```diff
@@ -179,6 +179,17 @@ func NewNfdWorker(opts ...NfdWorkerOption) (NfdWorker, error) {
...
+		return nil, err
+	}
+
+	if (nfd.args.Overrides.NoPublish != nil && *nfd.args.Overrides.NoPublish) || nfd.config.Core.NoPublish {
```
ArangoGutierrez, line 186: This ORs the flag and the config file, but `configure()` gives the flag precedence: an explicit `--no-publish=false` with `core.noPublish: true` in the file takes this branch (no client) while the effective runtime config publishes. Also `-options='core: {noPublish: true}'` never triggers the skip. Mirroring `configure()`'s precedence keeps the two in sync:
```suggestion
	noPublish := nfd.config.Core.NoPublish
	if nfd.args.Overrides.NoPublish != nil {
		noPublish = *nfd.args.Overrides.NoPublish
	}
	if noPublish {
```

pkg/nfd-worker/nfd-worker.go
```diff
@@ -179,6 +179,17 @@ func NewNfdWorker(opts ...NfdWorkerOption) (NfdWorker, error) {
...
+		// Don't connect to apiserver in case --no-publish is used.
+		// This is useful together with Oneshot option for quick
+		// out of cluster tests.
+		return nfd, nil
```
ArangoGutierrez, line 190: Returning here leaves `w.k8sClient` nil, but `Run()` -> `configure()` -> `configureCore()` unconditionally calls `setOwnerReference()`, which dereferences `w.k8sClient` whenever `POD_NAME` is set and owner refs are enabled. The stock kustomize/helm manifests set `POD_NAME`, so in-cluster `--no-publish` now panics. Skipping `setOwnerReference()` when no-publish is in effect (owner refs only go on the NodeFeature object, which is never created in that mode) or nil-guarding `w.k8sClient` there closes it.

ArangoGutierrez (commented):
This skips k8s client construction in `NewNfdWorker()` when no-publish is requested, so oneshot out-of-cluster runs work without a kubeconfig, and the config-file handling requested in the earlier review is in place via the `readConfig()` split. Two correctness gaps inline: in-cluster `--no-publish` now panics in `setOwnerReference()` (nil `k8sClient`; `POD_NAME` is set by the stock manifests), and the gate doesn't match `configure()`'s flag-over-file precedence. There's also no test for the new branch: the existing worker tests always inject a fake client via `WithKubernetesClient()`, so a case constructing the worker with `-no-publish` and no client option would pin this down. Prow hasn't run yet (needs-ok-to-test) and tide reports the PR not mergeable pending lgtm/approve. The nil-client panic and a test need addressing before this is ready to merge.

ArangoGutierrez:
/ok-to-test

Reviewed the code — the refactor is correct and nil-safe (`nfd.config` is initialized before `readConfig`, `Overrides.NoPublish` is nil-checked), and the early-return correctly skips apiserver connection only when no-publish is set. Running CI.

One pre-merge ask: please add a unit test for the new path (`NoPublish ⇒ NewNfdWorker creates no k8sClient`) — the worker is test-constructible via opts. Minor: config file is now read twice on the publish path (harmless, but a brief comment would help).

ArangoGutierrez:
/retest

`build-image-cross-generic` failed on a transient Docker Hub auth flake (`auth.docker.io/token ... 404` while pulling `debian:bookworm-slim`), not anything in this change — `build-master`/`verify`/`e2e` all passed. Re-running.

PR #2465, opened 2026-03-17
source/memory: Detect total memory

ozhuraki:


marquiz (commented):
May I ask the rationale/use case for this feature? Memory is managed by K8s as native resources and information is available e.g. in Node status. Even though this is a relatively small change, we need a solid justification if we're to add it

ArangoGutierrez (commented):
Adds MemTotal from /proc/meminfo as a new `memory.memory` attribute and also publishes it as a default node label. The attribute half is the right shape (NodeFeatureRule can match it with Gt/Lt), but the default label is high-cardinality (see inline), the hardcoded /proc path bypasses the hostmount prefix, the docs row doesn't match what the code publishes, and there's no unit test even though a meminfo fixture would be cheap. Besides the inline comments, @marquiz's use-case question from March is still unanswered and the PR description is empty — that's the gating discussion; CI is otherwise just waiting on ok-to-test. My take: drop the label, keep the attribute, document it in the customization guide, and answer the rationale question — then this is a small, easy change to land.

source/memory/memory.go
```diff
@@ -72,6 +77,9 @@ func (s *memorySource) GetLabels() (source.FeatureLabels, error) {
 	labels := source.FeatureLabels{}
 	features := s.GetFeatures()
 
+	// Memory
+	labels["memory"] = s.features.Attributes[MemoryFeature].Elements["total"]
```
ArangoGutierrez, line 81: This publishes `feature.node.kubernetes.io/memory-memory` with the raw kB count as the value — effectively unique per node, so it can't be used for nodeSelector/affinity matching and only adds label churn. The attribute you set in Discover() is already matchable via NodeFeatureRule (Gt/Lt), which is the right exposure for a numeric quantity, so I'd drop the default label entirely. If it stays, it needs an `ok` guard like every other label in this function: when detectMemory() fails, `Elements` here is a nil map and this publishes an empty-valued label.

source/memory/memory.go
```diff
@@ -141,6 +156,43 @@ func (s *memorySource) GetFeatures() *nfdv1alpha1.Features {
 	return s.features
 }
 
+func detectMemory() (map[string]string, error) {
+	// Open /proc/meminfo file
+	file, err := os.Open("/proc/meminfo")
```
ArangoGutierrez, line 161: Hardcoding `/proc/meminfo` bypasses the hostpath convention: the images are built with `hostpath.pathPrefix=/host-` and the worker deployments mount host proc files explicitly (cf. `host-proc-swaps` in the helm/kustomize templates), so this reads the container's own meminfo — wrong under Kata/gVisor/lxcfs. Use `hostpath.ProcDir.Path("meminfo")` like `detectSwap` below, and add the corresponding mount to the deployment templates. This also makes the function unit-testable against a testdata fixture (incl. malformed input), the way `TestDetectHugePages` does it via `hostpath.SysfsDir`.

```suggestion
	file, err := os.Open(hostpath.ProcDir.Path("meminfo"))
```

docs/usage/features.md
```diff
@@ -186,6 +186,7 @@ configuration options for details.
 
 | Feature              | Value | Description                                               |
 | --------------------| ----- | --------------------------------------------------------- |
+| **`memory-memory.total`**    | int  | Amount of RAM memory in a system in kilobytes, kB (decimal) |
```
ArangoGutierrez, line 189: Two mismatches: the code adds label key `memory`, so the published label is `feature.node.kubernetes.io/memory-memory`, not `memory-memory.total`; and /proc/meminfo's kB unit is 1024 bytes (KiB), not decimal kilobytes. If this ends up attribute-only, the documentation to update is the `memory.memory` feature table in docs/usage/customization-guide.md (next to `memory.numa`/`memory.swap`) rather than this default-labels table.

source/memory/memory.go
```diff
@@ -141,6 +156,43 @@ func (s *memorySource) GetFeatures() *nfdv1alpha1.Features {
...
+	// Check for errors
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("error reading /proc/meminfo: %w", err)
+	}
+
+	re := regexp.MustCompile(`\d+`)
+
+	matches := re.FindStringSubmatch(totalMemory)
+	if len(matches) != 1 {
+		return nil, fmt.Errorf("failed to parse total memory total from /proc/meminfo")
```
ArangoGutierrez, line 188: This error path conflates "MemTotal line not found" (`totalMemory == ""`) with a genuine parse failure, and the message has a typo ("total memory total"). `strings.Fields(totalMemory)` + `strconv.ParseUint(fields[1], 10, 64)` would be stricter (validates the value is actually numeric), distinguishes the two failures, drops the `regexp` import, and avoids recompiling the pattern on every Discover cycle.

source/memory/memory.go
```diff
@@ -141,6 +156,43 @@ func (s *memorySource) GetFeatures() *nfdv1alpha1.Features {
 	return s.features
 }
 
+func detectMemory() (map[string]string, error) {
+	// Open /proc/meminfo file
+	file, err := os.Open("/proc/meminfo")
+	if err != nil {
+		return nil, fmt.Errorf("failed to open /proc/meminfo: %w", err)
+	}
+	defer file.Close()
```
ArangoGutierrez, line 165: This is the `verify-master`/`ci-lint` failure: `errcheck` flags the unchecked `file.Close()`. Simplest fix is to switch `detectMemory` to `os.ReadFile("...meminfo")` (consistent with the rest of this source and drops the scanner); otherwise handle the close error explicitly.

source/memory/memory.go
ArangoGutierrez, line 81: This sets key `"memory"`, so the emitted label is `feature.node.kubernetes.io/memory-memory` — but the docs add `memory-memory.total`. They disagree; the intended key is almost certainly `"memory.total"`. Please align code and docs.

source/memory/memory.go
ArangoGutierrez, line 161: Use `hostpath.ProcDir.Path("meminfo")` rather than hardcoding `/proc/meminfo`, to match `detectSwap` and respect NFD's host-mount prefix convention (and testability).

ArangoGutierrez (changes requested):
Thanks — useful to surface total memory. A few things, and CI is currently red:

**CI blocker (inline):** `defer file.Close()` trips `errcheck` → `ci-lint` fails (`source/memory/memory.go:165`). Simplest fix is to read via `os.ReadFile` like the rest of this source (also drops the bufio scanner); otherwise handle/log the close error.

**Code/docs mismatch (inline):** `GetLabels` sets `labels["memory"]` → renders `memory-memory`, but the docs document `memory-memory.total`. Intended key is almost certainly `"memory.total"`. Please make code and docs agree.

**Host-path convention (inline):** use `hostpath.ProcDir.Path("meminfo")` instead of a hardcoded `/proc/meminfo`, matching `detectSwap` and respecting the host-mount prefix.

**Minor:** lift `regexp.MustCompile` to a package var; error string says "memory total total"; and a fixture-based test for the `MemTotal` parse would be easy and worthwhile.

Marking request-changes. Thanks!

PR #2469, opened 2026-03-26
nfd-master: Add command line unit test

ozhuraki:


cmd/nfd-master/main_test.go
```diff
@@ -0,0 +1,66 @@
+/*
+Copyright 2019-2021 The Kubernetes Authors.
```
marquiz: nit: we could use the correct year here

ozhuraki:
@marquiz
Thanks! Updated, PTAL.

marquiz (approved):
Thank you @ozhuraki 

/assign @ArangoGutierrez @fmuyassarov

marquiz:
@ozhuraki please fix the gofmt (ci failure)

marquiz (approved):
Looks good to me

/assign @fmuyassarov

cmd/nfd-master/main.go
```diff
@@ -41,22 +41,60 @@ func main() {
...
+		if len(a) > 0 {
+			filtered = append(filtered, a)
+		}
+	}
+
+	if len(filtered) == 0 {
+		return args
+	}
+
+	_ = flags.Parse(filtered)
```
ArangoGutierrez, line 90: The other binaries' `parseArgs` (nfd-worker, nfd-gc, nfd-topology-updater) is a plain `flags.Parse(osArgs)` with no filtering, and the filter changes behavior: an empty argv element (`nfd-master ""`) previously hit the unknown-argument check below and exited 2, now it is silently dropped (the `osArgs[:0]` trick also compacts `os.Args` in place). It only exists so the test can pass `""`; with the paired test change this whole block reduces to:

```suggestion
	_ = flags.Parse(osArgs)
```

Note this needs the test tweak first, otherwise the `""` case exits the test binary.

cmd/nfd-master/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	"sigs.k8s.io/node-feature-discovery/pkg/utils"
+)
+
+func TestParseArgs(t *testing.T) {
+	Convey("When parsing command line arguments", t, func() {
+		flags := flag.NewFlagSet(ProgramName, flag.ExitOnError)
+
+		Convey("When no override args are specified", func() {
+			args := parseArgs(flags, "")
```
ArangoGutierrez, line 34: Pairs with the `main.go` comment: `parseArgs` is variadic, so the no-args case can pass nothing — no empty-string placeholder needed. This also passes against the current code, so it can land regardless.

```suggestion
			args := parseArgs(flags)
```

cmd/nfd-master/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+				"-no-publish",
+				"-resync-period=60s",
+				"-informer-page-size=100",
+				"-extra-label-ns=ns",
+				"-deny-label-ns=denied",
+				"-label-whitelist=foo",
+				"-enable-taints",
+				"-nfd-api-parallelism=5")
+
+			Convey("args.sources is set to appropriate values", func() {
```
ArangoGutierrez, line 57: Copy-paste from the nfd-worker test — nfd-master has no `sources`.

```suggestion
			Convey("overrides should be set to the specified values", func() {
```

cmd/nfd-master/main_test.go
```diff
@@ -0,0 +1,69 @@
...
+		Convey("When no override args are specified", func() {
+			args := parseArgs(flags, "")
+
+			Convey("overrides should be nil", func() {
+				So(args.Overrides.ExtraLabelNs, ShouldBeNil)
+				So(args.Overrides.LabelWhiteList, ShouldBeNil)
+				So(args.Overrides.EnableTaints, ShouldBeNil)
+				So(args.Overrides.NoPublish, ShouldBeNil)
+				So(args.Overrides.ResyncPeriod, ShouldBeNil)
+				So(args.Overrides.InformerPageSize, ShouldBeNil)
```
ArangoGutierrez, line 42: `DenyLabelNs` and `NfdApiParallelism` are missing from the nil checks (both are covered in the second case). Nil-ness is what keeps cmdline overrides from clobbering config-file values — `nfd-master.go` only applies an override when non-nil — so all eight fields are worth asserting:

```suggestion
				So(args.Overrides.ExtraLabelNs, ShouldBeNil)
				So(args.Overrides.DenyLabelNs, ShouldBeNil)
				So(args.Overrides.LabelWhiteList, ShouldBeNil)
				So(args.Overrides.EnableTaints, ShouldBeNil)
				So(args.Overrides.NoPublish, ShouldBeNil)
				So(args.Overrides.ResyncPeriod, ShouldBeNil)
				So(args.Overrides.InformerPageSize, ShouldBeNil)
				So(args.Overrides.NfdApiParallelism, ShouldBeNil)
```

ArangoGutierrez (commented):
This extracts `parseArgs` from `main()` and adds `TestParseArgs` for the override flags, bringing nfd-master in line with nfd-worker/-gc/-topology-updater which all have this test already. The override mapping is asserted against literal values and the test runs in the verify job, so the coverage is real. One thing to fix before merge: the empty-argument filtering in `parseArgs` diverges from the other three binaries and silently accepts an empty argv element that previously exited 2 — it only exists to support the test's `""` placeholder (see inline comments; the two suggestions go together). The rest are test nits: two override fields missing from the nil checks and a copy-pasted subtest name; optionally, one non-override flag like `-instance` in the second case would also cover the plain `args` bindings, the way `-oneshot` does in the worker test. With the filtering block dropped this is ready for lgtm — CI is green and tide is only waiting on the label.

ArangoGutierrez:
/lgtm

Reviewed: behavior-preserving refactor extracting CLI parsing into a testable `parseArgs`; the empty-arg filter is a production no-op, ordering is unchanged, and `TestParseArgs` is meaningful (exercises both the no-override and full-override paths, verified green under `make test`).

PR #2474, opened 2026-03-30
kubectl-nfd: Add NodeFeatureGroup dryrun, validation and test

ozhuraki:
Support NodeFeatureGroup for dryrun, validation and test.

--nodefeaturerule-file flag is renamed to --rule-file,
now supports both rules and groups.

### DIFF
--- a/cmd/kubectl-nfd/subcmd/dryrun.go
+++ b/cmd/kubectl-nfd/subcmd/dryrun.go
@@ -26,34 +26,31 @@ import (
 
 var dryrunCmd = &cobra.Command{
 	Use:   "dryrun",
-	Short: "Process a NodeFeatureRule file against a NodeFeature file",
-	Long:  `Process a NodeFeatureRule file against a local NodeFeature file to dry run the rule against a node before applying it to a cluster`,
+	Short: "Process a NodeFeatureRule or NodeFeatureGroup file against a NodeFeature file",
+	Long:  `Process a NodeFeatureRule or NodeFeatureGroup file against a local NodeFeature file to dry run the rule against a node before applying it to a cluster`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Evaluating NodeFeatureRule %q against NodeFeature %q\n", nodefeaturerule, nodefeature)
-		err := kubectlnfd.DryRun(nodefeaturerule, nodefeature)
-		if len(err) > 0 {
-			fmt.Printf("NodeFeatureRule %q is not valid for NodeFeature %q\n", nodefeaturerule, nodefeature)
-			for _, e := range err {
+		fmt.Printf("Evaluating %q against NodeFeature %q\n", rule, nodefeature)
+		errs := kubectlnfd.DryRun(rule, nodefeature)
+		if len(errs) > 0 {
+			fmt.Printf("%q is not valid for NodeFeature %q\n", rule, nodefeature)
+			for _, e := range errs {
 				cmd.PrintErrln(e)
 			}
-			// Return non-zero exit code to indicate failure
 			os.Exit(1)
 		}
-		fmt.Printf("NodeFeatureRule %q is valid for NodeFeature %q\n", nodefeaturerule, nodefeature)
+		fmt.Printf("%q is valid for NodeFeature %q\n", rule, nodefeature)
 	},
 }
 
 func init() {
 	RootCmd.AddCommand(dryrunCmd)
 
-	dryrunCmd.Flags().StringVarP(&nodefeaturerule, "nodefeaturerule-file", "f", "", "Path to the NodeFeatureRule file to validate")
-	dryrunCmd.Flags().StringVarP(&nodefeature, "nodefeature-file", "n", "", "Path to the NodeFeature file to validate against")
-	err := dryrunCmd.MarkFlagRequired("nodefeaturerule-file")
-	if err != nil {
+	dryrunCmd.Flags().StringVarP(&rule, "rule-file", "f", "", "Path to the NodeFeatureRule or NodeFeatureGroup file to dry run")
+	dryrunCmd.Flags().StringVarP(&nodefeature, "nodefeature-file", "n", "", "Path to the NodeFeature file to dry run against")
+	if err := dryrunCmd.MarkFlagRequired("rule-file"); err != nil {
 		panic(err)
 	}
-	err = dryrunCmd.MarkFlagRequired("nodefeature-file")
-	if err != nil {
+	if err := dryrunCmd.MarkFlagRequired("nodefeature-file"); err != nil {
 		panic(err)
 	}
 }
--- a/cmd/kubectl-nfd/subcmd/validate.go
+++ b/cmd/kubectl-nfd/subcmd/validate.go
@@ -26,29 +26,27 @@ import (
 
 var validateCmd = &cobra.Command{
 	Use:   "validate",
-	Short: "Validate a NodeFeatureRule file",
-	Long:  `Validate a NodeFeatureRule file to ensure it is valid before applying it to a cluster`,
+	Short: "Validate a NodeFeatureRule or NodeFeatureGroup file",
+	Long:  `Validate a NodeFeatureRule or NodeFeatureGroup file to ensure it is valid before applying it to a cluster`,
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Validating NodeFeatureRule %s\n", nodefeaturerule)
-		err := kubectlnfd.ValidateNFR(nodefeaturerule)
-		if len(err) > 0 {
-			fmt.Printf("NodeFeatureRule %s is not valid\n", nodefeaturerule)
-			for _, e := range err {
+		fmt.Printf("Validating %s\n", rule)
+		errs := kubectlnfd.Validate(rule)
+		if len(errs) > 0 {
+			fmt.Printf("%s is not valid\n", rule)
+			for _, e := range errs {
 				cmd.PrintErrln(e)
 			}
-			// Return non-zero exit code to indicate failure
 			os.Exit(1)
 		}
-		fmt.Printf("NodeFeatureRule %s is valid\n", nodefeaturerule)
+		fmt.Printf("%s is valid\n", rule)
 	},
 }
 
 func init() {
 	RootCmd.AddCommand(validateCmd)
 
-	validateCmd.Flags().StringVarP(&nodefeaturerule, "nodefeaturerule-file", "f", "", "Path to the NodeFeatureRule file to validate")
-	err := validateCmd.MarkFlagRequired("nodefeaturerule-file")
-	if err != nil {
+	validateCmd.Flags().StringVarP(&rule, "rule-file", "f", "", "Path to the NodeFeatureRule or NodeFeatureGroup file to validate")
+	if err := validateCmd.MarkFlagRequired("rule-file"); err != nil {
 		panic(err)
 	}
 }
--- a/docs/usage/kubectl-plugin.md
+++ b/docs/usage/kubectl-plugin.md
@@ -22,7 +22,8 @@ sort: 10
 ## Overview
 
 The `kubectl` plugin `kubectl nfd` can be used to validate/dryrun and test
-NodeFeatureRule objects. It can be installed with the following command:
+NodeFeatureRule and NodeFeatureGroup objects. It can be installed with the
+following command:
 
 ```bash
 git clone https://github.com/kubernetes-sigs/node-feature-discovery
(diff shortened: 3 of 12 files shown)


cmd/kubectl-nfd/subcmd/dryrun.go
```diff
@@ -26,34 +26,50 @@ import (
...
+		return nil
+	},
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Evaluating NodeFeatureRule %q against NodeFeature %q\n", nodefeaturerule, nodefeature)
-		err := kubectlnfd.DryRun(nodefeaturerule, nodefeature)
-		if len(err) > 0 {
-			fmt.Printf("NodeFeatureRule %q is not valid for NodeFeature %q\n", nodefeaturerule, nodefeature)
-			for _, e := range err {
-				cmd.PrintErrln(e)
+		if nodefeaturerule != "" {
```
marquiz: How about auto-detect the Kind?

marquiz (commented):
/assign @ArangoGutierrez

cmd/kubectl-nfd/subcmd/dryrun.go
ozhuraki: @marquiz: Thanks, updated for dryrun, PTAL. If I misunderstood you, I can revert.

marquiz (commented):
@ozhuraki my review was a bit too concise. What I tried to say, don't add new commanline flags, just detect what type of object (NFR or NFG or if something else, error out) was given, probably renaming the longname of the flag to something more generic

ozhuraki:
@marquiz Thanks, done. PTAL.

pkg/kubectl-nfd/validate.go
```diff
@@ -28,22 +28,82 @@ import (
...
+		return validateNFRFromBytes(file)
+	case "NodeFeatureGroup":
+		return validateNFGFromBytes(file)
+	default:
+		return []error{fmt.Errorf("unsupported resource kind %q: must be NodeFeatureRule or NodeFeatureGroup", typeMeta.Kind)}
+	}
+}
+
+// ValidateNFG reads a file and checks if it is a valid NodeFeatureGroup file.
+func ValidateNFG(filepath string) []error {
```
marquiz: These are stale, dead code? Drop?

pkg/kubectl-nfd/validate.go
```diff
@@ -28,22 +28,82 @@ import (
...
-func ValidateNFR(filepath string) []error {
-	var err error
+// Validate reads a file and checks if it is a valid NodeFeatureRule or NodeFeatureGroup file.
+// The kind is detected automatically from the file content.
+func Validate(filepath string) []error {
+	file, err := os.ReadFile(filepath)
+	if err != nil {
+		return []error{fmt.Errorf("error reading file: %w", err)}
+	}
+	var typeMeta struct {
```
marquiz: nit: use `metav1.TypeMeta` type (from `k8s.io/apimachinery/pkg/apis/meta/v1`)

pkg/kubectl-nfd/test.go
```diff
@@ -67,19 +64,43 @@ func Test(nodefeaturerulepath, nodeName, kubeconfig string) []error {
...
+	}
+	if err = yaml.Unmarshal(resourceFile, &typeMeta); err != nil {
+		return []error{fmt.Errorf("error reading resource kind: %w", err)}
+	}
 
-	return errs
+	switch typeMeta.Kind {
+	case "NodeFeatureRule":
+		nfr := nfdv1alpha1.NodeFeatureRule{}
+		if err = yaml.Unmarshal(resourceFile, &nfr); err != nil {
```
marquiz: There's lot of repetition of the read-file-unmarshal-to-apitype. Could that be split into a common function like `func parseRuleFile(filepath string) metav1.Type`

Then in the Test/Validate/Dryrun use type switch to select the correct processor func like:

```golang
    switch  o := t.(type) {
    case fdv1alpha1.NodeFeatureRule:
        return processNodeFeatureRule(o, *features)
    case nfdv1alpha1.NodeFeatureGroup:
...
```

WDYT?

cmd/kubectl-nfd/subcmd/dryrun.go
```diff
@@ -26,34 +26,34 @@ import (
 
 var dryrunCmd = &cobra.Command{
 	Use:   "dryrun",
-	Short: "Process a NodeFeatureRule file against a NodeFeature file",
-	Long:  `Process a NodeFeatureRule file against a local NodeFeature file to dry run the rule against a node before applying it to a cluster`,
+	Short: "Process a NodeFeatureRule or NodeFeatureGroup file against a NodeFeature file",
+	Long:  `Process a NodeFeatureRule or NodeFeatureGroup file against a local NodeFeature file to dry run the rule against a node before applying it to a cluster`,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if nodefeaturerule == "" {
+			return fmt.Errorf("--nodefeaturerule-file must be specified")
```
marquiz: I'd suggest `s/nodefeaturerule-file/rule-file/` all over

cmd/kubectl-nfd/subcmd/dryrun.go
```diff
@@ -26,34 +26,34 @@ import (
...
+		}
+		return nil
+	},
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Printf("Evaluating NodeFeatureRule %q against NodeFeature %q\n", nodefeaturerule, nodefeature)
-		err := kubectlnfd.DryRun(nodefeaturerule, nodefeature)
-		if len(err) > 0 {
-			fmt.Printf("NodeFeatureRule %q is not valid for NodeFeature %q\n", nodefeaturerule, nodefeature)
-			for _, e := range err {
+		fmt.Printf("Evaluating %q against NodeFeature %q\n", nodefeaturerule, nodefeature)
```
marquiz: Similarly, e.g. 

```suggestion
		fmt.Printf("Evaluating %q against NodeFeature %q\n", rule, nodefeature)
```

marquiz (commented):
Thanks for the update @ozhuraki, looks quite good. Some further suggestions below

ozhuraki:
@marquiz Thanks for the help. Done, PTAL.

pkg/kubectl-nfd/validate.go
```diff
@@ -92,3 +117,30 @@ func ValidateNFR(filepath string) []error {
 
 	return validationErr
 }
+
+func parseRuleFile(filepath string) (obj interface{}) {
```
marquiz: nit: this is used by multiple subcommands so I'd suggest to move to a separate file, e.g. `common.go` or smth. WDYT?

marquiz (commented):
Thanks @ozhuraki looks pretty clean to me. One small wish below

ozhuraki:
@marquiz Thanks, done, PTAL.

pkg/kubectl-nfd/common.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"os"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/yaml"
+
+	nfdv1alpha1 "sigs.k8s.io/node-feature-discovery/api/nfd/v1alpha1"
+)
+
+func parseRuleFile(filepath string) (obj interface{}) {
```
ArangoGutierrez, line 29: When this fails, the `[]error` travels through the `interface{}` into each caller's type switch and lands in the `default` branch, so a plain read/parse failure is reported as `unsupported resource [error reading file: ...]: must be NodeFeatureRule or NodeFeatureGroup` — and a real kind mismatch gets the "must be ..." suffix twice. Returning two values (`func parseRuleFile(path string) (any, []error)`), or adding `case []error: return o` to the three switches, keeps the original message intact.

pkg/kubectl-nfd/common.go
```diff
@@ -0,0 +1,54 @@
...
+	file, err := os.ReadFile(filepath)
+	if err != nil {
+		return []error{fmt.Errorf("error reading file: %w", err)}
+	}
+	typeMeta := metav1.TypeMeta{}
+	if err = yaml.Unmarshal(file, &typeMeta); err != nil {
+		return []error{fmt.Errorf("error reading resource kind: %w", err)}
+	}
+	switch typeMeta.Kind {
```
ArangoGutierrez, line 38: Kind detection is now the entry point for all three subcommands and pkg/kubectl-nfd still has no unit tests; a small table-driven test over parseRuleFile (NFR, NFG, unsupported kind, malformed YAML, missing file) would pin these error paths down cheaply.

cmd/kubectl-nfd/subcmd/dryrun.go
```diff
@@ -26,34 +26,34 @@ import (
...
-	Short: "Process a NodeFeatureRule file against a NodeFeature file",
-	Long:  `Process a NodeFeatureRule file against a local NodeFeature file to dry run the rule against a node before applying it to a cluster`,
+	Short: "Process a NodeFeatureRule or NodeFeatureGroup file against a NodeFeature file",
+	Long:  `Process a NodeFeatureRule or NodeFeatureGroup file against a local NodeFeature file to dry run the rule against a node before applying it to a cluster`,
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		if rule == "" {
+			return fmt.Errorf("--rule-file must be specified")
+		}
+		return nil
+	},
```
ArangoGutierrez: Is PreRunE needed here? `MarkFlagRequired("rule-file")` like nodefeature-file below keeps one mechanism per command and marks the flag as required in --help output. Same pattern in test.go, and validate.go dropped MarkFlagRequired entirely.

ArangoGutierrez (commented):
This extends kubectl-nfd validate/dryrun/test to NodeFeatureGroup: the kind is auto-detected from TypeMeta behind a single --rule-file flag, file parsing is consolidated in parseRuleFile, and the NFG paths are built on ExecuteGroupRule/GroupRuleOutput, mirroring the existing NFR ones. The structure follows what was agreed in the earlier rounds; CI is green and the branch has no conflicts. The one thing I'd fix before merge is error propagation out of parseRuleFile (inline below) — a plain read/parse failure currently surfaces as "unsupported resource". Also, the PR description is empty: the --nodefeaturerule-file to --rule-file rename is user-visible, so a short description or release note for plugin users would help. With the error-path fix this is ready for lgtm.

ozhuraki:
@ArangoGutierrez
Thanks for the input, updated, PTAL.

ArangoGutierrez (commented):
Nice cleanup overall — kind auto-detection reads well and the parseRuleFile split addresses the earlier review comments. Six things worth a look before merge, mostly in the new test and the docs. I built the plugin from this branch and ran it, so the notes below come from actual output rather than from reading the diff.

1. This fixture uses `spec.rules`, but `NodeFeatureGroupSpec.Rules` is tagged `json:"featureGroupRules"`, so it decodes to a NodeFeatureGroup with zero rules. Running `validate` on this exact YAML prints no `Validating rule:` line, while the same file with `featureGroupRules` prints one. (pkg/kubectl-nfd/common_test.go:57)
2. These success cases assert only the returned Go type and never the decoded contents, so the test cannot tell a working parser from a broken one. I removed both `yaml.Unmarshal` calls from `parseRuleFile` so it returned empty structs, and all five subtests still passed — asserting something like `len(nfg.Spec.Rules) == 1` would cover the parse and also catch the fixture key above. (pkg/kubectl-nfd/common_test.go:97)
3. This loop never feeds `ruleOut.Vars` back into the features, so a group rule that backreferences an earlier one via `rule.matched` answers differently here than in the cluster — nfd-master.go:877 and node-validator.go:91 both call `features.InsertAttributeFeatures(RuleBackrefDomain, RuleBackrefFeature, ruleOut.Vars)` for that reason. On a two-rule group I get `did not match` from dryrun where the nfd-master loop matches; `processNodeFeatureRule` has the same gap, so it may be worth fixing both paths together. (pkg/kubectl-nfd/dryrun.go:39)
4. All three error cases assert only that a non-empty `[]error` came back, so a subtest passes regardless of which guard fired. Asserting the expected message per case would keep the missing-file, malformed-YAML and unsupported-kind paths distinct. (pkg/kubectl-nfd/common_test.go:104)
5. This line says the `kernel version` rule did not match, but `examples/nodefeature.yaml` carries `kernel.version` major `6` and the rule matches `major: {op: In, value: ["6"]}`. Running the documented command prints `Rule "kernel version" matched`. (docs/usage/kubectl-plugin.md:90)
6. The long-flag rename is not reflected in `docs/reference/plugin-commandline-reference.md`, which still documents `-f / --nodefeaturerule-file` for validate, test and dryrun (lines 28, 30, 52, 54, 61, 63). Following that page now gives `Error: unknown flag: --nodefeaturerule-file`, so it needs an update here or a deprecated alias for the old name. (cmd/kubectl-nfd/subcmd/validate.go:48)

pkg/kubectl-nfd/common_test.go
```diff
@@ -0,0 +1,108 @@
...
+			expected: &nfdv1alpha1.NodeFeatureRule{},
+		},
+		{
+			name: "NodeFeatureGroup",
+			content: `apiVersion: nfd.k8s-sigs.io/v1alpha1
+kind: NodeFeatureGroup
+metadata:
+  name: test-group
+spec:
+  rules:
```
ArangoGutierrez: This fixture uses `spec.rules`, but `NodeFeatureGroupSpec.Rules` is tagged `json:"featureGroupRules"`, so it decodes to a NodeFeatureGroup with zero rules. Running `validate` on this exact YAML prints no `Validating rule:` line, while the same file with `featureGroupRules` prints one.

pkg/kubectl-nfd/common_test.go
```diff
@@ -0,0 +1,108 @@
...
+			} else {
+				path = filepath.Join(t.TempDir(), "rule.yaml")
+				assert.NoError(t, os.WriteFile(path, []byte(test.content), 0644))
+			}
+
+			obj := parseRuleFile(path)
+
+			switch test.expected.(type) {
+			case *nfdv1alpha1.NodeFeatureRule:
+				assert.IsType(t, &nfdv1alpha1.NodeFeatureRule{}, obj)
```
ArangoGutierrez, line 97: These success cases assert only the returned Go type and never the decoded contents, so the test cannot tell a working parser from a broken one. I removed both `yaml.Unmarshal` calls from `parseRuleFile` so it returned empty structs, and all five subtests still passed — asserting something like `len(nfg.Spec.Rules) == 1` would cover the parse and also catch the fixture key above.

pkg/kubectl-nfd/dryrun.go
```diff
@@ -31,34 +31,53 @@ import (
...
 	var errs []error
-	nfr := nfdv1alpha1.NodeFeatureRule{}
-	nf := nfdv1alpha1.NodeFeature{}
 
-	nfrFile, err := os.ReadFile(nodefeaturerulepath)
-	if err != nil {
-		return []error{fmt.Errorf("error reading NodeFeatureRule file: %w", err)}
+	for _, rule := range nodeFeatureGroup.Spec.Rules {
+		fmt.Println("Processing rule: ", rule.Name)
+		ruleOut, err := nodefeaturerule.ExecuteGroupRule(&rule, &nodeFeature.Features, true)
```
ArangoGutierrez, line 39: This loop never feeds `ruleOut.Vars` back into the features, so a group rule that backreferences an earlier one via `rule.matched` answers differently here than in the cluster — nfd-master.go:877 and node-validator.go:91 both call `features.InsertAttributeFeatures(RuleBackrefDomain, RuleBackrefFeature, ruleOut.Vars)` for that reason. On a two-rule group I get `did not match` from dryrun where the nfd-master loop matches; `processNodeFeatureRule` has the same gap, so it may be worth fixing both paths together.

pkg/kubectl-nfd/common_test.go
```diff
@@ -0,0 +1,108 @@
...
+			case *nfdv1alpha1.NodeFeatureGroup:
+				assert.IsType(t, &nfdv1alpha1.NodeFeatureGroup{}, obj)
+			case []error:
+				assert.IsType(t, []error{}, obj)
+				errs, ok := obj.([]error)
+				assert.True(t, ok)
+				assert.NotEmpty(t, errs)
```
ArangoGutierrez: All three error cases assert only that a non-empty `[]error` came back, so a subtest passes regardless of which guard fired. Asserting the expected message per case would keep the missing-file, malformed-YAML and unsupported-kind paths distinct.

docs/usage/kubectl-plugin.md
```diff
@@ -34,37 +35,60 @@ mv ./bin/kubectl-nfd ${KUBECTL_PATH}
...
 vendor.io/my-sample-feature=true
-NodeFeatureRule "examples/nodefeaturerule.yaml" is valid for NodeFeature "examples/nodefeature.yaml"
+"examples/nodefeaturerule.yaml" is valid for NodeFeature "examples/nodefeature.yaml"
+```
+
+```bash
+$ kubectl nfd dryrun -f examples/nodefeaturegroup.yaml -n examples/nodefeature.yaml
+Evaluating "examples/nodefeaturegroup.yaml" against NodeFeature "examples/nodefeature.yaml"
+Processing rule:  kernel version
+Rule "kernel version" did not match
```
ArangoGutierrez: This line says the `kernel version` rule did not match, but `examples/nodefeature.yaml` carries `kernel.version` major `6` and the rule matches `major: {op: In, value: ["6"]}`. Running the documented command prints `Rule "kernel version" matched`.

cmd/kubectl-nfd/subcmd/validate.go
```diff
@@ -26,29 +26,27 @@ import (
...
 	},
 }
 
 func init() {
 	RootCmd.AddCommand(validateCmd)
 
-	validateCmd.Flags().StringVarP(&nodefeaturerule, "nodefeaturerule-file", "f", "", "Path to the NodeFeatureRule file to validate")
-	err := validateCmd.MarkFlagRequired("nodefeaturerule-file")
-	if err != nil {
+	validateCmd.Flags().StringVarP(&rule, "rule-file", "f", "", "Path to the NodeFeatureRule or NodeFeatureGroup file to validate")
```
ArangoGutierrez, line 48: The long-flag rename is not reflected in `docs/reference/plugin-commandline-reference.md`, which still documents `-f / --nodefeaturerule-file` for validate, test and dryrun (lines 28, 30, 52, 54, 61, 63). Following that page now gives `Error: unknown flag: --nodefeaturerule-file`, so it needs an update here or a deprecated alias for the old name.

ozhuraki:
@ArangoGutierrez
Thanks for the input, updated, PTAL.

Issue #2046, opened 2025-02-14
Publish x86-64 microarchitecture level instead of cpuid features


boniek83:
**What would you like to be added**:
add x86-64  microarchitecture level label and blacklist all other cpuid features by default except for cpuid features newer than x86-64v4.
**Why is this needed**:
Less clutter, more useful labels.

marquiz:
@boniek83 I agree with the issue about clutter/bloat. Not sure how we should address this, though. I'm not sure we can meaningfully map cpu flags to certain uarch level. Also, I'm not very fond of starting to maintain a mapping/database of cpu features.

Some options come to my mind:
 
1. We already have the default denylist -> extend it to drop more features
2. Change from default denylist to allowlist -> only enable some hand-picked cpuid labels by default
3. Additionally, we could support globbing for the deny/allow list to make configuration a tad easier

Thoughts? 
/cc @ozhuraki

boniek83:
Wiki has decent article about it: https://en.wikipedia.org/wiki/X86-64 (Microarchitecture levels) and that is how I do it now using nfd rules. It seems like a minimal and one time effort per uarch level and it's not like there is new uarch level every year. I personally don't are at all about cpu features if I have uarch levels. I wonder if adding all flags to default denylist won't make my uarch level rules stop working though.

marquiz:
The list on the wiki page is not exhaustive and not all all cpuid flags map to any particular uarch level. Sleeping over this I think @boniek83 your original suggestion would probably be good, just add the old uninteresting features to the default blacklist.

One possible bonus how to improve the usability would be to make it possible to users more easily to disable/enable certain flags, like

```yaml
attributeList:
 - "-AVX2"
 - "+TDX_GUEST"
```
...probably a bad ui example but you get the idea.

One possiblity (for the bonus feature) would be to deprecate the `attributeBlacklist` and `attributeWhitelist` configs and replace with deny list` and  allow list configs which could be used to modify the built-in default list of allowed cpuid attributes.

```yaml
deny:
- "AVX2"
allow:
- "TDX_GUEST"
```

jcpunk:
/remove-lifecycle rotten

I'd still like to see some version of this.

faganihajizada:
Hi @marquiz @ozhuraki 

I can help with this one.

[...]

PR #2490, opened 2026-04-15
feat(cpu): add x86-64 level CPUID flags

faganihajizada:
## Summary

Adds x86-64 microarchitecture level (v1-v4) as cumulative CPUID flags using the `klauspost/cpuid/v2` `X64Level()` API that is already vendored at v2.3.0.
A v4-capable node produces:

```
feature.node.kubernetes.io/cpu-cpuid.X86_64_V1=true
feature.node.kubernetes.io/cpu-cpuid.X86_64_V2=true
feature.node.kubernetes.io/cpu-cpuid.X86_64_V3=true
feature.node.kubernetes.io/cpu-cpuid.X86_64_V4=true
```

This enables scheduling workloads that require a minimum microarchitecture level
without numeric comparisons:

```yaml
nodeSelector:
  feature.node.kubernetes.io/cpu-cpuid.X86_64_V3: "true"
```

The flags are also available for NodeFeatureRule matching via cpu.cpuid.

## Why

Addresses #2046. Individual CPUID feature flags create clutter and don't map cleanly to the x86-64 psABI-defined microarchitecture levels that users actually care about for scheduling. The cumulative flag approach was discussed and agreed upon in the issue thread.

### DIFF
--- a/source/cpu/cpu_test.go
+++ b/source/cpu/cpu_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package cpu
 
 import (
+	"fmt"
+	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
@@ -31,5 +33,31 @@ func TestCpuSource(t *testing.T) {
 
 	assert.Nil(t, err, err)
 	assert.Empty(t, l)
+}
 
+func TestGetCpuidFlags_X86_64Levels(t *testing.T) {
+	if runtime.GOARCH != "amd64" {
+		t.Skip("x86-64 microarchitecture levels only apply on amd64")
+	}
+
+	flags := getCpuidFlags()
+	flagSet := make(map[string]bool, len(flags))
+	for _, f := range flags {
+		flagSet[f] = true
+	}
+
+	// Every x86-64 CPU is at least level 1
+	assert.True(t, flagSet["X86_64_V1"], "expected X86_64_V1 flag on amd64")
+
+	// Levels must be cumulative: if V(n) is present, all V(1)..V(n-1) must be too
+	for level := 4; level >= 2; level-- {
+		flag := fmt.Sprintf("X86_64_V%d", level)
+		if !flagSet[flag] {
+			continue
+		}
+		for i := 1; i < level; i++ {
+			lower := fmt.Sprintf("X86_64_V%d", i)
+			assert.True(t, flagSet[lower], "X86_64_V%d present but %s missing", level, lower)
+		}
+	}
 }
--- a/source/cpu/cpu_test.go
+++ b/source/cpu/cpu_test.go
@@ -21,6 +21,7 @@ import (
 	"runtime"
 	"testing"
 
+	"github.com/klauspost/cpuid/v2"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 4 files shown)


ozhuraki (approved):
LGTM

ArangoGutierrez (commented):
Leaving the approval to @marquiz 
/lgtm

faganihajizada:
@marquiz could you please check this one? thanks!

source/cpu/cpuid_amd64.go
```diff
@@ -17,14 +17,23 @@ limitations under the License.
...
 	"github.com/klauspost/cpuid/v2"
 )
 
 // getCpuidFlags returns feature names for all the supported CPU features.
 func getCpuidFlags() []string {
-	return cpuid.CPU.FeatureSet()
+	flags := cpuid.CPU.FeatureSet()
+
+	if level := cpuid.CPU.X64Level(); level > 0 {
+		for i := 1; i <= level; i++ {
```
ArangoGutierrez: The v1 feature set (CMOV/CX8/FPU/FXSR/MMX/SSE/SSE2) is the amd64 baseline, so every amd64 node will carry `cpu-cpuid.X86_64_V1=true`, which adds no scheduling signal beyond `kubernetes.io/arch=amd64`. Given #2046 is motivated by label clutter, consider starting the loop at `i := 2`, or keep V1 deliberately and say so in the docs row (e.g. as a "level detection succeeded" marker).

source/cpu/cpu_test.go
```diff
@@ -31,5 +33,31 @@ func TestCpuSource(t *testing.T) {
...
+	for level := 4; level >= 2; level-- {
+		flag := fmt.Sprintf("X86_64_V%d", level)
+		if !flagSet[flag] {
+			continue
+		}
+		for i := 1; i < level; i++ {
+			lower := fmt.Sprintf("X86_64_V%d", i)
+			assert.True(t, flagSet[lower], "X86_64_V%d present but %s missing", level, lower)
+		}
+	}
```
ArangoGutierrez: The cumulative property holds by construction (the implementation appends 1..level unconditionally), so this loop only catches a future switch to emitting just the top level. The failure mode that matters, emitting a level above what `X64Level()` reports (e.g. a hardcoded 1..4 loop), passes this test on any machine. Asserting equality against the same API pins both presence and absence:

```go
level := cpuid.CPU.X64Level()
for i := 1; i <= 4; i++ {
	flag := fmt.Sprintf("X86_64_V%d", i)
	assert.Equal(t, i <= level, flagSet[flag], "flag %s vs detected level %d", flag, level)
}
```

(`github.com/klauspost/cpuid/v2` builds on all arches, and the GOARCH skip above already guards execution.)

ArangoGutierrez (commented):
Verified the level mapping: this delegates to the vendored klauspost/cpuid v2.3.0 `X64Level()`, whose sets match the psABI definition exactly (v2 = +CX16/LAHF/POPCNT/SSE3/SSE4.1/SSE4.2/SSSE3, v3 = +AVX/AVX2/BMI1/BMI2/F16C/FMA/LZCNT/MOVBE/OSXSAVE, v4 = +AVX512F/BW/CD/DQ/VL), and the descending 4-3-2-1 check order means an emitted level implies all lower ones, so the cumulative V1..Vn emission is correct and the synthetic names cannot collide with any real flag the library emits. The flag shape follows what #2046 converged on, with the denylist-trimming half of that issue left as the agreed follow-up. Non-amd64 builds are untouched (per-arch files plus stub keep the new test compiling everywhere), the new flags clear the default blacklist, and they reach NodeFeatureRule matching via `cpu.cpuid` as advertised. CI is green and the branch has no conflicts; merge is only waiting on approval. Two notes inline, a test-strength gap around over-claimed levels and a question on V1 label noise; neither is structural, so this looks mergeable once the test is tightened or the gap consciously accepted.

source/cpu/cpu_test.go
faganihajizada: Done in the latest commit. The `amd64` test now asserts equality against `X64Level()` as suggested

source/cpu/cpuid_amd64.go
faganihajizada: I kept V1 and documented it in the features row rather than starting at `i := 2`. Its presence doubles as a "level detection succeeded" marker: X64Level() returns 0 when it can't classify the CPU, so "nothing emitted" vs "V1 present" is a real distinction, and keeping it preserves the clean `emit exactly the levels X64Level() report` contract. The docs row now notes V1 is the `amd64` baselne and explains that rationale.

ArangoGutierrez:
/lgtm
/approve

Reviewed: cumulative X86_64_V1..V4 logic is correct (psABI model, matches the design agreed in #2046), the mapping is arch-neutral + unit-testable, and the tests are strong — they pin both presence and absence against `X64Level()` and add a host-independent table that would catch an over-claimed level. Docs are honest about V1's low signal. Nice work, @faganihajizada.

PR #2495, opened 2026-04-23
feat(network): add aggregated SR-IOV VF count labels

ashokpariya0:
Expose total SR-IOV VF capacity and configured VF count across all network devices on the node.

Current SR-IOV labels(https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/docs/usage/features.md?plain=1#L194-L200)
 (sriov.capable, sriov.configured) are boolean and don't expose the actual VF counts.

**New Labels**

- **network-sriov.total_vfs**: Total SR-IOV VF capacity across all NICs
- **network-sriov.num_vfs**: Total configured (active) VFs across all NICs

**Behavior Changes**

Before this PR Only boolean labels were exposed:

- feature.node.kubernetes.io/network-sriov.capable: "true"
- feature.node.kubernetes.io/network-sriov.configured: "true"

**After this PR**

- feature.node.kubernetes.io/network-sriov.capable=true
- feature.node.kubernetes.io/network-sriov.configured=true
- feature.node.kubernetes.io/network-sriov.total_vfs=254
- feature.node.kubernetes.io/network-sriov.num_vfs=4

**Key Points**

- **total_vfs** is capacity (sum of all NICs)
- **num_vfs** is runtime configured VFs
- num_vfs is aggregated independently of total_vfs
- Labels are only set when meaningful:
        - capable → if any NIC supports SR-IOV
        - configured → if any VF is configured

**Tests**

Added table-driven tests covering:

- Single NIC
- Multi-NIC aggregation
- Devices with sriov_totalvfs=0
- Malformed sysfs values
- Mixed valid/invalid devices
- No SR-IOV present

**Docs**

Updated docs/usage/features.md to include:
network-sriov.total_vfs
network-sriov.num_vfs

with a note that values are aggregated across all physical NICs.

**Notes**
Backward compatible: existing labels unchanged
New labels are additive

### DIFF
--- a/source/network/network.go
+++ b/source/network/network.go
@@ -76,24 +76,54 @@ func (s *networkSource) Priority() int { return 0 }
 func (s *networkSource) GetLabels() (source.FeatureLabels, error) {
 	labels := source.FeatureLabels{}
 	features := s.GetFeatures()
+	totalVFsAll := 0
+	numVFsAll := 0
+	sriovCapable := false
 
 	for _, dev := range features.Instances[DeviceFeature].Elements {
 		attrs := dev.Attributes
-		for attr, feature := range map[string]string{
-			"sriov_totalvfs": "sriov.capable",
-			"sriov_numvfs":   "sriov.configured"} {
 
-			if v, ok := attrs[attr]; ok {
-				t, err := strconv.Atoi(v)
-				if err != nil {
-					klog.ErrorS(err, "failed to parse sriov attribute", "attributeName", attr, "deviceName", attrs["name"])
-					continue
-				}
-				if t > 0 {
-					labels[feature] = true
-				}
+		totalVFs := 0
+		numVFs := 0
+
+		// Parse total VFs
+		if v, ok := attrs["sriov_totalvfs"]; ok {
+			t, err := strconv.Atoi(v)
+			if err != nil {
+				klog.ErrorS(err, "failed to parse sriov_totalvfs", "deviceName", attrs["name"], "value", v)
+			} else {
+				totalVFs = t
 			}
 		}
+
+		// Parse configured VFs
+		if v, ok := attrs["sriov_numvfs"]; ok {
+			t, err := strconv.Atoi(v)
+			if err != nil {
+				klog.ErrorS(err, "failed to parse sriov_numvfs", "deviceName", attrs["name"], "value", v)
+			} else {
+				numVFs = t
+			}
+		}
+
+		if totalVFs > 0 {
+			sriovCapable = true
+			totalVFsAll += totalVFs
+		}
+
+		if numVFs > 0 {
+			numVFsAll += numVFs
+		}
+	}
+
+	if sriovCapable {
+		labels["sriov.capable"] = true
+		labels["sriov.total_vfs"] = totalVFsAll
+	}
+
+	if numVFsAll > 0 {
+		labels["sriov.configured"] = true
+		labels["sriov.num_vfs"] = numVFsAll
 	}
 	return labels, nil
 }
--- a/source/network/network_test.go
+++ b/source/network/network_test.go
@@ -20,6 +20,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	nfdv1alpha1 "sigs.k8s.io/node-feature-discovery/api/nfd/v1alpha1"
 )
 
 func TestNetworkSource(t *testing.T) {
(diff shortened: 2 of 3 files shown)


source/network/network.go
```diff
@@ -77,24 +77,55 @@ func (s *networkSource) GetLabels() (source.FeatureLabels, error) {
...
+				numVFs = t
+			}
+		}
+
+		// Detect SR-IOV capable NIC
+		if totalVFs > 0 {
+			sriovCapable = true
+			totalVFsAll += totalVFs
+			numVFsAll += numVFs
+		}
```
ArangoGutierrez, line 116: Small behavioral note: previously `sriov.configured` was set per-device when `sriov_numvfs > 0`, independent of `sriov_totalvfs`. Here `numVFsAll` is only accumulated when this device's `totalVFs > 0`. In practice sysfs should guarantee both together so it's probably fine — but could we either mention the tightening in the PR description, or accumulate `numVFsAll` independently of the `totalVFs > 0` guard?

source/network/network.go
```diff
@@ -77,24 +77,55 @@ func (s *networkSource) GetLabels() (source.FeatureLabels, error) {
...
+
+		totalVFs := 0
+		numVFs := 0
+
+		// Parse total VFs
+		if v, ok := attrs["sriov_totalvfs"]; ok {
+			t, err := strconv.Atoi(v)
+			if err != nil {
+				klog.ErrorS(err, "failed to parse sriov_totalvfs", "deviceName", attrs["name"])
+			} else {
```
ArangoGutierrez, line 94: Minor: including the raw value in the error would help triage a malformed sysfs read — `klog.ErrorS(err, "...", "deviceName", attrs["name"], "value", v)`.

source/network/network.go
```diff
@@ -77,24 +77,55 @@ func (s *networkSource) GetLabels() (source.FeatureLabels, error) {
...
+
+	// Apply labels ONLY if SR-IOV is supported
+	if sriovCapable {
+		labels["sriov.capable"] = true
+		labels["sriov.total_vfs"] = totalVFsAll
+
+		if numVFsAll > 0 {
+			labels["sriov.configured"] = true
+			labels["sriov.vf_count"] = numVFsAll
+		}
```
ArangoGutierrez: Naming question, non-blocking: sysfs uses `sriov_totalvfs` and `sriov_numvfs`. `sriov.total_vfs` mirrors the kernel name nicely — would `sriov.num_vfs` (instead of `sriov.vf_count`) be worth considering for parallel naming before these labels become API?

ArangoGutierrez (changes requested):
Thanks for this — the aggregation is a useful addition and the code is close.

A few things I'd like to see before I can approve:

- Docs: please add rows for `network-sriov.total_vfs` and `network-sriov.vf_count` in `docs/usage/features.md`, ideally with a note that they're aggregated across all physical NICs.
- Tests: `source/network/network_test.go` only covers the empty-features case today. A small table-driven test exercising single-NIC aggregation, multi-NIC sum, a device with `sriov_totalvfs=0` excluded, and a malformed-value error path would catch a lot.
- `sriov.configured` semantics — see inline note.

Left a small naming question inline as well — non-blocking, but worth discussing before the new labels become API.

source/network/network.go
ashokpariya0: updated as suggested.

source/network/network.go
ashokpariya0: good point.

I’ve updated the implementation to accumulate numVFsAll independently of the totalVFs > 0 check, so sriov.configured semantics remain consistent with the previous behavior (i.e., based solely on sriov_numvfs > 0).

source/network/network.go
ashokpariya0: updated as suggested.

ashokpariya0:
> Thanks for this — the aggregation is a useful addition and the code is close.
> 
> A few things I'd like to see before I can approve:
> 
> * Docs: please add rows for `network-sriov.total_vfs` and `network-sriov.vf_count` in `docs/usage/features.md`, ideally with a note that they're aggregated across all physical NICs.
> * Tests: `source/network/network_test.go` only covers the empty-features case today. A small table-driven test exercising single-NIC aggregation, multi-NIC sum, a device with `sriov_totalvfs=0` excluded, and a malformed-value error path would catch a lot.
> * `sriov.configured` semantics — see inline note.
> 
> Left a small naming question inline as well — non-blocking, but worth discussing before the new labels become API.

>
Thanks for the detailed review — I’ve addressed all the points
update pr descriptions with all the details.

**Docs:** Added entries for network-sriov.total_vfs and network-sriov.num_vfs in docs/usage/features.md, including a note that both values are aggregated across all physical NICs.

**Tests:** Added a table-driven test in source/network/network_test.go covering:
single NIC aggregation
multi-NIC aggregation
devices with sriov_totalvfs=0
malformed sysfs values
mixed valid/invalid devices
no SR-IOV case

**sriov.configured semantics:** Updated the implementation to accumulate numVFsAll independently of the totalVFs > 0 check, preserving previous behavior (based solely on sriov_numvfs > 0). Also added test coverage and clarified this in the PR description.

**Naming:** Based on the suggestion, went with sriov.num_vfs (instead of vf_count) to keep naming consistent with sriov.total_vfs and closer to sysfs (sriov_numvfs).

Please take another look

ashokpariya0:
@ArangoGutierrez Please take another look

ArangoGutierrez (commented):
This swaps the per-attribute boolean loop for per-device aggregation and adds `sriov.total_vfs` / `sriov.num_vfs` alongside the existing `capable`/`configured` labels. I checked the previous round's items and they're all addressed: `configured` accumulates independently of the `totalvfs` guard, the error logs carry the raw value, docs rows are in, and the tests are table-driven; int label values also flow cleanly through `GetFeatureLabels` (stringified and validated). Two small items from me, both in the test file: it isn't gofmt-clean so `make gofmt-verify` fails, and the `sriov_numvfs` parse-error branch is untested. The real CI hasn't run yet (`needs-ok-to-test`), so the gofmt failure is currently invisible on the checks tab. Fix the formatting, optionally add the one test case, and this is ready for ok-to-test and lgtm consideration.

source/network/network_test.go
```diff
@@ -33,3 +34,153 @@ func TestNetworkSource(t *testing.T) {
...
+	tests := []struct {
+		name     string
+		devices  []map[string]string
+		expected map[string]interface{}
+	}{
+		{
+			name: "single NIC",
+			devices: []map[string]string{
+				{
+					"name":            "eth0",
```
ArangoGutierrez: `make gofmt-verify` fails on this file: the `"name":` keys here and on lines 64, 69, 85, 99, 113 and 118 have one extra space (gofmt aligns the value column to the longest key in the literal). A single `gofmt -w source/network/network_test.go` fixes them all — best done now since the verify job will go red once this gets ok-to-test.

source/network/network_test.go
```diff
@@ -33,3 +34,153 @@ func TestNetworkSource(t *testing.T) {
...
+					"sriov_numvfs":   "4",
+				},
+			},
+			expected: map[string]interface{}{
+				"sriov.configured": true,
+				"sriov.num_vfs":    4,
+			},
+		},
+		{
+			name: "malformed totalvfs",
```
ArangoGutierrez, line 96: The `sriov_numvfs` parse-error branch in network.go is the one path without coverage — every malformed case here targets `sriov_totalvfs`. One more entry with a valid `sriov_totalvfs` and `sriov_numvfs: "garbage"`, expecting only `sriov.capable` and `sriov.total_vfs`, would close that gap.

ashokpariya0:
Thanks @ArangoGutierrez. Both addressed in the latest push:

Ran gofmt -w source/network/network_test.go — make gofmt-verify is clean now.
Added a malformed numvfs case (valid sriov_totalvfs, sriov_numvfs: "garbage") asserting only sriov.capable and sriov.total_vfs are set, which covers the previously-untested sriov_numvfs parse-error branch.

PTAL.

source/network/network_test.go
ashokpariya0: Done.

source/network/network_test.go
ashokpariya0: Updated as suggested, [change](https://github.com/kubernetes-sigs/node-feature-discovery/compare/faf22f3e486c6648efdfed8c597b2b8a67891192..8ad291de7a3291b6f68e3abca89cc7764599cedd)

ashokpariya0:
@ArangoGutierrez could you please check, all review comments are addressed.

ashokpariya0:
@ArangoGutierrez All tests have passed.

ashokpariya0:
can we  please merge this PR? @ArangoGutierrez

ashokpariya0:
Hi @ArangoGutierrez, just following up on this PR. All tests have passed, and the PR is ready for merge

PR #2507, opened 2026-05-13
template + custom/api: minor input-validation tweaks

ArangoGutierrez:
A few small input-validation tweaks I noticed while reading around the
template and custom-source decoding code.

**`pkg/apis/nfd/template`:**

1. Cap `labelsTemplate` / `varsTemplate` at 64 KiB in `NewHelper` before
   calling `text/template.Parse`. `text/template` parses recursively and
   very large inputs do a lot of recursive work before any rule even
   matches. Adds a typed `ErrTemplateTooLarge` so callers can tell this
   apart from generic parse errors. Real templates in our examples are
   tens to hundreds of bytes; 64 KiB is well above realistic use.

2. Drop `env` / `expandenv` / `getHostByName` from sprig's FuncMap.
   These three are the only sprig builtins that read process
   environment or outbound DNS, and aren't useful for `NodeFeatureRule`
   label expansion. The rest of sprig (date, random, string, list,
   etc.) stays available. Templates calling `{{env ...}}` /
   `{{expandenv ...}}` / `{{getHostByName ...}}` will now fail to parse
   with function-not-defined — worth a release note.

**`source/custom/api`:**

3. Cap JSON nesting depth in `MatchExpression` / `MatchExpressionSet`
   `UnmarshalJSON`. Go 1.26's json decoder caps depth internally (~10k)
   so the immediate risk is gone on current toolchains, but we fail
   earlier with a clearer error tied to the NFD type and stay covered
   against older Go toolchains. Gated at 100 levels (real custom-source
   configs are 2-3 levels deep at most). Adds typed `ErrJSONTooDeep`.

Tests in `pkg/apis/nfd/template/template_test.go` and
`source/custom/api/expression_test.go`. `make test` clean.

### DIFF
--- a/pkg/apis/nfd/template/template.go
+++ b/pkg/apis/nfd/template/template.go
@@ -18,6 +18,7 @@ package template
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"strings"
 	"text/template"
@@ -29,8 +30,36 @@ type Helper struct {
 	template *template.Template
 }
 
+// funcMap returns sprig's text-template FuncMap with the host-environment
+// functions removed. env / expandenv / getHostByName are the only sprig
+// builtins that read process state or outbound DNS — leaving the rest
+// (date, random, string, list, etc.) available for legitimate use in
+// label expansion.
+func funcMap() template.FuncMap {
+	m := sprig.TxtFuncMap()
+	for _, name := range []string{"env", "expandenv", "getHostByName"} {
+		delete(m, name)
+	}
+	return m
+}
+
+// MaxTemplateSize bounds the input to text/template.Parse to prevent stack-
+// overflow aborts from recursive descent in text/template/parse. Per-action
+// parser stack cost is ~1.7 KiB and minimum action width is ~5 bytes, so a
+// 64 KiB cap bounds total stack growth at ~22 MiB — well below Go's 1 GiB
+// goroutine stack ceiling.
+const MaxTemplateSize = 64 * 1024
+
+// ErrTemplateTooLarge is returned by NewHelper when the input template
+// exceeds MaxTemplateSize. Exported so callers can distinguish this
+// from generic parse errors via errors.Is.
+var ErrTemplateTooLarge = errors.New("template exceeds maximum allowed size")
+
 func NewHelper(name string) (*Helper, error) {
-	tmpl := template.New("").Funcs(sprig.FuncMap()).Option("missingkey=error")
+	if len(name) > MaxTemplateSize {
+		return nil, fmt.Errorf("%w: %d bytes > %d limit", ErrTemplateTooLarge, len(name), MaxTemplateSize)
+	}
+	tmpl := template.New("").Funcs(funcMap()).Option("missingkey=error")
 	tmpl, err := tmpl.Parse(name)
 	if err != nil {
 		return nil, fmt.Errorf("invalid template: %w", err)
--- a/pkg/apis/nfd/template/template_test.go
+++ b/pkg/apis/nfd/template/template_test.go
@@ -1,0 +1,111 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package template
+
+import (
+	"errors"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+	stdtemplate "text/template"
+	"time"
+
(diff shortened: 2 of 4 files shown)


PiotrProkop:
@ArangoGutierrez great catch, should harden NFD against corrupted inputs 👍 
/lgtm

pkg/apis/nfd/template/template.go
```diff
@@ -29,8 +30,36 @@ type Helper struct {
...
 }
 
+// funcMap returns sprig's text-template FuncMap with the host-environment
+// functions removed. env / expandenv / getHostByName are the only sprig
+// builtins that read process state or outbound DNS — leaving the rest
+// (date, random, string, list, etc.) available for legitimate use in
+// label expansion.
+func funcMap() template.FuncMap {
+	m := sprig.TxtFuncMap()
+	for _, name := range []string{"env", "expandenv", "getHostByName"} {
```
ArangoGutierrez, line 40: Noting for reviewers: sprig also ships `HermeticTxtFuncMap()`, which drops these three plus the nondeterministic group (`date*`, `now`, `rand*`, `uuidv4`). I went with the narrower deny-list so existing templates using `rand*`/`date` keep working; if the hermetic preset reads as the safer default I can switch, at the cost of a wider compat break.

ArangoGutierrez (commented):
Self-review notes for reviewers. The user-visible piece is the FuncMap trim: templates calling `env`/`expandenv`/`getHostByName` now fail at parse -- in nfd-master the rule errors out and its labels are dropped, and `kubectl-nfd` validate reports it too since `pkg/apis/nfd/validate` shares `NewHelper`. The customization guide still says Sprig functions "and all their functionality" are available (`docs/usage/customization-guide.md`, templating section), so I owe a doc tweak covering the three removed functions and the 64 KiB cap -- will push it to this PR unless a follow-up is preferred. The JSON depth gate should be behavior-neutral for real configs (2-3 levels deep): the pre-scan only rejects nesting beyond 100 levels and defers every other decode error to `json.Unmarshal`. CI: no conflicts, everything green except `pull-node-feature-discovery-build-image-cross-generic`, which has failed across several retests -- still checking whether that is infra flake or something this change tickles, so holding merge until it is resolved.

pkg/apis/nfd/template/template_test.go
```diff
@@ -0,0 +1,111 @@
...
+	// is what keeps parse latency predictable. The child is expected to NOT
+	// complete the parse cleanly and to exit quickly rather than hang. If a
+	// future Go release changes that behavior, this test will flip and the
+	// application-layer cap may become tunable rather than load-bearing.
+	cmd := exec.Command(os.Args[0], "-test.run=^$")
+	cmd.Env = append(os.Environ(), "NFD_TEMPLATE_LARGE_INPUT=1")
+	start := time.Now()
+	_, err := cmd.CombinedOutput()
+	elapsed := time.Since(start)
+	require.Error(t, err, "child should not complete the parse cleanly")
```
ArangoGutierrez, line 83: For reviewers weighing CI cost: this canary pins current upstream behavior -- the child grows its goroutine stack toward the 1 GiB ceiling before the runtime aborts, making it the most expensive test in the package (hence the `-short` skip). The boundary tests above already cover `NewHelper`'s contract, so if this looks too heavy for CI I am fine dropping it.

pkg/apis/nfd/template/template_test.go
```diff
@@ -0,0 +1,111 @@
...
+func TestNewHelper_DropsHostEnvFuncs(t *testing.T) {
+	// env / expandenv / getHostByName read process env and outbound DNS;
+	// not useful for NodeFeatureRule label expansion.
+	for _, tmpl := range []string{
+		`{{env "PATH"}}`,
+		`{{expandenv "$PATH"}}`,
+		`{{getHostByName "localhost"}}`,
+	} {
+		_, err := NewHelper(tmpl)
+		require.Error(t, err, "expected %q to fail", tmpl)
```
ArangoGutierrez, line 96: `require.Error` here passes for any parse failure, not specifically function-not-defined, so it would stay green if these templates ever started failing for an unrelated reason. Pinning the message ties the test to the FuncMap removal:

```suggestion
		require.ErrorContains(t, err, "not defined", "expected %q to fail", tmpl)
```

source/custom/api/expression_test.go
```diff
@@ -17,11 +17,44 @@ limitations under the License.
...
 	"github.com/stretchr/testify/assert"
 )
 
+func TestMatchExpression_RejectsDeepNesting(t *testing.T) {
+	deep := strings.Repeat("[", MaxJSONDepth+1) + strings.Repeat("]", MaxJSONDepth+1)
+	var m MatchExpression
+	err := json.Unmarshal([]byte(deep), &m)
+	assert.Error(t, err)
+	assert.True(t, errors.Is(err, ErrJSONTooDeep),
+		"expected ErrJSONTooDeep, got %v", err)
```
ArangoGutierrez, line 34: Self-nit: testify's `ErrorIs` collapses the `Error` + manual `errors.Is` pair -- same pattern applies to the set test below, `assert.False(errors.Is(...))` -> `assert.NotErrorIs` in the at-cap test, and the `require.True(errors.Is(...))` pairs in `template_test.go`.

```suggestion
	assert.ErrorIs(t, err, ErrJSONTooDeep)
```

PR #2508, opened 2026-05-13
custom/api: cap JSON nesting depth in MatchExpression(.Set) Unmarshal

ArangoGutierrez:
While reading `expression.go` I noticed `MatchExpression` and
`MatchExpressionSet` `UnmarshalJSON` hand the raw bytes straight to
`encoding/json` with an `interface{}` target.

Go 1.26's json decoder has its own internal depth cap (~10k), so the
immediate risk is gone on current toolchains, but we can fail earlier
with a clearer error tied to the NFD type, and stay covered if anyone
builds against an older Go toolchain.

Adds a small depth-counting `Token` pre-pass before `json.Unmarshal`,
gated at 100 levels (real custom-source configs are 2-3 levels deep at
most). Returns a typed `ErrJSONTooDeep` so callers can distinguish it
from a generic parse error.

Tests in `source/custom/api/expression_test.go`.

### DIFF
--- a/source/custom/api/expression.go
+++ b/source/custom/api/expression.go
@@ -17,12 +17,58 @@ limitations under the License.
 package api
 
 import (
+	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"regexp"
 	"strconv"
 	"strings"
 )
+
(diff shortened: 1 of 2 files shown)


ArangoGutierrez:
Folded into #2507.

PR #2509, opened 2026-05-15
topology-updater: serve per-pod lookups from a node-scoped Pod informer

alaypatel07:
PodResourcesScanner.isWatchable() previously issued a live GET pods to the API server for every pod returned by the kubelet podresources API, on every scan. In a production cluster with 700 nodes, and 25k pods, this resulted in sustained ~900 GET pods/s anti-pattern against the apiservers.

Replace the direct client GET with a lister backed by a SharedInformer that is scoped to this node via fieldSelector=spec.nodeName, so both the initial LIST and the ongoing WATCH only see this node's pods. Steady-state apiserver traffic from nfd-topology-updater drops to one LIST + one WATCH per process, with WATCH events only for pods that actually change on this node.

The kubelet podresources API can return a pod a moment before the apiserver WATCH stream delivers it to the informer, so NotFound from the lister falls back to a live GET to keep noderesourcetopology accurate. Other lister errors are propagated unchanged (Scan() still treats them as fatal for the scan). RBAC for the topology-updater ClusterRole is widened from get to get,list,watch on pods.

### Release Notes:
```
The nfd-topology-updater ClusterRole gains list,watch on pods. Helm users get it on upgrade, plain-kustomize users will need to re-apply the manifest or topology-updater will hang on cache sync.
```

Assisted-by: Cursor <cursor@cursor.com>

### DIFF
--- a/pkg/nfd-topology-updater/nfd-topology-updater.go
+++ b/pkg/nfd-topology-updater/nfd-topology-updater.go
@@ -24,14 +24,18 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync"
 	"time"
 
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/informers"
 	k8sclient "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
 
@@ -49,6 +53,11 @@ import (
 	"sigs.k8s.io/node-feature-discovery/pkg/version"
 	"sigs.k8s.io/yaml"
 )
+
+// nodeScopedPodInformerResync is zero because topology scans read pods
+// from the cached lister only. This value only makes sense when event
+// handlers are registered to the informer.
+const nodeScopedPodInformerResync = 0
 
 const (
 	// TopologyManagerPolicyAttributeName represents an attribute which defines
@@ -200,9 +209,50 @@ func (w *nfdTopologyUpdater) Run() error {
 		return fmt.Errorf("faild to configure Node Feature Discovery Topology Updater: %w", err)
 	}
 
+	syncCtx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+
+	// Build a node-scoped Pod informer so per-pod lookups in
+	// PodResourcesScanner.isWatchable() are served from a local cache
+	// avoiding cluster-wide pod traffic.
+	podInformerFactory := informers.NewSharedInformerFactoryWithOptions(
+		k8sClient,
+		nodeScopedPodInformerResync,
+		informers.WithTweakListOptions(func(opts *metav1.ListOptions) {
+			opts.FieldSelector = fields.OneTermEqualSelector("spec.nodeName", w.nodeName).String()
+		}),
+	)
+	podInformer := podInformerFactory.Core().V1().Pods()
+	podLister := podInformer.Lister()
+
+	podInformerStop := make(chan struct{})
+	var stopPodInformerOnce sync.Once
+	stopPodInformer := func() {
+		stopPodInformerOnce.Do(func() {
+			close(podInformerStop)
+		})
+	}
+	defer stopPodInformer()
+
+	go func() {
+		select {
+		case <-w.stop:
+			cancel()
+			stopPodInformer()
+		case <-podInformerStop:
+		}
+	}()
+
+	podInformerFactory.Start(podInformerStop)
+	klog.InfoS("waiting for node-scoped Pod informer cache sync", "nodeName", w.nodeName)
+	if !cache.WaitForCacheSync(syncCtx.Done(), podInformer.Informer().HasSynced) {
+		return fmt.Errorf("timed out waiting for node-scoped Pod informer cache sync for node %q", w.nodeName)
+	}
+	klog.InfoS("node-scoped Pod informer cache synced", "nodeName", w.nodeName)
+
 	var resScan resourcemonitor.ResourcesScanner
 
-	resScan, err = resourcemonitor.NewPodResourcesScanner(w.resourcemonitorArgs.Namespace, podResClient, k8sClient, w.resourcemonitorArgs.PodSetFingerprint)
+	resScan, err = resourcemonitor.NewPodResourcesScanner(w.resourcemonitorArgs.Namespace, podResClient, podLister, k8sClient, w.resourcemonitorArgs.PodSetFingerprint)
 	if err != nil {
 		return fmt.Errorf("failed to initialize ResourceMonitor instance: %w", err)
 	}
--- a/pkg/resourcemonitor/podresourcesscanner.go
+++ b/pkg/resourcemonitor/podresourcesscanner.go
@@ -22,8 +22,10 @@ import (
 	"strconv"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	client "k8s.io/client-go/kubernetes"
+	corev1listers "k8s.io/client-go/listers/core/v1"
 	"k8s.io/klog/v2"
 	podresourcesapi "k8s.io/kubelet/pkg/apis/podresources/v1"
 	"k8s.io/kubernetes/pkg/apis/core/v1/helper/qos"
@@ -35,15 +37,25 @@ import (
 type PodResourcesScanner struct {
 	namespace         string
 	podResourceClient podresourcesapi.PodResourcesListerClient
+	podLister         corev1listers.PodLister
 	k8sClient         client.Interface
 	podFingerprint    bool
 }
 
 // NewPodResourcesScanner creates a new ResourcesScanner instance
-func NewPodResourcesScanner(namespace string, podResourceClient podresourcesapi.PodResourcesListerClient, k8sClient client.Interface, podFingerprint bool) (ResourcesScanner, error) {
+func NewPodResourcesScanner(namespace string, podResourceClient podresourcesapi.PodResourcesListerClient, podLister corev1listers.PodLister, k8sClient client.Interface, podFingerprint bool) (ResourcesScanner, error) {
+	if k8sClient == nil {
+		return nil, fmt.Errorf("kubernetes client is required")
+	}
+
+	if podLister == nil {
+		return nil, fmt.Errorf("pod lister is required")
+	}
+
 	resourcemonitorInstance := &PodResourcesScanner{
 		namespace:         namespace,
 		podResourceClient: podResourceClient,
+		podLister:         podLister,
 		k8sClient:         k8sClient,
 		podFingerprint:    podFingerprint,
 	}
@@ -59,14 +71,26 @@ func NewPodResourcesScanner(namespace string, podResourceClient podresourcesapi.
 // isWatchable tells if the the given namespace should be watched.
 // In Scan(), if watchable is false, this pods scan will skip
 // so we can return directly if pod's namespace is not watchable
-func (resMon *PodResourcesScanner) isWatchable(podResource *podresourcesapi.PodResources) (bool, bool, error) {
+func (resMon *PodResourcesScanner) isWatchable(ctx context.Context, podResource *podresourcesapi.PodResources) (bool, bool, error) {
 	if resMon.namespace != "*" && resMon.namespace != podResource.GetNamespace() {
 		return false, false, nil
 	}
 
-	pod, err := resMon.k8sClient.CoreV1().Pods(podResource.GetNamespace()).Get(context.TODO(), podResource.Name, metav1.GetOptions{})
+	ns := podResource.GetNamespace()
+	name := podResource.GetName()
+
+	pod, err := resMon.podLister.Pods(ns).Get(name)
 	if err != nil {
-		return false, false, err
+		if !apierrors.IsNotFound(err) {
+			return false, false, err
+		}
+		// The informer cache can lag behind the kubelet PodResources API. Use a
+		// live GET to preserve topology correctness.
+		klog.InfoS("pod not in informer cache, falling back to GET", "namespace", ns, "podName", name)
+		pod, err = resMon.k8sClient.CoreV1().Pods(ns).Get(ctx, name, metav1.GetOptions{})
+		if err != nil {
+			return false, false, err
+		}
 	}
 
 	isPodGuaranteed := qos.GetPodQOS(pod) == corev1.PodQOSGuaranteed
@@ -137,7 +161,7 @@ func (resMon *PodResourcesScanner) Scan() (ScanResponse, error) {
 
 	for _, podResource := range respPodResources {
 		klog.V(1).InfoS("scanning pod", "podName", podResource.GetName())
-		isWatchable, isExclusiveCPUs, err := resMon.isWatchable(podResource)
+		isWatchable, isExclusiveCPUs, err := resMon.isWatchable(ctx, podResource)
 		if err != nil {
 			return ScanResponse{}, fmt.Errorf("checking if pod in a namespace is watchable, namespace:%v, pod name %v: %w", podResource.GetNamespace(), podResource.GetName(), err)
 		}
--- a/pkg/resourcemonitor/podresourcesscanner_test.go
+++ b/pkg/resourcemonitor/podresourcesscanner_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package resourcemonitor
 
 import (
+	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"sort"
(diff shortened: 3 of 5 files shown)


alaypatel07:
### Scale test results:
The perf-scale tests were performed without the fix and with the fix on 100 node GPU cluster, with 800 pods being created. Comparing the two runs here are the benefits:

#### APIserver benefits:

1. removed get pod requests.
1. used 20% less CPU at peak
1. API priority and fairness has very low quota usage

#### etcd benefits:
1. 26% reduction in RPC rate at peak
1. 20% reduction in CPU usage at peak
1. 30% reduction in network usage at peak

etcdbenefits

APIServer benefits

alaypatel07:
/test pull-node-feature-discovery-build-image-cross-generic

looks like that test is flaking

ffromani:
/cc

not sure I'll have the bandwidth to do a proper review anytime soon, but as general thought rather than optimize the `isWatchable` path I'd deeply revisit if we (still) need it and if we can just reorganize the flow to avoid it.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -200,9 +210,32 @@ func (w *nfdTopologyUpdater) Run() error {
...
+		informers.WithTweakListOptions(func(opts *metav1.ListOptions) {
+			opts.FieldSelector = fields.OneTermEqualSelector("spec.nodeName", w.nodeName).String()
+		}),
+	)
+	podInformer := podInformerFactory.Core().V1().Pods()
+	podLister := podInformer.Lister()
+
+	podInformerFactory.Start(w.stop)
+	klog.InfoS("waiting for node-scoped Pod informer cache sync", "nodeName", w.nodeName)
+	if !cache.WaitForCacheSync(w.stop, podInformer.Informer().HasSynced) {
```
ArangoGutierrez: No deadline here — if the apiserver is slow or unreachable at startup, `Run()` hangs silently after the "waiting for..." log and the pod still looks healthy to the kubelet. Mind wrapping a `context.WithTimeout` (1-2 min) and returning a distinct error so liveness can act?

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -50,6 +53,13 @@ import (
 	"sigs.k8s.io/yaml"
 )
 
+// nodeScopedPodInformerResync controls the resync period for the
+// node-scoped Pod informer used by nfd-topology-updater. The informer
+// is the source of truth for per-pod QoS / CPU lookups during topology
+// scans (see PodResourcesScanner.isWatchable); it replaces what used
+// to be one GET pods per pod per scan.
+const nodeScopedPodInformerResync = 10 * time.Minute
```
ArangoGutierrez: Since no `AddEventHandler` is registered, resync only re-fires update deltas to handlers from the local store — it doesn't re-LIST the apiserver. For a pure-lister consumer, `0` is the idiomatic value (skips the scheduled goroutine). Not a correctness issue, just dead work.

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -64,9 +68,32 @@ func (resMon *PodResourcesScanner) isWatchable(podResource *podresourcesapi.PodR
...
 	}
 
-	pod, err := resMon.k8sClient.CoreV1().Pods(podResource.GetNamespace()).Get(context.TODO(), podResource.Name, metav1.GetOptions{})
+	ns := podResource.GetNamespace()
+	name := podResource.GetName()
+
+	pod, err := resMon.podLister.Pods(ns).Get(name)
 	if err != nil {
-		return false, false, err
+		if !apierrors.IsNotFound(err) {
```
ArangoGutierrez, line 84: Under burst churn (rolling Job of N pods, short-lived sandbox workloads, CronJob bursts), one `Scan()` can hit NotFound for N pods and serialize N live GETs against the apiserver — briefly re-creating the pattern this PR kills. The comment further down notes "rare" but that depends on workload, not on us being optimistic.

Cheapest fix: skip the pod on NotFound and pick it up next scan — topology is eventually consistent already. Alternatives: cap fallback GETs per scan, or negative-cache the miss for one cycle.

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -64,9 +68,32 @@ func (resMon *PodResourcesScanner) isWatchable(podResource *podresourcesapi.PodR
...
+			klog.ErrorS(err, "pod not in informer cache and no apiserver client configured to fall back to", "namespace", ns, "podName", name)
+			return false, false, err
+		}
+		// pod informer can be slow, it is possible that the informer does not
+		// have the pod but PodResources api in kubelet returns it. In such case,
+		// the error is checked and Get against the apiserver is issued to make
+		// sure that pod is actually present. This avoids mis-reporting in
+		// noderesourcetopology. This should be a rare occurrence.
+		klog.V(2).InfoS("pod not in informer cache, falling back to live GET", "namespace", ns, "podName", name)
+		pod, err = resMon.k8sClient.CoreV1().Pods(ns).Get(context.Background(), name, metav1.GetOptions{})
```
ArangoGutierrez: `Scan()` already builds a `defaultPodResourcesTimeout` context further down — worth plumbing it through `isWatchable` so this GET inherits cancellation.

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -64,9 +68,32 @@ func (resMon *PodResourcesScanner) isWatchable(podResource *podresourcesapi.PodR
...
+	ns := podResource.GetNamespace()
+	name := podResource.GetName()
+
+	pod, err := resMon.podLister.Pods(ns).Get(name)
 	if err != nil {
-		return false, false, err
+		if !apierrors.IsNotFound(err) {
+			return false, false, err
+		}
+		if resMon.k8sClient == nil {
```
ArangoGutierrez: The "unreachable in production" comment doesn't quite hold: the exported `NewPodResourcesScanner` accepts any `client.Interface`, including nil. Either guard it in the constructor (`return nil, fmt.Errorf(...)`) or soften the comment. The guard is cleaner.

pkg/resourcemonitor/podresourcesscanner_test.go
```diff
@@ -170,8 +183,7 @@ func TestPodScanner(t *testing.T) {
 				},
 			}
 
-			fakeCli := fakeclient.NewClientset(pod)
-			resScan.(*PodResourcesScanner).k8sClient = fakeCli
+			resScan.(*PodResourcesScanner).podLister = newFakePodLister(pod)
```
ArangoGutierrez, line 311: Heads up: every modified Convey pre-populates the lister with the pod, so the new NotFound → live-GET branch is never exercised. If I delete lines 75-97 in `podresourcesscanner.go`, all these tests still pass — and that's the riskiest part of the change.

Worth adding direct tests for `isWatchable` covering:
- lister NotFound + live GET succeeds (hits both paths)
- lister NotFound + `k8sClient == nil` returns NotFound
- lister NotFound + live GET also fails (propagates)
- lister non-NotFound error propagates

Probably cleanest to test `isWatchable` directly on a `*PodResourcesScanner{}` literal rather than rework the Convey scaffolding.

ArangoGutierrez (changes requested):
Hey Alay, the approach is right and the motivation is solid. I checked whether the kubelet podresources gRPC response carries `QOSClass` directly (which would avoid the informer entirely) — the v1 proto has no such field, and `checkPodExclusiveCPUs` also needs `pod.Spec.Containers[].Resources.Requests`, which is also absent. The informer is the right shape.

A few things before this merges:

1. Cache-sync timeout — `WaitForCacheSync` can hang `Run()` indefinitely if the apiserver is slow at startup (inline).
2. Bounded fallback — the NotFound → live-GET path can re-create the GET storm under burst churn (inline).
3. Tests for the fallback — the new branch isn't exercised by any test (inline on the test file).
4. Commit GPG signature — `git log --format='%G?'` reports `N`; needs `-S`.

Also worth a release note: the ClusterRole gains `list,watch` on pods. Helm users get it on upgrade, plain-kustomize users will need to re-apply the manifest or topology-updater will hang on cache sync.

Verified locally on the PR head: `go vet ./pkg/resourcemonitor/... ./pkg/nfd-topology-updater/...` clean, `go test -race -count=1 ./pkg/resourcemonitor/...` passes (4.8s).

pkg/resourcemonitor/podresourcesscanner.go
alaypatel07: The cheapest fix is much more uncertain, there is no guarantee what scan will be make the update concistent.

> Cheapest fix: skip the pod on NotFound and pick it up next scan — topology is eventually consistent already. 

The client-go already ships with client side rate-limiter, users can configure this to be a small value to make sure this get call is rate-limited. 
> Alternatives: cap fallback GETs per scan, or negative-cache the miss for one cycle.

Considering the above options, I would much rather keep the fall back get and add a warning log. If  this is problematic, its okay to get a bug and fix it with real values rather then pre-optimizing this path.

pkg/nfd-topology-updater/nfd-topology-updater.go
alaypatel07: updated.

ffromani:
> /cc
> 
> not sure I'll have the bandwidth to do a proper review anytime soon, but as general thought rather than optimize the `isWatchable` path I'd deeply revisit if we (still) need it and if we can just reorganize the flow to avoid it.

Found some time. Here are some thoughs: 
In the [resource-topology-exporter](https://github.com/k8stopologyawareschedwg/resource-topology-exporter) (RTE) project,
which is these days somehow the distant ancestor of nfd-topology-updater, we solved this same problem differently and I think it's worth sharing how.

## A tiny amount of history

The nfd-topology-updater was bootstrapped from RTE code (commit `b4c92e4ee`, PR #525, "Introducing NFD Topology Updater").
The `isWatchable` function was added later in the same PR (`aa7ae9265`, "topologyupdater: watch/consider only guaranteed pods for accounting").
RTE never adopted the approach we have in NFD using isWatchable; instead it was rewritten to avoid the need entirely (commit [`d0d4523b`](https://github.com/k8stopologyawareschedwg/resource-topology-exporter/commit/d0d4523b), "resourcemonitor: rewrite and simplify", July 2021).

## What `isWatchable` does today in NFD

`isWatchable` (in `pkg/resourcemonitor/podresourcesscanner.go`) answers two questions for each pod, on each scan:

1. **Should this pod be included in resource accounting?** True if Guaranteed QoS OR has device allocations.
2. **Does this pod have exclusive CPUs?** True if Guaranteed AND any container has integral CPU requests.

Both checks require the full `pod.Spec` because:
- QoS class is computed from `Requests`/`Limits` across all containers.
- Integral CPU detection inspects `container.Resources.Requests[cpu]`.

## How RTE avoids the API server call

RTE's key insight: **the kubelet PodResources API already contains the answers to both questions.**

The kubelet's CPU Manager only populates `container.CpuIds` in the PodResources API response
for containers that received dedicated CPU cores. This only happens for Guaranteed pods with
integer CPU requests and a `static` CPU Manager policy. If `CpuIds` is non-empty, the container
has exclusive CPUs. And that's it.

In addition:
- `container.GetDevices()` tells us if a pod has device allocations (NFD's own `hasDevice()`
  function already checks this from PodResources data, not from the pod spec).
- `container.GetMemory()` with topology info tells us about NUMA-pinned memory.

### For completeness' sake: the shared CPU pool edge case

There it was one legitimate edge case: kubelet issue
[kubernetes/kubernetes#102190](https://github.com/kubernetes/kubernetes/issues/102190),
where the PodResources API reports shared pool CPUs to every container that uses them.
In this scenario, i fwe trust `CpuIds` naively, we'll count shared pool CPUs as exclusive.

RTE handles this with a middleware (`pkg/podres/middleware/sharedcpuspool/`), introduced in commit
[`75d23aae`](https://github.com/k8stopologyawareschedwg/resource-topology-exporter/commit/75d23aae).
It identifies shared pool CPUs via a known reference container (the RTE DaemonSet's pause container)
and strips them from every container's `CpuIds` before the scan sees the data:

```go
sharedPoolCPUs := findSharedPoolCPUsInListResponse(fc.refCnt, resp.GetPodResources())
for _, podRes := range resp.GetPodResources() {
    for _, cntRes := range podRes.GetContainers() {
        cntRes.CpuIds = removeCPUs(cntRes.CpuIds, sharedPoolCPUs)
    }
}
```

After this filtering, any remaining `CpuIds` are truly exclusive.
Good news though: this bug in kubelet is long fixed and what we have in RTE is mostly legacy code.
Can be useful as historical context, but should not be really relevant since few years.

## What does all his means for NFD?

We can replace `isWatchable` with logic equivalent to RTE's approach:

1. **A pod is "watchable" if any of its containers have `CpuIds`, `Devices`, or `Memory` with
   topology in the PodResources response.** This is a pure in-memory check on data already fetched.

2. **A pod has "exclusive CPUs" if any container has non-empty `CpuIds`.** The kubelet already
   made this determination — re-deriving it from the pod spec is wasted work.

A legitimate question would be if the podresources API clearly document these properties.
This is a good question and would call for a review of u/s docs.
We touched this topic in https://github.com/kubernetes/kubernetes/pull/97415 - and over the years adjusted
the semantics consistently, but this guarantee never made to the official docs. I guess it's time to fill this gap.
[edit] **important note though**: RTE has relied on this property since 2021 and it was designed this way intentionally (PR #97415). The missing docs are a important but paperwork gap, not a reliability concern AFAICT.

pkg/resourcemonitor/podresourcesscanner.go
ArangoGutierrez: Fair. The client-go rate-limiter is real, and a warning log plus a real-incident signal beats pre-emptive logic nobody has profiled. Skipping on NotFound would just push the bug to "wrong topology for one cycle", which isn't actually better. Withdrawn.

ArangoGutierrez:
Thanks for the quick turnaround. The cache-sync timeout with the `sync.Once` shutdown bridge is nicer than what I had in mind, and the fallback-path tests cover all four branches.

Two small things visible on CI right now:

- `verify-master` is failing on logcheck: `pkg/resourcemonitor/podresourcesscanner.go:86` uses `klog.Warningf` but the project standard is structured (`klog.WarningS` or equivalent).
- The new commit `8c25fb3dc` is still unsigned (`git log --format='%G?'` reports `N`). Sign-off is fine; just missing GPG. `git commit --amend -S --no-edit && git push -f` covers both fixes.

`build-image-cross-generic` looks like the flake you flagged on the 15th. e2e still pending. I'll dismiss CHANGES_REQUESTED once those are green.

alaypatel07:
@ffromani TBH I like talking to kubelet pod resources API much better then setting up the informer.

But this PR is small isolated change, I was hoping to put in place relatively cheaply.

How intrusive will it be to switch on pod resources API? Do the test cases and e2e test have enough coverage to do this safely? If yes, I can attempt to do this in this PR, otherwise I can follow it up with a much bigger change in a separate PR

ffromani:
> @ffromani TBH I like talking to kubelet pod resources API much better then setting up the informer.
> 
> But this PR is small isolated change, I was hoping to put in place relatively cheaply.
> 
> How intrusive will it be to switch on pod resources API? Do the test cases and e2e test have enough coverage to do this safely? If yes, I can attempt to do this in this PR, otherwise I can follow it up with a much bigger change in a separate PR

Your call @alaypatel07 It's all up to the your timeline - and the project's of course! I'm not arguing for a pivot in this PR. And your very numbers demonstrate the gains. I wanted to provide some historical data but how to consume it most effectively - I can't say.

That said, I only have vague memories of nfd-topology-updater. The code drifted from RTE quite a bit over the years, and despite best intentions we never reconciled. Changing the internals can require some refactoring, even if the core logic is actually quite simpler. The testing storycan be different though. We may want to review the e2e coverage to ensure no gaps. This also I can't say atm.

scripts/test-infra/logcheck.conf
```diff
@@ -1,2 +1,3 @@
 -structured sigs.k8s.io/node-feature-discovery/pkg/utils/grpc_log.go
 -structured sigs.k8s.io/node-feature-discovery/pkg/utils/featuregate/featuregate.go
+-structured sigs.k8s.io/node-feature-discovery/pkg/resourcemonitor/podresourcesscanner.go
```
ArangoGutierrez: This exempts the whole file from the structured-logging check to admit one new call. klog has no `WarningS`; the structured equivalent for warnings is `klog.InfoS` — e.g. `klog.InfoS("pod not found in informer cache, falling back to live GET", "pod", klog.KRef(ns, name))` at podresourcesscanner.go:86. That lets you drop this exemption and keeps future code in the file under the check.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -200,9 +209,49 @@ func (w *nfdTopologyUpdater) Run() error {
...
+		select {
+		case <-w.stop:
+			stopPodInformer()
+		case <-podInformerStop:
+		}
+	}()
+
+	podInformerFactory.Start(podInformerStop)
+	klog.InfoS("waiting for node-scoped Pod informer cache sync", "nodeName", w.nodeName)
+	if !cache.WaitForCacheSync(syncCtx.Done(), podInformer.Informer().HasSynced) {
```
ArangoGutierrez, line 248: Minor: if `w.stop` closes during this window (SIGTERM while the apiserver is slow), the wait still runs the full 2 minutes and then reports a timeout that didn't happen. The bridge goroutine above already reacts to `w.stop` — having it also call `cancel()` makes shutdown prompt, with the error text adjusted to "canceled or timed out".

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -200,9 +209,49 @@ func (w *nfdTopologyUpdater) Run() error {
 		return fmt.Errorf("faild to configure Node Feature Discovery Topology Updater: %w", err)
 	}
 
+	syncCtx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	defer cancel()
+
+	// Build a node-scoped Pod informer so per-pod lookups in
+	// PodResourcesScanner.isWatchable() are served from a local cache
+	// avoiding cluster-wide pod traffic.
+	podInformerFactory := informers.NewSharedInformerFactoryWithOptions(
```
ArangoGutierrez, line 218: Optional: the cache retains full Pod objects while `isWatchable()` only reads metadata and container requests/limits. client-go 0.35 supports `informers.WithTransform` on this factory; stripping `.metadata.managedFields` (often a sizable share of the encoded object) trims the steady-state footprint of the per-node cache. Fits the efficiency goal of this PR but doesn't block it.

pkg/resourcemonitor/podresourcesscanner_test.go
```diff
@@ -50,9 +189,8 @@ func TestPodScanner(t *testing.T) {
 	Convey("When I scan for pod resources using fake client and no namespace", t, func() {
 		mockPodResClient := new(mockpodres.PodResourcesListerClient)
 
-		fakeCli := fakeclient.NewClientset()
 		computePodFingerprint := true
-		resScan, err := NewPodResourcesScanner("*", mockPodResClient, fakeCli, computePodFingerprint)
+		resScan, err := NewPodResourcesScanner("*", mockPodResClient, newFakePodLister(), fakeclient.NewSimpleClientset(), computePodFingerprint)
```
ArangoGutierrez: `NewSimpleClientset` is deprecated in current client-go in favor of `NewClientset`, which this file used before — same applies to the other new call sites in this file.
```suggestion
		resScan, err := NewPodResourcesScanner("*", mockPodResClient, newFakePodLister(), fakeclient.NewClientset(), computePodFingerprint)
```

ArangoGutierrez (commented):
The informer swap is the right call and the scale-test numbers in the thread make the case; everything from the first review round (sync timeout, ctx plumbing, nil-client guard, fallback tests, RBAC in kustomize and helm, release note, signed commit) is addressed in dc13c69. Two small items inline: the logcheck.conf exemption silences the structured-logging check for the whole file when converting the single new Warningf to klog.InfoS would do, and the tests swap the non-deprecated fake NewClientset for the deprecated NewSimpleClientset. There is also a minor note on shutdown during the cache-sync window and an optional WithTransform idea to shrink the cached Pods. The only red check is the build-image-cross-generic flake already flagged in the thread; needs a /retest. Fix the log call and the fake-clientset swap and this is good to merge from my side.

ArangoGutierrez:
@alaypatel07 — ready to move; you've addressed the round-1 items. Two small things before I `/lgtm` (CI is green):

- Convert the lone `klog.Warningf` at `pkg/resourcemonitor/podresourcesscanner.go:86` to `klog.InfoS` and drop the whole-file `logcheck.conf` exemption.
- The tests use the deprecated `fakeclient.NewSimpleClientset` — switch to `NewClientset` (the non-deprecated form).

Non-blocking: also `cancel()` the sync context in the shutdown bridge — as-is, a SIGTERM during the initial cache-sync window waits out the full 2-minute timeout.

@ffromani — the PodResources-`List` direction is the right long-term shape: deriving the watchable/exclusive-CPU decision from `CpuIds`/`Devices`/`Memory` drops the informer, the `pods` RBAC, and the cache-sync startup dependency. Filed #2529 to track it (with the one edge case noted). Keeping it out of this PR — it reworks the accounting path and needs its own e2e coverage.

PTAL once the two are in and I'll `/lgtm` + `/approve`.

alaypatel07:
@ArangoGutierrez Addressed the outstanding comments. PTAL

ArangoGutierrez (commented):
Almost there. One ask before lgtm, plus one item still open from last round; neither is large.

1. The constructor guards a nil k8sClient but not a nil podLister, and isWatchable now dereferences podLister on every call before the k8sClient fallback, so a nil lister panics on the first Scan. Worth the same guard here, like the k8sClient one. (pkg/resourcemonitor/podresourcesscanner.go:47)
2. Still open from last round: on SIGTERM during this wait the bridge goroutine stops the informer but doesn't cancel syncCtx, so WaitForCacheSync runs the full 2 minutes and then reports a timeout that didn't happen. Having the goroutine also call cancel() makes shutdown prompt. Non-blocking. (pkg/nfd-topology-updater/nfd-topology-updater.go:247)

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -35,15 +37,21 @@ import (
...
 	podResourceClient podresourcesapi.PodResourcesListerClient
+	podLister         corev1listers.PodLister
 	k8sClient         client.Interface
 	podFingerprint    bool
 }
 
 // NewPodResourcesScanner creates a new ResourcesScanner instance
-func NewPodResourcesScanner(namespace string, podResourceClient podresourcesapi.PodResourcesListerClient, k8sClient client.Interface, podFingerprint bool) (ResourcesScanner, error) {
+func NewPodResourcesScanner(namespace string, podResourceClient podresourcesapi.PodResourcesListerClient, podLister corev1listers.PodLister, k8sClient client.Interface, podFingerprint bool) (ResourcesScanner, error) {
+	if k8sClient == nil {
```
ArangoGutierrez, line 47: The constructor guards a nil k8sClient but not a nil podLister, and isWatchable now dereferences podLister on every call before the k8sClient fallback, so a nil lister panics on the first Scan. Worth the same guard here, like the k8sClient one.

pkg/nfd-topology-updater/nfd-topology-updater.go
ArangoGutierrez, line 248: Still open from last round: on SIGTERM during this wait the bridge goroutine stops the informer but doesn't cancel syncCtx, so WaitForCacheSync runs the full 2 minutes and then reports a timeout that didn't happen. Having the goroutine also call cancel() makes shutdown prompt. Non-blocking.

pkg/resourcemonitor/podresourcesscanner.go
alaypatel07: added a nil check guard.

pkg/nfd-topology-updater/nfd-topology-updater.go
alaypatel07: goroutine now calls cancel before stopPodInformer() call.

pkg/resourcemonitor/podresourcesscanner.go
```diff
@@ -59,14 +71,26 @@ func NewPodResourcesScanner(namespace string, podResourceClient podresourcesapi.
...
+
+	pod, err := resMon.podLister.Pods(ns).Get(name)
 	if err != nil {
-		return false, false, err
+		if !apierrors.IsNotFound(err) {
+			return false, false, err
+		}
+		// The informer cache can lag behind the kubelet PodResources API. Use a
+		// live GET to preserve topology correctness.
+		klog.InfoS("pod not in informer cache, falling back to GET", "namespace", ns, "podName", name)
```
ArangoGutierrez, line 89: Non-blocking: this fallback fires for any pod the informer cache hasn't caught up with yet, which is routine right after scheduling, so at default verbosity it logs once per affected pod per scan. An earlier revision had this gated at klog.V(2); we can restore the gate in the follow-up.

ArangoGutierrez (approved):
All my earlier concerns are addressed - the informer lifecycle is clean, RBAC matches the informer's needs in both kustomize and Helm, and TestIsWatchable covers the fallback branches. One non-blocking nit inline; happy to pick it up together with a few small hardening items (nil-lister guard test, graceful-stop error path during cache sync, resync doc comment) in a follow-up PR alongside #2529.

1. Non-blocking: this fallback fires for any pod the informer cache hasn't caught up with yet, which is routine right after scheduling, so at default verbosity it logs once per affected pod per scan. An earlier revision had this gated at klog.V(2); we can restore the gate in the follow-up. (pkg/resourcemonitor/podresourcesscanner.go:89)

Issue #2511, opened 2026-05-19
CI cross-build fix — post-merge confirmation gate


ArangoGutierrez:
Tracking the post-merge gate for the multi-arch cross-compile fix (PR forthcoming).

## Closure criteria

- [ ] `pull-node-feature-discovery-build-image-cross-generic` is green on **5 consecutive PRs** against master after merge.
- [ ] Recorded baseline: `nfd-worker --oneshot --no-publish` output from a real `linux/arm64` node, attached as `arm64-baseline.txt` on this issue.
- [ ] (Best effort) For ppc64le and s390x: record a `feature.node.kubernetes.io/cpu-cpuid.*` label list from a real-hardware run if available; otherwise mark as 'qemu-validated only' in release notes.

## Rollback triggers

If any of the criteria fail in the gate window, pick the rollback path documented in the design spec (`.agents/plans/2026-05-15-multiarch-buildx-cross-compile-design.md`):

1. Narrow release scope: set `IMAGE_ALL_PLATFORMS=linux/amd64,linux/arm64` in `scripts/test-infra/push-image.sh`.
2. Per-arch fallback: override `--platform=$TARGETPLATFORM` on the builder stage for the broken arch only.

`git revert` is **NOT** a safe rollback — it reintroduces the QEMU memory-corruption crash class.

## References

- Design (in fork): `.agents/plans/2026-05-15-multiarch-buildx-cross-compile-design.md` (commit `891d63aa0` on `ArangoGutierrez/node-feature-discovery@agents-workbench`)
- Plan (in fork):   `.agents/plans/2026-05-15-multiarch-buildx-cross-compile-plan.md`
- Test-infra context: kubernetes/test-infra#37041
- tonistiigi/xx digest: `sha256:010d4b66aed389848b0694f91c7aaee9df59a6f20be7f5d12e53663a37bd14e2`

ArangoGutierrez:
Status update: the fix merged 2026-05-21 via #2512 (commit 2634886fe). Gate progress: 1/5 — the first post-merge cross-generic run is green on #2517 (base 26aef0e1). The failures on #2507 and #2509 were runs against pre-fix base 5fbdf913; retests have been triggered to collect post-fix data points. arm64 baseline capture is still pending. Closed #2419 in favor of this gate.

PR #2512, opened 2026-05-19
build(image): cross-compile from $BUILDPLATFORM via tonistiigi/xx

ArangoGutierrez:
## Problem

The presubmit `pull-node-feature-discovery-build-image-cross-generic` crashes inside the QEMU-emulated `linux/arm64` builder stage during `go mod download` with Go-runtime memory-corruption signatures:

- `runtime: pointer 0x... to unallocated span`
- `fatal error: SIGSEGV at PC=0`
- `runtime: marked free object in span`

kubernetes/test-infra#37041 (merged 2026-05-15) bumped the failing job's memory from 8Gi to 16Gi; the next run with the same workload still hit the same signature. Memory headroom is not the fix.

## Root cause

The Dockerfile's `builder` stage had no `--platform=$BUILDPLATFORM` declaration, so the entire builder — including `go mod download` and `go build` / `make install` — ran under QEMU user-mode emulation for every non-host target architecture. QEMU's translation cache and Go's runtime allocator interact pathologically under emulation.

## Approach

Move the builder stage to `$BUILDPLATFORM` and cross-compile per `$TARGETPLATFORM` via `tonistiigi/xx` (pinned by sha256 digest). Only runtime stages (`COPY` / `USER` / `ENV`) still run under emulation — trivial work QEMU handles reliably.

CGO is retained because `source/cpu/cpuid_linux_{arm,arm64,ppc64le,s390x}.go` use `getauxval`. A build-time silent-CGO-disable guard (`go tool nm | grep source/cpu._Cfunc_gethwcap`) catches the regression class where the cross-libc install silently fails and `CGO_ENABLED=0` falls back, producing binaries with empty CPU feature labels. amd64 is excluded from the guard because it uses pure-Go cpuid via `github.com/klauspost/cpuid/v2`.

`go install` redirects cross-compile output to `/go/bin/$GOOS_$GOARCH/<name>/` and refuses `GOBIN` override, so the production stage uses explicit per-binary `go build -o /go/bin/$bin ./cmd/$bin`. Runtime `COPY` enumerates `BUILD_BINARIES` explicitly (defensive against any future xx-tool leakage).

A new `make image-cross-test` target gives reviewers and future debuggers a one-command local reproducer for the full 5-platform release matrix.

## CI scope vs release scope

| Surface | Script | Platforms |
|---|---|---|
| Prow presubmit | `scripts/test-infra/build-image-cross.sh` | `linux/amd64,linux/arm64` |
| Release push / local `make image-all` | Makefile default | `linux/amd64,linux/arm64,linux/arm/v7,linux/s390x,linux/ppc64le` |

## Testing done

Local verification on `linux/arm64` host (Docker Desktop, Apple Silicon, buildkit v0.29.0):

- `make image-cross-test` green for all 5 platforms.
- `make image-all` green for `full` + `minimal` targets × 5 platforms (single buildx invocation each).
- Per-arch ELF assertions on the resulting `nfd-worker`:
  - `file` reports statically linked + correct arch for each.
  - `readelf -d` shows no NEEDED entries (no dynamic deps).
  - `readelf -A` shows `Tag_CPU_arch=v7` on `linux/arm/v7`.
  - `go version -m` confirms `CGO_ENABLED=1` and correct `GOARCH`/`GOARM` on all five.
- The build-time CGO guard (unstripped probe + `go tool nm` symbol grep) fired and passed for arm64, arm/v7, s390x, ppc64le during the cross-build.

## Tracking issue

Post-merge gate: #2511

## Rollback

`git revert` is **NOT** a safe rollback — it reintroduces the QEMU memory-corruption crash class. If a regression appears post-merge:

1. Narrow release scope: set `IMAGE_ALL_PLATFORMS=linux/amd64,linux/arm64` in `scripts/test-infra/push-image.sh`.
2. Per-arch fallback: override `--platform=$TARGETPLATFORM` on the builder stage for the broken arch only.

Detail in the design spec on the agents-workbench branch (`.agents/plans/2026-05-15-multiarch-buildx-cross-compile-design.md`).

## References

- Test-infra: kubernetes/test-infra#37041
- tonistiigi/xx digest: `sha256:010d4b66aed389848b0694f91c7aaee9df59a6f20be7f5d12e53663a37bd14e2`

/sig node
/area node-feature-discovery
/cc @marquiz

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,47 +1,111 @@
+# BUILDER_IMAGE MUST be Debian-family (xx-apt-get requires apt-get).
+# Project's Makefile pins it to golang:1.26-trixie (matches go.mod
+# minimum Go version) which satisfies this. Override with a non-Debian
+# image only if you also swap xx-apt-get for the matching package manager.
+#
+# tonistiigi/xx is pinned by sha256 digest below. Minimum acceptable
+# version is 1.5.0 (arm/v7 + $TARGETVARIANT handling). To update:
+#   docker pull tonistiigi/xx:1.x.y
+#   docker inspect --format='{{index .RepoDigests 0}}' tonistiigi/xx:1.x.y
+# Track in Renovate / Dependabot alongside the runtime base-image pins.
 ARG BUILDER_IMAGE
 ARG BASE_IMAGE_FULL
 ARG BASE_IMAGE_MINIMAL
+ARG XX_IMAGE=tonistiigi/xx@sha256:010d4b66aed389848b0694f91c7aaee9df59a6f20be7f5d12e53663a37bd14e2
 
-# Build node feature discovery
-FROM ${BUILDER_IMAGE:-golang} AS builder
+# Cross-build helper (host arch; never shipped)
+FROM --platform=$BUILDPLATFORM ${XX_IMAGE} AS xx
 
-# Get (cache) deps in a separate layer
+# Build node-feature-discovery on the build host, cross-compile to TARGET.
+# Running here on $BUILDPLATFORM avoids the QEMU translation-cache instability
+# that crashed go mod download under emulation (see PR commit body for the
+# crash signatures).
+FROM --platform=$BUILDPLATFORM ${BUILDER_IMAGE:-golang} AS builder
+COPY --from=xx / /
+
+ARG TARGETPLATFORM
+ARG TARGETARCH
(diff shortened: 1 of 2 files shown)


PiotrProkop:
@ArangoGutierrez thanks for looking into this!
/lgtm

PR #2515, opened 2026-06-02
fix(helm): render prune job resources at container level

zewelor:
## Summary

I believe the Helm `post-delete` prune Job currently renders `.Values.master.resources` at the PodSpec level instead of the `nfd-master` container level.

The regular `nfd-master` Deployment applies `master.resources` to the container, but the `post-delete` cleanup Job appears to place the same values under:

```yaml
spec:
  template:
    spec:
      resources:
```

instead of:

```yaml
spec:
  template:
    spec:
      containers:
        - name: nfd-master
          resources:
```

This PR moves the resources block into the prune Job's `nfd-master` container.

## Why I believe this is wrong

In `deployment/helm/node-feature-discovery/templates/post-delete-job.yaml`, the resources block is rendered after `restartPolicy`, `nodeSelector`, `affinity`, and `tolerations`:

```yaml
      restartPolicy: Never
      {{- with .Values.master.resources }}
      resources:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

That indentation places `resources` under `spec.template.spec`, not under the container.

When rendering the chart with non-empty `master.resources`, the prune Job ends up with pod-level resources and no container-level resources:

```bash
helm template nfd ./deployment/helm/node-feature-discovery \
  --namespace kube-system \
  --set master.resources.requests.cpu=10m \
  --set master.resources.requests.memory=64Mi \
  --set master.resources.limits.memory=128Mi \
  | yq 'select(.kind == "Job" and .metadata.name == "nfd-node-feature-discovery-prune") | {
      pod_resources: .spec.template.spec.resources,
      container_resources: .spec.template.spec.containers[0].resources
    }'
```

```json
{
  "pod_resources": {
    "limits": {
      "memory": "128Mi"
    },
    "requests": {
      "cpu": "10m",
      "memory": "64Mi"
    }
  },
  "container_resources": null
}
```

The `nfd-master` container in the prune Job therefore does not receive the configured CPU/memory requests and limits.

One reason this can be easy to miss is that Kubernetes 1.32 introduced `PodSpec.resources` as an alpha `PodLevelResources` field. On clusters whose API exposes that field, the manifest may pass validation even though the chart is applying `master.resources` to a different API field than the regular master container resources.

## How to reproduce

Render the chart with non-empty `master.resources` and inspect the prune Job:

```bash
helm template nfd ./deployment/helm/node-feature-discovery \
  --namespace kube-system \
  | yq 'select(.kind == "Job" and .metadata.name == "nfd-node-feature-discovery-prune") | {
      pod_resources: .spec.template.spec.resources,
      container_resources: .spec.template.spec.containers[0].resources
    }'
```

Before this change, the output shows resources at the pod level and no resources on the container:

```json
{
  "pod_resources": {
    "limits": {
      "memory": "4Gi"
    },
    "requests": {
      "cpu": "100m",
      "memory": "128Mi"
    }
  },
  "container_resources": null
}
```

After this change, the resources are rendered under the `nfd-master` container:

```json
{
  "pod_resources": null,
  "container_resources": {
    "limits": {
      "memory": "4Gi"
    },
    "requests": {
      "cpu": "100m",
      "memory": "128Mi"
    }
  }
}
```

## Testing

Rendered the Helm chart and verified that the prune Job now places `.Values.master.resources` under:

```yaml
spec.template.spec.containers[0].resources
```

instead of:

```yaml
spec.template.spec.resources
```

Also ran:

```bash
make helm-lint

helm template nfd deployment/helm/node-feature-discovery --namespace kube-system \
  | yq 'select(.kind == "Job" and .metadata.name == "nfd-node-feature-discovery-prune")' \
  | kubectl create --dry-run=server --validate=strict -f - -o json \
  | jq '{kind, name: .metadata.name, pod_resources: .spec.template.spec.resources, container_resources: .spec.template.spec.containers[0].resources}'
```

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/post-delete-job.yaml
+++ b/deployment/helm/node-feature-discovery/templates/post-delete-job.yaml
@@ -82,6 +82,10 @@ spec:
             {{- if .Values.master.instance | empty | not }}
             - "-instance={{ .Values.master.instance }}"
             {{- end }}
+          {{- with .Values.master.resources }}
+          resources:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
       restartPolicy: Never
       {{- with .Values.master.nodeSelector }}
       nodeSelector:
@@ -95,9 +99,4 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
       {{- end }}
-      {{- with .Values.master.resources }}
-      resources:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
 {{- end }}
-


ArangoGutierrez (commented):
The fix is correct: `resources` belongs on the container, and the new block lines up with the chart's existing pattern — `resources:` at the same level as the container's `securityContext`/`image`/`args`, children via `nindent 12`, exactly how master.yaml renders `master.resources`. The stale pod-level block is cleanly removed, and no other template in the chart has this issue. Since values.yaml ships non-empty `master.resources` defaults, every default install hit this: the prune container ran with no requests/limits while the chart emitted a field that API servers without `PodLevelResources` just drop. Same values key as before, so nothing for users to migrate, and the repro in the description checks out. Fix is correct and complete — just needs `/ok-to-test` and then lgtm/approve from a maintainer to merge.

ArangoGutierrez:
/lgtm
/approve

Reviewed: correctly moves `master.resources` from the (invalid) pod-spec level into the prune Job's container (`nindent 12`, sibling of the other container fields), matching the main master Deployment. No schema/helm-docs regen needed; affects only the ephemeral post-delete Job. Thanks @zewelor.

PR #2518, opened 2026-06-10
Add optional aggregate role for CRD resources

jcpunk:
On my test cluster it is helpful to have an aggregate role for viewing CRD resources.

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/clusterrole-aggregate.yaml
+++ b/deployment/helm/node-feature-discovery/templates/clusterrole-aggregate.yaml
@@ -1,0 +1,44 @@
+{{- if and .Values.master.enable .Values.master.rbac.create .Values.master.rbac.createAggregateRoles }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "node-feature-discovery.fullname" . }}-view
+  labels:
+    {{- include "node-feature-discovery.labels" . | nindent 4 }}
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+rules:
+- apiGroups:
+  - nfd.k8s-sigs.io
+  resources:
+  - nodefeatures
+  - nodefeaturerules
+  - nodefeaturegroups
+  verbs:
+  - get
+  - list
+  - watch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "node-feature-discovery.fullname" . }}-edit
+  labels:
+    {{- include "node-feature-discovery.labels" . | nindent 4 }}
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+- apiGroups:
+  - nfd.k8s-sigs.io
+  resources:
+  - nodefeatures
+  - nodefeaturerules
+  - nodefeaturegroups
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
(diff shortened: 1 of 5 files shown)


ArangoGutierrez (commented):
This adds two default-off aggregated ClusterRoles for the NFD CRDs: a `-view` role aggregated into `view`, and a `-edit` role with full write verbs aggregated into `edit` and `admin`, wired through values.yaml, the schema and both docs tables, matching the chart's existing gating and label conventions. The mechanics look good; the open question is RBAC scope: write on `nodefeatures` flows to every namespace editor while nfd-master consumes NodeFeature objects from all namespaces by default, and the option description says "view access" although an edit/admin role is also created. CI has not run yet (needs-ok-to-test), no merge conflicts. Once the edit-aggregate scope is settled and the description regenerated, this is good to go.

deployment/helm/node-feature-discovery/templates/clusterrole-aggregate.yaml
```diff
@@ -0,0 +1,44 @@
...
+  name: {{ include "node-feature-discovery.fullname" . }}-edit
+  labels:
+    {{- include "node-feature-discovery.labels" . | nindent 4 }}
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+- apiGroups:
+  - nfd.k8s-sigs.io
+  resources:
+  - nodefeatures
```
ArangoGutierrez, line 33: NodeFeature is the internal conveyor object written by nfd-worker (or third-party feature sources), not really an end-user API. With the default empty `restrictions.nodeFeatureNamespaceSelector`, nfd-master processes NodeFeature objects from every namespace and takes the target node from the `nfd.node.kubernetes.io/node-name` label, so aggregating create/update/patch/delete into `edit`/`admin` lets any namespace editor influence labels and extended resources of arbitrary nodes. I'd keep `nodefeatures` in the `-view` role only and restrict the write aggregate to `nodefeaturerules` and `nodefeaturegroups`, or at minimum spell this implication out in the option description.

deployment/helm/node-feature-discovery/templates/clusterrole-aggregate.yaml
```diff
@@ -0,0 +1,44 @@
...
+  - nodefeaturerules
```
ArangoGutierrez, line 34: NodeFeatureRule is cluster-scoped, so this entry is inert in the common case of `edit`/`admin` granted through namespaced RoleBindings; it only takes effect where those roles are bound with a ClusterRoleBinding, and there it grants cluster-wide write on an object that can set node taints and labels. Worth deciding deliberately whether rule writes belong in the edit aggregate at all; view-only would be the safer default.

deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -203,6 +203,9 @@ master:
     # -- Create [RBAC](https://kubernetes.io/docs/reference/access-authn-authz/rbac/) configuration for nfd-master.
     # @section -- NFD-Master
     create: true
+    # -- Create aggregate roles that grant view access to NFD CRD resources.
```
ArangoGutierrez: The description only mentions view access, but the template also creates an edit/admin-aggregated role with full write verbs. Since the chart README.md, docs/deployment/helm.md and values.schema.json are all generated from this comment, reword it here and re-run `make helm-generate` to keep them in sync.
```suggestion
    # -- Create aggregate roles that grant view and edit access to NFD CRD resources.
```

jcpunk: Done

Issue #2526, opened 2026-06-22
ci: migrate Codecov coverage upload from the deprecated legacy uploader to the Codecov CLI


ArangoGutierrez:
## Background

`scripts/test-infra/verify.sh` uploads coverage using the **legacy Codecov bash uploader** (`https://uploader.codecov.io/latest/linux/codecov`) and previously imported its PGP key from `https://keybase.io/codecovsecurity/pgp_keys.asc`.

Both are problematic:
- The Keybase PGP key URL now returns a 32-byte `SELF-SIGNED PUBLIC KEY NOT FOUND` stub (Keybase shutdown), which broke `verify-master` for every PR (fixed as a stopgap by making the upload non-fatal — see the linked PR).
- The legacy uploader itself is **deprecated** by Codecov in favor of the [Codecov CLI](https://docs.codecov.com/docs/codecov-uploader). It will eventually stop working entirely.

## Proposal

Migrate the coverage upload in `verify.sh` to the supported **Codecov CLI** (`cli.codecov.io`), including its integrity-verification flow, while keeping the upload best-effort (non-gating). Validate it works in the Prow `verify-master` job (this is not a GitHub Actions context, so the `codecov-action` is not applicable).

## Why separate

The immediate breakage was fixed minimally (make coverage upload non-fatal) to unblock merges. This modernization is a larger change with more review surface and is tracked here per one-concern-per-PR.

PR #2527, opened 2026-06-22
fix(ci): make Codecov coverage upload non-fatal in verify.sh

ArangoGutierrez:
## Problem

`pull-node-feature-discovery-verify-master` is failing on **every** PR, blocking all merges (including open Dependabot PRs #2510, #2520, #2521, #2522).

## Root cause

`scripts/test-infra/verify.sh` runs under `bash -e`. It imports the Codecov uploader's PGP key from `https://keybase.io/codecovsecurity/pgp_keys.asc`, which now returns a 32-byte `SELF-SIGNED PUBLIC KEY NOT FOUND` stub following the Keybase shutdown. `gpg --import` then exits non-zero and **aborts the whole script before any verification runs** — the job log dies at the gpg step (the `build`/`e2e`/`cross` jobs are unaffected, which is why only `verify-master` fails).

That imported key was never actually used to verify anything (there is no `gpgv` step), so it provided no real integrity guarantee.

## Fix

Coverage reporting is informational and must never gate merges. This consolidates the Codecov steps into a single best-effort `upload_coverage` function invoked via `if ! ...`, so any failure logs a **loud warning** and is non-fatal. The real verification steps (gofmt, golangci-lint, helm-lint, logcheck, unit tests, and the template/schema/README sync checks) remain **fatal and unchanged**.

## Testing

- `bash -n` and `shellcheck` (no new findings; my change adds no shellcheck warnings).
- Verified the guarded construct is non-fatal under `bash -e`: a failing `upload_coverage` logs the warning and execution continues (exit 0), whereas the original un-guarded form aborts. 
- This PR's own `verify-master` run is the end-to-end proof.

## Follow-up

The legacy Codecov uploader is itself deprecated; migrating to the supported Codecov CLI is tracked in #2526.

## Breaking changes

None. Coverage reporting behavior is unchanged when the uploader works; when it fails, the job now warns and continues instead of failing the gate.

### DIFF
--- a/scripts/test-infra/verify.sh
+++ b/scripts/test-infra/verify.sh
@@ -18,10 +18,6 @@ kubectl="$gobinpath/kubectl"
 curl -L https://dl.k8s.io/release/$KUBECTL_VERSION/bin/linux/amd64/kubectl -o "$kubectl"
 chmod 755 "$kubectl"
 
-curl https://keybase.io/codecovsecurity/pgp_keys.asc | gpg --no-default-keyring --keyring trustedkeys.gpg --import
-curl -Os https://uploader.codecov.io/latest/linux/codecov
-chmod +x codecov
-
 # TODO: update logcheck version when there is a new release (newer than v0.9.0)
 go install sigs.k8s.io/logtools/logcheck@v0.9.1-0.20251007102500-d35c84c015fe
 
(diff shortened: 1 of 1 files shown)


ArangoGutierrez:
/approve

Self-approving the OWNERS portion (this touches CI tooling I own). Still needs a peer `/lgtm` to merge. This unblocks #2510, #2520, #2521, #2522 and every other PR currently failing `verify-master`.

marquiz (approved):
/lgtm

ArangoGutierrez:
/cc @PiotrProkop

Hey Piotr — small CI fix that unblocks every currently-stuck PR. `verify-master` was dying under `bash -e` on the dead Keybase Codecov key URL; this makes the coverage upload best-effort/non-fatal. `verify-master` is green here and it's already /approve'd — just need your /lgtm to merge. It then unblocks the 4 open Dependabot PRs (#2510, #2520, #2521, #2522). Thanks!

PR #2530, opened 2026-06-25
feat(nfd-worker): add core.noPublishFeatures to slim NodeFeature objects

PiotrProkop:
NodeFeature objects can be dominated by large feature sets - e.g. pci.device on nodes with many SR-IOV virtual functions - that few or no labels are actually derived from. Because nfd-master caches every NodeFeature in its informer (one object per node), this bloat scales with cluster size and inflates nfd-master's memory, etcd storage and apiserver/watch traffic.

Add a new core.noPublishFeatures config option (and matching -no-publish-features flag) listing feature keys to omit from the published NodeFeature object. Feature discovery is unaffected and runs before label computation, so the omitted features remain available to label sources and inline sources.custom rules; only the published Spec.Features is trimmed.

Each entry matches the "<source>.<feature>" key exactly, or as a prefix when it ends with "*" (e.g. "pci.device" or "pci.*").

The option is opt-in and defaults to empty, so behaviour is unchanged unless configured. Note that it must only be used for features that no cluster-side NodeFeatureRule or NodeFeatureGroup consumes, since those are evaluated by nfd-master against the published Spec.Features.

### DIFF
--- a/pkg/nfd-worker/nfd-worker.go
+++ b/pkg/nfd-worker/nfd-worker.go
@@ -35,6 +35,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 	k8sclient "k8s.io/client-go/kubernetes"
 	"k8s.io/klog/v2"
@@ -84,7 +85,13 @@ type coreConfig struct {
 	FeatureSources []string
 	Sources        *[]string
 	LabelSources   []string
-	SleepInterval  utils.DurationVal
+	// NoPublishFeatures lists feature keys that are discovered (so they remain
+	// available to label sources and inline custom rules) but are omitted from
+	// the published NodeFeature object to reduce its size. Each entry is matched
+	// against the "<source>.<feature>" key, exactly or, when it ends with "*",
+	// as a prefix (e.g. "pci.device" or "pci.*").
+	NoPublishFeatures []string
+	SleepInterval     utils.DurationVal
 }
 
 type sourcesConfig map[string]source.Config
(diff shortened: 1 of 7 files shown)


pkg/nfd-worker/nfd-worker.go
```diff
@@ -646,6 +656,45 @@ func (w *nfdWorker) advertiseFeatures(labels Labels) error {
 	return nil
 }
 
+// removeNoPublishFeatures deletes the feature keys matched by patterns from the
+// Features that will be published in the NodeFeature object. Discovery is not
+// affected, so the features remain available to label sources and inline custom
+// rules; they are only omitted from the published object. Each pattern matches
+// the "<source>.<feature>" key exactly, or as a prefix when it ends with "*"
+// (e.g. "pci.device" or "pci.*").
+func removeNoPublishFeatures(features *nfdv1alpha1.Features, patterns []string) {
```
ArangoGutierrez: Consider logging when a configured key matches no feature (usually a typo), mirroring the existing "skipping unknown source specified in core.featureSources" warning in configureCore. A silent no-match leaves the object full-size with no signal that the config took effect.

ArangoGutierrez (commented):
Clean, well-scoped change - accurate docs, real discriminating tests, and override precedence consistent with the existing feature-sources/label-sources handling. The strip is safe: GetAllFeatures() builds a fresh Features each call and labels are computed first via GetLabels(), so deleting keys only trims the published object and never touches the source stores or label computation. One optional suggestion below.

1. Consider logging when a configured key matches no feature (usually a typo), mirroring the existing "skipping unknown source specified in core.featureSources" warning in configureCore. A silent no-match leaves the object full-size with no signal that the config took effect. (pkg/nfd-worker/nfd-worker.go:665)

pkg/nfd-worker/nfd-worker.go
PiotrProkop: fixed in latest push

ArangoGutierrez (approved):
The published-object stripping is safe: GetAllFeatures returns freshly-allocated top-level maps, so the deletes only shrink that per-call copy and never touch the shared feature store or the labels already computed earlier in the cycle. Flag-over-config precedence and the once-per-pattern no-match warning both behave as documented, and the test covers exact key, `*` prefix, multiple patterns, and the unmatched-in-config-order reporting. Reads well.

PR #2531, opened 2026-06-25
perf(nodefeaturerule): cache compiled MatchInRegexp patterns

PiotrProkop:
evaluateMatchExpression recompiled every MatchInRegexp pattern with regexp.Compile on each call -- and it is invoked per feature element, per rule, per node, on every reconcile. Compiled regexps are heavy objects, so repeatedly recompiling the same rule patterns is an avoidable source of heap allocation and CPU work under nfd-master's reconcile churn.

Memoize compiled patterns in a package-level sync.Map keyed by pattern string. Matching behavior is unchanged: a term's patterns are still all compiled before matching, so an invalid pattern errors exactly as before. sync.Map is safe for the concurrent node updater pool.

A size-capped LRU cache was considered but rejected. The cache key is the pattern string, which comes only from NodeFeatureRule / NodeFeatureGroup specs -- operator-authored objects that change infrequently -- never from NodeFeature objects or the feature values being matched. The set of distinct patterns is therefore small and slowly-changing, so a plain unbounded map stays bounded in practice and avoids the extra dependency and eviction bookkeeping. An LRU can be swapped in behind the same compileRegexp signature if some workload ever starts producing many unique patterns (e.g. node names or UUIDs baked into a regex).

### DIFF
--- a/pkg/apis/nfd/nodefeaturerule/expression.go
+++ b/pkg/apis/nfd/nodefeaturerule/expression.go
@@ -23,6 +23,7 @@ import (
 	"sort"
 	"strconv"
 	strings "strings"
+	"sync"
 
 	"maps"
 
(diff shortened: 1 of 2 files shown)


pkg/apis/nfd/nodefeaturerule/expression.go
```diff
@@ -56,6 +57,43 @@ var matchOps = map[nfdv1alpha1.MatchOp]struct{}{
...
+// bounded in practice (a handful of KB), and sync.Map is safe for the concurrent
+// node updater pool.
+//
+// Entries are never evicted -- including when a rule is modified or deleted, so a
+// retired pattern lingers until the process restarts. This is intentional: tying
+// eviction to rule lifecycle would need reference counting (a pattern may be
+// shared across rules) for negligible benefit. The only way this grows unbounded
+// is a workload that produces many *unique* patterns over time (e.g. rules with
+// node names or UUIDs baked into the regex). If that ever becomes a concern,
+// replace the sync.Map with a size-capped LRU (e.g. hashicorp/golang-lru) behind
```
PiotrProkop, line 79: @ArangoGutierrez wdyt? should we use some LRU cache here?

ArangoGutierrez (commented):
The nfd-master-side counterpart to #2530, and the two are orthogonal - no interaction in matching behavior.

The memoization is correct and safe:
- Single call site: the MatchInRegexp branch holds the only regexp.Compile in this package, so nothing is left recompiling.
- Bounded cache key: keys are MatchExpression.Value patterns from NodeFeatureRule/NodeFeatureGroup specs (operator-authored), never feature values from NodeFeature objects, so the unbounded sync.Map stays small in practice; the never-evict tradeoff is documented and reasonable.
- Sound concurrency: the sync.Map Load/LoadOrStore double-compile is the idiomatic memoization pattern, and *regexp.Regexp is safe to share across the node-updater pool.
- Behavior preserved: patterns are still all compiled before matching, so an invalid pattern errors exactly as before; the test covers instance reuse, correct matching, distinct instances, and the error path.

No issues found.

ArangoGutierrez (approved):
The cache key is the operator-authored pattern string from the rule/group specs, never feature data, so cardinality stays bounded the way the comment lays out, and sync.Map is the right fit for the concurrent updater pool. The LoadOrStore path also handles the benign race where two goroutines compile the same pattern at once. Clear win.

PR #2532, opened 2026-06-25
perf(nfd-master): reconcile only on NodeFeature/Rule/Group spec changes

PiotrProkop:
The NodeFeature, NodeFeatureRule and NodeFeatureGroup informer UpdateFuncs ignored the old object and triggered a reconcile on every update - including metadata-only writes, status-only writes, and the informer's periodic resync (default 1h), which replays every cached object as a no-op update.

This is wasteful: for NodeFeatureRule each update fans out into a full-cluster reconcile (nfdAPIUpdateAllNodes), and for NodeFeatureGroup the master writes the object's own status, so the status write fed back as another reconcile.

Gate each UpdateFunc on an actual Spec change via a small generic helper (specChanged) using apiequality.Semantic.DeepEqual - the same "skip if unchanged" idiom already used in nfd-worker. We compare Spec directly rather than metadata.generation because NodeFeature/NodeFeatureRule have no status subresource, so generation's spec-only semantics are not guaranteed. Add/Delete handlers and the resync safety net are unchanged; the helper fails open on a type-assertion failure so a needed reconcile is never skipped.

### DIFF
--- a/pkg/nfd-master/nfd-api-controller.go
+++ b/pkg/nfd-master/nfd-api-controller.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"time"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	k8sclient "k8s.io/client-go/kubernetes"
@@ -69,6 +70,22 @@ type nfdApiControllerOptions struct {
 
 func init() {
 	utilruntime.Must(nfdv1alpha1.AddToScheme(nfdscheme.Scheme))
+}
+
+// specChanged reports whether the reconcile-relevant part of an object — its
+// .Spec — differs between the old and new versions delivered to an informer
+// UpdateFunc. Metadata-only updates (annotations, labels, ownerReferences,
+// resourceVersion), status-only updates, and resync no-ops (oldObj == newObj)
+// therefore return false, letting the handler skip an unnecessary reconcile.
+// It fails open (returns true) when the objects cannot be type-asserted, so a
+// needed reconcile is never wrongly skipped.
+func specChanged[T any](oldObj, newObj interface{}, specOf func(*T) any) bool {
+	oldT, ok1 := oldObj.(*T)
+	newT, ok2 := newObj.(*T)
+	if !ok1 || !ok2 {
+		return true
+	}
+	return !apiequality.Semantic.DeepEqual(specOf(oldT), specOf(newT))
 }
 
 func newNfdController(config *restclient.Config, nfdApiControllerOptions nfdApiControllerOptions) (*nfdController, error) {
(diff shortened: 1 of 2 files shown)


ArangoGutierrez (commented):
Third in the series with #2530 and #2531, and they reinforce each other - this cuts nfd-master reconcile churn, and #2530 shrinks the very NodeFeature.Spec.Features that specChanged now DeepEquals.

The gate is correct:
- Status feedback loop fixed: NodeFeatureGroup's master-written status no longer re-triggers a reconcile, since only Spec is compared (covered by the status-only test case). This is the main win.
- Spec-only is the right granularity: the handlers derive their result from Spec, and the only mutable metadata any of them reads is the NodeFeature node-name label via getNodeNameForObj (object names are immutable). The worker rewrites that label to a constant on every update, so it can't change while Spec stays equal - worth keeping that invariant in mind if the worker's labeling ever changes.
- Fails open on a type-assertion failure, so a needed reconcile is never skipped (covered by the nil / mismatched-type cases).
- DeepEqual cost on the hourly resync stays well below the full reconcile it avoids, and #2530 reduces it further.

The test is real and discriminating - metadata-only, status-only, feature, spec, and fail-open cases all flip the expected result. No issues found.

ArangoGutierrez (commented):
Nice optimization, and TestSpecChanged is thorough (the fail-open and status-only cases especially).

One thing worth confirming on the record: this looks like it removes the only periodic reconcile path. The informer resync (1h default) is the sole periodic trigger here, there's no separate ticker or wait.Until loop in nfd-master, and the worker only pushes a NodeFeature update when the object actually changed (the apiequality.Semantic.DeepEqual guard in updateNodeFeatureObject). So once specChanged starts filtering the resync replays, externally-drifted node labels/taints/extended-resources are re-applied only on a real spec change or a master restart rather than continuously.

That's very likely the intended tradeoff given the scale win, but since it changes a controller-correctness property it'd be good to call out in the PR description (or track a follow-up). Not blocking.

PiotrProkop:
> Nice optimization, and TestSpecChanged is thorough (the fail-open and status-only cases especially).
> 
> One thing worth confirming on the record: this looks like it removes the only periodic reconcile path. The informer resync (1h default) is the sole periodic trigger here, there's no separate ticker or wait.Until loop in nfd-master, and the worker only pushes a NodeFeature update when the object actually changed (the apiequality.Semantic.DeepEqual guard in updateNodeFeatureObject). So once specChanged starts filtering the resync replays, externally-drifted node labels/taints/extended-resources are re-applied only on a real spec change or a master restart rather than continuously.
> 
> That's very likely the intended tradeoff given the scale win, but since it changes a controller-correctness property it'd be good to call out in the PR description (or track a follow-up). Not blocking.

I haven't removed periodic resync of all nodes, check here -> https://github.com/PiotrProkop/node-feature-discovery/blob/783297855c63106e9a6072f34739303f47b9945b/pkg/nfd-master/nfd-master.go#L386 . This PR just limits updates based on NodeFeature/Rule/RuleGroup updates to spec changes.

ArangoGutierrez:
You're right, thanks for the logs - I had this backwards. The periodic full reconcile lives in the resync ticker in nfd-master.go (the `time.NewTicker(ResyncPeriod)` -> `nfdAPIUpdateAllNodes` path that emits those hourly "will process all nodes in the cluster" lines), and it's independent of the informer resync replays this PR gates. So self-heal is preserved: the hourly sweep re-applies every node regardless of `specChanged`, and the gate just drops the redundant resync-replay and metadata-only reconciles. My earlier note was off - I checked it against an out-of-date local tree that predated that ticker.

ArangoGutierrez (approved):
The periodic-reconcile ticker preserves the full hourly sweep, so gating the informer handlers cleanly drops the redundant resync-replay and metadata-only reconciles without losing self-heal. specChanged fails open on a type mismatch, and TestSpecChanged covers the resync no-op, metadata-only, spec-change, status-only, and fail-open cases. Reads well.

PR #2533, opened 2026-06-25
cpu/s390x: implement getCpuidAttributes() with machine generation mapping

Basavaraju-G:
Currently getCpuidAttributes() returns nil on s390x, leaving the cpuid attribute namespace unused. Implemented it to expose machine_type and generation (z13, z14, z15, z16, z17) by parsing /proc/cpuinfo.

This enables NodeFeatureRules to match on IBM Z generation, 
e.g.:
  matchExpressions:
    generation: { op: In, value: ["z17"] }

### DIFF
--- a/source/cpu/cpuid_linux_s390x.go
+++ b/source/cpu/cpuid_linux_s390x.go
@@ -16,6 +16,12 @@ limitations under the License.
 
 package cpu
 
+import (
+	"bufio"
+	"os"
+	"strings"
+)
+
 /*
 #include <sys/auxv.h>
 
@@ -96,4 +102,56 @@ func getCpuidFlags() []string {
 	return r
 }
 
-func getCpuidAttributes() map[string]string { return nil }
+func getCpuidAttributes() map[string]string {
+	attrs := make(map[string]string)
+
+	machineType := parseMachineType()
+	if machineType != "" {
+		attrs["machine_type"] = machineType
+		if gen, ok := machineGenerations[machineType]; ok {
+			attrs["generation"] = gen
+		}
+	}
+
+	return attrs
+}
+
+// machineGenerations maps s390x machine type codes to generation names.
+// Reference: Linux kernel arch/s390/kernel/processor.c setup_elf_platform()
+// https://github.com/torvalds/linux/blob/master/arch/s390/kernel/processor.c#L262
+var machineGenerations = map[string]string{
+	"2817": "z196",
+	"2818": "z196",
+	"2827": "zEC12",
+	"2828": "zEC12",
+	"2964": "z13",
+	"2965": "z13",
+	"3906": "z14",
+	"3907": "z14",
+	"8561": "z15",
+	"8562": "z15",
+	"3931": "z16",
+	"3932": "z16",
+	"9175": "z17",
+	"9176": "z17",
+}
+
+func parseMachineType() string {
+	f, err := os.Open("/proc/cpuinfo")
+	if err != nil {
+		return ""
+	}
+	defer f.Close()
+
+	scanner := bufio.NewScanner(f)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if !strings.HasPrefix(line, "processor") {
+			continue
+		}
+		if idx := strings.LastIndex(line, "machine = "); idx != -1 {
+			return strings.TrimSpace(line[idx+len("machine = "):])
+		}
+	}
+	return ""
+}


Basavaraju-G:
@adrianchiris @ArangoGutierrez would you please review this PR?

source/cpu/cpuid_linux_s390x.go
```diff
@@ -96,4 +102,56 @@ func getCpuidFlags() []string {
 	return r
 }
 
-func getCpuidAttributes() map[string]string { return nil }
+func getCpuidAttributes() map[string]string {
+	attrs := make(map[string]string)
+
+	machineType := parseMachineType()
+	if machineType != "" {
+		attrs["machine_type"] = machineType
```
ArangoGutierrez, line 110: This value is already published on s390x: getS390xAttributes() in cpu.go reads the same /proc/cpuinfo machine field into cpu-model.id (commit e9cc443f4, "Map s390x 'machine' type to 'cpu-model.id'"), so this adds a second permanent label for one fact. Note GetLabels applies the cpuid attributeWhitelist/attributeBlacklist to flags only, never to attributes, so the duplicate can't be filtered off either. Putting machine_type and generation under cpu-model alongside vendor_id/family/id/hypervisor would match where s390x identity data already lives.

source/cpu/cpuid_linux_s390x.go
```diff
@@ -96,4 +102,56 @@ func getCpuidFlags() []string {
...
+		}
+	}
+
+	return attrs
+}
+
+// machineGenerations maps s390x machine type codes to generation names.
+// Reference: Linux kernel arch/s390/kernel/processor.c setup_elf_platform()
+// https://github.com/torvalds/linux/blob/master/arch/s390/kernel/processor.c#L262
+var machineGenerations = map[string]string{
```
ArangoGutierrez, line 122: The package documents the opposite placement for exactly this shape of table: the comment above microarchLevelFlags in cpu.go says it is kept in the arch-neutral file "so the mapping can be unit tested on any GOARCH", with TestX86_64LevelFlags as the host-independent table test. Here the map and its parser sit behind linux/s390x plus cgo, and presubmits build only amd64 and arm64, so a typo in any of the 14 keys would ship unexercised. Moving the map (and a parseMachineType taking an io.Reader) into cpu.go would make both testable.

source/cpu/cpuid_linux_s390x.go
```diff
@@ -96,4 +102,56 @@ func getCpuidFlags() []string {
...
+	"8562": "z15",
+	"3931": "z16",
+	"3932": "z16",
+	"9175": "z17",
+	"9176": "z17",
+}
+
+func parseMachineType() string {
+	f, err := os.Open("/proc/cpuinfo")
+	if err != nil {
```
ArangoGutierrez, line 141: The os.Open error is dropped and scanner.Err() is never checked, so an unreadable or truncated /proc/cpuinfo is indistinguishable from "no machine type here" and nothing is logged. The sibling reader getHypervisorFromProcSysinfo in cpu.go reports the equivalent failure with klog.ErrorS.

ArangoGutierrez (commented):
The machine-type table itself checks out: the 14 entries match setup_elf_platform() in arch/s390/kernel/processor.c exactly (all 8 machine-type pairs, z196 through z17), and the line format matches what show_cpu_summary() emits ("processor %d: version = %02X,  identification = %06X,  machine = %04X"). Machine types are BCD, so the %04X rendering matches the map keys. Three things worth sorting out before merge.

1. This value is already published on s390x: getS390xAttributes() in cpu.go reads the same /proc/cpuinfo machine field into cpu-model.id (commit e9cc443f4, "Map s390x 'machine' type to 'cpu-model.id'"), so this adds a second permanent label for one fact. Note GetLabels applies the cpuid attributeWhitelist/attributeBlacklist to flags only, never to attributes, so the duplicate can't be filtered off either. Putting machine_type and generation under cpu-model alongside vendor_id/family/id/hypervisor would match where s390x identity data already lives. (source/cpu/cpuid_linux_s390x.go:110)
2. The package documents the opposite placement for exactly this shape of table: the comment above microarchLevelFlags in cpu.go says it is kept in the arch-neutral file "so the mapping can be unit tested on any GOARCH", with TestX86_64LevelFlags as the host-independent table test. Here the map and its parser sit behind linux/s390x plus cgo, and presubmits build only amd64 and arm64, so a typo in any of the 14 keys would ship unexercised. Moving the map (and a parseMachineType taking an io.Reader) into cpu.go would make both testable. (source/cpu/cpuid_linux_s390x.go:122)
3. The os.Open error is dropped and scanner.Err() is never checked, so an unreadable or truncated /proc/cpuinfo is indistinguishable from "no machine type here" and nothing is logged. The sibling reader getHypervisorFromProcSysinfo in cpu.go reports the equivalent failure with klog.ErrorS. (source/cpu/cpuid_linux_s390x.go:141)

PR #2534, opened 2026-06-26
fix(nfd-master): prevent stop() panic and informer leak on init failure

PiotrProkop:
nfdController.stop() dereferenced c.namespaceLister unconditionally, but that lister is only created when a NodeFeature namespace selector is configured. In the default configuration the field is nil, so stopping the master panicked with a nil-pointer dereference. Guard the call.

newNfdController() starts the shared informers and then returns an error if WaitForCacheSync fails, previously without stopping them and leaking those goroutines. Call stop() before returning so a failed construction cleans up after itself.


marquiz (approved):
Thanks @PiotrProkop for the fix, looks good to me

PiotrProkop:
@marquiz thanks!, @ArangoGutierrez can you review/lgtm?

ArangoGutierrez (approved):
/lgtm

PR #2535, opened 2026-06-26
fix(nfd-worker): trim whitespace in noPublishFeatures patterns

PiotrProkop:
StringSliceVal keeps the spaces around comma-separated entries, so a configured value like "pci.device, usb.*" yielded a " usb.*" pattern that never matched any feature key and was wrongly reported as unmatched. Trim each pattern once up front and ignore blank entries, both when matching and when collecting the unmatched patterns. Add tests covering padded and blank patterns.

Also drop the per-call []*regexp.Regexp slice in the MatchInRegexp branch of evaluateMatchExpression: now that compiled patterns are cached, compile and match in a single pass. Every pattern is still compiled, so an invalid regexp is reported even after a match, while the last per-call allocation on that hot path is removed.

### DIFF
--- a/pkg/utils/flags.go
+++ b/pkg/utils/flags.go
@@ -81,9 +81,16 @@ func (a *StringSetVal) UnmarshalJSON(data []byte) error {
 // StringSliceVal is a Value encapsulating a slice of comma-separated strings
 type StringSliceVal []string
 
-// Set implements the regexp.Value interface
+// Set implements the regexp.Value interface. Surrounding whitespace is trimmed
+// from each comma-separated entry and blank entries are dropped, so a value like
+// "cpu, pci" parses to ["cpu", "pci"] and "" parses to an empty slice.
 func (a *StringSliceVal) Set(val string) error {
-	*a = strings.Split(val, ",")
+	*a = nil
+	for s := range strings.SplitSeq(val, ",") {
+		if s = strings.TrimSpace(s); s != "" {
+			*a = append(*a, s)
+		}
+	}
 	return nil
 }
 
--- a/pkg/utils/flags_test.go
+++ b/pkg/utils/flags_test.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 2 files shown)


pkg/apis/nfd/nodefeaturerule/expression.go
```diff
@@ -144,19 +144,22 @@ func evaluateMatchExpression(m *nfdv1alpha1.MatchExpression, valid bool, value i
...
 				re, err := compileRegexp(v)
 				if err != nil {
 					return false, fmt.Errorf("invalid expressiom, 'value' field must only contain valid regexps for Op %q (have %v)", m.Op, m.Value)
 				}
-				valueRe[i] = re
-			}
-			for _, re := range valueRe {
 				if re.MatchString(value) {
-					return true, nil
+					matched = true
```
marquiz: Do we need that matched for something? Just return from here?

Also, this change is out-of-context so could you put it at least in a separate commit?

pkg/nfd-worker/nfd-worker.go
```diff
@@ -678,10 +678,21 @@ func removeNoPublishFeatures(features *nfdv1alpha1.Features, patterns []string)
 	if features == nil || len(patterns) == 0 {
 		return nil
 	}
+	// StringSliceVal preserves spaces around comma-separated entries, so
+	// "pci.device, usb.*" yields a " usb.*" entry; trim once up front so such a
+	// pattern still matches instead of being wrongly reported as unmatched.
+	// Blank entries are ignored.
```
marquiz: Hmm, I think it might be better to fix the StringSliceVal instead, i.e. strip the whitespace there. Avoid problems with other similar flags, too

pkg/apis/nfd/nodefeaturerule/expression.go
PiotrProkop: yeah, I'll move it to another commit. I am not returning here to keep the old behaviour compile/validate every pattern before returning. With a plain `return true, nil` on first match, a trailing invalid regexp goes unreported whenever an earlier pattern happens to match. Maybe we should revisit this mechanism?

pkg/nfd-worker/nfd-worker.go
PiotrProkop: makes sense, will move fix to `StringSliceVal`

pkg/apis/nfd/nodefeaturerule/expression.go
PiotrProkop: I'll drop this commit and prepare a new PR, with the changed behaviou.  To return on first match and stop checking followin patterns.

pkg/nfd-worker/nfd-worker.go
PiotrProkop: fixed

ArangoGutierrez (approved):
leaving lgtm to @marquiz

marquiz (approved):
Thanks @PiotrProkop for the fix

/lgtm

PR #2545, opened 2026-07-07
feat(nfd-worker): configure NodeFeature owner references

PiotrProkop:
Add ownerRefs configuration and the -owner-refs flag with support for node, pod, and ds owners. Preserve pod,ds as the default and retain noOwnerRefs as a deprecated compatibility option.

Replace existing owner references in place during worker publication and track the publishing Pod UID. Reconcile the target Node when ownership or publisher identity changes so labels are restored after Node rebuilds even when the NodeFeature spec is unchanged.

Add Node read permissions for workers and update Helm, Kustomize, documentation, and tests.


docs/reference/worker-configuration-reference.md
```diff
@@ -180,10 +180,52 @@ core:
...
+least one owner still exists. Thus, adding `ds` makes the DaemonSet the
+longest-lived owner during a Node rebuild.
+
+Selecting `node` requires the worker service account to have `get` permission
+for Nodes. The bundled Helm and Kustomize deployments include this permission;
+installations with custom RBAC must add it. The NFD garbage collector may still
+explicitly delete stale NodeFeature objects independently of native
+owner-reference garbage collection.
+
+When this option is changed, each running worker replaces the complete owner
```
ArangoGutierrez: setOwnerReference() runs once at startup and the worker has no config-reload path, so an ownerRefs change only takes effect when the worker pod restarts, not on the next publication of a running worker (the new e2e migration test restarts the pod for exactly this reason). Suggest rewording to say migration happens as workers restart, e.g. on the DaemonSet rollout that delivers the new config.

cmd/nfd-worker/main.go
```diff
@@ -123,11 +125,14 @@ func initFlags(flagset *flag.FlagSet) (*worker.Args, *worker.ConfigOverrideArgs)
 		FeatureSources:    &utils.StringSliceVal{},
 		LabelSources:      &utils.StringSliceVal{},
 		NoPublishFeatures: &utils.StringSliceVal{},
+		OwnerRefs:         &worker.OwnerRefSources{"pod", "ds"},
 	}
 	overrides.NoPublish = flagset.Bool("no-publish", false,
 		"Do not publish discovered features, disable connection to nfd-master and don't create NodeFeature object.")
 	overrides.NoOwnerRefs = flagset.Bool("no-owner-refs", false,
-		"Do not set owner references for NodeFeature object.")
+		"Do not set owner references for NodeFeature object (deprecated: use -owner-refs=).")
```
ArangoGutierrez, line 133: The deprecated -no-owner-refs / core.noOwnerRefs paths log nothing when used, while the core.sources deprecation in nfd-worker.go emits a klog warning ("usage of deprecated 'core.sources' config file option..."). A similar warning here would give users a migration signal before the option is eventually removed.

deployment/helm/node-feature-discovery/templates/clusterrole.yaml
```diff
@@ -57,6 +57,23 @@ rules:
   - update
 {{- end }}
 
+{{- if and .Values.worker.enable .Values.worker.rbac.create }}
```
ArangoGutierrez, line 60: This grants every worker cluster-wide get on nodes even though the default ownerRefs=[pod,ds] never exercises it, and worker RBAC has stayed off the Node API since the NodeFeature split. It is get-only so the risk is bounded, but consider gating the grant behind a dedicated toggle so default installs do not carry an unused cluster-scoped permission.

ArangoGutierrez (commented):
Found 3 issues: one docs-accuracy, one deprecation-convention, one RBAC-scope consideration. The core implementation checks out - precedence rules, owner-ref replacement semantics, the master's update filtering, and the e2e RBAC pattern all verified clean. One housekeeping note: this overlaps with the still-open #2041, which it effectively supersedes - worth cross-linking/closing that one.

1. setOwnerReference() runs once at startup and the worker has no config-reload path, so an ownerRefs change only takes effect when the worker pod restarts, not on the next publication of a running worker (the new e2e migration test restarts the pod for exactly this reason). Suggest rewording to say migration happens as workers restart, e.g. on the DaemonSet rollout that delivers the new config. (docs/reference/worker-configuration-reference.md:204)
2. The deprecated -no-owner-refs / core.noOwnerRefs paths log nothing when used, while the core.sources deprecation in nfd-worker.go emits a klog warning ("usage of deprecated 'core.sources' config file option..."). A similar warning here would give users a migration signal before the option is eventually removed. (cmd/nfd-worker/main.go:133)
3. This grants every worker cluster-wide get on nodes even though the default ownerRefs=[pod,ds] never exercises it, and worker RBAC has stayed off the Node API since the NodeFeature split. It is get-only so the risk is bounded, but consider gating the grant behind a dedicated toggle so default installs do not carry an unused cluster-scoped permission. (deployment/helm/node-feature-discovery/templates/clusterrole.yaml:60)

deployment/helm/node-feature-discovery/templates/clusterrole.yaml
PiotrProkop: fixed

cmd/nfd-worker/main.go
PiotrProkop: fixed

docs/reference/worker-configuration-reference.md
PiotrProkop: fixed

PiotrProkop:
@ArangoGutierrez thanks for the review! I've applied your suggestions. Please take a look.

ArangoGutierrez (commented):
All three points from the previous round are addressed: the deprecation warning for `-no-owner-refs` / `core.noOwnerRefs` now fires on both the flag and config paths (with a klog-capture unit test), the worker Node RBAC is gated behind `worker.ownerRefs` containing `node` in Helm plus an opt-in `worker-node-rbac` kustomize component (default installs carry no extra grant), and the `core.ownerRefs` docs now spell out the restart-based rollout and in-place migration semantics.

The new e2e coverage for Node-rebuild label restoration and owner migration is a nice addition. No further issues found.

PR #2403, opened 2025-12-27
Add NFD image compatibility scheduler proposal.

Xunli-Yang:
Add KEP: NFD image compatibility scheduler proposal.

What the proposal does?
Building upon the first phase of [KEP-1845 Proposal](https://github.com/kubernetes-sigs/node-feature-discovery/blob/master/enhancements/1845-nfd-image-compatibility/README.md), which completed node compatibility validation. This proposal introduces a compatibility scheduling plugin. The compatibility scheduler plugin automatically analyzes the compatibility requirements of container images, filters suitable nodes for scheduling, and ensures that containers run on compatible nodes.

Special notes for reviewer:
Based on the discussions on [node-feature-discovery Slack channel](https://app.slack.com/client/T09NY5SBT/C01A82FKSQK), this proposal has presented three solutions and intends to get consensus on the implementation direction.
Co-authored-by:  @ChaoyiHuang

PR #2559, opened 2026-07-10
fix(nfd-master): scope NodeFeatureGroup updates to their own namespace

### LINKED ISSUE #2558: nfd-master: NodeFeatureGroups outside nfd-master's namespace are listed and enqueued but silently never updated
### What happened

`nfdAPIUpdateAllNodeFeatureGroups` lists NodeFeatureGroups across all namespaces, but the update pipeline resolves and writes them only in nfd-master's own namespace:

- [`pkg/nfd-master/nfd-master.go#L816`](https://github.com/kubernetes-sigs/node-feature-discovery/blob/d50636b383452f04be58da034d5cfd2b001d48f7/pkg/nfd-master/nfd-master.go#L816) — cluster-wide list (`List(k8slabels.Everything())`)
- [`pkg/nfd-master/nfd-master.go#L823`](https://github.com/kubernetes-sigs/node-feature-discovery/blob/d50636b383452f04be58da034d5cfd2b001d48f7/pkg/nfd-master/nfd-master.go#L823) — enqueued by name only, namespace dropped
- [`pkg/nfd-master/updater-pool.go#L108`](https://github.com/kubernetes-sigs/node-feature-discovery/blob/d50636b383452f04be58da034d5cfd2b001d48f7/pkg/nfd-master/updater-pool.go#L108) — consumer re-resolves with `getNodeFeatureGroup(cli, u.nfdMaster.namespace, nfgName)`
- [`pkg/nfd-master/nfd-master.go#L887`](https://github.com/kubernetes-sigs/node-feature-discovery/blob/d50636b383452f04be58da034d5cfd2b001d48f7/pkg/nfd-master/nfd-master.go#L887) — status write likewise pinned to `m.namespace`

Net effect:

1. An NFG created in any other namespace is watched and enqueued, but the lookup NotFounds in nfd-master's namespace and the update is skipped with an info-level log — its `status.nodes` never populates.
2. Two NFGs with the same name in different namespaces collide in the queue: events for either trigger processing of whichever lives in nfd-master's namespace.

### What did you expect to happen

Either (a) NFGs namespace-scoped end-to-end — namespace/name key through the queue, `Get`/`UpdateStatus` against the object's own namespace — or (b) nfd-master explicitly watches only its own namespace (as the NodeFeature path does via `isNamespaceSelected`) so out-of-scope NFGs aren't silently half-processed. Today's behavior is an inconsistent middle: cluster-wide watch, own-namespace processing, silent skip.

### How to reproduce

1. Deploy NFD with `NodeFeatureGroupAPI` enabled (nfd-master in the `node-feature-discovery` namespace).
2. `kubectl create ns team-a && kubectl apply -n team-a -f examples/nodefeaturegroup.yaml`
3. The NFG in `team-a` never gets `status.nodes`; nfd-master logs "NodeFeatureGroup not found, skip update".

### Anything else we need to know

Found while reviewing #2403, whose pre-grouping design assumes admin-created NFGs work; related to the namespace-scoping discussion in #2540. Happy to send a fix — leaning (a): thread namespace/name through the queue, use the object's namespace at resolution and `UpdateStatus`, and add a regression test with same-name NFGs in two namespaces. Note (a) makes cross-namespace NFGs work, which touches the multi-tenancy scoping debate in #2540 — if we'd rather not pre-empt that, (b) namespace-filter plus a loud event is the conservative alternative.

**Version:** master (`d50636b`)

ArangoGutierrez:
**What this PR does / why we need it:**

nfd-master lists NodeFeatureGroups cluster-wide but processes them pinned to its own namespace: the update queue carried bare names (so same-name NFGs in different namespaces collide), and both the consumer lookup and `UpdateStatus` used nfd-master's namespace — NFGs in any other namespace were enqueued and then silently skipped, their `status.nodes` never populated.

This threads namespace/name through the NFG update queue (`cache.MetaNamespaceKeyFunc` / `cache.SplitMetaNamespaceKey` at every enqueue site, including the informer handlers), resolves and updates each NFG in its own namespace, and keeps the cluster-wide listing.

**Testing:** regression tests that fail on master — same-name NFGs in two namespaces must both get `status.nodes` populated independently, and `UpdateStatus` must target the object's namespace (asserted via fake-client actions). Full `pkg/nfd-master` suite green.

**Fixes #2558**

/cc @PiotrProkop @Xunli-Yang

### DIFF
--- a/pkg/nfd-master/nfd-master-nfg-namespace_test.go
+++ b/pkg/nfd-master/nfd-master-nfg-namespace_test.go
@@ -1,0 +1,229 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package nfdmaster
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	fakeclient "k8s.io/client-go/kubernetes/fake"
+	"k8s.io/client-go/tools/cache"
+
+	fakenfdclient "sigs.k8s.io/node-feature-discovery/api/generated/clientset/versioned/fake"
+	nfdscheme "sigs.k8s.io/node-feature-discovery/api/generated/clientset/versioned/scheme"
+	nfdinformers "sigs.k8s.io/node-feature-discovery/api/generated/informers/externalversions"
+	nfdv1alpha1 "sigs.k8s.io/node-feature-discovery/api/nfd/v1alpha1"
+)
+
+const (
+	nfgTestMasterNamespace = "node-feature-discovery"
+	nfgTestOtherNamespace  = "other-namespace"
+	nfgTestName            = "test-nfg"
+)
+
+// newFakeNfdAPIControllerWithNFG builds a controller whose NodeFeature and
+// NodeFeatureGroup listers are backed by (and synced from) the given fake
+// client. Unlike newFakeNfdAPIController it also wires the NodeFeatureGroup
+// lister, which nfdAPIUpdateAllNodeFeatureGroups needs.
+func newFakeNfdAPIControllerWithNFG(client *fakenfdclient.Clientset) *nfdController {
+	c := &nfdController{
+		stopChan:           make(chan struct{}),
+		updateAllNodesChan: make(chan struct{}, 1),
+		updateOneNodeChan:  make(chan string),
+	}
+
+	informerFactory := nfdinformers.NewSharedInformerFactory(client, 1*time.Hour)
+
+	featureInformer := informerFactory.Nfd().V1alpha1().NodeFeatures()
+	c.featureLister = featureInformer.Lister()
+
+	nodeFeatureGroupInformer := informerFactory.Nfd().V1alpha1().NodeFeatureGroups()
+	c.featureGroupLister = nodeFeatureGroupInformer.Lister()
+
+	informerFactory.Start(c.stopChan)
+	cache.WaitForCacheSync(c.stopChan,
+		featureInformer.Informer().HasSynced,
+		nodeFeatureGroupInformer.Informer().HasSynced,
+	)
+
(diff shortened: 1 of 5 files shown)


pkg/nfd-master/nfd-master-nfg-namespace_test.go
```diff
@@ -0,0 +1,229 @@
...
+const (
+	nfgTestMasterNamespace = "node-feature-discovery"
+	nfgTestOtherNamespace  = "other-namespace"
+	nfgTestName            = "test-nfg"
+)
+
+// newFakeNfdAPIControllerWithNFG builds a controller whose NodeFeature and
+// NodeFeatureGroup listers are backed by (and synced from) the given fake
+// client. Unlike newFakeNfdAPIController it also wires the NodeFeatureGroup
+// lister, which nfdAPIUpdateAllNodeFeatureGroups needs.
```
marquiz: Wonder could we reduce the boilerplate by extending the existing newFakeNfdAPIController?

marquiz (commented):
Thanks @ArangoGutierrez for a fix for this rather embarrassing bug. Looks very straightforward, just one minor comment re the tests

ArangoGutierrez:
/cc @PiotrProkop

Piotr, this one is small and now sits on the critical path for KEP-2403, so it would help to get it moving.

The bug: nfd-master lists NodeFeatureGroups cluster-wide but enqueues them by name only, and the consumer then resolves and writes them in nfd-master's own namespace. A NodeFeatureGroup in any other namespace is picked up and silently never updated, and two same-named groups in different namespaces collide. It is still present on master today at `nfd-master.go` in the `UpdateStatus` call pinned to `m.namespace`. The fix threads namespace/name queue keys via `cache.MetaNamespaceKeyFunc` and `cache.SplitMetaNamespaceKey` through every enqueue site including the informer handlers, resolves Get and UpdateStatus against the object's own namespace, and keeps the cluster-wide listing. Regression tests fail on revert.

Why now: KEP-2403 (#2403) builds node pre-grouping on NodeFeatureGroup, and that design assumes cluster-wide resolution that does not exist yet, so this wants to land before the Alpha implementation starts.

@Xunli-Yang tagging you as well since this is a dependency of your pre-grouping path, and you may want to look at it before you build on top of NodeFeatureGroup.

CI is green but ran against a base from July. Happy to rebase onto current master for a fresh run if either of you would rather review it that way.

pkg/nfd-master/nfd-master-nfg-namespace_test.go
ArangoGutierrez: Good call, folded it in. `newFakeNfdAPIController` now wires the NodeFeatureGroup lister alongside NodeFeature and NodeFeatureRule, and waits for all three caches to sync so the listers are usable when it returns. The duplicate helper is gone and both tests use the shared one, which drops about 30 lines and five imports.

The cache sync is new for the three existing callers; full `pkg/nfd-master` suite passes. I also re-checked that the two regression tests still fail with the fix commit reverted, so folding the helper did not make them pass vacuously.

Xunli-Yang (commented):
Thanks @ArangoGutierrez , looks good to me.  This is exactly the cluster-wide nodeFeatureGroup resolution KEP-2403's pre-grouping path assumes. It's reasonable that a NodeFeatureGroup in any namespace gets its status.nodes populated instead of being silently skipped. Landing this before Alpha will help me to build on top of it.

PR #2564, opened 2026-07-14
move get+update to a server side patch for topology updater

alaypatel07:
Before this PR, the NFD updater did a get+update to avoid resource
version conflicts. 

With this PR, the nfd updater becomes the owner of the field and uses 
server side apply to patch the field. This comes with performance benefits
as it reduces an unnecessary get before the update.

At scale this will help with reducing CPU usage in kube-apiserver and etcd.

### DIFF
--- a/deployment/base/rbac-topologyupdater/topologyupdater-clusterrole.yaml
+++ b/deployment/base/rbac-topologyupdater/topologyupdater-clusterrole.yaml
@@ -36,8 +36,7 @@ rules:
   - noderesourcetopologies
   verbs:
   - create
-  - get
-  - update
+  - patch
 - apiGroups:
   - apiextensions.k8s.io
   resources:
--- a/deployment/helm/node-feature-discovery/templates/clusterrole.yaml
+++ b/deployment/helm/node-feature-discovery/templates/clusterrole.yaml
@@ -118,8 +118,7 @@ rules:
   - noderesourcetopologies
   verbs:
   - create
-  - get
-  - update
+  - patch
 - apiGroups:
   - apiextensions.k8s.io
   resources:
--- a/pkg/nfd-topology-updater/nfd-topology-updater.go
+++ b/pkg/nfd-topology-updater/nfd-topology-updater.go
@@ -18,6 +18,7 @@ package nfdtopologyupdater
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -28,10 +29,11 @@ import (
 	"time"
 
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
+	metav1apply "k8s.io/client-go/applyconfigurations/meta/v1"
 	"k8s.io/client-go/informers"
 	k8sclient "k8s.io/client-go/kubernetes"
 	restclient "k8s.io/client-go/rest"
@@ -40,6 +42,7 @@ import (
 	kubeletconfigv1beta1 "k8s.io/kubelet/config/v1beta1"
 
 	"github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/apis/topology/v1alpha2"
+	applyconfigurationtopologyv1alpha2 "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/applyconfiguration/topology/v1alpha2"
 	topologyclientset "github.com/k8stopologyawareschedwg/noderesourcetopology-api/pkg/generated/clientset/versioned"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
@@ -60,6 +63,8 @@ import (
 const nodeScopedPodInformerResync = 0
 
 const (
+	topologyManagerInfoRefreshTimeout = 2 * time.Minute
+
 	// TopologyManagerPolicyAttributeName represents an attribute which defines
 	// Topology Manager Policy
 	TopologyManagerPolicyAttributeName = "topologyManagerPolicy"
@@ -68,7 +73,12 @@ const (
 	TopologyManagerScopeAttributeName = "topologyManagerScope"
 	// NodeResourceTopologyCRDName is the name of the NodeResourceTopology CRD
 	NodeResourceTopologyCRDName = "noderesourcetopologies.topology.node.k8s.io"
+	// nodeResourceTopologyFieldManager is used for server-side apply updates to
+	// NodeResourceTopology objects.
+	nodeResourceTopologyFieldManager = "node-feature-discovery-topology-updater"
 )
+
+var errTopologyManagerInfoUnavailable = errors.New("topology-manager information unavailable")
 
 // Args are the command line arguments
 type Args struct {
@@ -93,18 +103,22 @@ type NfdTopologyUpdater interface {
 }
 
 type nfdTopologyUpdater struct {
-	nodeName            string
-	args                Args
-	topoClient          topologyclientset.Interface
-	resourcemonitorArgs resourcemonitor.Args
-	stop                chan struct{} // channel for signaling stop
-	eventSource         <-chan kubeletnotifier.Info
-	configFilePath      string
-	config              *NFDConfig
-	kubernetesNamespace string
-	ownerRefs           []metav1.OwnerReference
-	k8sClient           k8sclient.Interface
-	kubeletConfigFunc   func() (*kubeletconfigv1beta1.KubeletConfiguration, error)
+	nodeName             string
+	args                 Args
+	topoClient           topologyclientset.Interface
+	resourcemonitorArgs  resourcemonitor.Args
+	stop                 chan struct{} // channel for signaling stop
+	eventSource          <-chan kubeletnotifier.Info
+	configFilePath       string
+	config               *NFDConfig
+	kubernetesNamespace  string
+	ownerRefs            []metav1.OwnerReference
+	k8sClient            k8sclient.Interface
+	kubeletConfigFunc    func(context.Context) (*kubeletconfigv1beta1.KubeletConfiguration, error)
+	discoverCpuCoresFunc func() v1alpha2.AttributeList
+	kubeletConfigMu      sync.RWMutex
+	kubeletConfig        *kubeletconfigv1beta1.KubeletConfiguration
+	cpuCoreAttributes    v1alpha2.AttributeList
 }
 
 // NewTopologyUpdater creates a new NfdTopologyUpdater instance.
@@ -123,15 +137,16 @@ func NewTopologyUpdater(args Args, resourcemonitorArgs resourcemonitor.Args) (Nf
 	}
 
 	nfd := &nfdTopologyUpdater{
-		args:                args,
-		resourcemonitorArgs: resourcemonitorArgs,
-		stop:                make(chan struct{}),
-		nodeName:            utils.NodeName(),
-		eventSource:         eventSource,
-		config:              &NFDConfig{},
-		kubernetesNamespace: utils.GetKubernetesNamespace(),
-		ownerRefs:           []metav1.OwnerReference{},
-		kubeletConfigFunc:   kubeletConfigFunc,
+		args:                 args,
+		resourcemonitorArgs:  resourcemonitorArgs,
+		stop:                 make(chan struct{}),
+		nodeName:             utils.NodeName(),
+		eventSource:          eventSource,
+		config:               &NFDConfig{},
+		kubernetesNamespace:  utils.GetKubernetesNamespace(),
+		ownerRefs:            []metav1.OwnerReference{},
+		kubeletConfigFunc:    kubeletConfigFunc,
+		discoverCpuCoresFunc: discoverCpuCores,
 	}
 	if args.ConfigFile != "" {
 		nfd.configFilePath = filepath.Clean(args.ConfigFile)
@@ -139,10 +154,52 @@ func NewTopologyUpdater(args Args, resourcemonitorArgs resourcemonitor.Args) (Nf
 	return nfd, nil
 }
 
-func (w *nfdTopologyUpdater) detectTopologyPolicyAndScope() (string, string, error) {
-	klConfig, err := w.kubeletConfigFunc()
+// refreshTopologyManagerInfo fetches and caches the inputs used to publish
+// topology-manager information.
+func (w *nfdTopologyUpdater) refreshTopologyManagerInfo(ctx context.Context) error {
+	klConfig, err := w.kubeletConfigFunc(ctx)
 	if err != nil {
-		return "", "", err
+		return err
+	}
+	cpuCoreAttributes := w.discoverCpuCoresFunc()
+
+	w.kubeletConfigMu.Lock()
+	w.kubeletConfig = klConfig
+	w.cpuCoreAttributes = cpuCoreAttributes
+	w.kubeletConfigMu.Unlock()
+	return nil
+}
+
+// getKubeletConfig returns the cached kubelet configuration, or nil if it has
+// not been fetched yet.
+func (w *nfdTopologyUpdater) getKubeletConfig() *kubeletconfigv1beta1.KubeletConfiguration {
+	w.kubeletConfigMu.RLock()
+	defer w.kubeletConfigMu.RUnlock()
+	return w.kubeletConfig
+}
+
+func (w *nfdTopologyUpdater) getCpuCoreAttributes() v1alpha2.AttributeList {
+	w.kubeletConfigMu.RLock()
+	defer w.kubeletConfigMu.RUnlock()
+	return append(v1alpha2.AttributeList(nil), w.cpuCoreAttributes...)
+}
+
+func (w *nfdTopologyUpdater) newTopologyManagerRefreshContext() (context.Context, context.CancelFunc) {
+	ctx, cancel := context.WithTimeout(context.Background(), topologyManagerInfoRefreshTimeout)
+	go func() {
+		select {
+		case <-w.stop:
+			cancel()
+		case <-ctx.Done():
+		}
+	}()
+	return ctx, cancel
+}
+
+func (w *nfdTopologyUpdater) detectTopologyPolicyAndScope() (string, string, error) {
+	klConfig := w.getKubeletConfig()
+	if klConfig == nil {
+		return "", "", fmt.Errorf("kubelet configuration is not cached")
 	}
 
 	return klConfig.TopologyManagerPolicy, klConfig.TopologyManagerScope, nil
@@ -207,6 +264,15 @@ func (w *nfdTopologyUpdater) Run() error {
 
 	if err := w.configure(); err != nil {
 		return fmt.Errorf("faild to configure Node Feature Discovery Topology Updater: %w", err)
+	}
+
+	// Cache topology-manager information once on startup. It is subsequently
+	// refreshed on interval-based events (see the event loop below).
+	refreshCtx, cancelRefresh := w.newTopologyManagerRefreshContext()
+	err = w.refreshTopologyManagerInfo(refreshCtx)
+	cancelRefresh()
+	if err != nil {
+		klog.ErrorS(err, "failed to cache topology-manager information on startup, will retry on next scan")
 	}
 
 	syncCtx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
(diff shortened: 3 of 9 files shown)


alaypatel07:
cc @PiotrProkop

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -327,58 +365,63 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
-	attributes := scanResponse.Attributes
+	data, err := json.Marshal(&nrtApply)
+	if err != nil {
+		return fmt.Errorf("failed to marshal NodeResourceTopology apply patch: %w", err)
+	}
 
-	if readKubeletConfig {
-		if err := w.updateNRTTopologyManagerInfo(nrtMutated); err != nil {
-			return err
+	nrtUpdated, err := w.topoClient.TopologyV1alpha2().NodeResourceTopologies().Patch(
```
alaypatel07: Ideally Apply function should be used here, but it is blocked by this PR: https://github.com/k8stopologyawareschedwg/noderesourcetopology-api/pull/38/

PiotrProkop: the specified PR was just merged, so we can probably use `Apply` func in this PR.

ArangoGutierrez (commented):
Found 5 issues.

1. This refresh runs before the NoPublish gate, so a --no-publish instance, which previously never contacted the kubelet from the event loop, now exits Run() on a transient /configz failure. The startup path tolerates the same failure with a 'will retry on next scan' log; consider treating the refresh failure the same way (log and keep the cached config) instead of returning. (pkg/nfd-topology-updater/nfd-topology-updater.go:322)
2. refreshKubeletConfig goes through the /configz fetcher, which builds its request with a zero Timeout and context.TODO(), so a hung kubelet blocks Run() here indefinitely, right next to but outside the 2-minute syncCtx. Worth bounding this call the same way. (pkg/nfd-topology-updater/nfd-topology-updater.go:248)
3. applyNRTTopologyManagerInfo now runs on every scan event, so discoverCpuCores re-globs and re-reads sysfs per event, where the old readKubeletConfig gate bounded it to creation and interval-based updates. The kubelet config is cached now but the sysfs work is not; with FSUpdate events firing on pod churn this reintroduces the per-event filesystem access the gate was added to avoid (see the #1737 discussion). (pkg/nfd-topology-updater/nfd-topology-updater.go:372)
4. With the updater now using server-side apply exclusively, get and update on noderesourcetopologies are dead privilege; create must stay for create-via-apply, but the other two can be dropped while this rule is being touched. Same applies to the Helm clusterrole template and test/e2e/utils/rbac.go. (deployment/base/rbac-topologyupdater/topologyupdater-clusterrole.yaml:40)
5. This fake's Apply can never fail, so the new IsNotFound branch with the 'CRD may not be installed' hint and the error-wrap path are unreachable in tests. The v0.1.3 bump in this PR ships a generated fake clientset with a real ObjectTracker-backed Apply that would exercise real apply semantics and make those branches testable. (pkg/nfd-topology-updater/nfd-topology-updater_test.go:356)

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -281,13 +318,14 @@ func (w *nfdTopologyUpdater) Run() error {
 			}
 			zones = resAggr.Aggregate(scanResponse.PodResources)
 			klog.V(1).InfoS("aggregated resources identified", "resourceZones", utils.DelayedDumper(zones))
-			readKubeletConfig := false
 			if info.Event == kubeletnotifier.IntervalBased {
-				readKubeletConfig = true
+				if err := w.refreshKubeletConfig(); err != nil {
```
ArangoGutierrez: This refresh runs before the NoPublish gate, so a --no-publish instance, which previously never contacted the kubelet from the event loop, now exits Run() on a transient /configz failure. The startup path tolerates the same failure with a 'will retry on next scan' log; consider treating the refresh failure the same way (log and keep the cached config) instead of returning.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -212,6 +241,14 @@ func (w *nfdTopologyUpdater) Run() error {
 	syncCtx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
 	defer cancel()
 
+	// Cache the kubelet configuration once on startup. It is subsequently
+	// refreshed on interval-based events (see the event loop below), so
+	// topology-manager info can be computed locally without per-scan
+	// kubelet lookups.
+	if err := w.refreshKubeletConfig(); err != nil {
```
ArangoGutierrez: refreshKubeletConfig goes through the /configz fetcher, which builds its request with a zero Timeout and context.TODO(), so a hung kubelet blocks Run() here indefinitely, right next to but outside the 2-minute syncCtx. Worth bounding this call the same way.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -327,58 +365,62 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
-				return fmt.Errorf("failed to create NodeResourceTopology: %w. "+
-					"The NodeResourceTopology CRD may not be installed. "+
-					"If using Helm, ensure 'topologyUpdater.createCRDs=true' is set",
-					err)
-			}
-			return fmt.Errorf("failed to create NodeResourceTopology: %w", err)
-		}
-		return nil
-	} else if err != nil {
+	if err := w.applyNRTTopologyManagerInfo(nrt); err != nil {
```
ArangoGutierrez, line 412: applyNRTTopologyManagerInfo now runs on every scan event, so discoverCpuCores re-globs and re-reads sysfs per event, where the old readKubeletConfig gate bounded it to creation and interval-based updates. The kubelet config is cached now but the sysfs work is not; with FSUpdate events firing on pod churn this reintroduces the per-event filesystem access the gate was added to avoid (see the #1737 discussion).

deployment/base/rbac-topologyupdater/topologyupdater-clusterrole.yaml
```diff
@@ -37,6 +37,7 @@ rules:
   verbs:
   - create
   - get
+  - patch
```
ArangoGutierrez, line 39: With the updater now using server-side apply exclusively, get and update on noderesourcetopologies are dead privilege; create must stay for create-via-apply, but the other two can be dropped while this rule is being touched. Same applies to the Helm clusterrole template and test/e2e/utils/rbac.go.

pkg/nfd-topology-updater/nfd-topology-updater_test.go
```diff
@@ -110,3 +127,236 @@ func findAttributeByName(attrList v1alpha2.AttributeList, name string) (v1alpha2
...
+func (c *fakeNodeResourceTopologyClient) Watch(ctx context.Context, opts metav1.ListOptions) (watch.Interface, error) {
+	return watch.NewEmptyWatch(), nil
+}
+
+func (c *fakeNodeResourceTopologyClient) Patch(ctx context.Context, name string, pt types.PatchType, data []byte, opts metav1.PatchOptions, subresources ...string) (*v1alpha2.NodeResourceTopology, error) {
+	c.patchCalls++
+	return &v1alpha2.NodeResourceTopology{}, nil
+}
+
+func (c *fakeNodeResourceTopologyClient) Apply(ctx context.Context, nrt *applyconfigurationtopologyv1alpha2.NodeResourceTopologyApplyConfiguration, opts metav1.ApplyOptions) (*v1alpha2.NodeResourceTopology, error) {
```
ArangoGutierrez: This fake's Apply can never fail, so the new IsNotFound branch with the 'CRD may not be installed' hint and the error-wrap path are unreachable in tests. The v0.1.3 bump in this PR ships a generated fake clientset with a real ObjectTracker-backed Apply that would exercise real apply semantics and make those branches testable.

deployment/base/rbac-topologyupdater/topologyupdater-clusterrole.yaml
alaypatel07: +1 good catch. Updated.

pkg/nfd-topology-updater/nfd-topology-updater.go
alaypatel07: updated the cache to include CPU information. Also renamed it accordingly.

pkg/nfd-topology-updater/nfd-topology-updater.go
alaypatel07: changed this to logging entry.

pkg/nfd-topology-updater/nfd-topology-updater.go
alaypatel07: added 2-minute timeout.

pkg/nfd-topology-updater/nfd-topology-updater_test.go
alaypatel07: Used the generated fake clientset for units, PTAL

ArangoGutierrez (commented):
Second-round review. The round-1 items are largely addressed: the kubelet fetch is now context-bounded, kubelet config and CPU-core attributes are cached (no per-event sysfs re-glob), and the new tests genuinely exercise the apply/error branches (v0.1.3 + real tracker). Two follow-ups below. Minor: the new server-side Apply still passes context.TODO(), so it stays unbounded even though bounding blocking API calls is the theme of this PR; threading a bounded context there would be consistent.

1. This chart's topology-updater ClusterRole still grants get and update on noderesourcetopologies alongside the new patch. The updater now only server-side applies (patch, plus create-via-apply) and no longer calls Get or Update, so get and update are dead privileges here. The base kustomize manifest and the e2e RBAC were reduced to create and patch in this PR; the Helm block was missed. (deployment/helm/node-feature-discovery/templates/clusterrole.yaml:122)
2. When the startup cache is empty (startup refresh failed) and this refresh also fails, returning the error propagates through updateNodeResourceTopology to Run() and exits the process. On a non-interval publish event with the kubelet briefly unreachable this crash-loops the pod, whereas the startup and interval-refresh paths now tolerate the same failure and keep the cached info. Consider logging and publishing without the topology-manager attributes (or skipping this publish) so it matches that tolerance; the startup 'will retry on next scan' log implies a graceful retry that this path does not provide. (pkg/nfd-topology-updater/nfd-topology-updater.go:439)

deployment/helm/node-feature-discovery/templates/clusterrole.yaml
```diff
@@ -119,6 +119,7 @@ rules:
   verbs:
   - create
   - get
+  - patch
```
ArangoGutierrez, line 121: This chart's topology-updater ClusterRole still grants get and update on noderesourcetopologies alongside the new patch. The updater now only server-side applies (patch, plus create-via-apply) and no longer calls Get or Update, so get and update are dead privileges here. The base kustomize manifest and the e2e RBAC were reduced to create and patch in this PR; the Helm block was missed.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -327,58 +382,65 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
-	nrtUpdated, err := w.topoClient.TopologyV1alpha2().NodeResourceTopologies().Update(context.TODO(), nrtMutated, metav1.UpdateOptions{})
-	if err != nil {
-		return fmt.Errorf("failed to update NodeResourceTopology: %w", err)
+func (w *nfdTopologyUpdater) applyNRTTopologyManagerInfo(nrt *v1alpha2.NodeResourceTopology) error {
+	if w.getKubeletConfig() == nil {
+		refreshCtx, cancel := context.WithTimeout(context.Background(), topologyManagerInfoRefreshTimeout)
+		err := w.refreshTopologyManagerInfo(refreshCtx)
+		cancel()
+		if err != nil {
+			return err
```
ArangoGutierrez, line 462: When the startup cache is empty (startup refresh failed) and this refresh also fails, returning the error propagates through updateNodeResourceTopology to Run() and exits the process. On a non-interval publish event with the kubelet briefly unreachable this crash-loops the pod, whereas the startup and interval-refresh paths now tolerate the same failure and keep the cached info. Consider logging and publishing without the topology-manager attributes (or skipping this publish) so it matches that tolerance; the startup 'will retry on next scan' log implies a graceful retry that this path does not provide.

deployment/helm/node-feature-discovery/templates/clusterrole.yaml
alaypatel07: not sure how this got missed, updated now.

pkg/nfd-topology-updater/nfd-topology-updater.go
alaypatel07: +1 logged the error at the caller site and added unit test for this PTAL

pkg/nfd-topology-updater/nfd-topology-updater_test.go
```diff
@@ -110,3 +126,237 @@ func findAttributeByName(attrList v1alpha2.AttributeList, name string) (v1alpha2
...
+
+func TestUpdateNodeResourceTopologyApplyErrors(t *testing.T) {
+	const resource = "noderesourcetopologies"
+
+	Convey("Given the NodeResourceTopology resource is unavailable", t, func() {
+		topoClient := faketopologyclientset.NewSimpleClientset()
+
+		err := newTestTopologyUpdater(topoClient).updateNodeResourceTopology(nil, resourcemonitor.ScanResponse{})
+
+		So(apierrors.IsNotFound(err), ShouldBeTrue)
```
ArangoGutierrez, line 348: The NotFound here comes from the legacy tracker's Apply doing a Get first and failing on the absent object, not from a missing CRD; a real apiserver creates the object on apply, so this asserts the opposite of production behaviour and leaves create-via-apply untested. A PrependReactor returning apierrors.NewNotFound, like the sibling case already uses for the generic error, exercises the branch honestly. Worth flagging that NewClientset is not an alternative here: v0.1.3 generates an empty applyconfiguration schema, so its field-managed tracker fails with "no type found matching ... NodeResourceTopology".

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -327,58 +382,66 @@ func (w *nfdTopologyUpdater) updateNodeResourceTopology(zoneInfo v1alpha2.ZoneLi
...
-			if errors.IsNotFound(err) {
-				return fmt.Errorf("failed to create NodeResourceTopology: %w. "+
-					"The NodeResourceTopology CRD may not be installed. "+
-					"If using Helm, ensure 'topologyUpdater.createCRDs=true' is set",
-					err)
-			}
-			return fmt.Errorf("failed to create NodeResourceTopology: %w", err)
-		}
+	if err := w.applyNRTTopologyManagerInfo(nrt); err != nil {
+		klog.ErrorS(err, "failed to populate topology-manager information, skipping NodeResourceTopology publish")
```
ArangoGutierrez: Skipping the publish is a fine resolution of the earlier round, but the skip is currently invisible: no metric is bumped even though scanErrors is right there in the same loop, /healthz keeps returning 200, and under -oneshot Run() then returns nil so the process exits 0 having published nothing. Bumping an error counter here, and letting -oneshot surface the failure, would make a node whose /configz is permanently unreachable detectable.

ArangoGutierrez (commented):
Two notes on the new tests and the skip-publish path. The RBAC trim, the cached kubelet config plus CPU attributes, and the /configz timeout all look right on this head, and I confirmed the four new unit tests pass.

1. The NotFound here comes from the legacy tracker's Apply doing a Get first and failing on the absent object, not from a missing CRD; a real apiserver creates the object on apply, so this asserts the opposite of production behaviour and leaves create-via-apply untested. A PrependReactor returning apierrors.NewNotFound, like the sibling case already uses for the generic error, exercises the branch honestly. Worth flagging that NewClientset is not an alternative here: v0.1.3 generates an empty applyconfiguration schema, so its field-managed tracker fails with "no type found matching ... NodeResourceTopology". (pkg/nfd-topology-updater/nfd-topology-updater_test.go:323)
2. Skipping the publish is a fine resolution of the earlier round, but the skip is currently invisible: no metric is bumped even though scanErrors is right there in the same loop, /healthz keeps returning 200, and under -oneshot Run() then returns nil so the process exits 0 having published nothing. Bumping an error counter here, and letting -oneshot surface the failure, would make a node whose /configz is permanently unreachable detectable. (pkg/nfd-topology-updater/nfd-topology-updater.go:390)

pkg/nfd-topology-updater/nfd-topology-updater_test.go
alaypatel07: Nice catch, updated, PTAL

pkg/nfd-topology-updater/nfd-topology-updater.go
alaypatel07: fixed this path. The code now:
1. check if it is not found error and if it is one shot and returns error, otherwise update scan errors count
2. the error is only caught here and reported.

pkg/nfd-topology-updater/nfd-topology-updater_test.go
```diff
@@ -110,3 +126,244 @@ func findAttributeByName(attrList v1alpha2.AttributeList, name string) (v1alpha2
...
+func TestUpdateNodeResourceTopologyApplyErrors(t *testing.T) {
+	const resource = "noderesourcetopologies"
+
+	Convey("Given the NodeResourceTopology resource is unavailable", t, func() {
+		topoClient := faketopologyclientset.NewSimpleClientset()
+		notFoundErr := apierrors.NewNotFound(
+			v1alpha2.SchemeGroupVersion.WithResource(resource).GroupResource(),
+			"test-node",
+		)
+		topoClient.PrependReactor("patch", resource, func(k8stesting.Action) (bool, runtime.Object, error) {
```
ArangoGutierrez, line 342: The injected NotFound reactor never discriminates: `NewSimpleClientset` has no field management, so its `Apply` does a `Get` first and already returns the same NotFound when the object is absent. Deleting lines 320-326 and keeping the two assertions byte-identical still passes, so this block guards the fake's missing create-on-apply rather than the CRD-not-installed branch.

pkg/nfd-topology-updater/nfd-topology-updater_test.go
```diff
@@ -110,3 +126,244 @@ func findAttributeByName(attrList v1alpha2.AttributeList, name string) (v1alpha2
 	}
 	return v1alpha2.AttributeInfo{}, fmt.Errorf("Attribute Not Found name:=%s", name)
 }
+
+func TestUpdateNodeResourceTopologyApplies(t *testing.T) {
+	Convey("When NodeResourceTopology is published", t, func() {
+		topoClient := faketopologyclientset.NewSimpleClientset(&v1alpha2.NodeResourceTopology{
```
ArangoGutierrez: Every fixture seeds an existing NodeResourceTopology because `NewSimpleClientset` has no field management and its apply path requires the object to already exist, so the PR's central change - apply creating the object in place of the removed `Create` - is never exercised at unit level. The e2e that would cover it returns `true` on NotFound inside the `Eventually` closure, so it passes without the object coming back; and `NewClientset` is not a drop-in here, it fails with `schema error: no type found matching: ...v1alpha2.NodeResourceTopology`.

pkg/nfd-topology-updater/nfd-topology-updater.go
```diff
@@ -281,13 +335,25 @@ func (w *nfdTopologyUpdater) Run() error {
 			}
 			zones = resAggr.Aggregate(scanResponse.PodResources)
 			klog.V(1).InfoS("aggregated resources identified", "resourceZones", utils.DelayedDumper(zones))
-			readKubeletConfig := false
 			if info.Event == kubeletnotifier.IntervalBased {
-				readKubeletConfig = true
+				refreshCtx, cancel := context.WithTimeout(context.Background(), topologyManagerInfoRefreshTimeout)
```
ArangoGutierrez: The 2-minute bound is a clear improvement on the old unbounded `context.TODO()`, but rooting at `context.Background()` means `Stop()` still cannot cancel an in-flight kubelet fetch, so shutdown can wait out the timeout. Deriving from a context that closing `w.stop` cancels would finish the job; the sites at lines 259 and 446 have the same shape.

ArangoGutierrez (commented):
The SSA move is the right direction, and the RBAC change lines up consistently across the kustomize base, the Helm chart, and the e2e helper. Three notes, mostly about what the new tests actually pin down.

1. The injected NotFound reactor never discriminates: `NewSimpleClientset` has no field management, so its `Apply` does a `Get` first and already returns the same NotFound when the object is absent. Deleting lines 320-326 and keeping the two assertions byte-identical still passes, so this block guards the fake's missing create-on-apply rather than the CRD-not-installed branch. (pkg/nfd-topology-updater/nfd-topology-updater_test.go:324)
2. Every fixture seeds an existing NodeResourceTopology because `NewSimpleClientset` has no field management and its apply path requires the object to already exist, so the PR's central change - apply creating the object in place of the removed `Create` - is never exercised at unit level. The e2e that would cover it returns `true` on NotFound inside the `Eventually` closure, so it passes without the object coming back; and `NewClientset` is not a drop-in here, it fails with `schema error: no type found matching: ...v1alpha2.NodeResourceTopology`. (pkg/nfd-topology-updater/nfd-topology-updater_test.go:132)
3. The 2-minute bound is a clear improvement on the old unbounded `context.TODO()`, but rooting at `context.Background()` means `Stop()` still cannot cancel an in-flight kubelet fetch, so shutdown can wait out the timeout. Deriving from a context that closing `w.stop` cancels would finish the job; the sites at lines 259 and 446 have the same shape. (pkg/nfd-topology-updater/nfd-topology-updater.go:339)

alaypatel07:
All three review comments are now addressed:

1. NotFound test seeds an existing NRT, so only the injected reactor can produce NotFound.
2. Create-via-apply coverage now starts without an NRT, and the E2E correctly retries on NotFound.
3. Kubelet refresh contexts are canceled when Stop() closes w.stop, with unit coverage.

ArangoGutierrez (approved):
/lgtm

ArangoGutierrez:
@alaypatel07 quick question on backporting, since we are looking at a v0.19.1 patch release this week.

My read is that this one should stay on master rather than come along. It swaps `get, update` for `patch` in both the topology-updater ClusterRole and the Helm template, so anyone who bumped the image without re-applying manifests would end up with the updater lacking the verb it now needs. That is a rough thing to hand someone in a patch release, and the change itself is a performance improvement rather than a fix, so there is no urgency forcing it in.

Planning to leave it for the next minor unless you are seeing apiserver or etcd load in the field that makes the backport worth the RBAC coordination. If you are, say so and I will reconsider, though we would want to call the RBAC change out loudly in the release notes.

PR #2577, opened 2026-08-31
helm: set explicit namespace on the PodMonitor

### LINKED ISSUE #2576: PodMonitor is missing metadata.namespace when with Helm + Kustomize 5.8.x
## What happened

When the NFD Helm chart is rendered through Kustomize’s Helm integration, the generated `PodMonitor` does not contain `metadata.namespace`.

The rendered resource looks like:

```yaml
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
  name: nfd-node-feature-discovery
  # namespace is missing
```

With Kustomize 5.8.x, the parent namespace is not reliably propagated to the PodMonitor [see kustomize/issues/6058](https://github.com/kubernetes-sigs/kustomize/issues/6058). I am using Argo CD to do my deployments, which then rejects the PodMonitor sync:

```text
InvalidSpecError: Namespace for nfd-node-feature-discovery
monitoring.coreos.com/v1, Kind=PodMonitor is missing.
```

## Environment

- NFD Helm chart: `1.1.4`
- Kustomize: `5.8.x`
- Deployment tool: Argo CD
- Resource: `monitoring.coreos.com/v1`, `PodMonitor`
- Resource name: `nfd-node-feature-discovery`

## Expected behavior

The Helm chart should render the `PodMonitor` with the release namespace, consistent with the namespace used by the NFD workloads:

```yaml
metadata:
  name: nfd-node-feature-discovery
  namespace: {{ include "node-feature-discovery.namespace" . }}
```

## How to reproduce

Create a new directory containing this `kustomization.yaml`:

```yaml
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

helmCharts:
  - name: node-feature-discovery
    repo: https://kubernetes-sigs.github.io/node-feature-discovery/charts
    version: 0.18.3
    releaseName: nfd
    valuesInline:
      prometheus:
        enable: true
```

Render the chart with Kustomize `v5.8.1`

```bash
kustomize build --enable-helm . > rendered.yaml
```

The rendered `PodMonitor` contains the expected name and labels, but no `metadata.namespace`:

The same directory rendered with Kustomize `v5.7.1` includes the namespace (no bug)

```yaml
metadata:
  name: nfd-node-feature-discovery
```

With Kustomize `v5.8.1`, the `metadata.namespace` field is absent. This matches the behavior described in [Kustomize issue #6058](https://github.com/kubernetes-sigs/kustomize/issues/6058), which was introduced by the namespace handling changes in [Kustomize PR #5940](https://github.com/kubernetes-sigs/kustomize/pull/5940).

## Additional context

A Helm installation without Kustomize will succeed because the release namespace is used. The issue is exposed only when the chart is rendered using the Kustomize Helm integration.

## Workaround

A post-render Kustomize patch can be applied:

```yaml
- op: add
  path: /metadata/namespace
  value: node-feature-discovery
```

However, this requires folks to maintain a chart-specific workaround.

## Proposed fix

Add an explicit `metadata.namespace` to the Helm `PodMonitor` template using the chart’s existing namespace helper.

trastle:
The PodMonitor was the only namespaced resource in the chart without an explicit metadata.namespace. Every other namespaced template (master, worker, topology-updater, the ConfigMaps, ServiceAccounts, Roles, RoleBindings, PDBs and NetworkPolicies) already sets it via the node-feature-discovery.namespace helper.

Plain `helm install` was unaffected, since Helm assigns the release namespace to any resource that omits one. Rendering the chart through Kustomize's Helm integration is not: as of Kustomize 5.8.x the parent kustomization's namespace is no longer propagated to resources inflated from a chart (kubernetes-sigs/kustomize#6058, introduced by kubernetes-sigs/kustomize#5940), so the PodMonitor came out with no namespace at all. Argo CD then refuses to sync it:

  InvalidSpecError: Namespace for nfd-node-feature-discovery
  monitoring.coreos.com/v1, Kind=PodMonitor is missing.

Set the namespace from the existing helper, which also keeps the PodMonitor's own namespace consistent with the spec.namespaceSelector just below it and honours namespaceOverride.

Fixes #2576

### DIFF
--- a/deployment/helm/node-feature-discovery/templates/prometheus.yaml
+++ b/deployment/helm/node-feature-discovery/templates/prometheus.yaml
@@ -4,6 +4,7 @@ apiVersion: monitoring.coreos.com/v1
 kind: PodMonitor
 metadata:
   name: {{ include "node-feature-discovery.fullname" . }}
+  namespace: {{ include "node-feature-discovery.namespace" . }}
   labels:
     {{- include "node-feature-discovery.selectorLabels" . | nindent 4 }}
     {{- with .Values.prometheus.labels }}


trastle:
I fixed up the commit message, which I think made the bot happy.
Please let me know if any other changes are needed.

ArangoGutierrez (approved):
Thanks for chasing this back to the Kustomize 5.8 change and fixing it in the template: the PodMonitor now carries the same namespace as the namespaceSelector right below it, which is what unblocks the Argo CD sync.

ArangoGutierrez:
/ok-to-test
/approve

Checked this one locally: rendering the chart at this commit with `--namespace monitoring` puts `namespace: monitoring` on the PodMonitor, and setting `namespaceOverride` moves it accordingly. On master the field comes out absent, which is the Argo CD sync error in #2576. The PodMonitor was the last namespaced template in the chart without it, so this finishes what 76df597c1 started when it added the helper.

/cc @PiotrProkop

Piotr, could you give this a look and drop an lgtm if it reads right to you?

marquiz (approved):
Looks good to me

PR #2586, opened 2026-09-13
helm: add updateStrategy to the topology-updater DaemonSet

### LINKED ISSUE #2523: NFD helmrelease upgrade times out for a cluster having large number of nodes
**What happened**: NFD helmrelease upgrade times out for a cluster having large number of nodes

**What you expected to happen**: NFD helmrelease upgrade should go through without any timeout issue

1. Create a cluster with large number of nodes (lets take 30)
2. Try helmrelease upgrade with default helmrelease timeout of 5 mins
3. Helmrelease update times out and rollbacks to old version

**Environment**:
- Kubernetes version (use `kubectl version`):
Client Version: v1.35.2+rke2r1
Kustomize Version: v5.7.1
Server Version: v1.35.2+rke2r1
- Cloud provider or hardware configuration: RKE2 cluster
- OS (e.g: `cat /etc/os-release`): Suse SLES 15SP3
- Kernel (e.g. `uname -a`): Linux 6.4.0-150600.23.53-default
- Install tools: fluxcd source-controller and helmcontroller
- Network plugin and version (if this is a network-related bug): NA
- Others: NA

shashankvarma499:
## What this PR does / why we need it

The nfd-topology-updater DaemonSet has no `updateStrategy` value, so on upgrade it
falls back to the Kubernetes default of rolling one node at a time
(`maxUnavailable: 1`). On clusters with many nodes this makes the rollout take long
enough that Helm and Flux HelmRelease upgrades time out and roll back
(see #2523).

This mirrors the existing `worker.updateStrategy` knob for the topology-updater,
defaulting to a rolling update with `maxUnavailable: "10%"`. The topology-updater is
stateless and recomputes NodeResourceTopology objects on restart, so a faster roll
is safe, and the default now bounds the number of rollout waves on clusters of any
size. Users who prefer the Kubernetes default can set `maxUnavailable: 1`.

## Which issue(s) this PR fixes

Fixes #2523

## Changes

- `values.yaml`: add `topologyUpdater.updateStrategy` (rolling, `maxUnavailable: "10%"`).
- `templates/topologyupdater.yaml`: render `updateStrategy` on the DaemonSet.
- Regenerated `values.schema.json`, `README.md`, and `docs/deployment/helm.md`.


deployment/helm/node-feature-discovery/values.yaml
```diff
@@ -897,6 +897,22 @@ topologyUpdater:
...
+  # -- Update strategy for the nfd-topology-updater DaemonSet. Defaults to a
+  # rolling update with `maxUnavailable: "10%"` so upgrades complete in a
+  # bounded number of waves on clusters of any size (the Kubernetes default of
+  # `maxUnavailable: 1` rolls one node at a time and makes Helm wait and
+  # Flux HelmRelease timeouts likely on large clusters). The topology-updater
+  # is stateless and recomputes NodeResourceTopology objects on restart, so a
+  # faster roll is safe. Set `maxUnavailable: 1` to restore the Kubernetes
+  # default. To use `type: OnDelete`, also set `rollingUpdate: null`
+  # (Helm deep-merges maps).
+  # [More info](https://kubernetes.io/docs/tasks/manage-daemon/update-daemon-set)
```
marquiz, line 910: I think we could be slightly less verbose here. At least not "hard code" the defaults in the description text as they're shown (and auto-updated) in the values table.

marquiz (commented):
Thanks @shashankvarma499 for the patch. Just one small nit about the description

/ok-to-test

